PR #44, opened 2014-06-14
Arabic Normalizer (c++)  + Buckwalter Romanization Converter

assem-ch:
(This pull request is for reviewing)


xapian-core/queryparser/arabic-normalizer.h
```diff
@@ -0,0 +1,98 @@
...
+#include "xapian/visibility.h"
+
+#include <string>
+
+/** Arabic Normalizer, a  class to normalize Arabic text.
+ * its role is:
+ * - normalize Arabic text
+ * - converted Arabic romanized texts back to Arabic (arabization)
+ */
+class XAPIAN_VISIBILITY_DEFAULT ArabicNormalizer {
```
assem-ch: @ojwb it is going to be  new API class, where to add external API headers?

xapian-core/queryparser/arabic-normalizer.cc
```diff
@@ -0,0 +1,122 @@
...
+        subst_ch = get_value(ch, mapping, mapping_length);
+        if ( subst_ch != 0 ) {
+            Xapian::Unicode::append_utf8(new_word, subst_ch);
+        }
+    }
+    return new_word;
+}
+
+int ArabicNormalizer::guess_romanization_system(const string& word) {
+    return 0 && LENGTH(word); // TODO to be implemented
```
assem-ch: Each randomization has a special set of characters to use and specific rules. The idea is to use  this character to tighten the guessing interval till get to one system (see Comparison table https://en.wikipedia.org/wiki/Romanization_of_Arabic)

assem-ch:
If you want to avoid introducing Normalizers as an independent class in Xapian, I may include some normalization cases within the snowball stemmer as for the case of French. 

>The big things that seem to be needed here still are sorting out the public API for this, and adding test cases to make sure it actually works as intended (and to help catch any breakage from future changes). The docs should also be extended to discuss using normalisers too (I'd suggest this is best done in the "getting started" guide, as the plan is to eventually slot that into xapian-core to replace the existing non-collated documentation).

xapian-core/queryparser/arabic-normalizer.h
ojwb: External API headers are in `xapian-core/include/xapian/`.

ojwb:

assem-ch:
@owjb The arabic stemmer in PR  #45  takes care of normalization so I see this PR should be closed unless you have a bigger vision to separate other linguistic analyzing phases  from stemmers.

PR #48, opened 2014-06-19
Pl2 optimization

aarshkshah1992:
Optimizing the PL2 weighting scheme.

### DIFF
--- a/xapian-core/weight/pl2weight.cc
+++ b/xapian-core/weight/pl2weight.cc
@@ -1,7 +1,7 @@
 /** @file pl2weight.cc
  * @brief Xapian::PL2Weight class - the PL2 weighting scheme of the DFR framework.
  */
-/* Copyright (C) 2013 Aarsh Shah
+/* Copyright (C) 2013,2014 Aarsh Shah
  * Copyright (C) 2013,2014 Olly Betts
  *
  * This program is free software; you can redistribute it and/or
--- a/xapian-core/weight/pl2weight.cc
+++ b/xapian-core/weight/pl2weight.cc
@@ -85,19 +85,23 @@ PL2Weight::init(double)
     double term2;
 
     // P = ((wdfn + 0.5) * log2(wdfn) + P1 - (P2 * wdfn)) / (wdfn + 1.0).
-    // We break the term into three parts with (wdfn + 1.0) as the denominator in each.
-    // The maximum of the first term, i.e (wdfn + 0.5) * log2(wdfn) / (wdfn + 1.0),
-    // will happen at wdfn_upper as it is an increasing function.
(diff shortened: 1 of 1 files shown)


ojwb:

xapian-core/weight/pl2weight.cc
```diff
@@ -78,13 +78,52 @@ PL2Weight::init(double)
 
     // Calculate the lower bound on the weight.
     double P_min =
-	P1 + (wdfn_lower + 0.5) * log2(wdfn_lower) - P2 * wdfn_lower;
+	P1 + (wdfn_lower + 0.5) * log2(wdfn_lower) - P2 * wdfn_upper;
```
ojwb: I've committed a change to address this part fully - we now calculate upper_bound correctly whatever sign P2 has (dd6b359).

I've also dropped the lower_bound approach (which doesn't actually work as intended in general) and instead clamp the final weight to be >= 0 in get_sumpart() (1039fab).

xapian-core/weight/pl2weight.cc
```diff
@@ -78,13 +78,52 @@ PL2Weight::init(double)
```
ojwb: I've proved that we do indeed want to maximise x in this case, and implemented this in 01f7233a19460ba555f3cc13d964e862250668ca.

xapian-core/weight/pl2weight.cc
```diff
@@ -78,13 +78,52 @@ PL2Weight::init(double)
...
+    // is negative.
+    // The third term will be minimum at wdfn_lower as it is an increasing
+    // function.
+    // In order to determine the nature of the second term, i.e P1 /
+    // (wdfn + 1.0) in colloboration with the first term, we differentiate the
+    // expression of the first and the second terms i.e (P1 + (wdfn  + 0.5) *
+    // log2(wdfn)) / (wdfn + 1.0).
+    // The numerator of the differentiation can be negative.
+    // So, we again differentiate the numerator and check the value of the
+    // second and the first differentiation at wdfn_upper.
```
ojwb: While the details here aren't quite right, the idea of splitting the equation up and maximising each part individually is certainly sound.

The best approach seems to be to split it into two parts - one with the `log` and one for the rest.  When `P1 + P2 > 0` (which is often true) then both are optimised at `wdfn_upper`.

I've implemented that in 0964391357ec249cf133c288685e7fab63425eee.

This makes a huge difference in the upper bound - in a simple test, the upper bounds are now just over double the highest weight actually achieved - previously they were several hundred times.  So many thanks for coming up with this approach.

ojwb:
Everything here has been addressed now, so closing.

PR #53, opened 2014-08-01
fixed-width format and skip-list format for postlist

ShangtongZhang:
Apply fixed-width format for postlist about document length and skip-list format for normal postlist.


ojwb:

ojwb:
I've been looking over again.

I tried this branch out on a sample database (from apt-xapian-index on my dev box) and here are the sizes for chert, brass as of the branch point for this branch, this branch, and glass on git master in KB, and % of the chert size:

```
100.00%  53800	tmp.axi.chertc.db
 99.96%  53780	tmp.axi.baseline.brassc.db
104.69%  56324	tmp.axi.skiplist.brassc.db
 95.10%  51164	tmp.axi.glassc.db
```

That shows this branch adds close to 5% size overhead, while glass saves about 5% size overhead.  I'm not sure what the combined effect would be for putting these encodings inside glass, but I'd expect somewhere between 5% and 10% size overhead.

The approach taken here adds a way to mix fixed width chunks and skip lists with the existing encoding.  One downside of this is added overhead to switch mode within a chunk, and extra complexity when modifying a chunk.  It also dulls the big advantage of a fixed-width encoding of being able to seek in O(1).  I think it makes more sense to select fixed-width or variable width on a per-chunk basis, or perhaps even per term - either will incur a little extra overhead (from more chunks or from encoding as a fixed width which is larger than some data items), but we'd save on the bytes needed to switch encoding.

The latest honey backend supports a fixed-width encoding for document lengths, which gain particularly from this as they're typically present for long ranges of consecutive docids.  It's generally rare for a term to index a long run of consecutive docids, but we could extend this to them.

I'm not convinced that the overhead of storing a within-chunk skip list is worthwhile.  We can efficiently skip to the correct chunk via the table cursor.  Then a fixed width encoding mean O(1) within the chunk, while for variable width the cost of the skip is bounded by the largest chunk size, which is fixed so that's actually O(1) too (just a worse O(1)).

So I think my conclusion was that this was an interesting and useful experiment, but not something we want to merge as-is, and the pieces that we might want (fixed width chunks for terms) are better adapted from the newer honey code, especially since the code here is entangled with BrightStation copyright code so would need rewriting anyway.

Therefore I'm going to close this now.

PR #57, opened 2014-08-28
Makesnippetqueryrelevant: display snippet showing the area surrounding the searched word / phrase text

samuelharden:
Added the cosine similarity as an addition factor with weight.

To get the query term in the snippet. Attached is the image from local testing.

Most of the document seems to get the query word but few of them are still not able to get the query words, but snippet is picked from a location very close to the query word. needs a little more tunning of parameters for that i guess.

2014-08-28-16 37 44-screenshot


ojwb:
I seem to have managed to close all open PRs 28 days ago, which wasn't intended.

However, as discussed elsewhere this one has now been superseded by the new snippet implementation which takes the best ideas from the various implementations we had.

PR #63, opened 2014-12-16
Pkgconfig

seb777:
Add pkg-config support to libxapian - fix ticket #540


ojwb:
Fixed in 82f20c1.

seb777:
Arggss...

:+1: 
Thanks Ojwb

PR #65, opened 2014-12-23
Improved Support for XAPIAN::DB_RETRY_LOCK in Windows

rand0wn:
* Used LockFileEx() in Windows to Lock Files

### DIFF
--- a/xapian-core/backends/flint_lock.cc
+++ b/xapian-core/backends/flint_lock.cc
@@ -42,6 +42,7 @@
 
 #ifdef __CYGWIN__
 #include <sys/cygwin.h>
+#include "safewindows.h"
 #endif
 
 #ifdef FLINTLOCK_USE_FLOCK
@@ -85,9 +86,7 @@ FlintLock::lock(bool exclusive, bool wait, string & explanation) {
     const char *fnm = filename.c_str();
 #endif
 retry:
-    // FIXME: Use LockFileEx() for locking, which would allow proper blocking
-    // and also byte-range locking for when we implement MVCC.  But is there a
-    // way to interwork with the CreateFile()-based locking while doing so?
+   
     hFile = CreateFile(fnm, GENERIC_WRITE, FILE_SHARE_READ,
 		       NULL, OPEN_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
     if (hFile != INVALID_HANDLE_VALUE) return SUCCESS;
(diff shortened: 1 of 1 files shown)


xapian-core/backends/flint_lock.cc
```diff
@@ -42,6 +42,7 @@
 
 #ifdef __CYGWIN__
 #include <sys/cygwin.h>
+#include "safewindows.h"
```
ojwb: The `safewindows.h` header already gets included by `flintlock.h` when `__CYGWIN__` is defined.

xapian-core/backends/flint_lock.cc
```diff
@@ -150,6 +149,23 @@ FlintLock::lock(bool exclusive, bool wait, string & explanation) {
...
+                               // Requires Exclusive access hFile rather than standard filename.c_str() from lockfd
+                                 
+     GetLastError();   // Lock Failed check for errors
+     return UNKNOWN;                 
+    }
+    fd = fnm;                  
+    return SUCCESS;          
+
+#endif
+
```
ojwb: You're adding this code to the `#else` branch, so it'll never actually get used, as the first condition checked is:

```
#if defined __CYGWIN__ || defined __WIN32__
```

The code you've added here wouldn't actually compile, which is what tipped me off - `fd = fnm;` can't work as `fd` is an `int` here, while `fnm` is `std::string`.

Your indentation is off too.

ojwb:
Any updates?

rand0wn:
I guess I m unable to understand the transformation of FD = fnm in windows
I m sure u will be able to help me with it.

rand0wn:
If maybe you could help me in understanding point of fd here which i assume
is a check point, we just need to derive it in windows with something
similar to check the lock.

ojwb:
I recall we discussed this on IRC, but progress seems to have stalled.  The proposed change isn't valid code, so I'm going to close this PR.

I'd certainly be happy to see this worked on though - if you are still interested in doing so (or indeed if anyone else is), I'm happy to try to explain what's needed more clearly.

rand0wn:
Hi Olly,

Indeed it's been a long time and I lost thought of this patch, and I
loosely remember the details, should not take long time to figure out and
since it came up now, i am interested, well there were two critical
problems,

1. Adding code to wrong '#if', i added it to else branch

2. Figuring out 'fd', holding the lock and 'fnm' as in windows 'fnm' differs

Help me out here, and Mention your IRC time ( As I am in India).

Regards,
Abhishek

PR #66, opened 2014-12-24
Replace atoi with strtol: Initial Patch

PriyankBhatt:
1.Modified atoi with strtol and strtoll depending upon the system that is 32bit or 64bit. This will help to represent more year's with 64bit system and atlast ensuring that conversion of second's at last occurs in time_t format by type casting it.
2.Also checked for underflow or overflow condition using errno variable.

### DIFF
--- a/xapian-applications/omega/date.cc
+++ b/xapian-applications/omega/date.cc
@@ -166,7 +166,7 @@ date_range_filter(const string & date_start, const string & date_end,
 	    }
 	    secs = time_t(str_to_long) * 24 * 60 * 60;
 	} else {
-	    long str_to_longlong;
+	    long long str_to_longlong;
 	    errno = 0;
 	    str_to_longlong = strtoll(date_span.c_str(),NULL,0);
 	    if (errno == ERANGE) {
--- a/xapian-applications/omega/date.cc
+++ b/xapian-applications/omega/date.cc
@@ -32,7 +32,7 @@
 #include <cstdlib>
 #include <ctime>
 #include <climits> // for detecting underflow or overflow e.g LONG_MIN
-#include <cstring> // for strerror
+#include <limits>
 #include <errno.h> // for errno variable
 
 using namespace std;
@@ -147,40 +147,19 @@ date_range_filter(const string & date_start, const string & date_end,
 {
     int y1, m1, d1, y2, m2, d2;
     if (!date_span.empty()) {
-    	time_t secs;
-	// For x32 system use long (32bit)
-	// For x64 system use long long (64bit)
-	if (sizeof(int *) < 8) {
-	    long str_to_long;
-	    errno = 0;
-	    str_to_long = strtol(date_span.c_str(),NULL,0);
-	    if (errno == ERANGE) {
-		switch(str_to_long) {
-		    case LONG_MIN:
-			throw (string("Date span underflow: ") + string(strerror(errno)));
-		    break;
-		    case LONG_MAX:
-			throw (string("Date span overflow: ") + string(strerror(errno)));
-		    break;
-		}
+	errno = 0;
+	long long str_to_longlong = strtoll(date_span.c_str(), NULL, 0);
+	if (errno == ERANGE) {
+	    switch (str_to_longlong) {
+		case LLONG_MIN:
+		    throw (string("Date span underflow:"));
+		case LLONG_MAX:
+		    throw (string("Date span overflow:"));
 	    }
-	    secs = time_t(str_to_long) * 24 * 60 * 60;
-	} else {
-	    long long str_to_longlong;
-	    errno = 0;
-	    str_to_longlong = strtoll(date_span.c_str(),NULL,0);
-	    if (errno == ERANGE) {
-		switch(str_to_longlong) {
-		    case LLONG_MIN:
-			throw (string("Date span underflow: ") + string(strerror(errno)));
-		    break;
-		    case LLONG_MAX:
-			throw (string("Date span overflow: ") + string(strerror(errno)));
-		    break;
-		}
-	    }
-	    secs = time_t(str_to_longlong) * 24 * 60 * 60;
 	}
+	if (time_t(str_to_longlong) > (numeric_limits<time_t>::max() / (24 * 60 * 60)))
+	    throw (string("Date span overflow:"));
+	time_t span = time_t(str_to_longlong) * (24 * 60 * 60);
 	if (!date_end.empty()) {
 	    parse_date(date_end, &y2, &m2, &d2);
 	    struct tm t;
--- a/xapian-applications/omega/date.cc
+++ b/xapian-applications/omega/date.cc
@@ -31,6 +31,9 @@
 #include <stdio.h>
 #include <cstdlib>
 #include <ctime>
+#include <climits> // for detecting underflow or overflow e.g LONG_MIN
+#include <cstring> // for strerror
+#include <errno.h> // for errno variable
 
 using namespace std;
 
@@ -144,7 +147,40 @@ date_range_filter(const string & date_start, const string & date_end,
 {
     int y1, m1, d1, y2, m2, d2;
     if (!date_span.empty()) {
-	time_t secs = atoi(date_span.c_str()) * (24 * 60 * 60);
+    	time_t secs;
+	// For x32 system use long (32bit)
+	// For x64 system use long long (64bit)
+	if (sizeof(int *) < 8) {
+	    long str_to_long;
+	    errno = 0;
+	    str_to_long = strtol(date_span.c_str(),NULL,0);
+	    if (errno == ERANGE) {
+		switch(str_to_long) {
+		    case LONG_MIN:
+			throw (string("Date span underflow: ") + string(strerror(errno)));
+		    break;
+		    case LONG_MAX:
+			throw (string("Date span overflow: ") + string(strerror(errno)));
+		    break;
+		}
+	    }
+	    secs = time_t(str_to_long) * 24 * 60 * 60;
+	} else {
+	    long str_to_longlong;
+	    errno = 0;
+	    str_to_longlong = strtoll(date_span.c_str(),NULL,0);
+	    if (errno == ERANGE) {
+		switch(str_to_longlong) {
+		    case LLONG_MIN:
+			throw (string("Date span underflow: ") + string(strerror(errno)));
+		    break;
+		    case LLONG_MAX:
+			throw (string("Date span overflow: ") + string(strerror(errno)));
+		    break;
+		}
+	    }
+	    secs = time_t(str_to_longlong) * 24 * 60 * 60;
+	}
 	if (!date_end.empty()) {
 	    parse_date(date_end, &y2, &m2, &d2);
 	    struct tm t;
--- a/xapian-applications/omega/date.cc
+++ b/xapian-applications/omega/date.cc
@@ -147,10 +147,12 @@ date_range_filter(const string & date_start, const string & date_end,
     if (!date_span.empty()) {
 	long long str_to_longlong = strtoll(date_span.c_str(), NULL, 0);
 	time_t secs;
-	if (str_to_longlong > (numeric_limits<time_t>::max() / (24 * 60 * 60))) {
-	    secs = numeric_limits<time_t>::max();
-	} else if (str_to_longlong < (numeric_limits<time_t>::min() / (24 * 60 * 60))) {
-	    secs = numeric_limits<time_t>::min();
+	time_t max_time = numeric_limits<time_t>::max();
+	time_t min_time = numeric_limits<time_t>::min();
+	if (str_to_longlong > (max_time / (24 * 60 * 60))) {
+	    secs = max_time;
+	} else if (str_to_longlong < (min_time / (24 * 60 * 60))) {
+	    secs = min_time;
 	} else {
 	    secs = time_t(str_to_longlong) * (24 * 60 * 60);
 	}
@@ -164,15 +166,17 @@ date_range_filter(const string & date_start, const string & date_end,
 	    t.tm_min = t.tm_sec = 0;
 	    t.tm_isdst = -1;
 	    time_t given_time = mktime(&t);
+	    if (secs == min_time)
+	    	secs += 1;
 	    time_t then = given_time - secs;
 	    bool flag = given_time < 0;
 	    //check for underflow
 	    if (flag == (secs > 0) && flag != (then < 0))
-		then = numeric_limits<time_t>::min();
+		then = min_time;
 	    flag = given_time > 0;
 	    //check for overflow
 	    if (flag == (secs < 0) && flag != (then > 0))
-		then = numeric_limits<time_t>::max();
+		then = max_time;
 	    struct tm *t2 = localtime(&then);
 	    y1 = t2->tm_year + 1900;
 	    m1 = t2->tm_mon + 1;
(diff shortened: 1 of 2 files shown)


xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,40 @@ date_range_filter(const string & date_start, const string & date_end,
 {
     int y1, m1, d1, y2, m2, d2;
     if (!date_span.empty()) {
-	time_t secs = atoi(date_span.c_str()) * (24 * 60 * 60);
+    	time_t secs;
+	// For x32 system use long (32bit)
+	// For x64 system use long long (64bit)
+	if (sizeof(int *) < 8) {
```
ojwb: The size of a pointer really isn't relevant, and the world isn't just "x32 or x64".

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,40 @@ date_range_filter(const string & date_start, const string & date_end,
...
+		switch(str_to_long) {
+		    case LONG_MIN:
+			throw (string("Date span underflow: ") + string(strerror(errno)));
+		    break;
+		    case LONG_MAX:
+			throw (string("Date span overflow: ") + string(strerror(errno)));
+		    break;
+		}
+	    }
+	    secs = time_t(str_to_long) * 24 * 60 * 60;
```
ojwb: You carefully check for overflow in the conversion, but then blindly multiply here, which could easily overflow (e.g. if `DATESPAN=99999` and the time_t is 32 bit).

I think here it's not actually helpful to raise an error in either case.  It means that `DATESPAN=9999999999` will always raise an exception on some architectures, but work on others.  It would be more helpful to clamp overflow to the largest possible `time_t` value - we can't represent a time after that, so it really is the appropriate limit in this case.

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,40 @@ date_range_filter(const string & date_start, const string & date_end,
...
-	time_t secs = atoi(date_span.c_str()) * (24 * 60 * 60);
+    	time_t secs;
+	// For x32 system use long (32bit)
+	// For x64 system use long long (64bit)
+	if (sizeof(int *) < 8) {
+	    long str_to_long;
+	    errno = 0;
+	    str_to_long = strtol(date_span.c_str(),NULL,0);
+	    if (errno == ERANGE) {
+		switch(str_to_long) {
```
ojwb: Xapian coding convention is a space before the `(` here.

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,40 @@ date_range_filter(const string & date_start, const string & date_end,
...
+		    case LONG_MIN:
+			throw (string("Date span underflow: ") + string(strerror(errno)));
+		    break;
```
ojwb: `break;` is mis-indented here, but there's no point having a `break;` after a throw anyway.

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,40 @@ date_range_filter(const string & date_start, const string & date_end,
...
+		    case LONG_MAX:
+			throw (string("Date span overflow: ") + string(strerror(errno)));
```
ojwb: There's no need to convert both arguments to a `std::string` - there's an `operator+` overload for std::string + const char *, and making use of that will avoid creating an extra temporary `std::string` object, i.e.:

```
    throw string("Date span overflow) + strerror(errno);
```

But we know that the error is `ERANGE` (we just checked that) and `Date span overflow: Numerical result out of range` doesn't provide useful additional information over just `Date span overflow`.

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,40 @@ date_range_filter(const string & date_start, const string & date_end,
...
     int y1, m1, d1, y2, m2, d2;
     if (!date_span.empty()) {
-	time_t secs = atoi(date_span.c_str()) * (24 * 60 * 60);
+    	time_t secs;
+	// For x32 system use long (32bit)
+	// For x64 system use long long (64bit)
+	if (sizeof(int *) < 8) {
+	    long str_to_long;
+	    errno = 0;
+	    str_to_long = strtol(date_span.c_str(),NULL,0);
```
ojwb: Xapian coding convention is to put a space after commas in parameter lists.

ojwb: Also, it's better style in C++ to define a variable where you initialise it, so:

```
    errno = 0;
    long str_to_long = strtol(date_span.c_str(), NULL, 0);
```

PriyankBhatt:
1.Main purpose for using size of pointer is to check for  operating system, For Unix/Linux which uses long and pointer are 64bit whereas for Windows pointer are 64bit and long is 32bit on 64bit operating system. And time_t depends on the kernel which is used. time_t can be 32bit for x32 operating system and 64bit for x64 operating system's. 
2.Yes i completely missed that for checking multiplication ,Maximum time  that can be supported depends on our operating system so we can use long for 32bit but for 64bit linux/unix system long is 64bit, if we use (strtol).
3.Sorry I completely forgot about that.

ojwb:
My key point is that checking the pointer size doesn't tell you what size `time_t` is.  Any platform which currently uses signed 32-bit time_t and wants to live beyond 2038 is going to have to migrate to 64-bit `time_t` (and realistically well before 2038 as the inability to represent dates in the approaching future will cause problems for some applications).

If you want to know what the highest value a `time_t` can represent is, it's much more robust to ask for it directly, rather than trying to infer it from testing the size of something else and applying knowledge of the subset of platforms you know about.  In C++, the `numeric_limits` template allows you to introspect on a type, and can tell you the maximum value it can represent like so: `numeric_limits<time_t>::max()` - see:

http://www.cplusplus.com/reference/limits/numeric_limits/

As this is specified by the C++ standard, it should work reliably for any platform, be it an existing one or one not yet created.

PriyankBhatt:
Hello Olly,
As you told to clamp the overflow value, lets say i clamp it on max value for time_t in 64bit. so i implemented something like this:
	time_t secs;
	errno = 0;
	long long str_to_longlong = strtoll(date_span.c_str(), NULL, 0);
	if (errno == ERANGE) {
	    switch(str_to_longlong) {
		case LLONG_MIN:
		    throw (string("Date span underflow:"));
		case LLONG_MAX:
		    throw (string("Date span overflow:"));
	    }
	}
	if (time_t(str_to_longlong) > (numeric_limits<time_t>::max() / (24 * 60 * 60))
		throw (string("Date span overflow:"));
	secs = time_t(str_to_longlong) * (24 * 60 * 60);
This would check for both multiplication overflow and strtoll overflow but still the problem is that for a particular value may work for some system and may not work in another system. 
So if we clamp our overflow value for max value of time_t in 32bit system. Then the implementation would be something like this:
time_t secs;
	errno = 0;
	long str_to_long = strtol(date_span.c_str(), NULL, 0);
	if (errno == ERANGE) {
	    switch(str_to_long) {
		case LONG_MIN:
		    throw (string("Date span underflow:"));
		case LONG_MAX:
		    throw (string("Date span overflow:"));
	    }
	}
	if (time_t(str_to_long) > (numeric_limits<time_t>::max() / (24 * 60 * 60))
		throw (string("Date span overflow:"));
         secs = time_t(str_to_long) * (24 * 60 * 60);
This would also create the same problem . As long is max value is different for different operating system so some value's may work on particular system and may not work on different system.

PriyankBhatt:
Hello Olly,
As you told to clamp the overflow value, lets say i clamp it on max value for time_t in 64bit. so i implemented something like this:
	time_t secs;
	errno = 0;
	long long str_to_longlong = strtoll(date_span.c_str(), NULL, 0);
	if (errno == ERANGE) {
	    switch(str_to_longlong) {
		case LLONG_MIN:
		    throw (string("Date span underflow:"));
		case LLONG_MAX:
		    throw (string("Date span overflow:"));
	    }
	}
	if (time_t(str_to_longlong) > (numeric_limits<time_t>::max() / (24 * 60 * 60))
		throw (string("Date span overflow:"));
	secs = time_t(str_to_longlong) * (24 * 60 * 60);
This would check for both multiplication overflow and strtoll overflow but still the problem is that for a particular value may work for some system and may not work in another system. 
So if we clamp our overflow value for max value of time_t in 32bit system. Then the implementation would be something like this:
time_t secs;
	errno = 0;
	long str_to_long = strtol(date_span.c_str(), NULL, 0);
	if (errno == ERANGE) {
	    switch(str_to_long) {
		case LONG_MIN:
		    throw (string("Date span underflow:"));
		case LONG_MAX:
		    throw (string("Date span overflow:"));
	    }
	}
	if (time_t(str_to_long) > (numeric_limits<time_t>::max() / (24 * 60 * 60))
		throw (string("Date span overflow:"));
         secs = time_t(str_to_long) * (24 * 60 * 60);
This would also create the same problem . As long is max value is different for different operating system so some value's may work on particular system and may not work on different system.

PriyankBhatt:
Hello Olly,
I have capped time_t value and as you said i have used numeric_limits for checking purpose.

ojwb:
Note: If you're quoting code in github comments, put a line with just ``` in before and after each block of code so it is readable and not reformatted.

Sorry for the delay - taking a look at this now.

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,19 @@ date_range_filter(const string & date_start, const string & date_end,
...
-	time_t secs = atoi(date_span.c_str()) * (24 * 60 * 60);
+	errno = 0;
+	long long str_to_longlong = strtoll(date_span.c_str(), NULL, 0);
+	if (errno == ERANGE) {
+	    switch (str_to_longlong) {
+		case LLONG_MIN:
+		    throw (string("Date span underflow:"));
+		case LLONG_MAX:
+		    throw (string("Date span overflow:"));
+	    }
```
ojwb: If somebody asks for all documents which are dated in a range of length `LLONG_MAX + 1` days then treating the range as `LLONG_MAX` days will return all such documents, as we can't represent dates that far away anyway.  So the clamping which `strtoll()` does is actually fine for our purposes.  The same is true for `LLONG_MIN`.

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,19 @@ date_range_filter(const string & date_start, const string & date_end,
...
+	}
+	if (time_t(str_to_longlong) > (numeric_limits<time_t>::max() / (24 * 60 * 60)))
```
ojwb: This check is incorrect for 32-bit time_t - e.g. when `str_to_longlong = 4294967296` it evaluates to `0 > 24855`.

xapian-applications/omega/date.cc
```diff
@@ -144,7 +147,19 @@ date_range_filter(const string & date_start, const string & date_end,
...
+	    throw (string("Date span overflow:"));
```
ojwb: Again, clamp rather than giving an error.

xapian-applications/omega/date.cc
ojwb: I think you can actually just remove the `time_t` conversion on the LHS.

PriyankBhatt:
Hello Olly,
I also think so that removing time_t in if condition would solve the issue you listed. And I didn't get you about the clamping the value in if condition . Can you please suggest some alternative ?
Thank You.

ojwb:
That `if` checks if the multiplication that follows would overflow.  If it would, we can just set `span = numeric_limits<time_t>::max();` instead of doing the multiplication (we can't represent dates beyond that max anyway, so a date span that's larger than that can be safely treated as the max).

We should also take care when adding or subtracting `span` later in this function - those can overflow or underflow and should be clamped to the max or min if they would.

(It turns out you've perhaps actually picked one of the more complex cases to start with as the current code blithely ignores overflow issues in several places, but you've got to the point where it's worth keeping on going I think.)

PriyankBhatt:
Hello Olly,
As you suggested i have capped off the value.
'    if (!date_span.empty()) {
	errno = 0;
	time_t secs;
	long long str_to_longlong = strtoll(date_span.c_str(), NULL, 0);
	if (errno == ERANGE) {
	    switch (str_to_longlong) {
		case LLONG_MAX:
		    secs = numeric_limits<time_t>::max();
	    }
	} else if (str_to_longlong > (numeric_limits<time_t>::max() / (24 * 60 * 60))) {
	    secs = numeric_limits<time_t>::max();
	} else {
	    secs = time_t(str_to_longlong) * (24 * 60 * 60);
	}'
but the problem is that now during building, make throws an exception saying that sec can be uninitialized.
And another thing i have omitted checking for LLONG_MIN . Do you think thats correct ? because in my opnion datespan shouldn't be negative.
And can you tell me the slot when are you at irc ? that would be helpful in clearing the doubts faster.

ojwb:
The compiler is correct, you don't initilalise secs when `errno == ERANGE` and `str_to_longlong != LLONG_MAX`.

But as I said above, you don't need to check for `ERANGE` - the result of `strtoll()` will be what we want even in the face of overflow or underflow.  You can just drop the whole first part of that multi-stage `if`.

ojwb:
The meaning of a negative date span isn't explicitly documented, but there's an obvious interpretation (it's a date span in the other direction).  I believe that's how the code will currently handle it, so people may be making use of them, and it seems better to keep them working.

I don't have fixed IRC hours.  If you have questions, just ask them in the channel and other people can probably help too.

PriyankBhatt:
Hello Olly,
1.Removed switch statement.
2.Added checks in latter cases in if !date.end.empty ,!date.start.empty and else case.
3.Removed extra header files which weren't necessary.

ojwb:
It would really be better to check for `mktime()` failing, but I'd suggest doing that as a separate change, as that check is already missing and isn't really related to the atoi()->strtol() change.

xapian-applications/omega/date.cc
```diff
@@ -154,7 +163,14 @@ date_range_filter(const string & date_start, const string & date_end,
...
 	    t.tm_isdst = -1;
-	    time_t then = mktime(&t) - secs;
+	    time_t then;
+	    // if secs is equal to time_t min value,
+	    // then = mktime + time_t min val hence which would do addition
+	    // so checking for that overflow.
+	    if (secs < (mktime(&t) - numeric_limits<time_t>::max()))
+		then = numeric_limits<time_t>::max();
+	    else
+		then = mktime(&t) - secs;
```
ojwb: Calling `mktime()` twice is probably better avoided, as it isn't a completely trivial calculation it's doing, and the compiler is unlikely to be able to tell that it can actually safely eliminate the second call.

I'm not convinced the underflow check here is correct - shouldn't we be comparing against the `min()` not `max()` (at least if secs is positive)?

Underflow happens when (comparison done in the full integer range):

```
mktime(&t) - secs < MIN
```

Which is the same as this condition, only this check won't underflow or overflow if `secs >= 0`):

```
mktime(&t) < MIN + secs
```

PriyankBhatt:
Hello Olly,
1. Checked for underflow and overflow in each case.
2. Also store output of mktime in variable.

PriyankBhatt:
Hello Olly,
1.  Replaced all numeric_limits calls to variables
2.  Added extra check for case when given_time - secs and secs if time_minimum in which secs would overflow while converting time_min to time_max

ojwb:

ojwb:
There's been no response for over two years, so I think we have to assume this patch is abandoned. 

We don't have the required licensing agreement as far as I can see (https://trac.xapian.org/browser/git/xapian-core/HACKING#L1376) so this patch is not even something that somebody else can pick up and take forward.

PR #67, opened 2015-02-19
Mantissa overflow check fix

Kronuz:
In the `double(**p) > dbl_max_mantissa` part, `*p` is (or could be) already beyond `end`. The mantissa (v) should is what should be checked against `dbl_max_mantissa` (so `ldexp` works properly without overflow).


ojwb:
I've fixed the over-read in 4e552fe, and enhanced unittest's serialisedouble to provide a regression test.

ojwb:
Added the missing cast in 845d1b6, and opened http://trac.xapian.org/ticket/673 for considering using IEEE-754 format on the wire, so closing.

PR #68, opened 2015-02-20
dbcheck coverage

karthikiyer:


ojwb:
So from IRC discussion of the first commit, the last bit isn't needed as the bug it is addressing is fixed in current master (after you forked); the change to know about `.glass` extensions is definitely useful so I've added some broader handling of that in 9360754 (also using the extension of the table to decide what backend it is for); and the other changes are the remains of debugging and development work.

ojwb:

ojwb:
This PR seems to have been abandoned so closing.

I've pushed a couple of commits which cover errors from finding the DB/table and for checking a valid DB and table (d6617dc7e6ba79aa8876dcc87b98576c3f52edb9 and 742b83a64ed5263fabfe29a502320bbb81e1f6be).

PR #69, opened 2015-03-01
SimpleSearch.java example: Fix build and update example.

hyori:
SimpleSearch.java would not build against the current SWIG-generated Java bindings, updated and tried to make the iterator feel a bit more natural.

Thanks for a cool library :-)


ojwb:
Thanks.  Are you OK to license the contribution as described here:
http://trac.xapian.org/browser/trunk/xapian-core/HACKING#L1364

hyori:
Yes, I am OK with it.

ojwb:
Great, merged.  Thanks again.

PR #70, opened 2015-04-01
Added missing msg_freqs remote command

Kronuz:
`MSG_FREQS` command was missing from the remote server, it's listed as a `message_type` in common/remoteprotocol.h but it's skipped in list of commands in the remote server's `run()` thus making both `MSG_FREQS` and `MSG_UNIQUETERMS` fail (o return unexpected mixed results).

### DIFF
--- a/xapian-core/net/remoteserver.cc
+++ b/xapian-core/net/remoteserver.cc
@@ -184,6 +184,7 @@ RemoteServer::run()
 		0, // MSG_GETMSET - used during a conversation.
 		0, // MSG_SHUTDOWN - handled by get_message().
 		&RemoteServer::msg_openmetadatakeylist,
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for catching this - I've merged the change.

It's bad that the testsuite didn't catch this, so I'm working on adding better test coverage for `get_unique_terms()`.

It's harder to regression test the other message, as it's only used during a match where there are both local and remote databases, which the test harness doesn't directly support.  Adding harness support for such a combination is already on my todo list though.

ojwb:
OK, I've added better test coverage for `get_unique_terms()` (which fails before your patch) and which uncovered a couple of other bugs in it - committed in 4807062.

PR #71, opened 2015-04-24
Reset value manager cursor on reopen

Kronuz:
Fixes issue with value manager cursor not being reset during reopen in chert


Kronuz:
This fixes http://trac.xapian.org/ticket/675 for chert

Kronuz:
I also moved `value_manager.reset();` below tables opening in `ChertDatabase::open_tables()`.

ojwb:
Addressed by 1900eca6d325e9185b9f9f5c75a4a4a545616d7f.

The issue turns out to actually be a bit more complex, but the reproducer and this patch were useful in solving this one, so thanks very much for both of them.

PR #72, opened 2015-06-28
FreeBSD needs unistd.h for unlink()

Kronuz:


ojwb:
Thanks, I've committed a slightly different fix - instead including `safeunistd.h` which provides various portability workarounds - the important one here is that it exists when building under WIN32 with MSVC).

ojwb:
Backported to 1.2 branch.

PR #73, opened 2015-08-25
Perl matchspy

omega:
This is a work in progress pull request

- [x] Can use ValueCountMatchSpy
- [ ] Can create own perl-based MatchSpy implementations

I tried to emulate the PerlStopper way of doing things, but I can't even figure out if that works properly, and the tests seems poor in that area.


ojwb:
Do you agree to license the patch as per http://trac.xapian.org/browser/trunk/xapian-core/HACKING#L1364 ?

omega:
Yes, and I have permission from my employer to do so

ojwb:
Great, thanks.

The history was rather tangled and didn't seem very informative, so I've squashed the changes into a single commit: d320cb5d

(There currently seems to be some networking issue preventing me pushing to the Xapian git repo right now, but that commit should appear later today).

This should be in the next release, probably 1.2.22.0 or maybe 1.2.21.1.

PR #74, opened 2015-09-30
Database encryption

dkopecek:
Hi, the set of patches in this pull request implement database encryption for the glass and chert backends. I've used the Botan library as a provider of crypto routines but there's no strong reasoning behind this choice (it seemed to have the best documentation among the C++ crypto libraries).

The encryption/decryption is implemented at block I/O level in `common/io_utils.cc`.

## Changes
 * New configure option to enable/disable encryption support: `--enable-encryption`
 * Public API functions for creating databases (read-only and read-write) were extended with two std::string parameters with default values: `encryption_key` (`NULL`) and `encryption_cipher` (`"Serpent/XTS"`)
 * Added simple test to check that the encryption works: `tests/api_encryption.cc`.

## Some known issues
 * Public API in xapian/database.h and xapian/dbfactory.h was modified/extended with two additional parameters with default values, which preserves source level API compatibility but **doesn't preserve binary API compatibility**
 * **Some tests are broken**, saying: `DatabaseLockError: Unable to get write lock on ...: already locked`. I'll probably need help with resolving this issue.
 * Not tested with remote backends yet

## Example usage (from tests/api_encryption.cc)
```c++
const std::string test_encryption_key = "pQHQNZrTghIryAukoxkavLQupfDqZTKnKqFceofiKlNJaCkBsCynRWxQnJlVjGBF";
const std::string test_encryption_cipher = "Serpent/XTS";
 
Xapian::WritableDatabase db(".chert_encrypted",
    Xapian::DB_CREATE_OR_OVERWRITE|Xapian::DB_BACKEND_CHERT,
    8192,
    &test_encryption_key, test_encryption_cipher);

// do stuff with db
```

### DIFF
--- a/xapian-core/backends/chert/chert_database.cc
+++ b/xapian-core/backends/chert/chert_database.cc
@@ -103,7 +103,9 @@ const int MAX_OPEN_RETRIES = 100;
  * to the tables.
  */
 ChertDatabase::ChertDatabase(const string &chert_dir, int flags,
-			     unsigned int block_size)
+			     unsigned int block_size,
+           const std::string * encryption_key,
+           const std::string& encryption_cipher)
 	: db_dir(chert_dir),
 	  readonly(flags == Xapian::DB_READONLY_),
 	  version_file(db_dir),
@@ -118,6 +120,8 @@ ChertDatabase::ChertDatabase(const string &chert_dir, int flags,
 	  max_changesets(0)
 {
     LOGCALL_CTOR(DB, "ChertDatabase", chert_dir | flags | block_size);
+
+    set_encryption(encryption_cipher, encryption_key);
 
     if (readonly) {
 	open_tables_consistent();
@@ -375,6 +379,17 @@ ChertDatabase::get_changeset_revisions(const string & path,
     if (!unpack_uint(&start, end, endrev))
 	throw Xapian::DatabaseError("Couldn't read a valid end revision for "
 				    "changeset at " + path);
+}
+
+void
+ChertDatabase::set_encryption(const std::string& cipher, const std::string * key)
+{
+  postlist_table.set_encryption(cipher, key);
+  position_table.set_encryption(cipher, key);
+  termlist_table.set_encryption(cipher, key);
+  synonym_table.set_encryption(cipher, key);
+  spelling_table.set_encryption(cipher, key);
+  record_table.set_encryption(cipher, key);
 }
 
 void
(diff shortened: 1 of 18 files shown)


ojwb:
Thanks for the patches, and sorry for taking a while to get to this.

First a small bit of admin - are you OK with the patch submission requirements?

Some initial comments on the patch itself:

Please follow the coding standards for the Xapian code base (as documented in `HACKING`).  The indentation is off is a number of places for a start.

What are the situations in which you see this being useful?  I think that's something important to think about - users need to understand what an encryption feature does and doesn't give them, and so we need to ensure that the implementation and user expectations are aligned.  My this thought was how this compares to using an encrypted filing system, and I suspect that would be an FAQ.

It isn't useful to update deprecated methods - these are only still in the API so that existing user code doesn't simply stop compiling - it also provides a grace period where code can work with the old and new versions of the API without needing a lot of conditional compilation.  Existing user code doesn't use the encryption feature, and would need conditional compilation to work with older versions (since they don't have encryption features).

Databases which are encrypted ought to be flagged as such - otherwise you can't reliably tell corruption from encryption.  It would also be useful to be able to know if the encryption key provided is correct, so you can tell corruption from not having the correct key (and when writing, specifying an incorrect key could theoretically corrupt the database if the data read happens to look enough like actual blocks).  Not sure exactly how best to achieve that without any risk of weakening the encryption.

I also don't think this is appropriate for chert - it's a stable backend and has been for a few years now.  In 1.4.x, glass will be the default, and chert deprecated.

With glass, freelist blocks probably ought not be encrypted as they are overwritten in place - when encrypted that means an incomplete write (e.g. interrupted by a power failure) could corrupt your database.  It's recoverable at least (by regenerating the freelist), but the freelist is just a list of block numbers and revision numbers, so rather pointless to encrypt - that's not sensitive information.

Baking a default for the encryption cipher into the API seems unwise.  It would be very hard to change the default later (since changing the cipher changes the meaning of the encryption key), but a weakness in the default might be found, or else it'll just become insecure due to increases in available computing power.  And then we're essentially recommending an insecure cipher to people.

I also wonder if the encryption key should be passed as a more basic pointer type - `const void*` or `const char *` probably.  That way the caller is able to easily use `mlock()` to ensure the encryption key doesn't get swapped out and a copy end up on disk, should they want to defend against that.  With `std::string`, I'm not sure how you'd do that - a custom allocator gives you the control, but would create a different type AFAICS.

ojwb:
Hmm, I seem to have managed to close all open PRs 28 days ago, which wasn't intended.

Reopening, but this is blocked waiting on a response from the submitter, and if we don't hear back it's not useful to keep open indefinitely.

dkopecek:
@ojwb Hi, sorry for the delay. I don't have much time to spent on this at the moment. Anyway, I plan to address the things you mentioned and create a new pull request. Thanks for the review!

ojwb:
I think it makes sense to close this pull request, as the submitter said they'd rework the patch and create a new PR.

(And nobody else can usefully fix it up as things stand because the submitter didn't agree to the patch submission requirements...)

PR #75, opened 2015-10-19
issue689 (python3-bindings): rename .pyo to .opt-1.pyc files for PEP 0488

ususdei:


ojwb:
Thanks for the patch.  Looks good - I think the indentation is off (or perhaps github's patch viewer is) but that's easy to fix when I apply it for a small change like this.

One small bit of admin first - are you OK with the patch submission requirements?

ususdei:
Oh, sorry. I guess, I continued to use 4 spaces instead of a tab in the autoconf file.

Sure, the code is mine and GPL/MIT is fine for me. Feel free to use it.

PR #76, opened 2015-10-19
Issue690 (python3-bindings tests): check exceptions using regex

ususdei:


ojwb:
Same question as for the other PR, are you OK with the patch submission requirements?

ususdei:
Same as other PR, I'm OK with it.

PR #77, opened 2015-11-06
Added skip action to log while ignoring the mimetype.

samuelharden:
I have not added SKIP_VERBOSE_FLAG to the skip command in this action. Since that would make it similar to ignore psuedo-mimetype in most cases.

Not sure if there is any doc which need to be updated with skip feature.

I was thinking to explain it in help, But that might clutter the help section, i guess that might have been the reason to not add ignore psuedo-mimetype as-well. Shall we add it to help ?

### DIFF
--- a/xapian-applications/omega/omindex.cc
+++ b/xapian-applications/omega/omindex.cc
@@ -106,8 +106,7 @@ index_file(const string &file, const string &url, DirectoryIterator & d,
 	string m = "skipping extension '";
 	m += ext;
 	m += "'";
-	skip(urlterm, file.substr(root.size()), m, d.get_size(), d.get_mtime(),
-	    SKIP_VERBOSE_ONLY);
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks, merged.

Conventionally we put a space between the comment character(s) and the text in Xapian codebase (a common convention, as it's more readable).  You also misspelled `quietly` - I fixed both in 9c9bacf9ec712002c0d3ca365d1a17dafa954b00.

> Not sure if there is any doc which need to be updated with skip feature.

`ignore` is documented in some generated documentation - I've added `skip` alongside it in 3dbb5fbf95a0270fb450f7535e8367123c104d6e.

> I was thinking to explain it in help, But that might clutter the help section, i guess that might have been the reason to not add ignore psuedo-mimetype as-well. Shall we add it to help ?

You mean `omindex --help` (and the manpage)?  I think you're right that it would add more clutter than utility to try to squeeze it in there.

PR #78, opened 2016-01-13
Improving the arabic stopwords list

assem-ch:
- sorting the list to make it easier for checking
- eliminate duplicates after normalizing shedda symbol
- eliminate some words that are rare and can come not as stop word having a full meaning.
- romove the todos list from the header comment
- add new stop words


ojwb:
Thanks!

PR #79, opened 2016-02-15
Add $contains{STRING1,STRING2}

ayushgp:


xapian-applications/omega/query.cc
```diff
@@ -1307,6 +1309,15 @@ eval(const string &fmt, const vector<string> &param)
...
+        case CMD_contains: {
+        size_t pos = args[1].find(args[0]);
+        if(pos!=string::npos){
+            value = str(pos);
+        } else {
+            value = "";
+        }
+        break;
+        }
 	    case CMD_csv:
```
ojwb: Generally looks good to me, but there are some minor issues:

The indentation is off here (and above).  I suspect you have tabs set to display with width 4 - you need to have them display as width 8 for Xapian code.

You also should put whitespace around operators, etc for readability.  Xapian coding conventions for the `if` would make it:

```
        if (pos != string::npos) {
```

And `value` will be empty by default, so there's no need for the `else` branch here.

ojwb:
Also, are you OK with the patch licensing requirements described here:

ayushgp:
I'll keep the style in mind next time I submit a patch. And yeah I'm OK
with the licensing requirements in the given document.

jaylett:
@ayushgp can you look at `git rebase`, in particular interactive rebase and the `--autosquash` option? You should be able to use that to collapse my "fixup" commit with the one it references; then you can force push your local branch up to github and this PR will update. You can incorporate @ojwb's point about the `else` branch at the same time.

Olly's just pointed out to me that we have a test rig for omega. Although a lot of it isn't tested, we'd like to ensure that all new features are, so it'd be great to get a couple of tests for `$contains{}`, probably based on the documentation examples. (It'd be good to check that variable expansion works as well.) The code to look at is `omegatest`, which will probably need a new function writing to test Omegascript commands.

ayushgp:
@jaylett I've added the tests and have squashed all commits.

ayushgp:
@jaylett Do I need to do anything else on this?

jaylett:
@ayushgp sorry, I've had to look at some other things. If you can rewrite your commit to match the (accurate) PR title, and maybe rebase onto latest master, I'll try to get that merged when I have a minute.

ayushgp:
@jaylett I have rebased it on current master, squashed the commits and changed the git message to be like the title of this PR.

jaylett:
@ayushgp you hadn't actually rebased onto _current_ master; my guess is you did `git rebase master` rather than `git fetch upstream && git rebase upstream/master`. (You can tell because on the PR page there's a "This branch has conflicts that must be resolved" message.)

I've sorted this out at my end, although I think olly will have to merge it because of <reasons>. Closing in favour of #80.

ayushgp:
I actually did a git fetch origin and then git rebase origin/master. I
didn't realize that it'll fetch from my repository. Thanks for helping out
on this PR though!

PR #80, opened 2016-03-06
Add $contains{STRING1,STRING2}

jaylett:


xapian-applications/omega/query.cc
```diff
@@ -1307,6 +1309,15 @@ eval(const string &fmt, const vector<string> &param)
 		value = str(collapsed);
 		break;
 	    }
+            case CMD_contains: {
+                size_t pos = args[1].find(args[0]);
+                if (pos != string::npos) {
+                    value = str(pos);
+                } else {
+                    value = "";
+                }
```
ojwb: @ayushgp As I mentioned in feedback on the original PR, `value` starts empty, so this `else` is unnecessary.  I'll merge and fix this in a follow-up commit.

ojwb:
As it was a single commit branch, I just cherry picked it onto master as 49144e2df18fa6ec4ed08eb413da3b13f82ac313.

Thanks @ayushgp and @jaylett for your work on this.

PR #81, opened 2016-03-08
Add max-wdf normalization scheme to TfIdfWeight

Nishad94:
- Added code to track the max-wdf for each document. This was achieved by adding functionality to the PostList and Database API to compute max wdf for a document.
- Added an additional test in api_weights to verify results.
- The code builds successfully.
- Code passes all tests.

### DIFF
--- a/xapian-core/api/leafpostlist.cc
+++ b/xapian-core/api/leafpostlist.cc
@@ -60,6 +60,7 @@ LeafPostList::set_termweight(const Xapian::Weight * weight_)
     weight = weight_;
     need_doclength = weight->get_sumpart_needs_doclength_();
     need_unique_terms = weight->get_sumpart_needs_uniqueterms_();
+    need_wdfdocmax = weight->get_sumpart_needs_wdfdocmax_();
 }
 
 double
@@ -72,14 +73,16 @@ double
 LeafPostList::get_weight() const
 {
     if (!weight) return 0;
-    Xapian::termcount doclen = 0, unique_terms = 0;
+    Xapian::termcount doclen = 0, unique_terms = 0, wdfdocmax = 0;
     // Fetching the document length and number of unique terms is work we can
     // avoid if the weighting scheme doesn't use them.
     if (need_doclength)
 	doclen = get_doclength();
     if (need_unique_terms)
 	unique_terms = get_unique_terms();
-    double sumpart = weight->get_sumpart(get_wdf(), doclen, unique_terms);
+    if (need_wdfdocmax)
+	wdfdocmax = get_wdfdocmax();
+    double sumpart = weight->get_sumpart(get_wdf(), doclen, unique_terms, wdfdocmax);
     AssertRel(sumpart, <=, weight->get_maxpart());
     return sumpart;
 }
--- a/xapian-core/backends/database.h
+++ b/xapian-core/backends/database.h
@@ -149,6 +149,13 @@ class Database::Internal : public Xapian::Internal::intrusive_base {
 	 */
 	virtual	Xapian::termcount get_unique_terms(Xapian::docid did) const = 0;
 
+	/** Get the max wdf of terms in document.
+	 *
+	 *  @param did  The document id of the document whose max-wdf is
+	 *		being requested.
+	 */
+	virtual	Xapian::termcount get_wdfdocmax(Xapian::docid did) const = 0;
+
 	/** Returns frequencies for a term.
 	 *
 	 *  @param term		The term to get frequencies for
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -62,7 +62,9 @@ class XAPIAN_VISIBILITY_DEFAULT Weight {
 	/// Sum of wdf over the whole collection for the current term.
 	COLLECTION_FREQ = 4096,
 	/// Number of unique terms in the current document.
-	UNIQUE_TERMS = 8192
+	UNIQUE_TERMS = 8192,
+	/// Maximum wdf in the current document.
+	WDF_DOC_MAX = 16384
     } stat_flags;
 
     /** Tell Xapian that your subclass will want a particular statistic.
@@ -221,7 +223,8 @@ class XAPIAN_VISIBILITY_DEFAULT Weight {
      */
     virtual double get_sumpart(Xapian::termcount wdf,
 			       Xapian::termcount doclen,
-			       Xapian::termcount uniqterms) const = 0;
+			       Xapian::termcount uniqterms,
+ 			       Xapian::termcount wdfdocmax) const = 0;
 
     /** Return an upper bound on what get_sumpart() can return for any document.
      *
@@ -313,6 +316,16 @@ class XAPIAN_VISIBILITY_DEFAULT Weight {
 	return stats_needed & UNIQUE_TERMS;
     }
 
+    /** @private @internal Return true if the maximum WDF for a document is needed.
+     *
+     *  If this method returns true, then the max WDF will be
+     *  fetched and passed to @a get_sumpart().  Otherwise 0 may be passed for
+     *  the max wdf.
+     */
+    bool get_sumpart_needs_wdfdocmax_() const {
+	return stats_needed & WDF_DOC_MAX;
+    }
+
   protected:
     /** Don't allow copying.
      *
@@ -392,7 +405,8 @@ class XAPIAN_VISIBILITY_DEFAULT BoolWeight : public Weight {
 
     double get_sumpart(Xapian::termcount wdf,
 		       Xapian::termcount doclen,
-		       Xapian::termcount uniqterms) const;
+		       Xapian::termcount uniqterms,
+ 		       Xapian::termcount wdfdocmax) const;
     double get_maxpart() const;
 
     double get_sumextra(Xapian::termcount doclen,
@@ -415,7 +429,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     /* When additional normalizations are implemented in the future, the additional statistics for them
        should be accessed by these functions. */
-    double get_wdfn(Xapian::termcount wdf, char c) const;
+    double get_wdfn(Xapian::termcount wdf, char c, Xapian::termcount wdfdocmax = 0) const;
     double get_idfn(Xapian::doccount termfreq, char c) const;
     double get_wtn(double wt, char c) const;
 
@@ -435,8 +449,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     @li 'b': Boolean    wdfn=1 if term in document else wdfn=0
      *     @li 's': Square     wdfn=wdf*wdf
      *     @li 'l': Logarithmic wdfn=1+log<sub>e</sub>(wdf)
+     *     @li 'm': Max_wdf     wdfn=wdf/max_wdf
      *
-     *     The Max-wdf and Augmented Max wdf normalizations haven't yet been
+     *     The Augmented Max_wdf normalization hasn't yet been
      *     implemented.
      *
      * @li The second character indicates the normalization for the idf.  The
@@ -475,7 +490,8 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     double get_sumpart(Xapian::termcount wdf,
 		       Xapian::termcount doclen,
-		       Xapian::termcount uniqterm) const;
+		       Xapian::termcount uniqterm,
+ 		       Xapian::termcount wdfdocmax) const;
     double get_maxpart() const;
 
     double get_sumextra(Xapian::termcount doclen,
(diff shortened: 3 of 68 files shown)


ojwb:
Thanks for the PR.

One bit of admin first - are you OK with the patch licensing requirements described here:

Generally the changes look good, but there's a stylistic issue which needs fixing - the indentation is off in several places, which can be seen in the github diff viewer.  My guess is your editor is set to display a tab character with a different width than 8.  When working with Xapian code, the tab character display width should be 8, with each block indented 4 from the parent block, or else things will end up misidented.

xapian-core/api/leafpostlist.cc
```diff
@@ -72,14 +73,16 @@ double
...
+    Xapian::termcount doclen = 0, unique_terms = 0, wdfdocmax = 0;
     // Fetching the document length and number of unique terms is work we can
     // avoid if the weighting scheme doesn't use them.
     if (need_doclength)
 	doclen = get_doclength();
     if (need_unique_terms)
 	unique_terms = get_unique_terms();
-    double sumpart = weight->get_sumpart(get_wdf(), doclen, unique_terms);
+    if (need_wdfdocmax)
+    wdfdocmax = get_wdfdocmax();
```
ojwb: Here's an example of mis-identation - the line after the `if` should be indented more.

xapian-core/backends/database.h
```diff
@@ -149,6 +149,13 @@ class Database::Internal : public Xapian::Internal::intrusive_base {
 	 */
 	virtual	Xapian::termcount get_unique_terms(Xapian::docid did) const = 0;
 
+	/** Get the max wdf of terms in document.
+	 *
+	 *  @param did  The document id of the document whose number of terms is
+	 *		being requested.
```
ojwb: `number of terms` here is wrong (presumably due to cut+paste without changing enough).

xapian-core/matcher/synonympostlist.h
```diff
@@ -58,6 +58,12 @@ class SynonymPostList : public PostList {
      */
     bool want_unique_terms;
 
+    /** Flag indicating whether the weighting object needs the max wdf of
+     *  terms.
+     *  FIXME : Not implemented
```
ojwb: This really needs to be hooked up - without that either we'll end up calculating this statistic all the time, which is unnecessary overhead when it isn't wanted, or else we'll never actually calculate it, in which case your new code isn't actually used!

xapian-core/tests/api_weight.cc
```diff
@@ -586,6 +586,15 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
     mset_expect_order(mset, 2, 4);
     TEST_EQUAL_DOUBLE(mset[0].get_weight(), 8.0 * log(6.0 / 2));
 
+    Xapian::MSet mset3;
+    // Check for "mrn" when termfreq != N
+    enquire.set_query(query);
+    enquire.set_weighting_scheme(Xapian::TfIdfWeight("mtn"));
```
ojwb: Comment says `"mrn"` but code says `"mtn"`.

xapian-core/tests/api_weight.cc
```diff
@@ -766,6 +775,7 @@ class CheckStatsWeight : public Xapian::Weight {
 	TEST_REL(uniqueterms,>=,1);
 	TEST_REL(uniqueterms,<=,doclen);
 	TEST_REL(wdf,<=,wdf_upper);
+    	(void)wdfdocmax;
```
ojwb: The idea of this class is that it does sanity checks of the various statistics.  For `wdfdocmax` it's defined as the largest wdf in the document, so we can check that it's at least as large as the wdf we get passed.

xapian-core/include/xapian/weight.h
```diff
@@ -435,6 +449,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     @li 'b': Boolean    wdfn=1 if term in document else wdfn=0
      *     @li 's': Square     wdfn=wdf*wdf
      *     @li 'l': Logarithmic wdfn=1+log<sub>e</sub>(wdf)
+     *     @li 'm': Max-wdf     wdfn=wdf/max-wdf
```
ojwb: `max-wdf` in the formula could be mis-read as subtraction - `max_wdf` would avoid that.

xapian-core/matcher/synonympostlist.h
Nishad94: I have set it to false, in the constructor of SynonymPostList. Is that right? I am not very clear about what a SynonymPostList is and how it will affect the functioning of the max_wdf statistic.

Nishad94:
Thanks for reviewing the PR, I have fixed some issues you pointed out. I am okay with the patch licensing requirements. Do I need to add anything to the code in this regard?

jaylett:
@Nishad94 — good work! :-)

I've renamed this PR to fit within 50 characters (this is good practice because it makes git logs read better). I've slightly changed it from talking about 'max-tf' to 'max-wdf', which I think is accurate. If not, please change that part back with my apologies!

I'd like to see you 'squash' your changes into fewer overall commits. A lot of your commits are 'fixups', just correcting mistakes in earlier commits. We don't want to include them in our final history; we'd like a straightforward git history that just shows 'positive' changes. It may be that there's a couple of pieces of functionality here: tracking the `max_wdf` and then using that to implement the new normalization for `TfIdfWeight`. Let me know if you need any help or pointers with this.

You'll probably find it helpful to read through [our developer notes](https://xapian-developer-guide.readthedocs.org/en/latest/contributing/new-contributors.html#do-some-work), particularly the section on how to 'make commits out of your changes', which has some suggestions on how to write good commit messages, and a link or two to read up on some of the details of using git.

xapian-core/include/xapian/weight.h
```diff
@@ -435,8 +449,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     @li 'b': Boolean    wdfn=1 if term in document else wdfn=0
      *     @li 's': Square     wdfn=wdf*wdf
      *     @li 'l': Logarithmic wdfn=1+log<sub>e</sub>(wdf)
+     *     @li 'm': Max_wdf     wdfn=wdf/max_wdf
      *
-     *     The Max-wdf and Augmented Max wdf normalizations haven't yet been
+     *     The Max_wdf and Augmented Max_wdf normalizations haven't yet been
```
jaylett: Doesn't this PR implement `Max_wdf`? In which case it's only the augmented version that needs mentioning here now.

xapian-core/matcher/synonympostlist.h
ojwb: A `SynonymPostList` has the same logical effect as a `OR` - i.e. it matches when any of its subqueries match.  But the weight it gives is as if it was a single term, and hence it's the `SynonymPostList` which gets the weight object, rather than the terms at the leaf of the query.

For example, you might want to treat `furze` and `gorse` as synonymous, so if the user searches for either term it will match documents indexed by either (or both), and a document which mentions `furze` 3 times would get the same weight as one which mentions `furze` twice, and `gorse` once, or `furze` one and `gorse` twice, etc (I think we may approximate slightly in some cases for efficiency, but the aim is that it works like this).

Calculating things like wdf, number of unique terms, or max_wdf for a `SynonymPostList` is potentially quite costly (since it requires fetching data from all its children) so we only do that for stats which are actually going to be used.

Looking at the code, it looks like `want_unique_terms` is currently always false here - looks like that was never hooked up properly either.  I think that just setting in `SynonymPostList::set_weight()`:

{{{
    want_unique_terms = wt->get_sumpart_needs_uniqueterms_();
}}}

That change compiles and passes the current testsuite, but we could do with a regression test to ensure that uniqueterms gets calculated for a `SynonymPostList` to ensure it's actually fixed by that change and that it stays fixed in the future.

Anyway, maxwdf should be handled similarly.

xapian-core/include/xapian/weight.h
ojwb: Yes, it does.

Nishad94:
@jaylett I have squashed the commits..

jaylett:
@Nishad94 thanks! I don't think you've yet addressed all the comments yet (although some of them are now against outdated commits, which does make them hard to see in github's PR view). It's probably easiest to address them using fixup commits (`git commit --fixup <original commit>`), and then once everything's in order to squash them one final time.

Nishad94:
@jaylett @ojwb I have implemented the changes suggested in the comments and squashed all of them into a single commit. I fumbled a bit, as I directly ran 'git rebase' without showing the fixup commits in the PR. Will take better care next time to do that. Thanks!

xapian-core/include/xapian/weight.h
```diff
@@ -415,7 +429,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     /* When additional normalizations are implemented in the future, the additional statistics for them
        should be accessed by these functions. */
-    double get_wdfn(Xapian::termcount wdf, char c) const;
+    double get_wdfn(Xapian::termcount wdf, char c, Xapian::termcount wdfdocmax = 0) const;
```
ojwb: A minor point, but the default parameter value here doesn't seem to be used anywhere, so I'd just drop it.

ojwb:
Thanks for the updates - it's looking a lot better.

However, `xapian-check-patch` reports some whitespace issues still:

```
olly@gemse:~/git/xapian-git$ git show | xapian-maintainer-tools/xapian-check-patch 
xapian-core/backends/inmemory/inmemory_database.cc:575: added/changed line uses spaces for indentation rather than tab: +                string(" not found"));
xapian-core/backends/inmemory/inmemory_database.h:159: added/changed line has space before tab: +	Xapian::termcount 	get_wdfdocmax() const; 	// Max wdf of terms in current document
xapian-core/backends/inmemory/inmemory_database.h:190: added/changed line has space before tab: +	Xapian::termcount 	get_wdfdocmax() const; 	// Max wdf of terms in current document
xapian-core/backends/remote/remote-database.cc:569: FIXME added: +    // FIXME : Needs to be implemented for this
xapian-core/include/xapian/weight.h:227: added/changed line has space before tab: + 			       Xapian::termcount wdfdocmax) const = 0;
xapian-core/include/xapian/weight.h:409: added/changed line has space before tab: + 		       Xapian::termcount wdfdocmax) const;
xapian-core/include/xapian/weight.h:494: added/changed line has space before tab: + 		       Xapian::termcount wdfdocmax) const;
xapian-core/matcher/synonympostlist.h:63: FIXME added: +     *  FIXME : Not implemented
xapian-core/tests/api_weight.cc:596: added/changed line has trailing whitespace: +    TEST_EQUAL_DOUBLE(mset3[0].get_weight(), (8.0/8.0) * log(6.0/2));    
xapian-core/tests/api_weight.cc:779: added/changed line has space before tab: +    	TEST_REL(wdf,<=,wdfdocmax);
Files patched:	68
Lines added:	478
Lines removed:	111
```

The `FIXME` in `xapian-core/matcher/synonympostlist.h` is for `want_wdfdocmax` which you've now implemented in response to earlier feedback.

The other `FIXME` is for remote backend support, which would be good to hook up - it should be very like the existing support for unique terms.

The new parameter to `get_sumpart()` isn't documented in `xapian/include/weight.h`.

Given we pass `doclen` and `uniqterms` to `get_sumextra()`, it would make sense to pass `wdfdocmax` to that too.

In `inmemory_database.cc` there's extra whitespace between the brackets and the expression here (Xapian coding style is `if (a > b)` not `if ( a > b )`):

```
        if ( i->wdf > max_wdf ) max_wdf = i->wdf;
```

The new method body in `extraweightpostlist.h` is misindented.

New code in `tfidfweight.cc` is misindented (constuctor and `get_wdfn()`).

And in `mergepostlist.cc` there are two blank lines before the new method - we conventionally separate methods with a single blank line.

jaylett:
@Nishad94 have you had a chance to look at this further? (Don't worry if you haven't because of other pressures!)

Nishad94:
Hey James, I have been very busy with my school project since the last few
days. So it will be a while till I can get back to this..

jaylett:
@Nishad94 that's fine, I just wanted to keep track of things!

ojwb:
Rebased onto current master, and fixed up most of the remaining issues.  Tests currently pass.  Pushed to the tfidf-maxwdf-norm branch:

https://github.com/xapian/xapian/tree/tfidf-maxwdf-norm

Because this changes the API of the Weight class (by adding a new parameter to `get_sumpart()`) this can't be merged in 1.4.x.  I think it's better to hold off merging to master while these issues from before remain:

 * Remote backend support
 * Given we pass `doclen` and `uniqterms` to `get_sumextra()`, it would make sense to pass `wdfdocmax` to that too.

I'm not 100% happy with the way we seem to need to add new parameters to `get_sumpart()` from time to time, because this means every `Weight` subclass needs updating (fixing those in the library is OK, but this also affects user-defined weighting schemes).  I wonder if there's a clean and efficient way to avoid this (it needs to be efficient as this method can get called a lot).  Or perhaps there are only so many per-doc stats, and this is only the second time we've needed to do this.

It'd also be nice to store the `wdfdocmax` stats (and the `uniqueterms` stats) for all the documents in a chunked stream (like how document lengths are stored) - the code to work them out in this patch is correct, but requires scanning the termlist of each document we need this stat for, which is quite a lot of work.

ojwb:
I've merged latest git master into the branch and fixed up a couple of whitespace irregularities that `xapian-check-patch` spotted.

The PR here has been merged (though not to git master yet) so I'm going to close this and have opened https://trac.xapian.org/ticket/744 to track further work on this.

PR #82, opened 2016-03-09
PL2-Modified

ankush1995:
Clamp the weights for each term in the query to be >= 0.

### DIFF
--- a/xapian-core/weight/pl2weight.cc
+++ b/xapian-core/weight/pl2weight.cc
@@ -70,7 +70,10 @@ PL2Weight::init(double)
     double base_change(1.0 / log(2.0));
     double mean = double(get_collection_freq()) / get_collection_size();
     P1 = mean * base_change + 0.5 * log2(2.0 * M_PI);
-    P2 = log2(mean) + base_change;
+    // If the get_collection_freq() is comparable with get_collection_size()
+    // or mean >= 1/e than P2 will be positive and can lead to a negative
+    // weight, hence take P2 >=0
+    P2 = (mean < (1/2.71828))?(log2(mean) + base_change):0;
 
     double wdfn_lower = log2(1 + cl / get_doclength_upper_bound());
     double wdfn_upper =
--- a/xapian-core/weight/pl2weight.cc
+++ b/xapian-core/weight/pl2weight.cc
@@ -71,9 +71,10 @@ PL2Weight::init(double)
     double mean = double(get_collection_freq()) / get_collection_size();
     P1 = mean * base_change + 0.5 * log2(2.0 * M_PI);
     // If the get_collection_freq() is comparable with get_collection_size()
-    // or mean >= 1/e than P2 will be positive and can lead to a negative
-    // weight, hence take P2 >=0
-    P2 = (mean < (1/2.71828))?(log2(mean) + base_change):0;
+    // than P2 will be positive and can lead to a negative weight
+    // hence take P2 < 0
+    P2 = log2(mean) + base_change;
+    P2 = (P2 < 0) ? P2 : 0;
 
     double wdfn_lower = log2(1 + cl / get_doclength_upper_bound());
     double wdfn_upper =


jaylett:
@ankush1995 can you tidy this PR up before we review it, please? It certainly still contains a large number of files you didn't intend to commit, so I can't tell right now if it's the second commit only that you want incorporated, or both. The PR should contain only the changes you're proposing.

Additionally, it would be good to write a more helpful commit message. 'pl2-modified' won't help people in the future understand what the change is, or why you've made it. Your comment on the PR:

> Clamping of the weights for each term in the query to be >= 0.

Would be a better start. It might be good to also explain in the commit message why negative weights don't make sense (even if it's just by noting that weights only make sense if they're non-negative).

We have some guidance on using git to make your commits in our [draft developer guide](https://xapian-developer-guide.readthedocs.org/en/latest/contributing/new-contributors.html#do-some-work); particularly the section "Make commits out of your changes", which also links to a good article on writing commit messages.

ankush1995:
@jaylett  , I have updated the commits and the commit message.

jaylett:
@ankush1995 thanks! This is a very, very minor point but it's good practice to write as if you're doing it now, or as an instruction. So "Clamp the weight…" rather than "Clamping the weight". (This matches the messages that git generates automatically for some things.) I'll find a way of adding some guidance on commit messages to the developer guide, because there's no way anyone is going to guess that :-)

xapian-core/weight/pl2weight.cc
```diff
@@ -70,7 +70,7 @@ PL2Weight::init(double)
     double base_change(1.0 / log(2.0));
     double mean = double(get_collection_freq()) / get_collection_size();
     P1 = mean * base_change + 0.5 * log2(2.0 * M_PI);
-    P2 = log2(mean) + base_change;
+    P2 = (mean>=(1/2.71828))?(log2(mean) + base_change):0;
```
jaylett: An inline comment here explaining what's going on might be helpful; it's not obvious just looking at it what this does.

ankush1995: I have explained it in the commit message, anyways I will update it.

jaylett: It's opaque enough that people working on it shouldn't have to git blame to find the commit message, I feel :)

ankush1995: @jaylett Done.

xapian-core/weight/pl2weight.cc
```diff
@@ -70,7 +70,10 @@ PL2Weight::init(double)
     double base_change(1.0 / log(2.0));
     double mean = double(get_collection_freq()) / get_collection_size();
     P1 = mean * base_change + 0.5 * log2(2.0 * M_PI);
-    P2 = log2(mean) + base_change;
+    // If the get_collection_freq() is comparable with get_collection_size()
+    // or mean >= 1/e than P2 will be positive and can lead to a negative
+    // weight, hence take P2 >=0
+    P2 = (mean < (1/2.71828))?(log2(mean) + base_change):0;
```
ojwb: Xapian coding style is to put spaces around operators such as `/`, `?` and `:` for readability.

The "or" in the comment doesn't match the code - the only check in the code is on the value of `mean`.  It's really unhelpful when comments don't match the code they describe - the reader is left in doubt as to whether the comment or code is correct.

But that check doesn't really match the comment (or intent) either.  `e` to the precision of a `double` is not `2.71828`, with the consequence that this actually clamps small positive values of `P2` to 0 as well.

Given this code runs once per weighted term or synonym in the query, the cost of computing the log here is not significant, so it's probably cleanest to simply compare the computed value of `P2` with zero.

ojwb:
Also, one bit of admin - are you OK with the patch licensing requirements described here:

xapian-core/weight/pl2weight.cc
ankush1995: @ojwb Yeah, small values also clamps, I have made changes in the latest commit.

ankush1995:
@ojwb I didn't understand that, how do I License my patch?

jaylett:
@ankush1995 you just need to state that you're dual licensing it in this way. Ideally an email to the mailing list, since that's archived permanently. (Sorry for the delay in clarifying this!)

ankush1995:
Okay. No problem.

ojwb:
Thanks for the PR.

Reviewing the issue it aims to address, I've ended up committing a pair of changes to address this issue in a slightly different way, which I think is better overall - instead of clamping `P2` to be positive, I've made it calculate `upper_bound` correctly when `P2 < 0` (dd6b35900358d49ec41a32578a651d4ee2bc0a6f).  And I've dropped the `lower_bound` approach (which doesn't actually work as intended in general) and instead clamp the final weight to be `>= 0` in `get_sumpart()` (1039fab8012a01568d5a0948746eb18746416de8).

PR #83, opened 2016-03-22
making ValueIterator::decref() function Inline

skj-777:
making ValueIterator::decref() inline will probably speed it up a little.

### DIFF
--- a/xapian-core/api/valueiterator.cc
+++ b/xapian-core/api/valueiterator.cc
@@ -30,7 +30,7 @@ using namespace std;
 
 namespace Xapian {
 
-void
+inline void
 ValueIterator::decref()
 {
     Assert(internal);


ojwb:
> making ValueIterator::decref() inline will probably speed it up a little.

Do you have any benchmarks to support this supposition?

The calls to `decref()` only happen on rare codepaths (mostly when we hit the end of iteration), so it seems unlikely this would make a measurable difference.

The compiler is also allowed to decide to inline even without an explicit `inline` marker if it thinks that's appropriate.

jaylett:
@skj-777 even if you don't want to pursue this particular patch, it's worth reading through a section in our [draft developer guide](https://xapian-developer-guide.readthedocs.org/en/latest/contributing/new-contributors.html#do-some-work), which includes some discussion of good commit messages (including a link to an article). Your current commit message could be more helpful and consistent for future developers. (And with single commit PRs, the PR title and message should match the commit, which is easiest to do by getting the commit message right to start off with.)

skj-777:
@Olly Betts, thank you sir for reminding me that the compiler is also allowed to decide to inline a function, I have closed this pull request and next time I will try to test it before sending a pull request.
@James Aylett, thank you sir, I will surely take care of the points mentioned in the guide in my next commit request.
Cheers,
Skj.

PR #84, opened 2016-03-26
bite-size projects-Initial patch

nirmalsinghania2008:
1)Replaced atoi,atol with strtol,strtoul
2)Reworked omega templates HTML code
3)Added opensearch 1.1 rss skeleton(Todo-Add description document)
4)Added OmegaScript command to do a regexp match
5)Added OmegaScript commands to support random numbers

### DIFF
--- a/xapian-applications/omega/ChangeLog
+++ b/xapian-applications/omega/ChangeLog
@@ -2934,7 +2934,7 @@ Thu Oct 09 10:16:05 GMT 2008  Olly Betts <olly@survex.com>
 Wed Oct 01 04:48:37 GMT 2008  Olly Betts <olly@survex.com>
 
 	* scriptindex.cc: Separate Action constructor cases to avoid
-	  pointlessly calling atoi() on an empty string.
+	  pointlessly calling strtol() on an empty string.
 
 Wed Oct 01 03:15:29 GMT 2008  Olly Betts <olly@survex.com>
 
--- a/xapian-applications/omega/ChangeLog
+++ b/xapian-applications/omega/ChangeLog
@@ -3666,7 +3666,7 @@ Fri May 04 13:30:36 BST 2007  Olly Betts <olly@survex.com>
 
 Fri May 04 04:10:23 BST 2007  Olly Betts <olly@survex.com>
 
-	* query.cc: Explicitly include stdlib.h since we use atoi().
+	* query.cc: Explicitly include stdlib.h since we use strtol().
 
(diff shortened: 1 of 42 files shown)


jaylett:
@nirmalsinghania2008 can you have a read of the section in our developer guide about commits? In addition to the notes there, a good commit encompasses a single change, and is well-described. For instance, what does the commit 'completed some mini size projects' contain? It's a fairly large commit, and seems to include some search and replace changes, some new omegascript commands (without documentation), some whitespace changes that we probably don't want, and possibly some other things that I haven't spotted. They should be split across a number of separate commits.

Specifically on one issue: you cannot simply change every instance of `atoi` to `stro[u]l` — firstly you've changed things in `Changelog` files, which should act like a register (ie you don't change entries once they're there); second, some of the instances shouldn't be changed at all. You really need to separate each one out so you can explain the reasoning (and, ideally, explain at source level if it isn't completely obvious — which with `atoi`, `stro[u]l` and error handling, amongst others, is rarely the case).

By presenting lots of little changes as one big PR, you make it much harder to get this merged. You'll have to correct all the problems (and I'm not keen on even trying to do a review until you've looked at the commit boundaries again). Smaller PRs, each doing one single thing, could be merged once they're each individually ready.

nirmalsinghania2008:
sorry James! actually i am new to open source development and github. 
I wasn't knowing all the complications. 
I will try to correct all the problems and will issue a separate PR for each change.

nirmalsinghania2008:
@jaylett "some of the instances should not be changed at all" what do you mean by this? 

strtol is always better than atoi as far as i know. correct me if I am wrong. 
one case where atoi makes sense is when we have to ignote trailing characters after a number

jaylett:
@nirmalsinghania2008 don't apologise — this is how most people learn :)

Sadly I haven't found any great resources on how to think of commits online. If you're learning git, then [this article](https://sethrobertson.github.io/GitBestPractices/) looks to have some helpful tips (although it goes into a lot of detail in some places!).

On `atoi` / `strto[u]l`, what I should have said was that in some cases the behaviour should not change (so implement `atoi` in terms of `strto[u]l`), but in some cases you need to think about unprocessed characters and have suitable error handling. The choice is very much dependent on what each bit of code is doing, so separating them out and explaining each one will make things easier to review, and better-documented for future developers :-)

(We also know from experience that there are some places where changing from `atoi` is going to be harder than anticipated. There's another PR that demonstrates how thorny some situations can get, although some aspects of that are trying to straighten out details that weren't quite right to start with.)

PR #85, opened 2016-03-29
Rework Omega templates

ivmarkp:
Change all the tags and attributes to lowercase.

### DIFF
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -6,29 +6,29 @@ ${
 $def{PREV,
 $if{$ne{$topdoc,0},<INPUT TYPE=image NAME="&lt;" ALT="&lt;"
 SRC="/icons/omega/prev.png" BORDER=0 HEIGHT=30 WIDTH=30>,
-<IMG ALT="" SRC="/icons/omega/prevoff.png" HEIGHT=30 WIDTH=30>}
+<img alt="" src="/icons/omega/prevoff.png" height=30 width=30>}
 }
 
 $def{NEXT,
-$if{$ne{$last,$msize},<INPUT TYPE=image NAME="&gt;" ALT="&gt;"
-SRC="/icons/omega/next.png" BORDER=0 HEIGHT=30 WIDTH=30>,
-<IMG ALT="" SRC="/icons/omega/nextoff.png" HEIGHT=30 WIDTH=30>}
+$if{$ne{$last,$msize},<input type=image name="&gt;" alt="&gt;"
+src="/icons/omega/next.png" border=0 height=30 width=30>,
+<img alt="" src="/icons/omega/nextoff.png" height=30 width=30>}
 }
 
-$def{P,<INPUT TYPE=image NAME="$1" VALUE="$1" SRC="/icons/omega/page-$2.png" BORDER=0$opt{a} ALT="$1">}
+$def{P,<input type=image name="$1" value="$1" src="/icons/omega/page-$2.png" border=0$opt{a} alt="$1">}
 $def{PAGE,$if{$gt{$1,9},$if{$gt{$1,99},$P{$1,$div{$1,100}}}$P{$1,$mod{$div{$1,10},10}}}$P{$1,$mod{$1,10}}}
 
-$def{S,<IMG SRC="/icons/omega/page-$2s.png"$opt{a} ALT=$1>}
+$def{S,<img src="/icons/omega/page-$2s.png"$opt{a} alt=$1>}
 $def{SPAGE,$if{$gt{$1,9},$if{$gt{$1,99},$S{$1,$div{$1,100}}}$S{$1,$mod{$div{$1,10},10}}}$S{$1,$mod{$1,10}}}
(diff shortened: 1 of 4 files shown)


xapian-applications/omega/templates/query
```diff
@@ -1,4 +1,5 @@
-$httpheader{Content-Type,text/html; charset=utf-8}<html>
+$httpheader{Content-Type,text/html; charset=utf-8}
+<html>
```
ojwb: There's a reason why there wasn't a newline before `<html>` here before.

`$httpheader{}` sets a header to send on the HTTP response and expands to nothing in the output, so your updated template will result in a blank line before the `<html>` in the output, which is a little undesirable, if only cosmetically.

ojwb: Another point it's useful to make here is what a commit changes ought to match what its commit message says it changed - the message is "Changed all the HTML tags and attributes to lowercase." and a whitespace change doesn't fit that.  Trying to track down the reason a change was made (e.g. "why does the HTML output start with a blank line?") is made much harder if commits mix in unrelated changes.

If the whitespace "fix" had been done as a separate "clean up whitespace" commit, it would be clearer what the author thought they were doing.

Also, commit messages are generally in the present tense - i.e. "Change all [,,,]" is preferred to "Changed all [...]"

ivmarkp: That newline was not intentional. I just removed it now.
And I'm sorry about the commit messages that I made. I need to be more specific and careful from the next time. I will take care of the points you made. Thanks.

jaylett:
@ivmarkp in this case I think a single commit that did all of these would be fine, because it's really all the same change. To do this you can use _interactive rebase_ to both collapse the commits and to rewrite the commit message as ojwb suggested. Then you can force push to your branch on github and this PR will update. (Force push should be used very sparingly; a good rule is to use it only when other people you're collaborating with ask for it.) If you Google for it, there are some good articles (I recommend the one from thoughtbot).

In the final commit message I'd like to see the reason why this change is being made as well as the summary of it. I'm not entirely clear what the purpose of changing to all lower-case on its own is. Are you intending to move to html5, for instance? In which case you should do that all in one PR. I have [a presentation on committing for git](http://www.slideshare.net/jaylett/on-committing-well) which you might find useful.

ivmarkp:
Thanks @jaylett for detailed explanation and resources to look into. I'm reading the article you mentioned and will go through the slides of the presentation. Will update the PR with suggested changes accordingly.

About the purpose of changing to all lowercase, I actually came across this task on bite-size projects list. So, I quickly did the same. Although, yes it will be helpful for transition to HTML5 and XHTML too in the future if need be.

ivmarkp:
@jaylett @ojwb Please have a look at the latest commit I just made. I hope it is in better shape now. :)

jaylett:
@ivmarkp even once you've fixed the commit message (both as per ojwb's point and having the summary be more than 'Update templates', which doesn't say what the update *does*) I'd be reluctant to merge this, because it doesn't achieve anything. Moving to html5 would, and should only require adding suitable DOCTYPE lines and then testing that everything works okay in a couple of recent browsers. Simply changing the case of all the elements and attributes is preparation for delivering something useful, but doesn't feel like the whole story.

(Beyond this we'd like to change the generated HTML to have suitable classes for styling, and perhaps an easy hook for adding a stylesheet reference. However that doesn't have to be part of this PR.)

ivmarkp:
@jaylett I'm relatively new to the OSS development. I apologize for the mistakes I made. Though, I'll try to keep improving myself. I've changed the commit summary that conveys what it does.

ojwb:
> I'd be reluctant to merge this, because it doesn't achieve anything

Alone it is mostly a cosmetic change, but I think it's beneficial by itself.  The HTML output isn't only consumed by machines, people will look at it to, and these days upper case tags in HTML look wrong to most people (anecdotal example: the old snowball website used a lot of upper case tags - when Andy Chilton reworked it to use bootstrap he commented on not having seen upper case tags for years).  So changing them would make the output look "better" in a somewhat intangible way.  It'll also mean people writing their own template can reuse parts of ours without having to have a mix of tag case, or to do the case conversion themselves.

I'd agree making it validate at HTML5 would be beneficial for many of the same reasons plus some additional ones (and in fact I added that to `ProjectIdeas` a few days ago), but I'm OK to merge this as it is.

And it was something we have had on the bite-size project ideas list for ages - it seems unfair to suggest someone do something, then after they do exactly what was suggested, say we don't want to merge it.

jaylett:
@ivmarkp sorry about the confusion, as ojwb says this is fine — I'd failed to spot that lowercasing was actually on our ideas list. This PR is all good now (except that it's probably not appropriate to mention XHTML because a. the attribute values aren't quoted, which we'd also have to do and b. XHTML is considered fairly outdated these days compared to html5).

jaylett:
This has been merged upstream, thanks @ivmarkp!

ivmarkp:
Thanks!! @ojwb @jaylett :)

PR #86, opened 2016-03-30
Created method to set stopper strategy

arnavkj1995:


xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -121,9 +121,9 @@ inline unsigned check_suffix(unsigned ch) {
 }
 
 // FIXME: add API for this:
```
jaylett: You shouldn't commit commented-out code; just remove it (and the above comment). git makes it easy enough to figure out what was there before if we ever need it.

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -241,15 +241,15 @@ TermGenerator::Internal::index_text(Utf8Iterator itor, termcount wdf_inc,
 {
     bool cjk_ngram = (flags & FLAG_CJK_NGRAM) || CJK::is_cjk_enabled();
 
-    int stop_mode = STOPWORDS_INDEX_UNSTEMMED_ONLY;
+    //int stop_mode = STOPWORDS_INDEX_UNSTEMMED_ONLY;
```
jaylett: Likewise, just remove this.

xapian-core/queryparser/termgenerator_internal.h
```diff
@@ -43,7 +44,7 @@ class TermGenerator::Internal : public Xapian::Internal::intrusive_base {
     WritableDatabase db;
 
   public:
-    Internal() : strategy(STEM_SOME), stopper(NULL), termpos(0),
+    Internal() : strategy(STEM_SOME), stop_mode(STOPWORDS_NONE), stopper(NULL), termpos(0),
```
jaylett: Is this correct? If I create a `TermGenerator` and set its stopper, do we not want the default behaviour to be `INDEX_UNSTEMMED_ONLY`, as before?

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -241,15 +241,15 @@ TermGenerator::Internal::index_text(Utf8Iterator itor, termcount wdf_inc,
...
 
-    if (!stopper) stop_mode = STOPWORDS_NONE;
+    if (!stopper) stop_mode = TermGenerator::STOPWORDS_NONE;
```
jaylett: This will change the stored `stop_mode`, which I don't think is the right thing to do. (Otherwise, if I set the stopping strategy but don't set the stopper, call `index_text()` and then set the stopper, it won't do quite what I expect.)

xapian-core/include/xapian/termgenerator.h
```diff
@@ -141,6 +144,15 @@ class XAPIAN_VISIBILITY_DEFAULT TermGenerator {
...
     void set_stemming_strategy(stem_strategy strategy);
 
+    /** Set the stopper strategy.
+     *
+     *  @param strategy The strategy to use - possible values are:
+     *   - STOPWORDS_NONE:   
+     *   - STOPWORDS_IGNORE:   
+     *   - STOPWORDS_INDEX_UNSTEMMED_ONLY:    
+     */
+    void set_stopper_strategy(stop_strategy strategy);
```
jaylett: The behaviour of these should be documented here in the docstring. We may also be able to come up with a better name than `INDEX_UNSTEMMED_ONLY` (perhaps `SOME`, as with stemming strategy, but that might be more confusing).

We may also want to use the prefix `STOP_` instead of `STOPWORDS_`, so it's more like an instruction: "set stopping strategy to stop none". But that may involve renaming all of these enum constants to make it read sensibly.

jaylett:
You also need to write some tests that appropriate stopping happens with the various options.

Separately to that, your commit message is very terse! It's a good idea to talk about the motivation (allowing people to set the stopping strategy during indexing) rather than just to repeat what the code does. There's some tips on writing commit messages in [this article](https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message), which is linked from our draft developer guide.

arnavkj1995:
@jaylett : I have made the changes. I will start writing some tests. 

Thank you for the article on how to commit.

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -241,15 +236,17 @@ TermGenerator::Internal::index_text(Utf8Iterator itor, termcount wdf_inc,
 {
     bool cjk_ngram = (flags & FLAG_CJK_NGRAM) || CJK::is_cjk_enabled();
 
-    //int stop_mode = STOPWORDS_INDEX_UNSTEMMED_ONLY;
-
-    if (!stopper) stop_mode = TermGenerator::STOPWORDS_NONE;
+	stop_strategy current_stop_mode;
```
jaylett: This now looks like the indentation no longer matches up.

xapian-core/include/xapian/termgenerator.h
```diff
@@ -141,6 +144,16 @@ class XAPIAN_VISIBILITY_DEFAULT TermGenerator {
      */
     void set_stemming_strategy(stem_strategy strategy);
 
+    /** Set the stopper strategy.
+     *
+     *  @param strategy The strategy to use - possible values are:
+     *   - STOP_NONE:     Don't ignore the word if it is a stop word.  
+     *   - STOP_IGNORE:   Ignore the word if it is a stop word. 
+     *   - STOP_INDEX_UNSTEMMED_ONLY: Ignore the stemmed form of the
+     *               word if it is a stop word.
```
jaylett: I suspect this would be easier to understand if it was expressed the other way round. So instead of:

> Don't ignore the word if it is a stop word.

we could have:

> Index all words (don't use the stopper).

"Don't index" is probably better than 'Ignore' in the second, and the third should probably be something like "Only index the unstemmed…".

I wonder if the last constant should be named `STOP_STEMMED_FORMS` or something. And maybe the middle one could be `STOP_ALL` (which matches `STOP_NONE` better than `STOP_IGNORE`, I think). @ojwb do you have thoughts on these names?

arnavkj1995: I have edited the documentation of the function and changed STOP_IGNORE to STOP_ALL and STOP_INDEX_UNSTEMMED_ONLY to STOP_STEMMED_FORM.

ojwb:
> @ojwb do you have thoughts on these names?

I'd agree the existing names could be improved.

`STOP_NONE` and `STOP_ALL` seems good, and match `STEM_NONE` and `STEM_ALL`.

Not sure about how best to name the other option, but `STOP_STEMMED_FORM` seems better than the name being used internally.  It's a very good thing to carefully consider names before adding a new API (consistent and logical names can make a huge difference in how easy it is to learn and use an API), but it matters most for classes and virtual methods, then for non-virtual methods, and least for constants in terms of how traumatic it is to change them later.

_"Index all words (don't use the stopper)"_ is a bit misleading - we don't index "words" over a certain length for example.  Just _"Disable stopping"_ maybe?

arnavkj1995:
@ojwb : Changed index to Disable stopping. Can we name the last constant STOP_STEMMED because it enables the stopper for stemmed form the word?

arnavkj1995:
@ojwb @jaylett : I have made the changes.

xapian-core/include/xapian/termgenerator.h
```diff
@@ -141,6 +144,17 @@ class XAPIAN_VISIBILITY_DEFAULT TermGenerator {
      */
     void set_stemming_strategy(stem_strategy strategy);
 
+    /** Set the stopper strategy.
+     *
+     *  @param strategy The strategy to use - possible values are:
+     *   - STOP_NONE:   Disable stopper for both the stemmed and unstemmed form 
+     *           of the word (Don't use the stopper).  
+     *   - STOP_ALL:    Don't disable stopper for both forms of the word.
```
jaylett: This would probably be clearer as "Enable stopped for both forms of the word". Generally a positive statement is easier for people to understand than a negative one.

xapian-core/include/xapian/termgenerator.h
```diff
@@ -141,6 +144,17 @@ class XAPIAN_VISIBILITY_DEFAULT TermGenerator {
...
     void set_stemming_strategy(stem_strategy strategy);
 
+    /** Set the stopper strategy.
+     *
+     *  @param strategy The strategy to use - possible values are:
+     *   - STOP_NONE:   Disable stopper for both the stemmed and unstemmed form 
+     *           of the word (Don't use the stopper).  
+     *   - STOP_ALL:    Don't disable stopper for both forms of the word. 
+     *   - STOP_STEMMED:  Disable stopper for the unstemmed form of the
+     *           word.
```
jaylett: It looks slightly neater if you align the second line with the description (likewise for `STOP_NONE`). (Look at the `STEM_` documentation above.) Also, rephrasing as positive as above would be helpful.

jaylett:
@arnavkj1995 this is mostly in good shape now, although you still need to write some automated tests for the newly-exposed options.

arnavkj1995:
@jaylett : Wrote the automated tests for stopper strategy and made changes in the documentation.

xapian-core/include/xapian/queryparser.h
```diff
@@ -90,7 +90,7 @@ class XAPIAN_VISIBILITY_DEFAULT SimpleStopper : public Stopper {
      * Xapian includes stop list files for many languages. You can initialise from a file like that:
      * @code
      * ifstream inFile ("stopwords/english/stop.txt");
-     * Xapian::SimplerStopper stopper(istream_iterator<string>(inFile), istream_iterator<string>());
+     * Xapian::SimpleStopper stopper(istream_iterator<string>(inFile), istream_iterator<string>());
```
ojwb: This is useful to fix, but shouldn't be part of a patch for something else.

xapian-core/tests/termgentest.cc
```diff
@@ -735,7 +755,22 @@ static bool test_termgen1()
 	    } else if (strncmp(o, "some", 4) == 0) {
 		o += 4;
 		termgen.set_stemming_strategy(termgen.STEM_SOME);
-	    } else if (strncmp(o, "prefix=", 7) == 0) {
+	    } else if(strncmp(o, "stop=", 5) == 0) {
+               o += 7;
+               vector<string> x = {"the", "a", "an"};
+               Xapian::SimpleStopper *stopper = new Xapian::SimpleStopper(x.begin(), x.end());
+               Xapian::Stopper *stop = static_cast<Xapian::Stopper *>(stopper);
+               termgen.set_stopper(stop);
```
ojwb: The cast here should be unnecessary - the compiler can implicitly convert a pointer to a derived class into a pointer to the base class.

However, you do have a memory leak here - you need to release ownership of the `Stopper` object so that the `TermGenerator` object will delete it later:

```
termgen.set_stopper(stopper->release());
```

xapian-core/tests/termgentest.cc
```diff
@@ -735,7 +755,22 @@ static bool test_termgen1()
 	    } else if (strncmp(o, "some", 4) == 0) {
 		o += 4;
 		termgen.set_stemming_strategy(termgen.STEM_SOME);
-	    } else if (strncmp(o, "prefix=", 7) == 0) {
+	    } else if(strncmp(o, "stop=", 5) == 0) {
```
ojwb: Xapian coding style is to put a space between `if` and `(`.

xapian-core/tests/termgentest.cc
```diff
@@ -735,7 +755,22 @@ static bool test_termgen1()
...
+               o += 7;
```
ojwb: `7` here doesn't match `5` in the line above.

xapian-core/tests/termgentest.cc
```diff
@@ -735,7 +755,22 @@ static bool test_termgen1()
...
+               vector<string> x = {"the", "a", "an"};
```
ojwb: `vector<string>` is overkill here - the size is constant and known - `array<const char*, 3>` should work just as well.

xapian-core/tests/termgentest.cc
arnavkj1995: Released the ownership of the Stopper object.

arnavkj1995:
@ojwb : Made the changes according to your comments.

xapian-core/tests/termgentest.cc
```diff
@@ -50,7 +50,7 @@ struct test {
     //    (this persists for subsequent tests until it's turned off).
     //  - some: Set stemming strategy to STEM_SOME.
     //    (this persists for subsequent tests until it's turned off).
-    //	- stop=FOO Change stemming algorithm to foo ("none" to turn off).
+    //	- stop=en: Add the stop words.
```
jaylett: I'd state the stop words explicitly, or perhaps "Add a small set of English stop words".

jaylett:
@arnavkj1995 when adding commits to a PR to deal with previous comments, it's often helpful to add them as 'fixup' commits (`git commit --fixup <sha>`) based on the previous commit they are correcting. This then makes it easy to use `git rebase` to "squash" them down into the commits that would have existed had you done all the work in one go. It also acts as a hint to someone reviewing the PR of where to look for context and the previous comments that are being addressed.

It's also useful when you notice problems yourself. For instance, `d0398c8` (where you remove the duplicate `set_stopper()` call in the tests) could have been `git commit --fixup 0c6e703`, which would have made it explicit that it was tidying up the previous commit.

arnavkj1995:
Sorry @jaylett ! Its my mistake! I have merged those two commits using `git rebase` and updated the comments of automated tests using `git commit --fixup <sha>` as suggested by you above.

jaylett:
@arnavkj1995 it's generally good practice not to rebase while a PR is in review, because it makes it harder to match up new changes with the comments they address. (This is mostly true because of how the github interface works.) So in future if you could just push the fixup commits until the PR is ready, then squash at the end just before we merge the patch, that'd be helpful.

arnavkj1995:
@jaylett : Sorry, I didn't realise that. From the next time onwards, I will keep that in mind :)

xapian-core/tests/termgentest.cc
```diff
@@ -152,6 +161,19 @@ static const test test_simple[] = {
...
+
+    { "stop_all",
+    	"The stop words.", "stop[1] words[2]" },
+
+    { "stop_stemmed",
+    	"The stop words.", "stop[2] the[1] words[3]" },
+
+    { "stem=en,all,stop_stemmed",
+    	"The stemmed words.", "stem[2] the[1] word[3]" },
+
```
jaylett: Given the stopper persists, shouldn't we end this set with the `stop_none` test, so it's equivalent to not having a stopper configured?

arnavkj1995: Should I add one new `stop_none` test below or change the order of the tests, by moving the `stop_none` test at the end ?

jaylett: Unless you have something else you want to test, just reorder them (and add a comment explaining that it's there for that reason).

arnavkj1995: Done :)

xapian-core/tests/termgentest.cc
```diff
@@ -171,8 +171,13 @@ static const test test_simple[] = {
     { "stop_stemmed",
     	"The stop words.", "stop[2] the[1] words[3]" },
```
jaylett: Should we be running a test with stemming and `stop_stemmed`? Because this has stemming disabled, right? (Because of the `none` at line 165.) Ideally we want to test the different combinations of stemming and stopping options.

arnavkj1995: Modified the tests. Now `STOP_STEMMED` is getting tested.

jaylett:
@arnavkj1995 have you compiled this final version? I'm having problems on OS X with clang, so this may be a compiler-specific issue.

arnavkj1995:
Yes, I have compiled the final version and ran the tests too.

Can you tell me on which part/file/function is the issue arising ?

jaylett:
@arnavkj1995 clang has problems with not including `<array>` for `std::array`, and also complains about brace elision with the `-W` options we have; it seems like the easiest solution is to just not use elision. (GCC had the same error, and now disables that warning, which seems drastic and worrisome to me. Clang has an open bug for the same.)

arnavkj1995:
@jaylett : I have complied the final version before pushing and recompiled after running `make distclean`. I am using Ubuntu on my workstation.

Can you tell me on which part/function/class you are getting the compiler issue/error ?

xapian-core/tests/termgentest.cc
```diff
@@ -735,7 +762,21 @@ static bool test_termgen1()
 	    } else if (strncmp(o, "some", 4) == 0) {
 		o += 4;
 		termgen.set_stemming_strategy(termgen.STEM_SOME);
-	    } else if (strncmp(o, "prefix=", 7) == 0) {
+	    } else if (strncmp(o, "stop=en", 7) == 0) {
+               o += 7;
+               array<const char *, 3> x = {"the", "a", "an"};
```
arnavkj1995: @jaylett : Should I change the above line to array<const char *, 3> x = {{"the", "a", "an"}};

I read on this link that Clang rejects list-initialization with single braces([link](http://stackoverflow.com/questions/16985687/brace-elision-in-stdarray-initialization)).

jaylett: @arnavkj1995 yep, this is brace elision. What you're actually doing is initialising with `{{"the", …}}`, ie initialising the first member of `x` to `{"the", …}`. (And the first member of `array<T>` is the actual array data.)

arnavkj1995: @jaylett : I have fixed it. It should work now.

jaylett:
@arnavkj1995 you also need to include `<array>`, as I previously indicated. (If you have space, it may be worth installing the clang toolchain so you can check whether this problem exists with mainline clang or just Apple's version. If it's on both, then a fix that works on Ubuntu should be fine for OS X also.)

arnavkj1995:
@jaylett : Can I use Travis-CI([link](https://travis-ci.org/)) for compiling in Clang?

jaylett:
@arnavkj1995 you can try (just add the travis configuration to your own branch). We won't pull travis config into master right now, although we're open to starting to use it for branch testing this year, so having a basic configuration that works would be really helpful :-)

arnavkj1995:
@jaylett : I think its compiling. I installed clang and configured using `GCC=clang g++=clang++ ../configure` and then run `make` to compile it. Is that the right way to compile using Clang ?

jaylett:
@arnavkj1995 that will compile the main library; you need to `(cd tests; make termgentest)` to build the relevant test program.

arnavkj1995:
Its compiling and i ran it too. All the test cases are passing.

jaylett:
@arnavkj1995 at this point it certainly should :-)

jaylett:
@arnavkj1995 that compiles here cleanly as well now. Excellent!

jaylett:
@arnavkj1995 okay, sorry for the tiny messages. Can you squash all these into a single commit against latest master? (`git rebase -i master` and change all but the first commit to 'fixup' should do it.) Then we should be able to get this merged :-)

arnavkj1995:
You mean squashing all the commits done till now to a single commit using `git rebase` ?

jaylett:
@arnavkj1995 yes please :-)

It's generally useful to keep separate commits during PR, and then squash them to a smaller number that will live permanently in the history. In this case, I think one commit that adds this new functionality is best.

arnavkj1995:
@jaylett : I merged the code to resolve the merge conflicts while I was working on this PR. And the commits are not consecutive now and I am not able to squash them into a single commit. Can you suggest something ?

ojwb:
> @jaylett : I think its compiling. I installed clang and configured using
GCC=clang g++=clang++ ../configure and then run make to compile it. Is that
the right way to compile using Clang ?

No, that is wrong in several ways - you want:

../configure CC=clang CXX=clang++

jaylett:
@arnavkj1995 you can squash even with the merge commit, but it's more work. (Generally, for projects that rebase rather than merge, you should do the same to make this easier.)

What I'd do first is to cherry-pick the pre-merge commits onto latest master. (That will involve doing the same merge resolution as you did previously.) Then you can rebase the post-merge commits onto your HEAD, and finally squash them all down.

(Alternatively you might have luck making a patch from your branch and applying it as one commit onto master. Again, you may need to resolve conflicts.)

arnavkj1995:
@jaylett : I will do it by this weekend. I have few tests and project submissions in this week :) Sorry for the late reply !

jaylett:
@arnavkj1995 no problem! Hope the academic work is going well :)

arnavkj1995:
@jaylett : I have squashed the commits using cherry-pick. Please have a look :)

jaylett:
@arnavkj1995 you haven't rebased onto latest master as I asked (the parent of the first commit is from February), and the second commit here is still a merge commit. Additionally, the first commit (which is the one you can probably now rebase easily onto latest master, and discard the other) shouldn't have in its commit messages all the mini-commits you've squashed together; just leave it as a good commit message that describes the actual final commit.

arnavkj1995:
@jaylett : Sorry for that mistake!

I have now rebased it into current master. Now there is a single final commit and commit message.

xapian-core/queryparser/termgenerator_internal.h
```diff
@@ -37,14 +37,15 @@ class TermGenerator::Internal : public Xapian::Internal::intrusive_base {
...
+    stop_strategy stop_mode;
     Document doc;
     termcount termpos;
     TermGenerator::flags flags;
     unsigned max_word_length;
     WritableDatabase db;
 
   public:
-    Internal() : strategy(STEM_SOME), stopper(NULL), termpos(0),
+    Internal() : strategy(STEM_SOME), stop_mode(STOP_STEMMED), stopper(NULL), termpos(0),
```
jaylett: Building this with clang and default warnings (and `-Werror`) fails because `stop_mode` and `stopper` are the other way round in the class definition. Does this build okay on gcc?

jaylett:
No response from original contributor, so rebased and moved to #109.

PR #87, opened 2016-04-02
Java Iterator implementation: MSetIterator

mkash32:
Created a class MatchDescriptor as an inner class of MSetIterator. MSetIterator now implements the Iterable<MatchDescriptor> interface and implements the method iterator() to enable the use of MSetIterator in a foreach loop.

### DIFF
--- a/xapian-bindings/java/SmokeTest.java
+++ b/xapian-bindings/java/SmokeTest.java
@@ -121,8 +121,20 @@ public class SmokeTest {
 		System.exit(1);
 	    }
 	    Enquire enq = new Enquire(db);
-	    enq.setQuery(new Query(Query.OP_OR, "there", "is"));
+	    enq.setQuery(new Query("dog"));
 	    MSet mset = enq.getMSet(0, 10);
+	    // MSet size is 0
+	    if (mset.size() == 0) {
+		for (MSetIterator.MatchDescriptor descriptor : mset) {
+		    System.err.println("Empty iterator entered invalid section (from MSet java iterator)");
+		    System.exit(1);
+		}
+	    } else {
+		System.err.println("Unexpected mset.size(), should be 0");
+		System.exit(1);
+	    }
+	    enq.setQuery(new Query(Query.OP_OR, "there", "is"));
+	    mset = enq.getMSet(0, 10);
 	    if (mset.size() != 1) {
 		System.err.println("Unexpected mset.size()");
 		System.exit(1);
@@ -131,7 +143,7 @@ public class SmokeTest {
 	    Document m_doc = null;
 	    long m_id;
 	    while(m_itor.hasNext()) {
-		m_id = m_itor.next();
+		m_id = m_itor.next().getDocId();
 		if(m_itor.hasNext()) {
 		    m_doc = mset.getDocument(m_id);
 		}
@@ -140,7 +152,16 @@ public class SmokeTest {
 	    // Only one doc exists in this mset
 	    if(m_doc != null && m_doc.getDocId() != 0) {
 		System.err.println("Unexpected docid");
-		    System.exit(1);
+		System.exit(1);
+	    }
+	    // New iterator, should start from the beginning
+	    int iterations = 0;
+	    for(MSetIterator.MatchDescriptor descriptor : mset) {
+		iterations++;
+	    }
+	    if(iterations != mset.size()) {
+		System.err.println("Unexpected number of iterations (from MSet java iterator)");
+		System.exit(1);
 	    }
 
 	    String term_str = "";
--- a/xapian-bindings/java/docs/examples/SimpleSearch.java
+++ b/xapian-bindings/java/docs/examples/SimpleSearch.java
@@ -52,16 +52,12 @@ public class SimpleSearch {
         Enquire enquire = new Enquire(db);
         enquire.setQuery(query);
         MSet matches = enquire.getMSet(0, 2500);    // get up to 2500 matching documents
-        MSetIterator itr = matches.begin();
 
         System.err.println("Found " + matches.size() + " matching documents using " + query);
-        while (itr.hasNext()) {
-            int percent = itr.getPercent();
-            long docID = itr.next();
-            // TODO:  Make this more like a Java Iterator by returning some
-            // kind of "MatchDescriptor" object
-            Document doc = db.getDocument(docID);
-            System.err.println(percent + "% [" + docID + "] " + doc.getValue(0));
+        for (MSetIterator.MatchDescriptor match : matches) {
+            System.err.println(match.getPercent() + "% "
+				+ "[" + match.getDocId() + "] "
+				+ match.getDocument().getValue(0));
         }
     }
 
--- a/xapian-bindings/java/java.i
+++ b/xapian-bindings/java/java.i
@@ -174,20 +174,68 @@ namespace Xapian {
     bool hasNext() const { return Xapian::iterator_valid(*self); }
 }
 
+%ignore MSetIterator::next();
+%javamethodmodifiers MSetIterator::next_helper() "private"
+
 %extend MSetIterator {
-    Xapian::docid next () {
-	Xapian::docid tmp;
+    void next_helper () {
 	if (Xapian::iterator_valid(*self)) {
-	    tmp = (**self);
 	    ++(*self);
-	} else {
-	    tmp = -1;
 	}
-	return tmp;
     }
 
     bool hasNext() const { return Xapian::iterator_valid(*self); }
 }
+
+%typemap(javaimports) Xapian::MSetIterator "import java.util.Iterator;"
+%typemap(javaimports) Xapian::MSetIterator %{
+import java.util.NoSuchElementException;
+import java.util.Iterator;
+%}
+%typemap(javainterfaces) Xapian::MSetIterator "Iterator<MSetIterator.MatchDescriptor>"
+%typemap(javacode) MSetIterator%{
+    @Override
+    public MatchDescriptor next() {
+        MatchDescriptor descriptor = new MatchDescriptor();
+        if(hasNext()) {
+            nextHelper();
+        } else {
+            throw new NoSuchElementException();
+        }
+        return descriptor;
+    }
+
+    public class MatchDescriptor {
+        private long docId,rank;
+        private int percent;
+        private double weight;
+        private Document document;
+
+        public MatchDescriptor() {
+            MSetIterator thisIterator = MSetIterator.this;
+            this.docId = thisIterator.getDocId();
+            this.rank = thisIterator.getRank();
+            this.percent = thisIterator.getPercent();
+            this.weight = thisIterator.getWeight();
+            this.document = thisIterator.getDocument();
+        }
+
+        public long getDocId() { return docId; }
+        public long getRank() { return rank; }
+        public int getPercent() { return percent; }
+        public double getWeight() { return weight; }
+        public Document getDocument() { return document; }
+    }
+
+%}
+
+%typemap(javaimports) Xapian::MSet "import java.util.Iterator;"
+%typemap(javainterfaces) Xapian::MSet "Iterable<MSetIterator.MatchDescriptor>";
+%typemap(javacode) MSet%{
+    public Iterator<MSetIterator.MatchDescriptor> iterator() {
+        return begin();
+    }
+%}
 
 }
 


xapian-bindings/java/java.i
```diff
@@ -189,6 +189,64 @@ namespace Xapian {
...
+          public MatchDescriptor next() {
+              MSetIterator thisIterator = MSetIterator.this;
+              long rank = thisIterator.getRank();
+              long docId = thisIterator.getDocId();
+              double weight = thisIterator.getWeight();
+              int percent = thisIterator.getPercent();
+              Document doc = thisIterator.getDocument();
+              MatchDescriptor descriptor = new MatchDescriptor(docId, rank, percent, weight, doc);
+	      thisIterator.next();
+              return descriptor;
```
ojwb: There's an indentation difference here (it'll look OK in an editor, but in the patch viewer the leading `+` means it's visible).  Looks like line 210 has a tab, but the other lines only use spaces.  It's better to use consistent indentation.

xapian-bindings/java/java.i
```diff
@@ -189,6 +189,64 @@ namespace Xapian {
...
+
+      public int getPercent() { return percent; }
+      public void setPercent(int percent) { this.percent = percent; }
+
+      public double getWeight() { return weight; }
+      public void setWeight(double weight) { this.weight = weight; }
+
+      public Document getDocument() { return document; }
+      public void setDocument(Document document) { this.document = document; } 
+  }
```
ojwb: I would suggest omitting the setters here.  They aren't needed for the expected use of the iterator, and if we don't have setters there's more scope for changing the implementation - for example, `MatchDescriptor` could hold an `MSetIterator` and just call that to lazily fetch the things which are actually asked for.

ojwb:
Generally this looks pretty good to me.

But what does this look like in use?  I'm not familiar with Java's iterator pattern, but would it be more natural for `MSet` to have this `iterator()` method?

It ought to have some feature tests, to make sure it actually works as intended (and that it stays working in the face of future changes to the code).

The Java bindings documentation ought to talk about this, and the example programs probably ought to use it.

xapian-bindings/java/java.i
mkash32: To align this part with the rest of SWIG generated java file, I intentionally added the spaces. I fixed the thisIterator.next() line. I originally used tabs but the output java file was unaligned, so I switched to spaces. The spacing looks odd here but in the generated java file it aligns neatly.

jaylett: @mkash32 we don't particularly care about the generated Java file, since almost no one will ever look at it. Indentation and alignment of source files, which Xapian developers have to work with, is more important.

jaylett:
@mkash32 don't we actually want `Iterator<MatchDescriptor> iterator()` on the `MSet`? Otherwise you have to do something like `mset.begin().iterator()` which looks weird to me.

I also wonder if it'd be possible to add `hasNext()` and `next()` to the wrapped `MSetIterator`, and make that implement `E<MatchDescriptor>`. (That way `MSet::iterator()` might even just be a rename of `MSet::begin()`.)

mkash32:
After looking at the Collections library, I think it would be more consistent with java collections if we make MSet implement Iterable. 
@jaylett Yeah that sound good. We could create the Iterator class to contain an instance of MSetIterator which can be initialized in the constructor of the Iterator with MSet.begin(). Then Iterator would just be a wrapping of MSetIterator.

jaylett:
@mkash32 is there a reason why we can't make the `MSetIterator` into the thing we want rather than having to wrap it? All the methods are there in some shape or other, so isn't it just getting the interface declaration in there? (It may be that SWIG can't achieve this, which would be annoying but a good reason!)

mkash32:
Oh sorry, I misunderstood your previous post. The implementation you suggested would be easiest and most logical, I'll proceed with that. (And SWIG is making this implementation surprisingly simple!)

xapian-bindings/java/java.i
```diff
@@ -174,8 +174,11 @@ namespace Xapian {
     bool hasNext() const { return Xapian::iterator_valid(*self); }
 }
 
+%ignore MSetIterator::next();
+%javamethodmodifiers MSetIterator::next(int) "private"
+
 %extend MSetIterator {
-    Xapian::docid next () {
+    Xapian::docid next (int pseudoArg) {
```
mkash32: The whole work around starts here. We want to make MSetIterator itself be an iterator. So we make MSetIterator implement the `Iterator<MSetIterator.MatchDescriptor>` interface. To implement this interface, we need to implement two methods, hasNext() and next(). `public boolean hasNext()` is already implemented so no problem there. We also need to implement `public MatchDescriptor next()`, but here we face the problem since we already have a next() function. The only way to overcome this is to overload the next functions, where the main next function will be the one that is required by the interface. We overload by adding an argument to the function signature of `long next()`. So I added a pseudo argument which doesn't serve any purpose, it just enables us to overload.
`long next(int pseudoArg)` will then be a confusing method to deal with and it should be hidden outside of the class, so I made it a private method.

I feel this is the best way to make MSetIterator implement Iterator. The only issue here would be that `long next()` can't be used anymore. We could keep the method as public and explain in the documentation to the user that the argument should be kept as 0. If we keep it as private the only issue would be that next is being used differently in different languages.

@jaylett @ojwb What do you guys think? Is this a good way to do it or should I just wrap the MSetIterator into a custom implementation of Iterator interfaces?

mkash32:
MSet mset = enquire.getMSet(0, 2500);    // get up to 2500 matching documents

        for (MatchDescriptor descriptor: mset) {
              int percent = descriptor.getPercent();
              long docID = descriptor.getDocId();

              Document doc = descriptor.getDocument();
              System.err.println(percent + "% [" + docID + "] " + doc.getValue(0));
        }

@ojwb It would look like that^ in use.

xapian-bindings/java/java.i
ojwb: You don't really want the existing `next()` method's current functionality anyway, as it dereferences the iterator before advancing it and returns that value, and you aren't interested in that.

But the current iterator wrappers weren't present in the hand-coded JNI Java bindings in 1.2.x, so compatibility with them isn't a big concern - it's fine to throw out the current `next()` behaviour to implement something smarter.

So I'd just replace the `next()` added by extend with a simpler helper method which does what you want.

Ideally it would check `Xapian::iterator_valid(*self)` as well as incrementing the iterator, but unfortunately that's too late as you already read values in the `MatchDescriptor` constructor. Really you want (in pseudo-code):

```
if (!iterator_valid()) {
    return null;
}

MatchDescriptor descriptor = new MatchDescriptor();
_next();
return descriptor;
```

It would probably work to use `%extend` to add an `iterator_valid()` method to the wrapped object.

There's probably scope for optimising this - `next()` then makes 7 JNI calls when there's a next element, and we could potentially just make one to check validity, read docid, rank, percent, etc, increment the iterator and return all the values.  But getting this working and done for all the iterators is most important - optimising the implementation can be done later (and profiling it to see where the time is actually spent would be wise first anyway).

mkash32: Is it necessary to add the check for `Xapian::iterator_valid(*self)` in next? In the foreach loop `for (MatchDescriptor descriptor: mset)`, `hasNext()` is called before every iteration and directly followed by a call to `next()`. It is equivalent to the pseudo code you suggested above, if we add a check to next then it would result in consecutive checks when we use the iterator.

ojwb: The Java documentation seems to say that `next()` should throw an exception in this case:

https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#next--

I suspect in practice people don't rely on that, but I don't know much about Java norms.  But running the C++ iterator past the end results in undefined behaviour, which is unhelpful in a case which Java defines the behaviour of.

At the C++ level, `iterator_valid()` is a very cheap check - it's just comparing an integer with zero:

```
inline bool iterator_valid(const Xapian::MSetIterator & it) {
    return it.off_from_end != 0;
}
```

mkash32: The exception will only be thrown if we take an iterator from MSet and use it in a while loop, forcing next() calls without checking the existence of next item with hasNext(). If we use the MSet in the for each loop we won't face any problem (next() can't even be called since there is no access to the iterator object) and no exceptions will be thrown.

What would you suggest? I could add a hasNext() check in the next() function.

ojwb: Ah yes, we already wrap `iterator_valid()` as `hasNext()`.

I do realise that the normal way to use the iterator doesn't exercise this case.  The Java docs seem to specify a behaviour, so the conservative option is to implement what that says, but I don't really know how important it is to do so for a case like this.

The current Java iterator wrappers that dcolish added just return a dummy value from `next()` in this case (e.g. `-1` for `MSetIterator`), so they don't follow that documentation, but they've not been in a stable release yet, so probably aren't widely used.

xapian-bindings/java/java.i
```diff
@@ -174,21 +174,63 @@ namespace Xapian {
...
 }
 
+%ignore MSetIterator::next();
+%javamethodmodifiers MSetIterator::next_helper() "private"
+
 %extend MSetIterator {
-    Xapian::docid next () {
-	Xapian::docid tmp;
+    void next_helper () {
 	if (Xapian::iterator_valid(*self)) {
```
mkash32: I changed the helper function to just check the validity of the iterator and move on to the next item.

xapian-bindings/java/java.i
```diff
@@ -174,21 +174,63 @@ namespace Xapian {
...
     bool hasNext() const { return Xapian::iterator_valid(*self); }
 }
 
+%typemap(javaimports) Xapian::MSetIterator "import java.util.Iterator;"
+%typemap(javainterfaces) Xapian::MSetIterator "Iterator<MSetIterator.MatchDescriptor>"
+%typemap(javacode) MSetIterator%{
+    @Override
+    public MatchDescriptor next() {
+        MatchDescriptor descriptor = new MatchDescriptor();
+        if(hasNext()) {
```
mkash32: I feel this check may not be necessary because we are already checking the iterator validity within the next_helper function before moving to the next item.

mkash32:
int iterations=0;
	    for(MatchDescriptor descriptor : mset) {
                iterations++;
        }
	    // Only one doc exists in this mset
	    if(iterations != mset.size()) {
		    System.err.println("Unexpected number of iterations (from MSet java iterator)");
		    System.exit(1);
	    }
Would this be a good test case?

jaylett:
@mkash32 for tests, you want to do two things: check that the expected normal case works, and check that boundary conditions (failures, mostly) also do the correct thing according to the API. Additionally, you never want to `System.exit(1)` within a test, because it will prevent subsequent tests from being run.

It's probably worth moving to jUnit for the Java tests, because then they'll be a more familiar structure for everyone who works on them. So you'll end up writing a number of test methods to cover the various different situations (eg: normal behaviour, empty MSet, re-using an iterator, …).

mkash32:
I was planning on migrating all the tests in SmokeTest.java to a test suite in jUnit. Would it be ok to add on to SmokeTest.java for now?
I have a few days before I get tied up with academic work. I would like to finish the MSet Java Iterator with test cases (and also change the examples to incorporate the Java Iterator) and hopefully move on to the other Iterators also in this time.

jaylett:
@mkash32 absolutely, if you can get some tests written in SmokeTest as it stands, that's better than having to put off finishing this up.

xapian-bindings/java/SmokeTest.java
```diff
@@ -142,6 +153,14 @@ public static void main(String[] args) throws Exception {
 		System.err.println("Unexpected docid");
 		    System.exit(1);
 	    }
+	    // New iterator, should start from the beginning 
+	    int iterations = 0;
+	    for(MSetIterator.MatchDescriptor descriptor : mset) {
```
mkash32: I wanted to create a test where there were more documents, but currently there is only 1 document. I didn't think it was a good idea to add more documents now because the tests are somewhat dependent on each other. After switching to jUnit we won't have to worry about the dependencies.

xapian-bindings/java/java.i
jaylett: This check is necessary, but if it fails you should throw `NoSuchElementException`.

xapian-bindings/java/java.i
jaylett: I'm not sure why this is a separate function; it's only used once, so why not just inline it in the `next()` implementation further down?

xapian-bindings/java/docs/examples/SimpleSearch.java
```diff
@@ -52,15 +52,12 @@ public static void main(String[] args) throws Exception {
...
-            int percent = itr.getPercent();
-            long docID = itr.next();
-            // TODO:  Make this more like a Java Iterator by returning some
-            // kind of "MatchDescriptor" object
-            Document doc = db.getDocument(docID);
+        for (MSetIterator.MatchDescriptor match : matches) {
+            int percent = match.getPercent();
+            long docID = match.getDocId();
+            Document doc = match.getDocument();
             System.err.println(percent + "% [" + docID + "] " + doc.getValue(0));
```
jaylett: I'd be inclined to not even bother using local variables, given each is used only once.

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,19 @@ public static void main(String[] args) throws Exception {
...
-	    enq.setQuery(new Query(Query.OP_OR, "there", "is"));
+	    enq.setQuery(new Query("dog"));
 	    MSet mset = enq.getMSet(0, 10);
+	    // MSet size is 0
+	    if(mset.size() != 0) {
+	    	for(MSetIterator.MatchDescriptor descriptor : mset) {
+	    		System.err.println("Empty iterator entered invalid section (from MSet java iterator)");
+	    		break;
+	    	}
+	    }else {
```
jaylett: We put spaces around `else`.

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,19 @@ public static void main(String[] args) throws Exception {
 		System.exit(1);
 	    }
 	    Enquire enq = new Enquire(db);
-	    enq.setQuery(new Query(Query.OP_OR, "there", "is"));
+	    enq.setQuery(new Query("dog"));
 	    MSet mset = enq.getMSet(0, 10);
+	    // MSet size is 0
+	    if(mset.size() != 0) {
```
jaylett: We put a space after `if`.

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,19 @@ public static void main(String[] args) throws Exception {
...
+	    	for(MSetIterator.MatchDescriptor descriptor : mset) {
```
jaylett: And after `for`.

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,19 @@ public static void main(String[] args) throws Exception {
...
+	    		System.err.println("Empty iterator entered invalid section (from MSet java iterator)");
```
jaylett: Will the test fail (overall program return non-zero) if this happens?

xapian-bindings/java/SmokeTest.java
jaylett: That seems like the right call, yes.

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,19 @@ public static void main(String[] args) throws Exception {
...
+	    		break;
+	    	}
+	    }else {
+	    	System.err.println("Unexpected mset.size()");
+	    }
```
jaylett: You could also (as well as the above) explicitly test the `hasNext()` and `next()` behaviour, so you're not just testing the full iterator integration. (It'll make it slightly more obvious if one part of this fails in future.)

jaylett:
@mkash32 I realise I've asked for spacing changes when the original file isn't well laid out already. It's worth at some point tidying the whole thing up so our Java style is closer to our C++ style in terms of whitespace, but it's probably best to leave that out of this PR.

xapian-bindings/java/java.i
mkash32: From my understanding we don't have any Java way of proceeding to the next element of the iterator, a helper function which makes a JNI call (which will execute  `++(*self)`) is needed to move to the next item.

xapian-bindings/java/SmokeTest.java
mkash32: The overall program wouldn't return non-zero. Should I just add a System.exit(1)? This would make the test cases consistent with the rest of SmokeTest. 
SmokeTest is exiting whenever a test fails (which we don't want). The tests are dependent so it would be difficult to modify it in such a manner that the remaining tests will run even when we encounter failed tests. These issues will be solved when we move to jUnit :).

jaylett: Smoketest should exit non-zero when it fails; because it's very simple that means a failure won't bother running the other tests. However that's better than having to read the output (for instance, it allows our buildbot to notice when tests fail, which helps keep track of problems).

jUnit will make this all more flexible and powerful, true :)

ojwb:
On IRC it was suggested we drop a copy of junit into the source tree, but I'm not keen on bundling dependencies.  It's a maintenance headache to have to update them - suddenly we have to care about security vulnerabilities in junit.  Once we go down the bundling route, it's hard to say "no" when we've said "yes" in other cases, and we end up with the source tree bloated by copies of third party code.

New tests cases can go in a new junit-based test suite, but my preference would be to keep the current `SmokeTest.java` as-is, to serve as what the name suggests - a [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_%28software%29).  Then developers making changes to xapian-core can still have confidence they aren't breaking the java-bindings without having to install junit.  The maintenance requirement for this should be very low.

xapian-bindings/java/java.i
ojwb: Indeed - we can't just inline one in the other since they're written in different languages (`next()` is in Java, `next_helper()` is in C++).

mkash32:
@ojwb @jaylett How is the PR looking? Are there any other things I should fix/change?

xapian-bindings/java/java.i
jaylett: D'oh. My mistake, confused between the two languages :-(

ojwb:
Not a comment on this PR directly, but I was thinking about this from IRC:

> <aakash> I was planning on making a test suite with the tests in smoke test. Apart from giving us the ability to write independent unit test, we would be able to run the tests on an android platform

Probably the cleanest way to achieve that is to slightly refactor SmokeTest.java so that it can be called from a junit testcase, or from a simple wrapper program if junit isn't available.  (And additional more sophisticated tests would just be written for junit).

mkash32:
Refactoring SmokeTest.java and calling it from a testcase sounds like a good idea. If we keep the smoke test and the more complicated tests separate then it would be easier to maintain later on also (compared to my original idea of porting all of the tests from SmokeTest into their own test case in jUni).

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,20 @@ public static void main(String[] args) throws Exception {
 		System.exit(1);
 	    }
 	    Enquire enq = new Enquire(db);
-	    enq.setQuery(new Query(Query.OP_OR, "there", "is"));
+	    enq.setQuery(new Query("dog"));
 	    MSet mset = enq.getMSet(0, 10);
+	    // MSet size is 0
+	    if (mset.size() != 0) {
```
jaylett: This test is inverted.

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,20 @@ public static void main(String[] args) throws Exception {
...
+	    	for (MSetIterator.MatchDescriptor descriptor : mset) {
+	    		System.err.println("Empty iterator entered invalid section (from MSet java iterator)");
+	    		System.exit(1);
+	    	}
+	    } else {
+	    	System.err.println("Unexpected mset.size()");
```
jaylett: Since this matches an error further down, it'd be useful to distinguish them (perhaps "…, should be 0" here).

xapian-bindings/java/SmokeTest.java
```diff
@@ -142,6 +154,15 @@ public static void main(String[] args) throws Exception {
 		System.err.println("Unexpected docid");
 		    System.exit(1);
 	    }
+	    // New iterator, should start from the beginning
```
jaylett: There's a trailing space on this line. Two ways of spotting these: `git show` will generally highlight trailing spaces (and tabs mixed with spaces); also running `git diff origin/master | ./xapian-maintainer-tools/xapian-check-patch` from the top of the checkout will check for these common whitespace problems. (You can also do `git show | ./xapian-maintainer-tools/xapian-check-patch` if you've already created a single commit.)

Right now, `xapian-check-patch` is showing about a dozen errors. If you can get them fixed, and correct the test (did you actually run the smoketest? make sure you do and that it passes) then this should be ready for merge.

If you can make changes as fixup commits, and once we're ready we can squash them down.

mkash32: Yikes, I apologize for all of the spacing issues. I will avoid them from now onward. I ran the test before committing but I guess I must've added that inverted test case afterward without checking. 
I'll fix it up and check it properly.

xapian-bindings/java/SmokeTest.java
```diff
@@ -121,8 +121,20 @@ public static void main(String[] args) throws Exception {
 		System.exit(1);
 	    }
 	    Enquire enq = new Enquire(db);
-	    enq.setQuery(new Query(Query.OP_OR, "there", "is"));
+	    enq.setQuery(new Query("dog"));
 	    MSet mset = enq.getMSet(0, 10);
+	    // MSet size is 0
+	    if (mset.size() == 0) {
+	    for (MSetIterator.MatchDescriptor descriptor : mset) {
+	    System.err.println("Empty iterator entered invalid section (from MSet java iterator)");
```
jaylett: This alignment still looks wrong to me; l128 should be indented from l127, and l129 indented even further. What editor are you using? I'm wondering if it's possible to tweak some configuration there so it'll get this right for you automatically :-)

mkash32: I kept tabs at those lines as you have mentioned but when I run `git diff origin/master | ./xapian-maintainer-tools/xapian-check-patch`, I kept getting the message "added/changed line has space before tab" so I just removed those tabs. 
The rest of the file also has indentation mistakes within if blocks, I wonder why those aren't showing up.

jaylett: Okay, you need to configure sublime to use 4 character offsets for each level of nesting, and to prefer tabs over spaces. That way it'll use tabs as much as possible (at the left end of the line), and switch to spaces later (just before the text starts). The problem comes if there are tabs after spaces, in the leading whitespace, because some editors do different things with that, so it's confusing.

For instance, line 128 should really start with two tabs, which would make it appear further indented than the surrounding `if`.

mkash32:
Looks like I've been giving bad commit messages ([using -m](https://xapian-developer-guide.readthedocs.org/en/latest/contributing/workflow.html#good-commit-messages)). Should I amend it after squashing?

jaylett:
@mkash32 yes please :)

If you use interactive rebase and squash them all together, you'll get to write a new commit message in the process.

xapian-bindings/java/docs/examples/SimpleSearch.java
```diff
@@ -52,16 +52,10 @@ public static void main(String[] args) throws Exception {
...
         System.err.println("Found " + matches.size() + " matching documents using " + query);
-        while (itr.hasNext()) {
-            int percent = itr.getPercent();
-            long docID = itr.next();
-            // TODO:  Make this more like a Java Iterator by returning some
-            // kind of "MatchDescriptor" object
-            Document doc = db.getDocument(docID);
-            System.err.println(percent + "% [" + docID + "] " + doc.getValue(0));
+        for (MSetIterator.MatchDescriptor match : matches) {
+            System.err.println(match.getPercent() + "% [" + match.getDocId() + "] " + match.getDocument().getValue(0));
```
jaylett: This might be slightly neater broken onto multiple lines.

ojwb:
We actually already have the licence declaration (as part of Aakash's GSoC 2016 proposal):

> Licensing contributions to Xapian
> I will dual-license all of my contributions to Xapian under the GNU GPL version 2 (and all later versions) and the MIT/X licence.

I believe the patch here has only not been applied yet because we wanted to make the corresponding changes to all the iterators together.

PR #88, opened 2016-04-02
Change response feeds to support opensearch 1.1

nirmalsinghania2008:
1)Change response feeds to support opensearch 1.1
http://www.opensearch.org/Specifications/OpenSearch/1.1#Examples_of_OpenSearch_responses

RSS-based search results in Omega previously used OpenSearch 1.0.
This patch adds the support for newer OpenSearch 1.1 standard.

The changes 1.0 -> 1.1 mostly include 
1)change of XML Namespaces URI
There are no changes in the tags and attributes used, as those are compatible with OpenSearch 1.1 standard.

Issue:
Currently Omega is using query.html for its search interface.
One possibility can be to add OpenSearch Description document replacing the same(query.html).(Devs have to check if this is feasible or not)
That could be done in separate project since there is some more work involved.

### DIFF
--- a/xapian-applications/omega/templates/opensearch
+++ b/xapian-applications/omega/templates/opensearch
@@ -1,5 +1,5 @@
-$httpheader{Content-Type,application/xml}<?xml version="1.0"?>
-  <rss version="2.0" xmlns:openSearch="http://a9.com/-/spec/opensearchrss/1.0/">
+$httpheader{Content-Type,application/xml}<?xml version="1.0" encoding="UTF-8"?>
+  <rss version="2.0" xmlns:openSearch="http://a9.com/-/spec/opensearch/1.1/">
     <channel>
       <title>Xapian search results</title>
       <link>$html{$env{SCRIPT_NAME}?DB=$url{$dbname}&P=$url{$query}}</link>


ojwb:
> Previously i've made a large commit containing several sub patches.

And this is another large commit intermingling three separate (though related) change - you've even identified them (1, 2 and 3 above - each of those ought to be a commit).

nirmalsinghania2008:
@ojwb 1 and 2 are part of same bite size project. and 3 is just change of namespace.
since changing the case of HTML tags,attributes doesn't achieve much so I've added 2.I think this whole should be a patch as the main work we are doing here is in "same" files and with same objective for improving obsolete code of omega templates based on new HTML standard.don't you think so?

jaylett:
@nirmalsinghania2008 a bite-sized project is probably going to be one pull request, but that doesn't mean one commit makes sense. Additionally, the commit messages don't tell me what you're trying to achieve with each, which makes it difficult to review. One commit should cover the conversion to html5 (and clearly state that, including notes about what you've done in the commit message), and one adding classes. A third commit then could cover the OpenSearch work.

However as I think I've suggested before, for your first review I'd go for a small patch. I'd suggest you just get the OpenSearch work into a PR with one well-formed commit. (Otherwise we might be ready to merge the OpenSearch work, but be unable to merge anything if we're still working on improving other commits in the PR.)

Note that you'll need to rebase against master now (there should be a "This branch has conflicts…" message on the PR screen for you). In particular, master now contains lowercasing from someone else (which was merged faster because it did less!), which simplifies what your patches will need to cover.

nirmalsinghania2008:
@jaylett @ojwb Please check the changes!

xapian-applications/omega/templates/opensearch
```diff
@@ -1,5 +1,5 @@
-$httpheader{Content-Type,application/xml}<?xml version="1.0"?>
-  <rss version="2.0" xmlns:openSearch="http://a9.com/-/spec/opensearchrss/1.0/">
+$httpheader{Content-Type,application/xml}<?xml version="1.0" encoding="UTF-8"?>
+  <rss version="2.0" xmlns:openSearch="http://a9.com/-/spec/opensearch/1.1/" xmlns:atom="http://www.w3.org/2005/Atom">
```
jaylett: You don't seem to use the `atom` XMLNS, so there's no need to declare it.

jaylett:
@nirmalsinghania2008 it may be worth having in the commit message a brief note stating that the changes 1.0 -> 1.1 are mostly not for elements or attributes we use (since I assume that's why the changes are minimal).

nirmalsinghania2008:
@jaylett done!

jaylett:
@nirmalsinghania2008 great! If you can squash the changes, and make the suggested alteration to the commit message at the same time (see [this article](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history) for more information), I think that's ready.

nirmalsinghania2008:
@jaylett I squashed and amended the commit message.please check!

jaylett:
@nirmalsinghania2008 you haven't done what I asked on the commit message at all; and that commit message is no longer a particularly good description of the change. What I suggest you do is to read back through our [developer guide](https://xapian-developer-guide.readthedocs.org/en/latest/) and write a new commit message based on what's in there. (I've added some new things today.)

I specifically want to see something in the commit message about why the changes are so small, which I think will help people looking at this commit in future have confidence that it is a well-formed change, without having to check the OpenSearch specification themselves.

nirmalsinghania2008:
@jaylett I've Added the description to the commit message and restored the previous commit title.
Is it right now?

jaylett:
@nirmalsinghania2008 that looks good — I'll get that queued up so it can be merged. Thanks!

nirmalsinghania2008:
@jaylett :)

nirmalsinghania2008:
@jaylett :+1:

PR #90, opened 2016-04-04
Update Omega templates to html5

nirmalsinghania2008:
This patch contributes to the improvement of omega templates.
Each of the Improvement is a "different" commit .
The description is also given in the commit messages

1)Double Quote Attribute Values
2)lowercase HTML tags,attributes and attribute values
3)Remove markup errors(Closing Tags,remove extra quotes)
4)Add HTML5 Doctype
5)Follow HTML5 validation rules 
     Used HTML5 validation to find some errors
    such as 
1) border attribute not allowed in <input> tag
2)language attribute of script tag is obsolete and no longer required.

### DIFF
--- a/xapian-applications/omega/templates/emptydocs
+++ b/xapian-applications/omega/templates/emptydocs
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <h1>Documents with no text</h1>
 <ul>
 $map{$emptydocs{$cgi{TERM}},
--- a/xapian-applications/omega/templates/emptydocs
+++ b/xapian-applications/omega/templates/emptydocs
@@ -1,7 +1,14 @@
 <!DOCTYPE html>
+<html>
+<head>
+<title>Documents with no text</title>
+</head>
+<body>
 <h1>Documents with no text</h1>
 <ul>
 $map{$emptydocs{$cgi{TERM}},
 <li> <a href="$html{$field{url,$_}}">$html{$or{$field{title,$_},$prettyurl{$field{url,$_}}}}</a>
 }
 </ul>
+</body>
+</html>
--- a/xapian-applications/omega/templates/inc/anyallradio
+++ b/xapian-applications/omega/templates/inc/anyallradio
@@ -1,2 +1,2 @@
-<input type="radio" name="DEFAULTOP" value="or" $if{$eq{$defaultop,or},CHECKED}> Matching any words
-<input type="radio" name="DEFAULTOP" value="and" $if{$eq{$defaultop,and},CHECKED}> Matching all words
+<input type="radio" name="DEFAULTOP" value="or" $if{$eq{$defaultop,or},checked}> Matching any words
+<input type="radio" name="DEFAULTOP" value="and" $if{$eq{$defaultop,and},checked}> Matching all words
--- a/xapian-applications/omega/templates/inc/anyallradio
+++ b/xapian-applications/omega/templates/inc/anyallradio
@@ -1,2 +1,2 @@
-<input type=radio name=DEFAULTOP value=or $if{$eq{$defaultop,or},CHECKED}> Matching any words
-<input type=radio name=DEFAULTOP value=and $if{$eq{$defaultop,and},CHECKED}> Matching all words
+<input type="radio" name="DEFAULTOP" value="or" $if{$eq{$defaultop,or},CHECKED}> Matching any words
+<input type="radio" name="DEFAULTOP" value="and" $if{$eq{$defaultop,and},CHECKED}> Matching all words
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -4,31 +4,31 @@ $set{thousand,$.}$set{decimal,.}
 $setmap{prefix,,$split{ S B F},title,S,author,A,topic,B}
 ${
 $def{PREV,
-$if{$ne{$topdoc,0},<INPUT TYPE=image NAME="&lt;" ALT="&lt;"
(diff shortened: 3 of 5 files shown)


xapian-applications/omega/templates/inc/anyallradio
```diff
@@ -1,2 +1,2 @@
-<input type="radio" name="DEFAULTOP" value="or $if{$eq{$defaultop,or}",CHECKED}> Matching any words
-<input type="radio" name="DEFAULTOP" value="and $if{$eq{$defaultop,and}",CHECKED}> Matching all words
+<input type="radio" name="DEFAULTOP" value="or" $if{$eq{$defaultop,or},checked}> Matching any words
```
jaylett: Ideally, moving the quote would have happened in either the quote attribute or fix errors commit.

xapian-applications/omega/templates/emptydocs
```diff
@@ -1,6 +1,11 @@
+<!DOCTYPE html>
+<html>
```
jaylett: You cannot omit the `head` element in html5 if it's empty. It's easiest to put one in with a `title` matching the `h1`.

jaylett:
@nirmalsinghania2008 you need to excise the first commit, which we'll merge separately. (Sorry it takes a while to do this; we have a private git repo behind the github one, and there are currently some minor problems which slow things down.)

I think this PR overall should be called 'Update Omega templates to html5', with the last commit broken into two: the doctype as one, and then validation changes as another. (Have you run the output through an html5 validator? It's difficult to do with a dynamic template, but checking the variants and running them through will probably throw up some other minor problems if you haven't already.)

The bodies of your commit messages are wrapping; they should be 70 or so characters maximum.

nirmalsinghania2008:
@jaylett my semester exams have started! i'll fix it soon.
is it okay?
also what do you mean by excise the first commit? i didn't get it

jaylett:
@nirmalsinghania2008 focus on your exams, and good luck!

Excise = remove. That commit shouldn't be in this PR.

nirmalsinghania2008:
@jaylett removed the first commit. and squashed the moving of quotes in first commit.
also divided last commit in two parts.
please check!

xapian-applications/omega/templates/emptydocs
```diff
@@ -1,7 +1,14 @@
 <!DOCTYPE html>
+<html>
```
jaylett: It's encouraged to include a `lang` attribute; in the case of Omega `en` or `en-gb` would be appropriate, because we have some fixed strings in English.

xapian-applications/omega/templates/query
```diff
@@ -4,31 +4,31 @@ $set{thousand,$.}$set{decimal,.}
...
-$if{$ne{$last,$msize},<input type=image name="&gt;" alt="&gt;"
-src="/icons/omega/next.png" border=0 height=30 width=30>,
-<img alt="" src="/icons/omega/nextoff.png" height=30 width=30>}
+$if{$ne{$last,$msize},<input type="image" name="&gt;" alt="&gt;"
+src="/icons/omega/next.png" border="0" height="30" width="30">,
+<img alt="" src="/icons/omega/nextoff.png" height="30" width="30">}
 }
 
-$def{P,<input type=image name="$1" value="$1" src="/icons/omega/page-$2.png" border=0$opt{a} alt="$1">}
+$def{P,<input type="image" name="$1" value="$1" src="/icons/omega/page-$2.png" border="$opt{a}" alt="$1">}
```
jaylett: You've misunderstood what `$opt{a}` is doing here — it's additional attributes, set (confusingly) further down in the document, at line 116. Quoting like this will result in something like `border=""` or `border=" HEIGHT=60 WIDTH=30"` or similar, neither of which is correct. You've made a similar mistake at line 100.

It's difficult to see how we can write automated tests that would catch this (although the one at line 100 would be easier). So what I'd suggest is that you write a manual test script which explains what to visually check for, ensuring that the different code paths are taken. (We might be able to automate some or all of this in future, with some cleverness.) Right now the changes are too many and too invasive to guarantee that code review is going to pick up any potential breakages, so I think we really need the manual test script for this.

jaylett:
@nirmalsinghania2008 given the difficulties with the quoting commit, I'd suggest moving the lowercasing and markup errors into their own PR, since they're ready to merge. We could then proceed to the html5 commits, since they have only a small change remaining to get them ready. Finally we could return to quoting. (Right now, the difficulties of verifying the quoting changes are going to hold up all the other changes from being merged. This is why we recommend doing as small changes as possible.)

nirmalsinghania2008:
@jaylett do you know some good method to move commits from one branch to another.

one thing i can think of is `git cherry-pick` but it is causing conflicts.
can you suggest something better based on your experience .

jaylett:
@nirmalsinghania2008 `cherry-pick` is the correct approach. If there are conflicts, then either those commits are dependent on earlier commits in your PR or your PR wouldn't merge (which it would, right now). So you're probably just going to have to fix the merge problems by hand.

jaylett:
@nirmalsinghania2008 if you either rebase this onto master and remove the commits that were in #92, or just cherry-pick the remaining commits on the top of master and then force push to the branch behind this PR, I can review the html5 changes now. Sorry it took a few days to get this done!

nirmalsinghania2008:
sure!

nirmalsinghania2008:
@jaylett i have created a PR for HTML5 changes. #96 
should i now create a PR(update this) for double quote changes?

jaylett:
@nirmalsinghania2008 you probably want to close this PR. If you have more work (such as the quoting changes) then opening a new PR from latest master will be much easier.

PR #91, opened 2016-04-05
Update subdirectory structure in xapian-letor

ayshtmr:
Discussion thread:
https://lists.xapian.org/pipermail/xapian-devel/2016-April/002859.html
https://lists.xapian.org/pipermail/xapian-devel/2016-April/002860.html

Through this patch, the intention is to restructure xapian-letor in a way similar to how xapian-core has been written. This change enables further pieces of work such as including the test harness and writing automated tests for the API.

Following changes were made:
1. Create subdirectories api/, ranker/, include/xapian-letor/, and move appropriate files to them
2. Create xapian-letor.h in include/ 
3. Make corresponding changes to configure.ac and Makefile.am


xapian-letor/ranker/svmranker.cc
```diff
@@ -41,8 +41,8 @@ namespace Xapian {
 
 class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
 
-    string model = null;
-    double[] weight;
+    string model = NULL;
+    double* weight;
```
jaylett: These changes should be in a separate commit, because they aren't related to moving the code around.

xapian-letor/ranker/Makefile.mk
```diff
@@ -0,0 +1,8 @@
+noinst_HEADERS +=\
+	ranker/svmranker.h
+
+EXTRA_DIST +=\
+	weight/Makefile
```
jaylett: Was this left over from copying something similar out of `xapian-core`?

xapian-letor/include/xapian-letor.h
```diff
@@ -0,0 +1,39 @@
...
+#include <xapian-letor/letor.h>
+#include <xapian-letor/letor_features.h>
+#include <xapian-letor/letor_internal.h>
+#include <xapian-letor/ranker.h>
+#include <xapian-letor/ranklist.h>
+
+#include <string>
+#include <map>
+
+#endif /* XAPIAN_INCLUDED_XAPIAN_LETOR_H */
```
jaylett: In github you should be able to see a small pair of red icons, indicating there's no newline at the end of the file. It's better to add one; it makes trailing diffs easier to read in future.

xapian-letor/configure.ac
```diff
@@ -34,9 +34,14 @@ dnl 0:0:0 1.3.1 First release with xapian-letor split out from xapian-core
...
 
+dnl Where xapian-letor.h, etc go.  In development release append "/xapian-1.3".
+dnl incdir=$includedir
+incdir=$includedir/xapian-1.3
+AC_SUBST([incdir])
+
 AM_INIT_AUTOMAKE([1.10.1 -Wportability tar-ustar])
 
-AC_CONFIG_SRCDIR([letor.cc])
+AC_CONFIG_SRCDIR([api/letor.cc])
```
jaylett: This is probably required in the earlier restructuring commit, or configure won't generate.

xapian-letor/configure.ac
```diff
@@ -34,9 +34,14 @@ dnl 0:0:0 1.3.1 First release with xapian-letor split out from xapian-core
 LIBRARY_VERSION_INFO=0:0:0
 AC_SUBST(LIBRARY_VERSION_INFO)
 
+dnl Where xapian-letor.h, etc go.  In development release append "/xapian-1.3".
+dnl incdir=$includedir
+incdir=$includedir/xapian-1.3
+AC_SUBST([incdir])
```
jaylett: The commit message talks about an install path, but this is really just about the include directory so it'd be good to adjust the message to match.

jaylett:
@ayshtmr that's beginning to look a lot more structured — nice! :-)

Your first four commits are merge commits that don't seem to introduce anything. You should be able to rebuild the branch without them, which will make things a little clearer.

You'll need to squash some of these commits together. For instance, having one commit that updates the `Makefile`, another which moves the files and then one that introduces the new core public API header is a problem, because it means there's a couple of states in the middle where the code can't possibly build. (This causes problems if someone's using `git bisect` in future to figure out where a bug was introduced, for instance.) So you really want a single commit that does that entire update in one go.

I think the next block of work needs to focus on getting the headers in better shape. Some care needs to be taken in distinguishing between headers that are part of the public API (which should be in `include/xapian-letor`) and those that are just part of the build system (internal headers that include implementation-private details). `xapian-core` is a good model here. We can get this merged before looking at that, though.

ayshtmr:
@jaylett I made the changes you suggested. Please review.

You are right to point out that headers need to be in better shape, but since many headers in current master function differently when compared to the ones in v-hasu's gsoc2014 branch, would it be better to get to that task while in the process of introducing code from his branch?

As the next block of work, how do you think placing the test harness and writing a few tests would go? Since the way featurevectors are being computed have been the same in current master and v-hasu's branch, it should be good to begin writing some tests for that.

jaylett:
@ayshtmr you're probably right that separating the headers (and ensuring the correct use of PIMPL &c) is best done when introducing functionality from v-hasu's branch. However if there's anything that can be tested, then presumably it has a reasonably stable public API to test with, so perhaps you'll want to tidy up the feature vector work following writing the initial tests.

ayshtmr:
@jaylett I get your point. Will rework the public API headers first and then proceed to write tests.

jaylett:
@ayshtmr it'd be good to also go through all the code that's on master so far and fix any compile warnings. (Certainly on clang I have to use `CXXFLAGS=-Wno-error` to get it to build at all. There are only two or three errors, so it shouldn't take long.)

At some point (whether you do this now or later), we need to tidy up the whitespace, get rid of 'dead' commented code, and generally bring the code quality up to the standard of xapian-core.

I'll merge this PR in a minute, so you should be able to work from master shortly.

PR #95, opened 2016-04-08
$match function for omegascript

richhiey1996:
Added the matching function in transform.cc and hooked it up in query.cc
Do let me know whether this is good enough and how I can improve.

### DIFF
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -337,6 +337,19 @@ $map{LIST,STUFF)
 	versions). If the tabs are a problem, use $list{$map{...},}
 	to get rid of them.
 
+$match{REGEX,STRING[,OPTIONS]}
+	peform a regex match using Perl-compatible regular expressions. Returns
+	true if a match is found, else it returns an empty string.
+
+	The optional OPTIONS argument can contain zero or more of the letters
+	``imsx``, which have the same meanings as the corresponding Perl regexp
+	modifiers:
+
+	* ``i`` - make the pattern matching case-insensitive
+	* ``m`` - make ``^``/``$`` match after/before embedded newlines
+	* ``s`` - allows ``.`` in the pattern to match a linefeed
+	* ``x`` - allow whitespace and ``#``-comments in the pattern
+
 $msize
 	estimated number of matches.
 
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -204,6 +204,22 @@ printf '$contains{$cgi{a},$cgi{b}}' > "$TEST_TEMPLATE"
 testcase '6' P=text a=fish b="Hello fish"
 testcase '' P=text a="Example" b="random"
 
+NEWLINE='
+'
+
+# Feature tests for $match.
+printf '$match{$cgi{a},$cgi{b},$cgi{c}}' > "$TEST_TEMPLATE"
+testcase 'true' P=text a="ab*c+" b="ac"
+testcase '' P=text a="acb" b="abc"
+testcase 'true' P=text a="[A-Z]bcD" b="abcd" c="i"
+testcase '' P=text a="[A-Z]bcD" b="abcd"
+testcase 'true' P=text a="^abc$" b="${NEWLINE}abc${NEWLINE}def" c="m"
+testcase '' P=text a="^abc$" b="${NEWLINE}abc${NEWLINE}def"
+testcase 'true' P=text a="abc." b="abc${NEWLINE}" c="s"
+testcase '' P=text a="abc." b="abc${NEWLINE}"
+testcase 'true' P=text a="    ABC #test_comment " b="ABC" c="x"
+testcase '' P=text a="    ABC #test_comment " b="ABC"
+
 # Feature tests for $terms.
 printf 'text : index\nhost : boolean=H\nfoo : boolean=XFOO' > "$TEST_INDEXSCRIPT"
 rm -f "$TEST_DB"
--- a/xapian-applications/omega/query.cc
+++ b/xapian-applications/omega/query.cc
@@ -928,6 +928,7 @@ CMD_lookup,
 CMD_lower,
 CMD_lt,
 CMD_map,
+CMD_match,
 CMD_max,
 CMD_min,
 CMD_mod,
@@ -1055,6 +1056,7 @@ T(lookup,	   2, 2, N, 0), // lookup in named cdb file
 T(lower,	   1, 1, N, 0), // convert string to lower case
 T(lt,		   2, 2, N, 0), // test <
 T(map,		   1, 2, 1, 0), // map a list into another list
+T(match,	   2, 3, N, 0), // regex match
 T(max,		   1, N, N, 0), // maximum of a list of values
 T(min,		   1, N, N, 0), // minimum of a list of values
 T(mod,		   2, 2, N, 0), // integer modulus
@@ -1765,6 +1767,9 @@ eval(const string &fmt, const vector<string> &param)
 		    }
 		}
 	        break;
+	    case CMD_match:
+		omegascript_match(value, args);
+		break;
 	    case CMD_max: {
 		vector<string>::const_iterator i = args.begin();
 		int val = string_to_int(*i++);
--- a/xapian-applications/omega/transform.cc
+++ b/xapian-applications/omega/transform.cc
@@ -55,6 +55,43 @@ get_re(const string & pattern, int options)
 }
 
 void
+omegascript_match(string & value, const vector<string> & args)
+{
+    int offsets[30];
+    int options=0;
+    if (args.size() > 2) {
+       const string &opts = args[2];
+       for (string::const_iterator i = opts.begin(); i != opts.end(); ++i) {
+	    switch (*i) {
+	        case 'i':
+		    options |= PCRE_CASELESS;
+		    break;
+		case 'm':
+		    options |= PCRE_MULTILINE;
+		    break;
+	        case 's':
+		    options |= PCRE_DOTALL;
+		    break;
+	        case 'x':
+		    options |= PCRE_EXTENDED;
+		    break;
+	        default: {
+		    string m = "Unknown $match option character: ";
+		    m += *i;
+		    throw m;
+	        }
+	     }
+        }
+    }
+    pcre * re = get_re(args[0], options);
+    int matches = pcre_exec(re, NULL, args[1].data(), args[1].size(),
+			    0, 0, offsets, 30);
+    if (matches > 0) {
+	value += "true";
+    }
+}
+
+void
 omegascript_transform(string & value, const vector<string> & args)
 {
     int offsets[30];
--- a/xapian-applications/omega/transform.h
+++ b/xapian-applications/omega/transform.h
@@ -27,4 +27,7 @@
 void omegascript_transform(std::string & value,
 			   const std::vector<std::string> & args);
 
+void omegascript_match(std::string &value,
+			   const std::vector<std::string> & args);
+
 #endif // OMEGA_INCLUDED_TRANSFORM_H


richhiey1996:
Thanks for closing the previous pull request. Sorry for the inconvenience with that!
Is there any way I can improve this patch?
Also, most of these above commits were due to indentation issues that came up after I pushed the code.
The spacing in my text editor and the spacing as seen when the code is online, are different in many ways.
Is there any way I could get around that?

jaylett:
@richhiey1996 there's a script `xapian-maintainer-tools/xapian-check-patch` which you can run over diffs to catch some indentation issues. (eg: `git diff origin/master | ./xapian-maintainer-tools/xapian-check-path` from the top of the source tree).

Once you've corrected this, it's probably best if you squash all the commits into one, then I can review the whole thing in one go. In future, you'll find `git commit --fixup <sha>` to make a commit that 'fixes up' a previous one makes life easier; then you can do `git rebase -i --autosquash master` and the commits will be marked as fixup in the interactive rebase (assuming you branched from your local master).

richhiey1996:
Thanks for closing the previous pull request. Sorry for the inconvenience with that!
Is there any way I can improve this patch?
Also, most of these above commits were due to indentation issues that came up after I pushed the code.
The spacing in my text editor and the spacing as seen when the code is online, are different in many ways.
Is there any way I could get around that?

richhiey1996:
I've squashed all of the above commits into a single commit. I hope it'll be easier to review it now.
This helped me learn use git well. Thanks for the pointers :)

xapian-applications/omega/query.cc
```diff
@@ -1765,6 +1767,9 @@ eval(const string &fmt, const vector<string> &param)
 		    }
 		}
 	        break;
+	    case CMD_match:
+		omegascript_match(value,args);
```
jaylett: Space after the comma, please.

jaylett:
@richhiey1996 you also need to add documentation (in `xapian-applications/omega/docs/omegascript.rst`) for your new command.

richhiey1996:
I've added the documentation and rectified the spacing mistake in query,cc.
Should I add documentation for the $match command in the omegascript documentation page on the website?

jaylett:
@richhiey1996 that page on the website is built from `xapian-applications/omega/docs/omegascript.rst`, so you're covered.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -337,6 +337,19 @@ $map{LIST,STUFF)
 	versions). If the tabs are a problem, use $list{$map{...},}
 	to get rid of them.
 
+$match{REGEX,STRING[,OPTIONS]}
+	peform a regex match using Perl-compatible regular expressions. This
+	command will return true if a match is found
+
+		The optional OPTIONS argument is supported by Omega 1.3.4 and later.
```
jaylett: This should align with the rest of the paragraph.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -337,6 +337,19 @@ $map{LIST,STUFF)
 	versions). If the tabs are a problem, use $list{$map{...},}
 	to get rid of them.
 
+$match{REGEX,STRING[,OPTIONS]}
+	peform a regex match using Perl-compatible regular expressions. This
+	command will return true if a match is found
```
jaylett: You need to say what happens if this doesn't match, but also I'd get rid of 'this command will…'. Something like:

> Return true if matches using Perl-compatible regular expressions. Else return an empty string.

richhiey1996:
Changed the $match command description and fixed the above stated alignment problem.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -341,14 +341,14 @@ $match{REGEX,STRING[,OPTIONS]}
...
-     	 * ``s`` - allows ``.`` in the pattern to match a linefeed
-     	 * ``x`` - allow whitespace and ``#``-comments in the pattern
+	The optional OPTIONS argument is supported by Omega 1.3.4 and later.
+    It can contain zero or more of the letters ``imsx``, which have the
+    same meanings as the corresponding Perl regexp modifiers:
+
+    * ``i`` - make the pattern matching case-insensitive
+    * ``m`` - make ``^``/``$`` match after/before embedded newlines
+    * ``s`` - allows ``.`` in the pattern to match a linefeed
+    * ``x`` - allow whitespace and ``#``-comments in the pattern
```
jaylett: This alignment is still wrong; I think all the lines should be indented by one tab. Can you run the `xapian-maintainer-tools/xapian-check-patch` script over your patch and correct anything that complains about, as well please?

jaylett:
It's also possible to write tests for omegascript functions (sorry, should have remembered that sooner). Have a look at `xapian-applications/omega/omegatest` to see how it works; it'd be good to have a couple of tests for normal behaviour of `$match{}`.

xapian-applications/omega/docs/omegascript.rst
richhiey1996: I'm sorry, just having problems with the indentation since yesterday. Will use the maintainer tool from now on before commits.

richhiey1996:
@jaylett Hopefully I've corrected the alignment problems now in the new commit. I ran xapian-check-patch and it didn't return any warnings.
Is there anything more I need to work on in this patch?
I was also looking at adding tests for $match (as you above mentioned) and a few more commands if possible. But I have exams going on so I'll try to contribute as much as I can to the testing part in the time I have.

jaylett:
@richhiey1996 I'll hold off on merging this to give you a chance to get through exams before writing the tests. It's always better to merge with tests, but we can wait a little to get this in.

richhiey1996:
Hello,
I just got back to completing this merge request.
I looked into omegatest and didn't really understand how to add more tests. Could you just give me a rough idea on how to do this?

jaylett:
@richhiey1996 if you skip to line 74 (ish) you'll see a lot of test cases, written something like:

```bash
qtestcase 'Zsimpl@1' P=simple
```

That one (which is the first in the file) means to run omega with `P=simple`, ie a query string of "simple", and check that it is parsed as `Zsimpl@1`, ie the term "Zsimpl" (which is "simple" stemmed and then stored with the prefix 'Z', which is used for stemmed terms) at position 1 in the query. Which is what we'd expect.

For what you're testing, you don't need to run queries, so you can use `testcase` instead of `qtestcase`. The clearest examples of this are around line 204, the feature tests for `$contains`. Like them, you'll need to overwrite the test template with something using `$match{}` and then run it with various different CGI parameters (which become things like `a=fish` in each `testcase` line) to test the various possible behaviours of `$match{}`.

richhiey1996:
Thanks for the pointers.
I realized this later after going through omega.cc. Helped a lot!
I've added a few test cases to omegatest and they seem to work fine. Should these test cases be enough or are there any more tests that we can add?

jaylett:
@richhiey1996 you don't need to test regular expressions, just that they are regular expressions. So I'd aim for a small number of tests:

 * regex match
 * regex non-match
 * non-regex (ie what's the error handling like)
 * test the options as well (including that you can set multiple options if they all make sense)

richhiey1996:
I've added tests for a regex match, a non regex match and a few cases using the options (imsx) that are provided with the function.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -337,6 +337,19 @@ $map{LIST,STUFF)
 	versions). If the tabs are a problem, use $list{$map{...},}
 	to get rid of them.
 
+$match{REGEX,STRING[,OPTIONS]}
+	peform a regex match using Perl-compatible regular expressions. Returns
+	true if a match is found, else it returns an empty string.
+
+	The optional OPTIONS argument is supported by Omega 1.3.4 and later.
```
jaylett: Only just noticed — this isn't true, because we're introducing `OPTIONS` as we introduce the command. Instead, just say "The optional OPTIONS argument can contain zero or more…".

xapian-applications/omega/omegatest
```diff
@@ -205,14 +205,13 @@ testcase '6' P=text a=fish b="Hello fish"
...
-printf '$match{$cgi{a},$cgi{b}}' > "$TEST_TEMPLATE"
-testcase 'true' P=text a="abcde" b="abcde"
+printf '$match{$cgi{a},$cgi{b},$cgi{c}}' > "$TEST_TEMPLATE"
 testcase 'true' P=text a="ab*bc" b="abbbbbbc"
 testcase '' P=text a="a(^bc)d" b="azd"
-testcase 'true' P=text a="(.*)c(.*)" b="abcde"
-testcase 'true' P=text a="(bc+d$|ef*g.|h?i(j|k))" b="bcccd"
-testcase 'true' P=text a="(?<!\\?)'(.*?)(?<!\\?)'" b="'ab?'cd'efg"
-testcase 'true' P=text a="^([ab]*?)(?!(b))c" b="ababc"
+testcase 'true' P=text a="abc.*" b="ABCab\n" c="i" c="m"
```
jaylett: This isn't how `$cgi{}` works; it will pick only one of those "arbitrarily". You actually want `c="im"`. You also aren't actually testing the `m` option here; you probably want something like:

```
testcase 'true' P=text a="abc$" b="ABC\nAB" c="im"
```

to test that.

xapian-applications/omega/omegatest
```diff
@@ -205,14 +205,13 @@ testcase '6' P=text a=fish b="Hello fish"
...
-testcase 'true' P=text a="abcde" b="abcde"
+printf '$match{$cgi{a},$cgi{b},$cgi{c}}' > "$TEST_TEMPLATE"
 testcase 'true' P=text a="ab*bc" b="abbbbbbc"
 testcase '' P=text a="a(^bc)d" b="azd"
-testcase 'true' P=text a="(.*)c(.*)" b="abcde"
-testcase 'true' P=text a="(bc+d$|ef*g.|h?i(j|k))" b="bcccd"
-testcase 'true' P=text a="(?<!\\?)'(.*?)(?<!\\?)'" b="'ab?'cd'efg"
-testcase 'true' P=text a="^([ab]*?)(?!(b))c" b="ababc"
+testcase 'true' P=text a="abc.*" b="ABCab\n" c="i" c="m" 
+testcase 'true' P=text a=".*$" b="a\n" c="m"
```
jaylett: This doesn't test the `m` option either, because `.*$` should match `a\n` whether you enable multiline or not.

richhiey1996: I think this is an older commit. All these problems were rectified in the newer commit.

xapian-applications/omega/omegatest
```diff
@@ -205,14 +205,13 @@ testcase '6' P=text a=fish b="Hello fish"
...
+testcase 'true' P=text a="^.*$" b="abc\n" c="s"
```
jaylett: Again, this passes even if you turn off `s`.

xapian-applications/omega/omegatest
```diff
@@ -205,14 +205,13 @@ testcase '6' P=text a=fish b="Hello fish"
...
+testcase 'true' P=text a="   abc#  " b="abc\n" c="x" c="m" c="i"
```
jaylett: Easiest to just test `x` alone here, I suspect.

richhiey1996: I think this is a slightly old commit.

jaylett:
@richhiey1996 you need to slow down a little and think and plan. You're still writing tests that pass either way, and you're still writing tests that don't actually check the things you think they're checking.

A large part of this is because you aren't actually generating `b` with embedded newlines. There's a good [explanation of why](http://stackoverflow.com/a/10618140/6302) on Stack Overflow. However I'm not entirely certain that you can make this work by passing things around through CGI parameters (because of how CGI works, or maybe omega's CLI CGI-like approach), so you may have to write a more complex template. For instance, you could have a constant string in the template, with embedded newlines, and just vary `a` (the regex) and `c` (the options) to check the various match behaviours.

Then you can write tests for the options in pairs, checking that they expand to `''` without the relevant options, and `'true'` with them.

richhiey1996:
Yes, you were right about the newlines.
That link was really helpful.
I think going according to the way you suggested is better. It saves me from writing wrong tests because of lack of planning and makes it a lot more planned.

jaylett:
@richhiey1996 it's generally a good idea when you get to something you didn't plan at the outset to stop and plan. And when you run into problems, too. Planning is basically a good thing :-)

richhiey1996:
I've written a few test cases after looking at the previous mistakes. Writing cases for options in pairs was really helpful. The one problem I faced was embedding newlines but I was able to overcome this by manually adding a newline. (This makes the tests look a little messy so I'll try to find a way to get around this)
The tests added are:
1) Checks for a simple regex match.
2) It checks for a regex non-match. 
3) Performs a caseless match using the option 'i'
4) Performs multiline regex match. When using ^ and $, a match is found immediately before or after a newline character, or at start or end. The pair without the modifier returns true as expected
5) If this modifier is set, a dot metacharacter in the pattern matches all characters, including newlines. In this test, (.)* matches '\nefgh\n', which it wouldn't do by default.
6) The extended modifier is tested by adding spaces and using space inside a character set for a match

xapian-applications/omega/omegatest
```diff
@@ -206,12 +206,27 @@ testcase '' P=text a="Example" b="random"
...
+abc
+def"
+testcase 'true' P=text a="ab(.)*cd" b="ab
+efgh
+cd" c="s"
+testcase '' P=text a="ab(.)*cd" b="ab
+efgh
+cd"
+testcase 'true' P=text a="    #test_comment1
+			[\sab] #test_comment2  " b="de " c="x"
```
jaylett: Again, the regex is sufficiently far from the test string that it's difficult to see what's going on. Why not test something much simpler, as you did previously? `a="   ABC# comment" b="ABC"` should fail without the `x` option. You certainly don't need newlines in this testcase.

richhiey1996: Won't it be necessary to check whether a space within a character set matches the input string? Because that's one of the main parts of this modifier.
The newlines weren't necessary though I feel now. I'll remove them out.

jaylett: The `x` modifier is about allowing whitespace within the pattern that doesn't match whitespace in the test string (and allowing comments).

richhiey1996: Oh yes im sorry. A character set would match irrespective. 
Got it. I'll make that change.

richhiey1996:
@jaylett I've added simpler regexes for the options 'x' and 's' as suggested by you and also used a NEWLINE variable to hold a newline character and used that within the strings to embed newlines, so that it looks clearer.

xapian-applications/omega/omegatest
```diff
@@ -204,28 +204,21 @@ printf '$contains{$cgi{a},$cgi{b}}' > "$TEST_TEMPLATE"
...
-efgh
-cd" c="s"
-testcase '' P=text a="ab(.)*cd" b="ab
-efgh
-cd"
-testcase 'true' P=text a="    #test_comment1
-			[\sab] #test_comment2  " b="de " c="x"
-testcase '' P=text a="    #test_comment1
-			[\sab] #test_comment2  " b="de "
+testcase 'true' P=text a="^abc$" b="def""$NEWLINE""abc""$NEWLINE""def" c="m"
```
jaylett: You can use `${NEWLINE}` rather than starting and stopping the strings, which should be more readable. This test also works if you drop the `def` from the beginning.

xapian-applications/omega/omegatest
```diff
@@ -204,28 +204,21 @@ printf '$contains{$cgi{a},$cgi{b}}' > "$TEST_TEMPLATE"
...
+testcase '' P=text a="^abc$" b="def""$NEWLINE""abc""$NEWLINE""def"
+testcase 'true' P=text a="abc." b="abc""$NEWLINE" c="s"
+testcase '' P=text a="abc." b="abc""$NEWLINE"
```
jaylett: Because there's nothing after it, you can just do `b="abc$NEWLINE"` here.

xapian-applications/omega/omegatest
richhiey1996: Thanks for this tip!
It makes it look much better now.

jaylett:
@richhiey1996 great! Can you squash that to a single commit (and write a good overall commit message for this), and rebase onto latest master?

richhiey1996:
@jaylett Fixed the readability issues for NEWLINE with the simple hack you provided.
It looks much better and easy to read now.
Is there anything more I'm missing out on?

richhiey1996:
@jaylett I've rebased it onto latest master. What is it that I have to do next?

jaylett:
@richhiey1996 you need to squash the commits into one.

jaylett:
Merged in b600b38.

PR #96, opened 2016-04-09
Update Omega Template to HTML5 v2

nirmalsinghania2008:
1)Add HTML5 Doctype
2)Remove HTML5 Validation Errors
Subpart of Previous PR #90

### DIFF
--- a/xapian-applications/omega/templates/emptydocs
+++ b/xapian-applications/omega/templates/emptydocs
@@ -1,6 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<title>Documents with no text</title>
+</head>
+<body>
 <h1>Documents with no text</h1>
 <ul>
 $map{$emptydocs{$cgi{TERM}},
 <li> <a href="$html{$field{url,$_}}">$html{$or{$field{title,$_},$prettyurl{$field{url,$_}}}}</a>
 }
 </ul>
+</body>
+</html>
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -1,24 +1,24 @@
-$httpheader{Content-Type,text/html; charset=utf-8}<html>
+$httpheader{Content-Type,text/html; charset=utf-8}<!DOCTYPE html><html lang="en">
 $set{flag_spelling,$ne{$cgi{SPELL},0}}
 $set{thousand,$.}$set{decimal,.}
 $setmap{prefix,,$split{ S B F},title,S,author,A,topic,B}
 ${
 $def{PREV,
 $if{$ne{$topdoc,0},<input type=image name="&lt;" alt="&lt;"
-src="/icons/omega/prev.png" border=0 height=30 width=30>,
+src="/icons/omega/prev.png" height=30 width=30>,
 <img alt="" src="/icons/omega/prevoff.png" height=30 width=30>}
 }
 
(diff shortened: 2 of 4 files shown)


xapian-applications/omega/templates/emptydocs
```diff
@@ -1,6 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<title>Documents with no text</title>
+</head>
+<body
```
jaylett: You need a `>` there.

xapian-applications/omega/templates/query
```diff
@@ -1,24 +1,24 @@
-$httpheader{Content-Type,text/html; charset=utf-8}<html>
+$httpheader{Content-Type,text/html; charset=utf-8}<!DOCTYPE html><html lang="en">
 $set{flag_spelling,$ne{$cgi{SPELL},0}}
 $set{thousand,$.}$set{decimal,.}
 $setmap{prefix,,$split{ S B F},title,S,author,A,topic,B}
 ${
 $def{PREV,
 $if{$ne{$topdoc,0},<input type=image name="&lt;" alt="&lt;"
-src="/icons/omega/prev.png" border=0 height=30 width=30>,
+src="/icons/omega/prev.png" style="border:0px;" height=30 width=30>,
```
jaylett: Although we don't have a standard for this, I'd like a space after the colon in inline CSS please :-)

(Better yet actually might be to put it in a `<style>` block, since it's for all `input[type=image]`.

nirmalsinghania2008:
@jaylett I made the changes.
Actually i was thinking that we should now have css classes and id supported in omega templates.
I think its really unstructured to have inline css,document css all at different places.
I will make a PR for double quotes and css id,classes asap.


Also, i wanted to ask if i should work on some other bite size project(parts of my first PR) since this CSS,HTML related things  are not much related to my GSOC proposal work.
or is it fine?

jaylett:
@nirmalsinghania2008 yes, we should move to using ids and CSS. (Having inline CSS in a `<style>` block is fine for this kind of use, at least initially.) I wouldn't do it in the same PR as double quotes, for the same reasons as before :-)

However as you say, it's probably not a bad idea to work on a different part of the code. Perhaps improve some test coverage, if you can't find anything else?

nirmalsinghania2008:
@jaylett I have implemented $match command and $random,$srandom for omegascript, previously.
but since richhiey is now working on $match, i can put a PR for random number generation.

also for test coverage work,can i get some pointers. :)

jaylett:
@nirmalsinghania2008 random numbers for omegascript would be great.

For test coverage, you need to generate a coverage report. There's information on that in `xapian-core/HACKING` (search for "lcov", I think). Then you're looking for someone with lots of untested lines or branches, and to craft tests that will exercise that code. It's a bit of an art, but generally just keep looking at files that have low coverage until you find something you think you can write a test for.

jaylett:
This PR looks good; can you squash it and ensure it's rebased onto latest master, then I'll merge it :-)

(It's often easier to review changes split into more commits than you actually want in the final history. In this case, I think the overall changes can just be summarised as moving the whole thing to html5, and the changeset is small enough that a single commit will be easier to work with in future.)

nirmalsinghania2008:
@jaylett :+1:

jaylett:
Merged in 213e636.

PR #97, opened 2016-04-10
Changes for Android Build

tdjogi010:
Changes required to cross compile for android.

### DIFF
--- a/xapian-core/common/errno_to_string.cc
+++ b/xapian-core/common/errno_to_string.cc
@@ -49,7 +49,7 @@ errno_to_string(int e, string & s) {
 	s += "Unknown error ";
 	s += str(e);
     }
-#elif HAVE_DECL_SYS_ERRLIST && HAVE_DECL_SYS_NERR
+#elif HAVE_DECL_SYS_ERRLIST && HAVE_DECL_SYS_NERR && !defined(__ANDROID__)
     if (e >= 0 && e < sys_nerr && sys_errlist[e]) {
 	s += sys_errlist[e];
     } else {
--- a/xapian-core/common/errno_to_string.cc
+++ b/xapian-core/common/errno_to_string.cc
@@ -49,7 +49,7 @@ errno_to_string(int e, string & s) {
 	s += "Unknown error ";
 	s += str(e);
     }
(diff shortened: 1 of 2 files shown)


ojwb:
It's odd that the configure probe finds these.  I wonder if they're in a different header to the ones we are including in the code.

If you grep in the ndk headers for `sys_nerr`, which file is it declared in?

tdjogi010:
What we are looking for are sys_nerr and sys_errlist. They can be found in stdio.h located in sysroot/usr/include/ in your standalone ndk toolchain folder and if you are using android-ndk-r10e and toolchain=arm-linux-androideabi-4.9 it can be found defined on line no 238.(if you don't have ndk it is similarly defined here https://android.googlesource.com/platform/development/+/73a5a3b/ndk/platforms/android-20/include/stdio.h#226 )

And I have also checked for their respective symbols in libc.so (for my x86 linux pc counterpart it can be found in libc.so ) using readelf -a and grep but no luck. I have also checked in other libs as well both in standalone toolchain as well in android ndk folder, but couldn't find them.

jaylett:
autoconf is using (deep within the M4 complexity) `ac_fn_cxx_check_decl`, which is compiling a line that I think looks like `(void) sys_nerr;`. That will compile, but not link, in running the test.

It's possible that something a little like [openssh's test for sys_nerr](https://github.com/openssh/openssh-portable/blob/master/configure.ac#L3953) would be more robust for us.

tdjogi010:
OK I will look into it.

ojwb:
> And I have also checked for their respective symbols in libc.so (for my x86 linux pc counterpart it can be found in libc.so ) using readelf -a and grep but no luck. I have also checked in other libs as well both in standalone toolchain as well in android ndk folder, but couldn't find them.

Ah I see - so the fundamental issue is that the headers declare it but there's no actual definition of it in the libraries, so it fails at link time.

Yes, using `AC_LINK_IFELSE` would solve this by actually ensuring that code using `sys_nerr` and `sys_errlist` can be linked.

tdjogi010:
@ojwb @jaylett  I have the required changes to configure.ac. Please review them.

jaylett:
Merged in 96d31c8.

tdjogi010:
Thanks! :)

PR #99, opened 2016-04-16
Add test code for check_if_db_table()

PoeLoren:
Test database table name ending with ".glass"

### DIFF
--- a/xapian-core/HACKING
+++ b/xapian-core/HACKING
@@ -313,7 +313,7 @@ a subdirectory to tell bootstrap to ignore it.
 You will need the following tools installed to build from git:
 
 * GNU m4 >= 4.6 (for autoconf)
-* perl >= 5.6 (for automake; also for various maintainer scripts)
+* perl 5 (for automake; also for various maintainer scripts)
 * python >= 2.3 (for generating the Python bindings)
 * GNU make (or another make which support VPATH for explicit rules)
 * GNU bison (for building SWIG, used for generating the bindings)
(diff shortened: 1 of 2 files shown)


jaylett:
@PoeLoren this really needs to be in its own test case, as if something is wrong with `check_if_db_table()` then we want the test case that fails to clearly point to that function (and ideally to what specifically is tested about it; in this case we're testing its normal behaviour with a correctly-formed table).

It's also worth noting that you've written code to test `Xapian::Document::check()`, rather than `check_if_db_table()` directly, which is what your commit message suggests. (It's actually impractical to directly test that function, so this is the right approach, but your commit message should mirror the code and explain the intent, rather than trying to conflate them with no explanation.)

What I suggest you do is to initially write a list of what behaviours need testing. They will include "returns 0 when testing a correctly-formed database", but also whatever the correct behaviour is when the database is missing, not a recognised Xapian database, has missing files and so forth. You can then figure out for each how you'd go about testing them, whether using a pre-built database via `get_database()` inside the test, or by building suitable files to check other correct behaviour.

PoeLoren:
Thanks for your suggestion. 

I want to know which file is suitable to add this test, api_backend.cc or the other one?

My thinking is as below:
I 'll test the database table name ending with ".glass" or ".DB" . So, I  will create a database first by  `get_writable_database()` , then get the path to the database table. Later, call `Xapian::Database::check(path)`. If check success, the function return 0.

jaylett:
@PoeLoren don't worry too much about which file to put it in.

What you've proposed will test behaviour with a good table file. What about a bad table file? Or a missing one? I suggested you think about these in my previous comment.

You've reverted one of Olly's commits, which you didn't mean to do. I'd start from master again, close this PR and open a new one once you have some code ready for review.

PR #100, opened 2016-04-16
Letor refactor

ayshtmr:
This patch does the following:

1. Move include/xapian-letor/letor_internal.h to api/. Since this header has private implementation details, include isn't the right place for it.
2. Remove evalmetric module since it is non-functional.
3. Use pimpl in featurevector to hide internal class variables
4. <del> Update type of qid and did from std::string to Xapian::docid</del>


xapian-letor/api/featurevector.cc
```diff
@@ -48,14 +48,33 @@ using namespace std;
 
 using namespace Xapian;
 
-FeatureVector::FeatureVector() {
+class FeatureVector::Feature_internal : public Xapian::Internal::intrusive_base
```
jaylett: You can just call this `FeatureVector::Internal` (similar to `Document::Internal`).

xapian-letor/api/featurevector.cc
```diff
@@ -48,14 +48,33 @@ using namespace std;
...
+	int fcount;
+	string did;
+
+  public:
+  	map<string, map<string, int> > load_relevance(const std::string & qrel_file);
+};
+
+FeatureVector::FeatureVector() : internal(new FeatureVector::Feature_internal) { }
+
+FeatureVector::FeatureVector(const FeatureVector & o) : internal(o.internal) { }
```
jaylett: We'd usually format these with the braces on their own lines (see `api/omdocument.cc`, for instance).

jaylett:
I'm not sure that a naive conversion of everything to `::docid` is going to work. Firstly, `qid` isn't a `docid`. Secondly, some thought has to be put into converting from input datasets (such as TREC), which use similar terms and have strings in them. (In the case of docids I think they're effectively Q-terms, or we could have a convention of Q-prefixing them or something.)

Can you separate that out from the rest of the PR (so the other commits can be merged sooner), and maybe at the same time as doing this write some doccomments for more of the members and functions, so it's clear what their purpose is? That'll probably need to be done in relation to the v-hasu branch rather than what's currently in master, to ensure it's how it's actually supposed to be used and not some weird intermediate state that I left it in after stopping the previous merge :-)

ayshtmr:
@jaylett I agree with you on naively updating ::docid. Will remove the commit from current PR. 
The reason I included it was because it was discussed that updating types should be in the next PR. However, apart from this, I couldn't actually find any other code whose type could be updated.

I'm getting started with adding doccomments to members and functions now. Will let you know once I'm done.

jaylett:
@ayshtmr I still think that getting the types right is a high priority, but I think we're now at the point where you're running out of little things you can reasonably get done without starting the bigger work of introducing/testing/documenting chunks of code from the other branch. (And fixing things like types in the meantime.)

If you push a commit which reverts that one, you can push fixup commits for the others until this PR is ready to squash.

ayshtmr:
@jaylett Removed the ::docid commit and made the changes you suggested. Please review and close if its fine now.

Is it better to open up a separate PR for doccomments? Alternatively, along with fixing types, doccomments can also be added in the meantime while introducing code from the other branch. That'll save the effort of cross checking the relation with v-hasu's branch that might be required right now. (As you rightly pointed out, any further patches in this project will make sense only once the process of introducing code from other branch and testing/documenting things has started).

jaylett:
@ayshtmr I'd suggest you create a new branch locally off this one and continue work there. Once this has been merged to master you can then rebase just that onto latest master and continue making PRs from there.

ayshtmr:
@jaylett Could you please merge this PR? I made pimpl changes featuremanager and added some documentation comments for api classes and methods. Would like to open a new PR for that.

ayshtmr:
@jaylett Alternatively, I can close this PR and create a new extended one.

jaylett:
@ayshtmr you need to have a little more patience, or you're going to sometimes get very frustrated — you pushed the last changes to this at the weekend, and so only one working day has gone by since then! I've got half a dozen or so PRs open I'm working with people on at the moment, and I _hope_ to get all of them done today.

jaylett:
@ayshtmr by the way, you never need to close a PR and create a new one that builds on it — you can just push the extra commits onto the original PR. But since PRs should be focussed around something specific, you should be cautious about doing so (also, there's a risk that commits that have already been through a review cycle may be ready but the new ones not).

ayshtmr:
@jaylett I understand. Sorry for my impatience.

jaylett:
@ayshtmr this PR is good on functionality, but has some whitespace issues. If you do `git diff origin/master | ./xapian-maintainer-tools/xapian-check-patch` you'll see a number of issues, and it'd be good to clean them up (particularly killing trailing whitespace). You may discover you have to fix quite a lot of whitespace in -letor, so again it's probably wise to clean as we touch things.

ojwb:
If there are many existing whitespace issues, perhaps we should do a mass-cleanup in a single commit.

Previously we've tended to resist doing that as it made `svn blame` much less useful, and backporting patches harder.  But `git blame -w` can "see through" whitespace changes (within a single line at least), and there's no backporting concerns from letor currently (indeed, it would make sense to clean it up before backporting becomes relevant).

jaylett:
Okay, that's a good point. I'm happy to merge this now then, and @ayshtmr maybe you can do a whitespace cleanup early on, and also tidy up as you introduce new code from the other branch.

jaylett:
Merged in afc2454.

PR #102, opened 2016-04-20
Add test code to Xapian::Database::check()

PoeLoren:
The database table name could end with ".DB"(for chert backend) or ".glass"(for glass backend). Xapian::Database::check() will reach check_if_db_file() by calling check_db_file() to check it. This PR tests two situations: the database not exist and exist.

### DIFF
--- a/xapian-core/tests/api_backend.cc
+++ b/xapian-core/tests/api_backend.cc
@@ -1401,3 +1401,35 @@ DEFINE_TESTCASE(embedded1, singlefile) {
 
     return true;
 }
+
+static string get_db_table_path(const string& db_path){
+    string db_table_path;
+    if(endswith(get_dbtype(), "glass")){
+        db_table_path = db_path + "/postlist.glass";
+    }else if(endswith(get_dbtype(), "chert")){
+        db_table_path = db_path + "/postlist.DB";
+    }
+    return db_table_path;
+}
+
+//check database table name. it could end with ".DB", or ".glass".
+DEFINE_TESTCASE(databasetablename, chert || glass){
+	//database not exist
+    const string & db_path_nonexist = get_named_writable_database_path("databasetablename");
+    string db_table_path = get_db_table_path(db_path_nonexist);
+    try{
+        Xapian::Database::check(db_table_path);
+    }catch(const Xapian::DatabaseOpeningError &){
+    }
+
+	//database exist
+    Xapian::WritableDatabase db = get_named_writable_database("databasetablename", string());
+    Xapian::Document doc;
+    doc.add_term("foo");
+    db.add_document(doc);
+    db.commit();
+    const string & db_path_exist = get_named_writable_database_path("databasetablename");
+    string db_table_path_exist = get_db_table_path(db_path_exist);
+    TEST_EQUAL(Xapian::Database::check(db_table_path), 0);
+    return true;
+}


xapian-core/tests/api_backend.cc
```diff
@@ -1401,3 +1401,35 @@ DEFINE_TESTCASE(embedded1, singlefile) {
...
     return true;
 }
+
+static string get_db_table_path(const string& db_path){
+    string db_table_path;
+    if(endswith(get_dbtype(), "glass")){
+        db_table_path = db_path + "/postlist.glass";
+    }else if(endswith(get_dbtype(), "chert")){
+        db_table_path = db_path + "/postlist.DB";
+    }
```
ojwb: I think you really just want equality tests for the db type - the multi and remote backends won't work with this test.

And something sensible ought to happen if the dbtype is not one that's expected.  Calling `FAIL_TEST` would do.

xapian-core/tests/api_backend.cc
```diff
@@ -1401,3 +1401,35 @@ DEFINE_TESTCASE(embedded1, singlefile) {
...
+//check database table name. it could end with ".DB", or ".glass".
+DEFINE_TESTCASE(databasetablename, chert || glass){
+	//database not exist
+    const string & db_path_nonexist = get_named_writable_database_path("databasetablename");
+    string db_table_path = get_db_table_path(db_path_nonexist);
+    try{
+        Xapian::Database::check(db_table_path);
+    }catch(const Xapian::DatabaseOpeningError &){
+    }
+
```
ojwb: This will also pass if the exception is not thrown.  The `TEST_EXCEPTION` macro provides a standard way to check that an expected exception is definitely thrown.

xapian-core/tests/api_backend.cc
```diff
@@ -1401,3 +1401,35 @@ DEFINE_TESTCASE(embedded1, singlefile) {
...
+	//database exist
+    Xapian::WritableDatabase db = get_named_writable_database("databasetablename", string());
+    Xapian::Document doc;
+    doc.add_term("foo");
+    db.add_document(doc);
+    db.commit();
+    const string & db_path_exist = get_named_writable_database_path("databasetablename");
```
ojwb: By creating a custom database like this, it has to be built every time, and can't be reused for other testcases.  It would be better to just use one of the standard test databases - you should just be able to do:

```
const string & db_path_exist = get_database_path("apitest_simpledata");
```

xapian-core/tests/api_backend.cc
```diff
@@ -1401,3 +1401,35 @@ DEFINE_TESTCASE(embedded1, singlefile) {
 
     return true;
 }
+
+static string get_db_table_path(const string& db_path){
```
ojwb: You not using the Xapian coding style consistently - especially with regards to whitespace which is missing in a lot of places.

There's a script to help with this, which catches the more common issues:

```
git diff origin/master | xapian-maintainer-tools/xapian-check-patch
```

ojwb:
This PR seems to have been abandoned so closing.

I've pushed a couple of commits which cover errors from finding the DB/table and for checking a valid DB and table (d6617dc7e6ba79aa8876dcc87b98576c3f52edb9 and 742b83a64ed5263fabfe29a502320bbb81e1f6be).

PR #103, opened 2016-06-01
Instructions for cross compiling for android

tdjogi010:
Write up on how to cross compile for android.


xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,90 @@
...
+
+2)
+As Xapian makes use of built-in UUID APIs and ndk does not have UUID APIs, we need to explicitly cross compile that as well. So download libuuid, inside the e2fsprogs package by doing following:
+
+git clone git://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git
+cd e2fsprogs/
+
+Now configure and install the library by doing:
+export $TOOLCHAIN=<path_to_ndk_standalone_toolchain>
+export PATH=$PATH:$TOOLCHAIN
```
ojwb: I'm working through this, to try to set it up for the buildbot.

There are no executable files in the `$TOOLCHAIN` directory, so this step doesn't seem useful.  Should this actually be:

```
export PATH=$PATH:$TOOLCHAIN/bin
```

I think then you shouldn't need to set all the variables below.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,90 @@
...
+<path_to_ndk_standalone_toolchain> is the path where your standalone toolchain will be installed.
+
+2)
+As Xapian makes use of built-in UUID APIs and ndk does not have UUID APIs, we need to explicitly cross compile that as well. So download libuuid, inside the e2fsprogs package by doing following:
+
+git clone git://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git
+cd e2fsprogs/
+
+Now configure and install the library by doing:
+export $TOOLCHAIN=<path_to_ndk_standalone_toolchain>
```
ojwb: The `$` here is wrong, should be `export TOOLCHAIN=...`

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,90 @@
...
+export PATH=$PATH:$TOOLCHAIN
+export CC=$TOOLCHAIN/bin/<target-arch>-gcc
+export CXX=$TOOLCHAIN/bin/<target-arch>-g++
+export CPP=$TOOLCHAIN/bin/<target-arch>-cpp
+export LD=$TOOLCHAIN/bin/<target-arch>-ld
+export AR=$TOOLCHAIN/bin/<target-arch>-ar
+export RANLIB=$TOOLCHAIN/bin/<target-arch>-ranlib
+
+./configure --host=<host> --prefix=<path_to_xapian>/xapian-bindings/android/built CFLAGS="-fPIC"
```
ojwb: I'd suggest `--prefix="$TOOLCHAIN/sysroot/usr"` here, which simplifies configuring xapian-core later (because libuuid will be found automatically).

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,90 @@
...
+
+export NDK=<path_to_ndk>
+export $TOOLCHAIN=<path_to_ndk_standalone_toolchain>
+export PATH=$PATH:$TOOLCHAIN
+export CC=$TOOLCHAIN/bin/<target-arch>-gcc
+export CXX=$TOOLCHAIN/bin/<target-arch>-g++
+export CPP=$TOOLCHAIN/bin/<target-arch>-cpp
+export LD=$TOOLCHAIN/bin/<target-arch>-ld
+export AR=$TOOLCHAIN/bin/<target-arch>-ar
+export RANLIB=$TOOLCHAIN/bin/<target-arch>-ranlib
```
ojwb: As above, there's a bogus `$` when setting `TOOLCHAIN`, and `/bin` is needed on the `PATH` entry, then none of these tools need to be explicitly specified.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,90 @@
...
+export OUTPUT_DIR=<path_to_xapian>/xapian-bindings/android/built
+cd <path_to_xapian>/xapian-core
+./configure --prefix=$OUTPUT_DIR CC=$CC CXX=$CXX --host=<host> --with-sysroot=$TOOLCHAIN/sysroot LDFLAGS="-L$OUTPUT_DIR/lib -L$NDK/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi/ -fPIE -pie" --disable-shared CXXFLAGS="-Wl,--no-undefined" --with-pic --disable-backend-chert
```
ojwb: So we don't need to specify `CC` or `CXX` here, as the defaults are fine once `PATH` is set correctly.

`--with-sysroot` doesn't seem to be needed, as the compiler default is correct:

```
bill@mx:~/android/xapian/xapian-core$ arm-linux-androideabi-gcc --print-sysroot
/home/bill/android/toolchain/bin/../sysroot
bill@mx:~/android/xapian/xapian-core$ arm-linux-androideabi-g++ --print-sysroot
/home/bill/android/toolchain/bin/../sysroot
```

The first entry in `LDFLAGS` isn't needed if we install libuuid into the sysroot, as I suggest above.  What's the second entry in `LDFLAGS` needed for?  I omitted it and it seemed to build OK...

What's do we need `-Wl,--no-undefined` for?  I omitted that too without any obvious problem.

And is there a particular reason to recommend people disable the chert backend?

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,90 @@
...
+
+adb push <path_of_file> <dst_on_device>
+
+<dst_on_device> is the destination where you have the read-write-execute permission. eg. /data/local/tmp
+
+After pushing all the files to the device, you can run test as follows:
+
+adb shell
+cd <dst_on_device>
+sh runtest ./apitest
```
ojwb: Will try this later - might be I then discover the reason for some of the options I queried above...

ojwb:
I've successfully set up an automated build based on these instructions:

https://buildbot.xapian.org/builders/xapian%20tar%20android

(Ignore the first failed build - I had failed to set `PATH` and it was trying to build the 1.2 branch not git master).

ojwb:
Cool - I still ought to confirm I can reproduce the testsuite passing on an android device, will try to get to that soon, but I'm new to android development so sorry if it takes me a bit of time.

tdjogi010:
@ojwb I have further added the instructions for building and testing the bindings for android. I have also added the modified SimpleIndex and SimpleSearch example. [Here](https://drive.google.com/open?id=0B0guswJrnvJLUElpSGs1VlZ0QzQ) is an .apk(arm-only) demonstrating the example.

xapian-bindings/android/example/java/SimpleIndex.java
```diff
@@ -0,0 +1,74 @@
...
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED
+ WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+ PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY
+ DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ THE POSSIBILITY OF SUCH DAMAGE.
+ **/
+package com.example.tj.xapiansample;
```
jaylett: Should we have the source files in the same layout as the package namespace? Also, you can use `org.xapian.examples.android` or something similar.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -55,14 +55,50 @@ make install
...
+===================================
+Now generate the bindings for android as follows:
+
+cd <path_to_xapian>/xapian-bindings/android
+mkdir -p org/xapian
+swig -Werror -I. -I../../xapian-core/include -c++ -java -module Xapian -package org.xapian -outdir org/xapian -o xapian_wrap.cc android.i
+
+This should generate xapian_wrap.cc,xapian_wrap.h and .java files in org/xapian. Now add the following line in org/xapian/XapianJNI.java at start of static block at line no 1114:
+
+System.loadLibrary("xapian_jni");
```
jaylett: Can we do this using `%pragma(java)` in SWIG, as described in [the SWIG manual section 25.4.1.1](http://www.swig.org/Doc3.0/Java.html#Java_imclass_pragmas)?

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -55,14 +55,50 @@ make install
...
+The 'source' and 'target' flags are required for android compactibility if you have java 1.8 or above. 
+
+
+Now create a proper file structure for .aar for android by:
+
+mkdir -p aar/jni/<proper_architecture> aar/res
+jar -cf aar/classes.jar org/xapian/*.class
+cp built/lib/libxapian_jni.so aar/jni/<proper_architecture>/
+cd aar
+jar -cf ../built/xapianlib.aar *
```
jaylett: The next stage of this should probably be to turn it all into make rules within the bindings automake system, so people can just type `make` rather than having to cut and paste commands. I'd guess that we should not try to build Android by default, and maybe should require `TOOLCHAIN` (probably as `ANDROID_TOOLCHAIN`) to be passed into the configure system.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -55,14 +55,50 @@ make install
...
+
+Now xapianlib.aar as a library contains libxapian_jni.so and jni bindings ready for use on android devices.
+
 Testing Xapian on android
 =========================
-Now to test your build do the following:
+1)Now to test your cross compiled build(libxapian-1.3.a) do the following:
```
jaylett: It won't be `libxapian-1.3.a` in future (because 1.4.0 is out, and the future development versions will probably be 1.5.x) :-)

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -75,3 +111,9 @@ After pushing all the files to the device, you can run test as follows:
 adb shell
 cd <dst_on_device>
 sh runtest ./apitest
+
+2)Now to test your bindings, create a empty project in your android studio and copy the files in the examples folder to your respective project folder. Add the xapianlib.aar in your project's libs folder and add the following line in your app's build.gradle file:
+
+compile(name:'xapianlib',ext:'aar')
```
jaylett: Can we put the Gradle file, and anything else required to build this, within `xapian-bindings/android` so people don't have to make a new project anywhere for the purpose of testing things? I don't know what other files are required to make it work, but ideally we'd drive that all from `make` instead of having to use Android Studio.

xapian-bindings/android/aar/.gitignore
```diff
@@ -0,0 +1,3 @@
+/jni/
+/res/
+classes.jar
```
jaylett: When you see a little symbol like this in github, it means the file doesn't have an ending newline. It's generally better to have one, where file formats allow it (which .gitignore, XML &c do).

xapian-bindings/android/android.i
```diff
@@ -0,0 +1,401 @@
+%module(directors="1") xapian
```
jaylett: Since this is (apparently) identical to `xapian-bindings/java/java.i`, is there a way we can share the two rather than having a copy? It'll be easier to keep updated in future.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -71,7 +71,7 @@ Save the file. Now generate the libxapian_jni.so by doing the following:
 
 export CXX=$TOOLCHAIN/bin/<host>-g++
 $CXX -c xapian_wrap.cc -I../../xapian-core/include
-$CXX -shared -o built/lib/libxapian_jni.so -Wl,-whole-archive libxapian-1.3.a  -Wl,-no-whole-archive xapian_wrap.o -lz -luuid -lc
+$CXX -shared -o built/lib/libxapian_jni.so -Wl,-whole-archive libxapian-x.x.a  -Wl,-no-whole-archive xapian_wrap.o -lz -luuid -lc
```
ojwb: For release versions (like 1.4.x), the library is just `libxapian.a`.  Only development versions have a suffix, and I think it makes more sense to make these instructions clearer for release versions.

But these steps (and the Java compilation below) seem to do by hand pretty much what's automated by `java/Makefile.am` - have you tried configuring xapian-bindings with something like `--with-java CXX=$TOOLCHAIN/bin/<host>-g++ JAVAC='javac -source 1.7 -target 1.7 -bootclasspath <path_to_jvm>/<your_jdk>/jre/lib/rt.jar'` ?

Was the issue libtool not managing to spot the static libxapian contained PIC code?

xapian-bindings/android/android.i
ojwb: My thought would be to just conditionalise any required differences between "normal" Java and Android in the existing Java bindings.  Otherwise we'll have to make sure any change to Java is also applied to the Android bindings (or isn't appropriate) which is a lot of fiddly patch tracking.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,121 @@
...
+export TOOLCHAIN=<path_to_ndk_standalone_toolchain>
+export PATH=$PATH:$TOOLCHAIN/bin
+
+./configure --host=<host> --prefix="$TOOLCHAIN/sysroot/usr" CFLAGS="-fPIC"
+cd lib/uuid
+make
+make install
+
+<host> is the target architecture. eg. arm-linux-androideabi
+<target-arch> is the same as in step 1)
```
ojwb: `<target-arch>` is no longer used in this step, so this line can go.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,121 @@
...
+
+Save the file. Now generate the libxapian_jni.so by doing the following:
+
+export CXX=$TOOLCHAIN/bin/<host>-g++
+$CXX -c xapian_wrap.cc -I../../xapian-core/include
+$CXX -shared -o built/lib/libxapian_jni.so -Wl,-whole-archive libxapian-x.x.a  -Wl,-no-whole-archive xapian_wrap.o -lz -luuid -lc
+
+This will generate libxapian_jni.so in built/lib/libxapian_jni.so. Now compile all the .java class using :
+
+javac -source 1.7 -target 1.7 -bootclasspath <path_to_jvm>/<your_jdk>/jre/lib/rt.jar org/xapian/*.java
```
ojwb: Having read up a little on this, it seems that the `rt.jar` pointed to by `-bootclasspath` needs to be from the same JRE version you specify to `-source` and `-target`, as that's the whole point of specifying it - the one from "your_jdk" would be used by default.

@tdjogi010 Have I misunderstood here?

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,121 @@
...
+
+<path_to_jvm> is the jvm installation path eg.usr/lib on Ubuntu
+The 'source' and 'target' flags are required for android compactibility if you have java 1.8 or above. 
+
+
+Now create a proper file structure for .aar for android by:
+
+mkdir -p aar/jni/<proper_architecture> aar/res
+jar -cf aar/classes.jar org/xapian/*.class
+cp built/lib/libxapian_jni.so aar/jni/<proper_architecture>/
```
ojwb: What should `<proper_architecture>` be?  You don't seem to say anywhere.

ojwb: Also, http://tools.android.com/tech-docs/new-build-system/aar-format implies there are some mandatory files we aren't creating here - ` AndroidManifest.xml` and`R.txt`.  Should we be generating those somehow?

xapian-bindings/android/CROSS_COMPILE
ojwb: Ooh, we can use that in the java bindings rather than modifying a SWIG-generated file with Perl.  Implemented that in 9f53a0a5761da1fb2f868c5da87cb374a6f5ce84.

xapian-bindings/android/CROSS_COMPILE
ojwb: I managed to run `apitest` on my phone, which is cool.

Many tests pass, but it seems any testcase involving the remote backend fails.  Is that just because programs run from adb don't have permission to use the network, even over localhost?

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,121 @@
...
+
+adb shell
+cd <dst_on_device>
+sh runtest ./apitest
+
+2)Now to test your bindings, set the location of android sdk and android ndk in example/local.properties file and copy the xapianlib.aar in example/app/libs folder. To build the example and install the app on device or emulator run the following:
+
+cd example
+./gradlew build
+./gradlew installDebug
```
ojwb: I'm looking at this step now - it seems there's binary code in `gradle/wrapper/gradle-wrapper.jar` without any corresponding source code, so I'm not at all keen to run the commands above - I don't trust binary code I can't rebuild from source for myself if I want to.  We also don't want to be adding jar files to xapian git.

The Debian `libgradle-core-java` package contains `/usr/share/java/gradle-wrapper.jar` - is there a way to use a system version like that instead of bundling jar files?

xapian-bindings/android/android.i
ojwb: I believe I've managed to build android bindings from the standard java bindings subdirectory by adding a custom rule to build `xapianlib.aar` and passing suitable options when building the bindings.  However, I'm now stuck at compiling the example app so I can actually test the `xapianlib.aar` I built.

ojwb:
I've added support for reading `/proc/sys/kernel/random/uuid` (which is present on Android) to generate UUIDs in 2d4b56b3bcda3006a8d2a4f234a06ae6784e0432 which should remove the need to have to cross-build libuuid - I've not yet tested this on Android though.

ojwb:
I've done a bit of tidying up of the branch:

* 2bf99cbdea24 Remove bundled source-less copy of gradle
* 657e07754a49 Use Xapian logo for app icon
* 91ea1e33a1c1 Update after running through using instructions
* 5b2edde08725 Remove "cross-build libuuid" step
* dcab2fb824a7 xapian-core is just libxapian.a in a release version
* 9bf38da4281e Add a make target to build xapianlib.aar
* 82cf615767ca Remove obsolete placeholder definition
* 8649b1d3e462 Fix documentation typo
* 84b3d694465a Relocate android stuff into java subdirectory

I've pushed a flattened version of this (flattened because of the bundled source-less copy of gradle): https://github.com/ojwb/xapian/tree/android-bindings-flattened

With this branch, xapian-core and the java bindings build fine following the updated instructions.  I haven't yet tried to run the testsuite on a device.

I haven't got the example to build yet.  I'm testing it in a VM, which I installed as Debian jessie (only because I had an ISO image for that already downloaded), but the gradle version jessie has packaged seems too old to understand `jcenter()`. I've set it to update to a Debian stretch and will retest later once that's completed.

ojwb:
Actually, Tejas already agreed to license contributions to Xapian suitably in his 2016 GSoC application:

> I agree to dual-license all my contributions to Xapian under the GNU GPL version 2 and all later versions, and the MIT/X licence. I don’t have any objections regarding this.

ojwb:
Also, this is now waiting on me to get back to it (or someone else to take over) not the submitter.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,121 @@
...
+
+Prerequisites
+=============
+1)
+You will need to download android ndk from https://developer.android.com/ndk/downloads/index.html 
+for your respective development platform. Extract it and now using the make-standalone-toolchain.sh
+in the ndk make a customized toolchain from the command line by doing the following:
+
+export NDK=<path_to_ndk>
+$NDK/build/tools/make-standalone-toolchain.sh --toolchain=<target-arch> --platform=<target-android-api> --install-dir=<path_to_ndk_standalone_toolchain>
```
ojwb, line 14: Trying to clean up our PRs, and looking into this one it seems this step is no longer needed - you can now just unzip the NDK and run the compiler from the resulting tree.  I've added an `android-cross` CI job which does this:

```
wget -Ondk.zip https://dl.google.com/android/repository/android-ndk-r25c-linux.zip             
unzip -q ndk.zip  
rm -f ndk.zip  
mv android-ndk-* ndk  
```
then:
```
PATH=$PWD/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin:$PATH \  
    ./configure --disable-documentation \  
        --host armv7a-linux-androideabi33 \  
        CC=$PWD/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi33-clang \  
        CXX=$PWD/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi33-clang++ \  
        PKG_CONFIG=/bin/false  
```
and:
```
PATH=$PWD/ndk/toolchains/llvm/prebuilt/linux-x86_64/bin:$PATH make 
```

This seems to successfully build xapian-core.

I'll try to get the java bindings to build in CI too.

xapian-bindings/android/CROSS_COMPILE
```diff
@@ -0,0 +1,121 @@
...
+
+<target-arch> is the architecture your standalone toolchain is going to target. eg. for ARM-based architecture can
+use arm-linux-androideabi-4.9. For more info refer https://developer.android.com/ndk/guides/standalone_toolchain.html
+
+<target-android-api> is the target Android platform/API level. eg android-16
+
+<path_to_ndk_standalone_toolchain> is the path where your standalone toolchain will be installed.
+
+2)
+As Xapian makes use of built-in UUID APIs and ndk does not have UUID APIs, we need to explicitly cross compile that as well. So download libuuid, inside the e2fsprogs package by doing following:
```
ojwb, line 24: We now support reading UUIDs from `/proc/sys/kernel/random/uuid` which works on android and means this step isn't needed.

PR #104, opened 2016-06-03
Add support for BM25+ weighting function

ivmarkp:
- The original `bm25weight.cc` file that contains implementation of BM25 weighting function is modified to add support for BM25+ weighting function without disturbing existing functionalities of BM25 function. 
- Also, modified existing test cases of BM25 to extend test coverage for BM25+ function.
- Added API support in `weight.h` for the same.

Issue marked as "FIXME" about dealing with negative weights in `bm25weight.cc` that Olly mentioned on IRC is fixed by default when BM25+ is used.  BM25+ can never assign a negative termweight as in BM25+ termweight is calculated as `N+1 / df(t)` which translates to `(total doc in collection + 1) / termfreq` in Xapian terminology. Since, term frequency can't possibly be greater than N so `log((total doc in collection + 1) / termfreq)` can never be negative. I'd say it can be regarded as one of the benefits of BM25+ function.

Please note that I haven't run any automated tests yet using `make check`. Just wanted to get the code reviewed first. Will make sure to run all the tests after getting feedback.

### DIFF
--- a/xapian-bindings/java/Makefile.am
+++ b/xapian-bindings/java/Makefile.am
@@ -26,8 +26,8 @@ SmokeTest: SmokeTest.class
 XAPIAN_SWIG_JAVA_SRCS=\
 	org/xapian/Auto.java\
 	org/xapian/BB2Weight.java\
-	org/xapian/BM25Weight.java\
 	org/xapian/BM25PlusWeight.java\
+	org/xapian/BM25Weight.java\
 	org/xapian/BoolWeight.java\
 	org/xapian/Chert.java\
 	org/xapian/Compactor.java\
--- a/xapian-bindings/java/Makefile.am
+++ b/xapian-bindings/java/Makefile.am
@@ -27,6 +27,7 @@ XAPIAN_SWIG_JAVA_SRCS=\
 	org/xapian/Auto.java\
 	org/xapian/BB2Weight.java\
 	org/xapian/BM25Weight.java\
+	org/xapian/BM25PlusWeight.java\
 	org/xapian/BoolWeight.java\
 	org/xapian/Chert.java\
 	org/xapian/Compactor.java\
--- a/xapian-bindings/python/doxy2swig.py
+++ b/xapian-bindings/python/doxy2swig.py
@@ -28,6 +28,7 @@ import sys
 import textwrap
 import types
 import os.path
+import io
 
 def my_open_read(source):
     if hasattr(source, "read"):
@@ -39,7 +40,7 @@ def my_open_write(dest):
     if hasattr(dest, "write"):
         return dest
     else:
-        return open(dest, 'w')
+        return io.open(dest, 'w', encoding = 'utf8')
 
 
 class Doxy2SWIG:
--- a/xapian-core/api/registry.cc
+++ b/xapian-core/api/registry.cc
@@ -39,271 +39,273 @@
 using namespace std;
 
 class Xapian::Registry::Internal : public Xapian::Internal::intrusive_base {
-    friend class Xapian::Registry;
+	friend class Xapian::Registry;
 
-    /// Registered weighting schemes.
-    std::map<std::string, Xapian::Weight *> wtschemes;
+	/// Registered weighting schemes.
+	std::map<std::string, Xapian::Weight *> wtschemes;
 
-    /// Registered external posting sources.
-    std::map<std::string, Xapian::PostingSource *> postingsources;
+	/// Registered external posting sources.
+	std::map<std::string, Xapian::PostingSource *> postingsources;
 
-    /// Registered match spies.
-    std::map<std::string, Xapian::MatchSpy *> matchspies;
+	/// Registered match spies.
+	std::map<std::string, Xapian::MatchSpy *> matchspies;
 
-    /// Registered lat-long metrics.
-    std::map<std::string, Xapian::LatLongMetric *> lat_long_metrics;
+	/// Registered lat-long metrics.
+	std::map<std::string, Xapian::LatLongMetric *> lat_long_metrics;
 
-    /// Add the standard subclasses provided in the API.
-    void add_defaults();
+	/// Add the standard subclasses provided in the API.
+	void add_defaults();
 
-    /// Clear all registered weighting schemes.
-    void clear_weighting_schemes();
+	/// Clear all registered weighting schemes.
+	void clear_weighting_schemes();
 
-    /// Clear all registered posting sources.
-    void clear_posting_sources();
+	/// Clear all registered posting sources.
+	void clear_posting_sources();
 
-    /// Clear all registered match spies.
-    void clear_match_spies();
+	/// Clear all registered match spies.
+	void clear_match_spies();
 
-    /// Clear all registered lat-long metrics.
-    void clear_lat_long_metrics();
+	/// Clear all registered lat-long metrics.
+	void clear_lat_long_metrics();
 
   public:
-    Internal();
-    ~Internal();
+	Internal();
+	~Internal();
 };
 
 template<class T>
 static inline void
 register_object(map<string, T*> & registry, const T & obj)
 {
-    string name = obj.name();
-    if (rare(name.empty())) {
-	throw Xapian::InvalidOperationError("Unable to register object - name() method returned empty string");
-    }
+	string name = obj.name();
+	if (rare(name.empty())) {
+		throw Xapian::InvalidOperationError("Unable to register object - name() method returned empty string");
+	}
 
-    pair<typename map<string, T *>::iterator, bool> r;
-    r = registry.insert(make_pair(name, static_cast<T*>(NULL)));
-    if (!r.second) {
-	// Existing element with this key, so replace the pointer with NULL
-	// and delete the existing pointer.
-	//
-	// If the delete throws, this will leave a NULL entry in the map, but
-	// that won't affect behaviour as we return NULL for "not found"
-	// anyway.  The memory used will be leaked if the dtor throws, but
-	// throwing exceptions from the dtor is bad form, so that's not a big
-	// problem.
-	T * p = NULL;
-	swap(p, r.first->second);
-	delete p;
-    }
+	pair<typename map<string, T *>::iterator, bool> r;
+	r = registry.insert(make_pair(name, static_cast<T*>(NULL)));
+	if (!r.second) {
+		// Existing element with this key, so replace the pointer with NULL
+		// and delete the existing pointer.
+		//
+		// If the delete throws, this will leave a NULL entry in the map, but
+		// that won't affect behaviour as we return NULL for "not found"
+		// anyway.  The memory used will be leaked if the dtor throws, but
+		// throwing exceptions from the dtor is bad form, so that's not a big
+		// problem.
+		T * p = NULL;
+		swap(p, r.first->second);
+		delete p;
+	}
 
-    T * clone = obj.clone();
-    if (rare(!clone)) {
-	throw Xapian::InvalidOperationError("Unable to register object - clone() method returned NULL");
-    }
+	T * clone = obj.clone();
+	if (rare(!clone)) {
+		throw Xapian::InvalidOperationError("Unable to register object - clone() method returned NULL");
+	}
 
-    r.first->second = clone;
+	r.first->second = clone;
 }
- 
+
 template<class T>
 static inline const T *
 lookup_object(map<string, T*> registry, const string & name)
 {
-    typename map<string, T*>::const_iterator i = registry.find(name);
-    if (i == registry.end()) {
-	return NULL;
-    }
-    return i->second;
+	typename map<string, T*>::const_iterator i = registry.find(name);
+	if (i == registry.end()) {
+		return NULL;
+	}
+	return i->second;
 }
 
 namespace Xapian {
 
 Registry::Internal::Internal()
 {
-    add_defaults();
+	add_defaults();
 }
 
 Registry::Internal::~Internal()
 {
-    clear_weighting_schemes();
-    clear_posting_sources();
-    clear_match_spies();
-    clear_lat_long_metrics();
+	clear_weighting_schemes();
+	clear_posting_sources();
+	clear_match_spies();
+	clear_lat_long_metrics();
 }
 
 void
 Registry::Internal::add_defaults()
 {
-    Xapian::Weight * weighting_scheme;
-    weighting_scheme = new Xapian::BB2Weight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::BM25Weight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::BoolWeight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::TradWeight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::TfIdfWeight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::InL2Weight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::IfB2Weight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::IneB2Weight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::DLHWeight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::PL2Weight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::DPHWeight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
-    weighting_scheme = new Xapian::LMWeight;
-    wtschemes[weighting_scheme->name()] = weighting_scheme;
+	Xapian::Weight * weighting_scheme;
+	weighting_scheme = new Xapian::BB2Weight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::BM25Weight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::BM25PlusWeight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::BoolWeight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::TradWeight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::TfIdfWeight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::InL2Weight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::IfB2Weight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::IneB2Weight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::DLHWeight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::PL2Weight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::DPHWeight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
+	weighting_scheme = new Xapian::LMWeight;
+	wtschemes[weighting_scheme->name()] = weighting_scheme;
 
-    Xapian::PostingSource * source;
-    source = new Xapian::ValueWeightPostingSource(0);
-    postingsources[source->name()] = source;
-    source = new Xapian::DecreasingValueWeightPostingSource(0);
-    postingsources[source->name()] = source;
-    source = new Xapian::ValueMapPostingSource(0);
-    postingsources[source->name()] = source;
-    source = new Xapian::FixedWeightPostingSource(0.0);
-    postingsources[source->name()] = source;
-    source = new Xapian::LatLongDistancePostingSource(0,
-	Xapian::LatLongCoords(),
-	Xapian::GreatCircleMetric());
-    postingsources[source->name()] = source;
+	Xapian::PostingSource * source;
+	source = new Xapian::ValueWeightPostingSource(0);
+	postingsources[source->name()] = source;
+	source = new Xapian::DecreasingValueWeightPostingSource(0);
+	postingsources[source->name()] = source;
+	source = new Xapian::ValueMapPostingSource(0);
+	postingsources[source->name()] = source;
+	source = new Xapian::FixedWeightPostingSource(0.0);
+	postingsources[source->name()] = source;
+	source = new Xapian::LatLongDistancePostingSource(0,
+		Xapian::LatLongCoords(),
+		Xapian::GreatCircleMetric());
+	postingsources[source->name()] = source;
 
-    Xapian::MatchSpy * spy;
-    spy = new Xapian::ValueCountMatchSpy();
-    matchspies[spy->name()] = spy;
+	Xapian::MatchSpy * spy;
+	spy = new Xapian::ValueCountMatchSpy();
+	matchspies[spy->name()] = spy;
 
-    Xapian::LatLongMetric * metric;
-    metric = new Xapian::GreatCircleMetric();
-    lat_long_metrics[metric->name()] = metric;
+	Xapian::LatLongMetric * metric;
+	metric = new Xapian::GreatCircleMetric();
+	lat_long_metrics[metric->name()] = metric;
 }
 
 void
 Registry::Internal::clear_weighting_schemes()
 {
-    map<string, Xapian::Weight*>::const_iterator i;
-    for (i = wtschemes.begin(); i != wtschemes.end(); ++i) {
-	delete i->second;
-    }
+	map<string, Xapian::Weight*>::const_iterator i;
+	for (i = wtschemes.begin(); i != wtschemes.end(); ++i) {
+		delete i->second;
+	}
 }
 
 void
 Registry::Internal::clear_posting_sources()
 {
-    map<string, Xapian::PostingSource *>::const_iterator i;
-    for (i = postingsources.begin(); i != postingsources.end(); ++i) {
-	delete i->second;
-    }
+	map<string, Xapian::PostingSource *>::const_iterator i;
+	for (i = postingsources.begin(); i != postingsources.end(); ++i) {
+		delete i->second;
+	}
 }
 
 void
 Registry::Internal::clear_match_spies()
 {
-    map<string, Xapian::MatchSpy *>::const_iterator i;
-    for (i = matchspies.begin(); i != matchspies.end(); ++i) {
-	delete i->second;
-    }
+	map<string, Xapian::MatchSpy *>::const_iterator i;
+	for (i = matchspies.begin(); i != matchspies.end(); ++i) {
+		delete i->second;
+	}
 }
 
 void
 Registry::Internal::clear_lat_long_metrics()
 {
-    map<string, Xapian::LatLongMetric *>::const_iterator i;
-    for (i = lat_long_metrics.begin(); i != lat_long_metrics.end(); ++i) {
-	delete i->second;
-    }
+	map<string, Xapian::LatLongMetric *>::const_iterator i;
+	for (i = lat_long_metrics.begin(); i != lat_long_metrics.end(); ++i) {
+		delete i->second;
+	}
 }
 
 Registry::Registry(const Registry & other)
-	: internal(other.internal)
+		: internal(other.internal)
 {
-    LOGCALL_CTOR(API, "Registry", other);
+	LOGCALL_CTOR(API, "Registry", other);
 }
 
 Registry &
 Registry::operator=(const Registry & other)
 {
-    LOGCALL(API, Xapian::Registry &, "Xapian::Registry::operator=", other);
-    internal = other.internal;
-    RETURN(*this);
+	LOGCALL(API, Xapian::Registry &, "Xapian::Registry::operator=", other);
+	internal = other.internal;
+	RETURN(*this);
 }
 
 Registry::Registry()
-	: internal(new Registry::Internal())
+		: internal(new Registry::Internal())
 {
-    LOGCALL_CTOR(API, "Registry", NO_ARGS);
+	LOGCALL_CTOR(API, "Registry", NO_ARGS);
 }
 
 Registry::~Registry()
 {
-    LOGCALL_DTOR(API, "Registry");
+	LOGCALL_DTOR(API, "Registry");
 
-    // Note - we don't need to do anything special in this destructor, but it
-    // does need to be explicitly defined because the definition of the
-    // internals is not visible externally, which results in an error if the
-    // compiler tries to generate a default destructor.
+	// Note - we don't need to do anything special in this destructor, but it
+	// does need to be explicitly defined because the definition of the
+	// internals is not visible externally, which results in an error if the
+	// compiler tries to generate a default destructor.
 }
 
 void
 Registry::register_weighting_scheme(const Xapian::Weight &wt)
 {
-    LOGCALL_VOID(API, "Xapian::Registry::register_weighting_scheme", wt.name());
-    register_object(internal->wtschemes, wt);
+	LOGCALL_VOID(API, "Xapian::Registry::register_weighting_scheme", wt.name());
+	register_object(internal->wtschemes, wt);
 }
 
 const Xapian::Weight *
 Registry::get_weighting_scheme(const string & name) const
 {
-    LOGCALL(API, const Xapian::Weight *, "Xapian::Registry::get_weighting_scheme", name);
-    RETURN(lookup_object(internal->wtschemes, name));
+	LOGCALL(API, const Xapian::Weight *, "Xapian::Registry::get_weighting_scheme", name);
+	RETURN(lookup_object(internal->wtschemes, name));
 }
 
 void
 Registry::register_posting_source(const Xapian::PostingSource &source)
 {
-    LOGCALL_VOID(API, "Xapian::Registry::register_posting_source", source.name());
-    register_object(internal->postingsources, source);
+	LOGCALL_VOID(API, "Xapian::Registry::register_posting_source", source.name());
+	register_object(internal->postingsources, source);
 }
 
 const Xapian::PostingSource *
 Registry::get_posting_source(const string & name) const
 {
-    LOGCALL(API, const Xapian::PostingSource *, "Xapian::Registry::get_posting_source", name);
-    RETURN(lookup_object(internal->postingsources, name));
+	LOGCALL(API, const Xapian::PostingSource *, "Xapian::Registry::get_posting_source", name);
+	RETURN(lookup_object(internal->postingsources, name));
 }
 
 void
 Registry::register_match_spy(const Xapian::MatchSpy &spy)
 {
-    LOGCALL_VOID(API, "Xapian::Registry::register_match_spy", spy.name());
-    register_object(internal->matchspies, spy);
+	LOGCALL_VOID(API, "Xapian::Registry::register_match_spy", spy.name());
+	register_object(internal->matchspies, spy);
 }
 
 const Xapian::MatchSpy *
 Registry::get_match_spy(const string & name) const
 {
-    LOGCALL(API, const Xapian::MatchSpy *, "Xapian::Registry::get_match_spy", name);
-    RETURN(lookup_object(internal->matchspies, name));
+	LOGCALL(API, const Xapian::MatchSpy *, "Xapian::Registry::get_match_spy", name);
+	RETURN(lookup_object(internal->matchspies, name));
 }
 
 void
 Registry::register_lat_long_metric(const Xapian::LatLongMetric &metric)
 {
-    LOGCALL_VOID(API, "Xapian::Registry::register_lat_long_metric", metric.name());
-    register_object(internal->lat_long_metrics, metric);
+	LOGCALL_VOID(API, "Xapian::Registry::register_lat_long_metric", metric.name());
+	register_object(internal->lat_long_metrics, metric);
 }
 
 const Xapian::LatLongMetric *
 Registry::get_lat_long_metric(const string & name) const
 {
-    LOGCALL(API, const Xapian::LatLongMetric *, "Xapian::Registry::get_lat_long_metric", name);
-    RETURN(lookup_object(internal->lat_long_metrics, name));
+	LOGCALL(API, const Xapian::LatLongMetric *, "Xapian::Registry::get_lat_long_metric", name);
+	RETURN(lookup_object(internal->lat_long_metrics, name));
 }
 
 }
--- a/xapian-core/api/registry.cc
+++ b/xapian-core/api/registry.cc
@@ -309,4 +309,3 @@ Registry::get_lat_long_metric(const string & name) const
 }
 
 }
-
--- a/xapian-core/api/registry.cc
+++ b/xapian-core/api/registry.cc
@@ -39,273 +39,274 @@
 using namespace std;
 
 class Xapian::Registry::Internal : public Xapian::Internal::intrusive_base {
-	friend class Xapian::Registry;
+    friend class Xapian::Registry;
 
-	/// Registered weighting schemes.
-	std::map<std::string, Xapian::Weight *> wtschemes;
+    /// Registered weighting schemes.
+    std::map<std::string, Xapian::Weight *> wtschemes;
 
-	/// Registered external posting sources.
-	std::map<std::string, Xapian::PostingSource *> postingsources;
+    /// Registered external posting sources.
+    std::map<std::string, Xapian::PostingSource *> postingsources;
 
-	/// Registered match spies.
-	std::map<std::string, Xapian::MatchSpy *> matchspies;
+    /// Registered match spies.
+    std::map<std::string, Xapian::MatchSpy *> matchspies;
 
-	/// Registered lat-long metrics.
-	std::map<std::string, Xapian::LatLongMetric *> lat_long_metrics;
+    /// Registered lat-long metrics.
+    std::map<std::string, Xapian::LatLongMetric *> lat_long_metrics;
 
-	/// Add the standard subclasses provided in the API.
-	void add_defaults();
+    /// Add the standard subclasses provided in the API.
+    void add_defaults();
 
-	/// Clear all registered weighting schemes.
-	void clear_weighting_schemes();
+    /// Clear all registered weighting schemes.
+    void clear_weighting_schemes();
 
-	/// Clear all registered posting sources.
-	void clear_posting_sources();
+    /// Clear all registered posting sources.
+    void clear_posting_sources();
 
-	/// Clear all registered match spies.
-	void clear_match_spies();
+    /// Clear all registered match spies.
+    void clear_match_spies();
 
-	/// Clear all registered lat-long metrics.
-	void clear_lat_long_metrics();
+    /// Clear all registered lat-long metrics.
+    void clear_lat_long_metrics();
 
   public:
-	Internal();
-	~Internal();
+    Internal();
+    ~Internal();
 };
 
 template<class T>
 static inline void
 register_object(map<string, T*> & registry, const T & obj)
 {
-	string name = obj.name();
-	if (rare(name.empty())) {
-		throw Xapian::InvalidOperationError("Unable to register object - name() method returned empty string");
-	}
+    string name = obj.name();
+    if (rare(name.empty())) {
+	throw Xapian::InvalidOperationError("Unable to register object - name() method returned empty string");
+    }
 
-	pair<typename map<string, T *>::iterator, bool> r;
-	r = registry.insert(make_pair(name, static_cast<T*>(NULL)));
-	if (!r.second) {
-		// Existing element with this key, so replace the pointer with NULL
-		// and delete the existing pointer.
-		//
-		// If the delete throws, this will leave a NULL entry in the map, but
-		// that won't affect behaviour as we return NULL for "not found"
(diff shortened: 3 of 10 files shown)


xapian-core/weight/bm25weight.cc
```diff
@@ -35,11 +35,23 @@ using namespace std;
 
 namespace Xapian {
 
+BM25Weight::BM25Weight(bool& bm25+)
+{
+	if(bm25+ != 0 && bm25+ != 1)
+	throw Xapian::InvalidArgumentError("Boolean input is invalid.");
```
jaylett: This looks like it isn't aligned properly; it's the body of the `if` so should be aligned further across than the `if` itself. Also, you're indenting to 8 spaces (as a TAB) rather than 4 as the rest of the file.

xapian-core/include/xapian/weight.h
```diff
@@ -526,15 +526,20 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+     *		   with a fixed δ = 1.0. (default 1.0)
+     * 
      *  @param min_normlen  A parameter specifying a minimum value for
      *		   normalised document length.  Normalised document length
      *		   values less than this will be clamped to this value, helping
      *		   to prevent very short documents getting large weights.
      *		   (default 0.5)
      */
-    BM25Weight(double k1, double k2, double k3, double b, double min_normlen)
+    BM25Weight(double k1, double k2, double k3, double b, double delta, double min_normlen)
```
jaylett: I think we should put the new parameter last; it'll make it easier for people to start using `delta` if they're putting a new number on the end rather than having to insert it in the middle of the list.

We also could do with an overloaded constructor that only has the old parameters, with `param_delta` being initialised to `1.0` as with the zero parameter constructor (so we don't break API compatibility with earlier releases).

xapian-core/weight/bm25weight.cc
```diff
@@ -35,11 +35,23 @@ using namespace std;
 
 namespace Xapian {
 
+BM25Weight::BM25Weight(bool& bm25+)
+{
+	if(bm25+ != 0 && bm25+ != 1)
```
jaylett: Please put a space after `if` (before the opening parenthesis).

xapian-core/include/xapian/weight.h
```diff
@@ -543,6 +548,7 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
 	    param_b = 0;
 	} else if (param_b > 1) {
 	    param_b = 1;
+	if (param_delta < 0) param_delta = 0;
```
jaylett: This is either not indented correctly, or is in the wrong place. I believe it should be after the closing parenthesis from the previous `if…else if`. (Unless `param_delta` should only be snapped to non-negative if `param_b` has been snapped to max-1, which seems unlikely.)

xapian-core/include/xapian/weight.h
```diff
@@ -557,11 +563,15 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
 	if (param_k1 != 0 && param_b != 0) need_stat(DOC_LENGTH);
 	if (param_k2 != 0) need_stat(QUERY_LENGTH);
 	if (param_k3 != 0) need_stat(WQF);
+	if (param_delta != 0) {
+	    need_stat(WQF);
+	    need_stat(DOC_LENGTH);
+	    need_stat(AVERAGE_LENGTH);
```
jaylett: This is missing a closing brace. Without it, none of this will compile.

xapian-core/weight/bm25weight.cc
```diff
@@ -35,11 +35,23 @@ using namespace std;
 
 namespace Xapian {
 
+BM25Weight::BM25Weight(bool& bm25+)
```
jaylett: You can't call a formal parameter `bm25+`!

xapian-core/weight/bm25weight.cc
jaylett: Oh, and `bool`s I think can be `true` or `false` and nothing else, so I'm not sure that this conditional should be there at all.

xapian-core/weight/bm25weight.cc
```diff
@@ -48,8 +60,8 @@ BM25Weight::init(double factor)
     Xapian::doccount tf = get_termfreq();
 
     double tw = 0;
-    if (get_rset_size() != 0) {
-	Xapian::doccount reltermfreq = get_reltermfreq();
+    if (get_rset_size() != 0 && !bm25+) {
```
jaylett: `bm25+` was a parameter to one of the constructors, not a member, so you can't access it here. You probably want to make a `bm25_plus` private member, and initialise that somehow.

xapian-core/weight/bm25weight.cc
```diff
@@ -48,8 +60,8 @@ BM25Weight::init(double factor)
...
+       	Xapian::doccount reltermfreq = get_reltermfreq();
```
jaylett: This line has come up as changed because the tab has been changed to spaces. If possible, try to configure your editor to either leave unedited lines as is, or to use tabs where possible. (Tabs followed by spaces is fine.)

xapian-core/weight/bm25weight.cc
```diff
@@ -71,8 +83,13 @@ BM25Weight::init(double factor)
 	double numerator = (reltermfreq + 0.5) * (Q - tf + 0.5);
 	double denom = (reldocs_not_indexed + 0.5) * (nonreldocs_indexed + 0.5);
 	tw = numerator / denom;
-    } else {
-	tw = (get_collection_size() - tf + 0.5) / (tf + 0.5);
+	}
+	else if(bm25+) {
```
jaylett: `else if` should be on the same line as the closing brace, and there should be a space after `if`. Also, the alignment is wrong on your new lines (which is why this is difficult to read in github as effectively adding a few lines).

xapian-core/weight/bm25weight.cc
```diff
@@ -103,11 +120,16 @@ BM25Weight::init(double factor)
...
-	termweight *= (param_k3 + 1) * wqf_double / (param_k3 + wqf_double);
+    if (tw < 2 && !bm25+) {
+        tw = tw * 0.5 + 1;
+        termweight = log(tw) * factor;
+        if (param_k3 != 0) {
+            double wqf_double = get_wqf();
+            termweight *= (param_k3 + 1) * wqf_double / (param_k3 + wqf_double);
+        }
+    }
+    else if (bm25+) {
```
jaylett: I'm not convinced this does the right thing for non-BM25+. If `tw >= 2` then `termweight` isn't set, while previously it would have been `log(tw) * factor`. What's probably simplest is an outer `if (bm25_plus) { termweight = log(tw) * factor; } else { … }` with the existing code inside.

xapian-core/weight/bm25weight.cc
```diff
@@ -170,7 +194,15 @@ BM25Weight::get_sumpart(Xapian::termcount wdf, Xapian::termcount len,
     double wdf_double = wdf;
     double denom = param_k1 * (normlen * param_b + (1 - param_b)) + wdf_double;
     AssertRel(denom,>,0);
-    RETURN(termweight * (wdf_double / denom));
+    if(!bm25+) {
+        RETURN(termweight * (wdf_double / denom));
+    }
+    else if(bm25+) {
```
jaylett: You should just have `if (bm25_plus) { … } else { … }`. Testing the negation, then using an `if else` is less readable for people trying to understand what's going on.

xapian-core/weight/bm25weight.cc
```diff
@@ -170,7 +194,15 @@ BM25Weight::get_sumpart(Xapian::termcount wdf, Xapian::termcount len,
...
+        //Parameter delta (δ) is a pseudo tf value to control the scale of the
+        //tf lower bound. δ can be tuned for e.g from 0.0 to 1.5 but BM25+ can
+        //still work effectively across collections with a fixed δ = 1.0
```
jaylett: Please put a single space after the `//` on each line of this comment block.

xapian-core/weight/bm25weight.cc
```diff
@@ -221,7 +258,7 @@ BM25Weight::get_maxextra() const
 {
     LOGCALL(WTCALC, double, "BM25Weight::get_maxextra", NO_ARGS);
     if (param_k2 == 0.0)
-	RETURN(0.0);
+    RETURN(0.0);
```
jaylett: This has been re-indented to the wrong place.

jaylett:
@ivmarkp when you work on the changes I've suggested, it's also a good idea to run the final diff through our `xapian-check-diff` script, which will look for common formatting errors (mostly around spacing). The easiest way to do that is:

```
$ git diff github/master HEAD | ./xapian-maintainer-tools/xapian-check-patch
```

Assuming your remote to Xapian on github is called `github` and you're sitting at the top of the repository.

xapian-core/include/xapian/weight.h
ivmarkp: You mean an overloaded constructor with `param_delta` initialized to `1.0`and rest of the parameters initialized to values received in variables k1, k2 etc.? 
So, initializing `param_delta` to `1.0` seem to be overriding `delta` value with `1.0` which calls additional stats required for BM25+ but we can avoid calling stats, `WQF`, `DOC_LENGTH` and `AVERAGE_LENGTH`, when `delta` is zero or less than zero. 
In that case, we'll have a new constructor looking like :-
```C++
BM25Weight(double k1, double k2, double k3, double b, double min_normlen, double delta)
    : param_k1(k1), param_k2(k2), param_k3(k3), param_b(b),
      param_min_normlen(min_normlen), param_delta(delta)
    {
        param_delta = 1.0;
        need_stat(WQF);
	    need_stat(DOC_LENGTH);
	    need_stat(AVERAGE_LENGTH);
	    if (param_k1 < 0) param_k1 = 0;
	    if (param_k2 < 0) param_k2 = 0;
	    if (param_k3 < 0) param_k3 = 0;
	    if (param_b < 0) {
	        param_b = 0;
	        } else if (param_b > 1) {
	        param_b = 1;
	    }
	    need_stat(COLLECTION_SIZE);
	    need_stat(RSET_SIZE);
	    need_stat(TERMFREQ);
	    need_stat(RELTERMFREQ);
	    need_stat(WDF);
	    need_stat(WDF_MAX);
	    if (param_k2 != 0 || (param_k1 != 0 && param_b != 0)) {
	        need_stat(DOC_LENGTH_MIN);
	        need_stat(AVERAGE_LENGTH);
	    }
	    if (param_k1 != 0 && param_b != 0) need_stat(DOC_LENGTH);
	    if (param_k2 != 0) need_stat(QUERY_LENGTH);
	    if (param_k3 != 0) need_stat(WQF);
    }
```
Does this look good to you?

xapian-core/weight/bm25weight.cc
ivmarkp: What name do you suggest for this parameter? I was also thinking of `bm25_plus` earlier.

jaylett: I really don't mind, although right now I don't see how you can tune the other parameters while using BM25+, which suggests that this is the wrong API. An alternative would be to either have it set on the class (to `false`), and override to `true` on a `BM25PlusWeight` subclass, or to factor out the bits that need to change in the calculations, and override them.

xapian-core/include/xapian/weight.h
jaylett: I'd just refactor the old constructor to have an extra parameter at the end (as you have done), and implement the backwards compatible one in terms of that. Note that this is [a C++11 feature](https://en.wikipedia.org/wiki/C%2B%2B11#Object_construction_improvement), which we now accept (which is why the current one isn't done like that, I guess):

```c++
    BM25Weight(double k1, double k2, double k3, double b, double min_normlen)
        : BM25Weight(k1, k2, k3, b, min_normlen, 1.0) {}
```

ojwb: C++11 is OK inside the library, but we're trying to avoid it in the API headers for 1.4.x.  This is just for practical reasons, as at least the older versions of the compilers we aim to support have C++11 off by default - if our API headers use it, then users are forced to pass the appropriate option to their compiler, which there's no portable way to do, so you end up having to hard code for one compiler, or probe in `configure` or equivalent.

So far there's not been anything from C++11 that's really needed in the API headers - for this case, I'd just add a second implementation of the constructor for now (the obvious alternative is adding a new `delta` parameter and giving it a default value, but that would be an ABI-incompatible change so we couldn't include that in 1.4.x - user code built against 1.4.0 would fail to find the implementation of the 4-parameter form in a shared library built from 1.4.x with this change).  But beyond 1.4.x, adding a new `delta` parameter with default value 1.0 to the end of the parameter list should achieve this without needing C++11 enabled when building code using the library).

ojwb:
> Please note that I haven't run any automated tests yet using make check. Just wanted to get the code reviewed first. Will make sure to run all the tests after getting feedback.

There's something to be said for early review, but your code here doesn't even compile (e.g. `bm25+` isn't a valid parameter name).  It's inefficient for you to open a PR and wait for us to review only to tell you about errors which the compiler could have found in seconds (even with lucky timing, you're unlikely to get a review from us in under 15 minutes, and it's more likely to be at least half a day).

For similar reasons, at least run the exist testcases before opening a PR - if your changes make existing tests fail, there's something wrong.  It can happen that the issue is an existing testcase legitimately needs adjusting (it's not the normal situation, but it happens), but even then we can't merge a patch which makes existing tests fail - in this situation, those testcases need adjusting.

Running `xapian-check-patch` before submitting and fixing any issues it highlights is similarly a good idea - it's much faster feedback than waiting for a human to point out those same issues (which is why it exists).

xapian-core/weight/bm25weight.cc
ojwb: I realise we're not keeping this constructor, but as a general point you don't pass parameters by non-`const` reference unless they are used to return a value - the parameter here should just be `bool`.

You wouldn't pass a primitive type like `bool` by `const` reference either - passing by `const` reference is a way to avoid copying the value, which is useful when copying is potentially expensive (e.g. for a `std::vector`, `std::map`, etc) but copying a `bool` (or `int`, `double`, etc) is cheap as it fits in a single CPU register, and dealing with the reference is likely to add some overhead so could well actually be slower).

xapian-core/weight/bm25weight.cc
ojwb: Indeed, this check is pointless.

I'd avoid explicitly comparing boolean variables against an explicit value - instead of:

```
if (bm25_plus == true) {
    // Do something
}
if (flag == false) {
    // Do something else
}
```

write like so:

```
if (bm25_plus) {
    // Do something
}
if (!flag) {
    // Do something else
}
```

The intent seems clearer (at least to me) without the explicit test, and the behaviour is less surprising if the type of `bm25_plus` is `int` not `bool` - the first version checks in the value is `1`, the second if it is non-zero.

If you're dealing with C code using `int` for flags with something like:

```
#define FALSE 0
#define TRUE 1
```

then `if (value)` generally behaves more helpfully than `if (value == TRUE)` if `value` is set to (say) `2`.

ivmarkp:
@ojwb @jaylett Sorry, like you said I should have had compiled and run all the tests beforehand. 
Although, I refactored the code following the previous comments and suggestions (thanks so much). Code is finally compiling but tests are failing with errors messages like `SIGFPE at 0x80435b676` which are currently beyond my understanding as I have never encountered such errors before. Is there anything that can be done to find what exactly in the code causing the tests to fail? I'd commit the changes so you can look at the code if that's okay.

samuelharden:
@ivmarkp  SIGFPE is the easiest runtime error to debug - it is a floating point error. It is virtually always caused by a division by 0, so check any divisions or modulo operations in your code carefully.

It's always advisable to google the error you get, i generally tells you a lot about error and reason.

To debug the following error is will suggest using gdb.

Run your tests under gdb and see where are you getting the exception.

1. Run your program under gdb.
2. Add a breakpoint to catch exception using(catch throw)
3. see the exact point when compilar stops at exception (bt)
4. You can change stack level to further debug what causes the error.

If you have not worked with gdb, try reading a tutorial for gdb(it might be very helpful) http://www.tutorialspoint.com/gnu_debugger/

ivmarkp:
These errors have given me a couple of sleepless nights. I'm kind of stuck at this point. Thanks for the helpful links for gdb though. Hopefully, I'll locate the bugs and remove them and be able to move further according the project timeline.

samuelharden:
@ivmarkp I am happy to help/assist you in debugging these errors, If you have done small amount of research into error your self on IRC(It's easy ). I am generally around my laptop from 11AM IST to 6-7PM IST, hence higher probability of replying . 

If you want we can create a account on quassel core on VM, which can help you stay connected on IRC  and see all messages when you are back online

ivmarkp:
@samuelharden I've read about it as you advised. I found this [link](http://www.gnu.org/software/libc/manual/html_node/Program-Error-Signals.html) that has information on various program errors including SIGFPE. I'm currently verifying test cases manually for division by zero error in the code.

Yes, it'd be great to have an account on quassel. I have been wanting to setup my IRC account on quassel from past few months but couldn't due to unavailability of server (24x7 running PC)  at home.

ivmarkp:
@ojwb @jaylett I have implemented a new sub class (BM25PlusWeight) for BM25+ weighting function. All the tests are now passing. Please have a look again at the code and let me know things I can improve upon. Thanks!

P.S. Update on what I have been doing -- Last few days, I was busy trying to modify the `bm25weight.cc` to incorporate functionalities for BM25+ weighting function but that didn't turn out very well as `make check` tests were failing. I tried verifying the tests manually. A couple of days were spent trying to refactor and debug the same code but no success. Finally, I decided to implement a new subclass for BM25+ function which seems to have worked well.

xapian-core/include/xapian/weight.h
```diff
@@ -588,6 +588,114 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+ * This class implements the BM25+ Probabilistic Weighting scheme which is
+ * described by Lv-Zhai in their paper.  BM25+ is useful when very-long
+ * documents are present in document collection.
+ *
+ * BM25+ uses an extra parameter delta(δ). Formally, BM25PlusWeight(delta)
+ * should be equivalent to BM25Weight(k1, k2, k3, b, min_normlen, delta).
+ */
+class XAPIAN_VISIBILITY_DEFAULT BM25PlusWeight : public BM25Weight {
+	/// Factor to multiply the document length by.
+    mutable Xapian::doclength len_factor;
```
jaylett: Do you actually need to repeat these members, given you're inheriting from `BM25Weight`?

xapian-core/include/xapian/weight.h
```diff
@@ -588,6 +588,114 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+  public:
+    /** Construct a BM25PlusWeight.
+     *
+     *  @param delta  A parameter for pseudo tf value to control the scale
+     *		   of the tf lower bound. Delta(δ) can be tuned for e.g from 0.0
+     *		   to 1.5 but BM25+ can still work effectively across collections
+     *		   with a fixed δ = 1.0. (default 1.0)
+     */
+    BM25PlusWeight(double k1, double k2, double k3, double b, double min_normlen, double delta)
+    : param_k1(k1), param_k2(k2), param_k3(k3), param_b(b),
```
jaylett: We normally indent the entire initialiser list, including the colon ("`:`").

xapian-core/include/xapian/weight.h
```diff
@@ -588,6 +588,114 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+ 	if (param_k3 != 0) need_stat(WQF);
+	if (param_delta < 0) param_delta = 0;
+	if (param_delta != 0) {
+	    need_stat(AVERAGE_LENGTH);
+	    need_stat(DOC_LENGTH);
+		need_stat(WQF);
+	}
+    }
+
+	BM25PlusWeight()
```
jaylett: This should be aligned further left.

xapian-core/tests/api_weight.cc
```diff
@@ -134,6 +134,49 @@ DEFINE_TESTCASE(bm25weight5, backend) {
 }
 
 // Test exception for junk after serialised weight.
+DEFINE_TESTCASE(bm25plusweight3, !backend) {
```
jaylett: It would probably make more sense to call this `bm25plusweight1`.

xapian-core/include/xapian/weight.h
```diff
@@ -588,6 +588,114 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+		       Xapian::termcount doclen,
+		       Xapian::termcount uniqterm) const;
+    double get_maxpart() const;
+
+    double get_sumextra(Xapian::termcount doclen,
+			Xapian::termcount uniqterms) const;
+    double get_maxextra() const;
+
+  protected:
+	//BM25+ is enabled.
```
jaylett: Space at the start of the comment, please :)

jaylett:
I'm not at home, so I can't easily build and play with this, but you don't seem to have rolled back your previous changes to the `BM25Weight` implementation, so this may still not compile.

You also seem (beyond serialisation checks) to only have one test for the BM25+ implementation, which only uses one configuration. I'm not sure if this is enough without having the facilities to dig a bit deeper, but it may be worth thinking about whether there are corner cases you aren't currently testing.

xapian-core/tests/api_weight.cc
ivmarkp: There is another test case named `bm25plusweight1` inside `api_nodb.cc` file. Actually, this is the third test case for `BM25PlusWeight` so I thought `bm25plusweight3` would make sense overall.

xapian-core/include/xapian/weight.h
ivmarkp: These data members are used for various calculations in member functions of `BM25PlusWeight` subclass and they ,being private members by default, can't be accessed from `BM25Weight`. So, I repeated them here.

ivmarkp:
@jaylett Thanks for the review. I'll write more test cases to improve overall test coverage. And I made last commits from Github interface after making changes to a fresh clone locally and building it. I forgot to rollback the previous changes I made in `bm25weight.cc` here.  I'm sorry about that. Working on it now.

xapian-core/tests/api_weight.cc
jaylett: Ah, sorry I missed that. Maybe we can come up with names for them that describe what about BM25+ they're testing?

xapian-core/include/xapian/weight.h
jaylett: So we should probably promote them to `protected`, otherwise each `BM25PlusWeight` object has lots of members it's not using.

jaylett:
@ivmarkp you have a merge commit in this; generally you shouldn't submit PRs with merge commits at all; this one in particular is weird because the message suggests you've merged from the bm25+ plus into itself, when the merge I assume you intended was from upstream master into bm25+. You may want to review [the article I wrote](https://tartarus.org/james/diary/2016/06/05/git-remotes) on remotes and rebasing, which I posted to the list recently.

Additionally, in the merge commit you seem to have committed at least one conflicted file without resolving it. (You may get the same issue when rebasing to make a merge-free branch, so you should be careful about that.)

ivmarkp:
@jaylett Sorry, this seems all messed. Everything looks just fine in my local repo but now all the changes have extra indentation and spaces here. I had checked the patches for any indentation issues using `git diff origin/bm25+ | xapian-maintainer-tools/xapian-check-patch` and no issues were reported. What has gone wrong here? :-/

 Actually, the local clone was not up-to-date with remote repo (github.com/ivmarkp/xapian) so I pulled the changes and merged them. Then, made the new changes and pushed them. All of the past commits have also come up in this PR. This has become confusing. How do I fix this?

jaylett:
What went wrong was you merged (probably via `git pull`, which you should almost never use because it doesn't give you an opportunity to spot these things in advance; in my article I suggest using `git fetch` followed by `git show-branch` so you know what's going on before you make any changes), and the merge contained errors which you didn't resolve properly before proceeding.

You can fix this by rebasing just your changes onto master. You also want to fetch from the upstream repo (xapian/xapian on github), and rebase onto that rather than onto ivmarkp/xapian, which may be behind the 'official' state.

Preparing for this is fairly easy:

```
git add remote upstream https://github.com/xapian/xapian.git
git fetch upstream
git checkout bm25+
git checkout -b SAVE
```

The last line creates a new branch that you can get back to easily if things go wrong. (It's not strictly necessary, but recovering otherwise is more of a pain.)

The rebase itself is harder, because you aren't rebasing everything since your branch and the upstream master diverged:

```
git rebase --onto upstream/master 756c156 bm25+
```

What that will do is:

1. Save everything on bm25+ _after_ 756c156 (ie all the changes except the merge commit) into a temporary area
2. Reset the bm25+ branch to upstream/master
3. Apply the commits from the temporary area in turn

I believe that will do what you need. (Alternatively, you could do the preparation, reset bm25+ to upstream/master yourself, and then cherry-pick your commits one by one. That basically does the same thing.)

If there are merge conflicts, you'll have to resolve them individually; you'll have to edit each file (find the problems using `git diff`) to remove the conflict markers (things like `<<<<<`) and leave the files in the correct state, then add the changes (using `git add <filename>`) and finally `git rebase --continue` to keep going with the rebase.

ivmarkp:
While trying  to rebase using `git rebase --onto upstream/master 756c156 bm25+` , I'm getting this error:
```
CONFLICT (modify/delete): xapian-core/weight/bm25plusweight.cc deleted in
59b28e7dce8281bebafe6f8d20312a98be0894a2 and modified in Fixup! Implement BM25PlusWeight subclass for BM25+. 
Version Fixup! Implement BM25PlusWeight subclass for BM25+ of xapian-core/weight/bm25plusweight.cc left in tree.
error: Failed to merge in the changes.
Patch failed at 0001 Fixup! Implement BM25PlusWeight subclass for BM25+
The copy of the patch that failed is found in: .git/rebase-apply/patch.```

Do I need to delete `bm25plusweight.cc` to continue?

**Comment by ivmarkp** [comment]
@jaylett Wondering if it'll be a good idea to close this PR, make changes afresh to a new branch and open another PR from new branch? Not sure if that will go smoothly.

**Comment by ivmarkp** [comment]
@jaylett  Rebasing is done. Thanks for helping out. 
Just to be sure `git status` shows:
```
vivek@Gnome-PC:~/Documents/xapian$ git status
On branch bm25+
Your branch and 'origin/bm25+' have diverged,
and have 7 and 23 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
nothing to commit, working directory clean
```
bm25+ is the name for both -- local and remote branch.
Is it now safe to push from bm25+ branch? I pray nothing goes wrong this time.

**Comment by jaylett** [comment]
You don't need to close PRs unless you're withdrawing the changes; in this situation you can force push to replace what's already on the branch behind the PR.

`git status` isn't really the best tool for checking what's going on, because it summarises quite aggressively. `git show-branch` can tell you what commits exist from the different branches back to a common ancestor (and you can include upstream/master to get an idea of what will happen when we apply your changes).

**Comment by ivmarkp** [comment]
`git show-branch

**Comment by ivmarkp** [comment]
`git show-branch` gives this [output](http://pastebin.com/sdfj1hDJ). `git show-branch upstream/master` shows:
```
vivek@Gnome-PC:~/Documents/xapian$ git show-branch upstream/master
[upstream/master] Suppress compiler warnings about pointer alignment
```
And, `git show-branch bm25+` shows:
```
vivek@Gnome-PC:~/Documents/xapian$ git show-branch bm25+
[bm25+] Remove previous BM25+ changes.
```
So, the most recent commits on upstream/master branch is "Suppress compiler warnings about pointer alignment" and from the output of `git show-branch`, recent commits on bm25+ branch are:
```
 *  [bm25+] Remove previous BM25+ changes.
 *  [bm25+^] Improve test coverage for BM25PlusWeight.
 *  [bm25+~2] Add Xapian::BM25PlusWeight in add_defaults()
 *  [bm25+~3] Promote BM25PlusWeight private members to protected.
 *  [bm25+~4] Fixup! Implement BM25PlusWeight subclass for BM25+
 *  [bm25+~5] Suppress compiler warnings about pointer alignment
 *  [bm25+~6] Update to the beta version of Unicode 9.0.0
```
Not sure if I'm correct but this should mean bm25+ branch is ahead of upstream/master by top 5 commits and this is what I want to go ahead and push (force push in this case) ?

**Comment by ivmarkp** [comment]
@jaylett I have force pushed the commits from bm25+ branch after rebasing and PR looks clean now. Now I can move on to setting up xapian-evaluation as the next thing to do is evaluate BM25+ and see how it performs. Please do build and test on your system to validate the changes whenever you have some time.

File: xapian-core/api/registry.cc [hunk 19]
```diff
@@ -39,271 +39,273 @@
...
-    /// Clear all registered lat-long metrics.
-    void clear_lat_long_metrics();
+	/// Clear all registered lat-long metrics.
+	void clear_lat_long_metrics();
 
   public:
-    Internal();
-    ~Internal();
+	Internal();
+	~Internal();
```
ojwb: Lots of unwanted whitespace changes here and above.

xapian-core/include/xapian/weight.h
```diff
@@ -573,21 +573,122 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+
+	/// Additional parameter delta in the BM25+ formula.
+	double param_delta;
+
+	BM25PlusWeight * clone() const;
+
+	void init(double factor);
+
+	// BM25+ is enabled.
+	bool bm25_plus = 1;
```
ojwb: How is this flag useful?

If we are just adding BM25+ support as an option to the existing `BM25Weight` class, then adding a flag to that class would make sense.  But subclassing is an alternative to that.  Doing both isn't useful - methods of `BM25PlusWeight` know they're methods of `BM25PlusWeight`, and methods of the parent class called on a `BM25PlusWeight` will see the member variable of the parent class, not the subclass.

I still feel making this an option in `BM25Weight` is the best way to go, but if we're going for a subclass, can we get rid of this flag?

xapian-bindings/java/Makefile.am
```diff
@@ -27,6 +27,7 @@ XAPIAN_SWIG_JAVA_SRCS=\
 	org/xapian/Auto.java\
 	org/xapian/BB2Weight.java\
 	org/xapian/BM25Weight.java\
+	org/xapian/BM25PlusWeight.java\
```
ojwb: This list is sorted, but your insertion is in the wrong place.

xapian-core/include/xapian/weight.h
ivmarkp: > I still feel making this an option in BM25Weight is the best way to go, but if we're going for a subclass, can we get rid of this flag?

Yes, this flag should go now. Sorry, I probably overlooked it

Making an option to switch to BM25+ functionalities using this flag was causing lots of `make check` tests to fail including the unrelated tests too with messages indicating floating point errors probably due to division by zero errors and memory overflows. 

I spent a day or two trying to debug the code and also manually verified the test cases for those errors but wasn't able to resolve them completely as some of the unrelated tests were still failing. In the end, I took the subclass approach to keep moving according to the project timeline and avoid getting stuck at one point. 

Although, if it is a better way to go I can try that approach afresh once more and see if it passes tests but that might cost some time?

ivmarkp: Yes, this flag should go now. Sorry, I probably overlooked it. Actually, it was used to control what `get_sumpart()` and `get_maxpart()` should return i.e. when BM25+ is enabled and when its not.

Making an option to switch to BM25+ functionalities using this flag was causing lots of make check tests to fail including the unrelated tests too with messages indicating floating point errors probably due to division by zero errors and memory overflows.

I spent a day or two trying to debug the code and also manually verified the test cases for those errors but wasn't able to resolve them completely as some of the unrelated tests were still failing. In the end, I took the subclass approach to keep moving according to the project timeline and avoid getting stuck at one point.

Although, if it is a better way to go I can try that approach afresh once more and see if it passes tests but that might cost some time?

xapian-bindings/java/Makefile.am
ivmarkp: Oh, it should be above BM25Weight. Will fix that.

xapian-core/weight/bm25weight.cc
```diff
@@ -228,3 +228,4 @@ BM25Weight::get_maxextra() const
 }
 
 }
+
```
jaylett: You can kill this now (just a tiny fixup commit will do it).

xapian-core/tests/api_weight.cc
jaylett: Looking at the PR overall, I can't see these other tests in `api_nodb.cc` -- I can see some serialisation checks in the `weight1` testcase, which use variables `bm25plusweight` and `bm25plusweight2`, but these shouldn't conflict with the testcases you're adding here.

ivmarkp: You are correct. Sorry, didn't really catch that earlier. I'm naming the testcases in `api_weight.cc` as `bm25plusweight1`, `bm25plusweight2` and so on.

xapian-core/weight/bm25weight.cc
ivmarkp: Done.

xapian-core/api/registry.cc
```diff
@@ -307,3 +309,4 @@ Registry::get_lat_long_metric(const string & name) const
 }
 
 }
+
```
jaylett: Uh-oh — there's another one here to remove :)

xapian-core/include/xapian/weight.h
```diff
@@ -1239,3 +1334,4 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
 }
 
 #endif // XAPIAN_INCLUDED_WEIGHT_H
+
```
jaylett: And here…

jaylett:
@ojwb other than a couple of trailing newlines, this looks good to me, but I'd appreciate your eyes over it if you have a moment. Builds cleanly here, and tests pass.

ivmarkp:
@jaylett Done :)

xapian-bindings/python/doxy2swig.py
```diff
@@ -39,7 +40,7 @@ def my_open_write(dest):
     if hasattr(dest, "write"):
         return dest
     else:
-        return open(dest, 'w')
+        return io.open(dest, 'w', encoding = 'utf8')
 
 
 class Doxy2SWIG:
```
ojwb: This is a useful change, but not really related to the stated purpose of the PR ("Add support for BM25+ weighting function") - it just happens this change adds the first instance of a non-ASCII character to the C++ documentation comments (or at least to those which are taken through to Python).  So it would arguably be better to deal with separately.

It's also incomplete - the same issue affects Python3.

And for Python2, even with the fix in this commit I get an error when the Python interpreter tries to import the wrapper due to the default encoding for Python2 being ASCII (since 2.5 - before that it was latin-1 which wouldn't give an error, but wouldn't work as intended):

```
/usr/bin/python2 -c 'import xapian'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "xapian/__init__.py", line 3099
SyntaxError: Non-ASCII character '\xc3' in file xapian/__init__.py on line 3100, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

Python3's default source encoding if UTF-8, so no equivalent change is needed.

So I've cherry-picked this commit as 215c8cce2df208a0b3cea3ed5f2ec410b340de90, with the identical fix applied for Python3, and then committed c277f4422cfa15aeefa4c7692c16f3763c0af68e to specify the source ending for Python2.

xapian-core/api/registry.cc
```diff
@@ -105,7 +105,7 @@ register_object(map<string, T*> & registry, const T & obj)
 
     r.first->second = clone;
 }
- 
+
```
ojwb: It's generally unhelpful to include whitespace changes (even if they are to conform to what's expected) amongst functional changes as it makes those functional changes harder to review.  So please avoid doing this in future patches - send a separate patch/PR for the whitespace cleanup instead.

xapian-core/include/xapian/weight.h
```diff
@@ -590,6 +590,101 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
 
+/// Xapian::Weight subclass implementing the BM25+ probabilistic formula.
+class XAPIAN_VISIBILITY_DEFAULT BM25PlusWeight : public BM25Weight {
+
+  public:
+    /** Construct a BM25PlusWeight.
+     *
+     *  @param delta  A parameter for pseudo tf value to control the scale
+     *		      of the tf lower bound. Delta(δ) can be tuned for e.g
+     *		      from 0.0 to 1.5 but BM25+ can still work effectively
```
ojwb: "e.g." is from the Latin "exempli gratia" and means "for example", so "for e.g." would be "for for example"...

xapian-core/include/xapian/weight.h
```diff
@@ -590,6 +590,101 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+	need_stat(WDF);
+	need_stat(WDF_MAX);
+	if (param_k2 != 0 || (param_k1 != 0 && param_b != 0)) {
+	    need_stat(DOC_LENGTH_MIN);
+	    need_stat(AVERAGE_LENGTH);
+	}
+	if (param_k1 != 0 && param_b != 0) need_stat(DOC_LENGTH);
+	if (param_k2 != 0) need_stat(QUERY_LENGTH);
+	if (param_k3 != 0) need_stat(WQF);
+	if (param_delta < 0) param_delta = 0;
```
ojwb: The pattern of the existing code is to first clamp all the parameters to valid values, and then do all the unconditional `need_stat()` calls, and then do all the conditional `need_stat()` calls, so this clamping of `param_delta` would fit better higher up.

xapian-core/tests/api_weight.cc
```diff
@@ -134,6 +134,66 @@ DEFINE_TESTCASE(bm25weight5, backend) {
...
+    Xapian::BM25PlusWeight * b2 = b.unserialise(wt.serialise() + "X");
+    // Make sure we actually use the weight.
+    bool empty = b2->name().empty();
+    delete b2;
+    if (empty)
+	FAIL_TEST("Serialised BM25PlusWeight with junk appended unserialised to empty name!");
+    FAIL_TEST("Serialised BM25PlusWeight with junk appended unserialised OK");
+    } catch (const Xapian::SerialisationError &) {
+    // Good!
+    }
```
ojwb: The bodies of both the `try` and `catch` blocks are under-indented here (unlike the code of the `bm25weight3` testcase which this seems to be based on).

xapian-core/weight/bm25plusweight.cc
```diff
@@ -0,0 +1,157 @@
+/** @file bm25plusweight.cc
+ * @brief Xapian::BM25PlusWeight class - the BM25+ probabilistic formula
+ */
+/* Copyright (C) 2016  Vivek Pal
```
ojwb: This file is clearly adapted from `bm25weight.cc`, which is fine except that means it's not correct to list only yourself as the copyright holder here.

xapian-core/weight/bm25plusweight.cc
```diff
@@ -0,0 +1,157 @@
...
+    termweight = log(tw) * factor;
+    if (param_k3 != 0) {
+	double wqf_double = get_wqf();
+	termweight *= (param_k3 + 1) * wqf_double / (param_k3 + wqf_double);
+    }
+    termweight *= (param_k1 + 1);
+
+    LOGVALUE(WTCALC, termweight);
+
+}
```
ojwb: It would be better not to have a blank line at the end of the function body here.

xapian-core/weight/bm25plusweight.cc
```diff
@@ -0,0 +1,157 @@
...
+    double min_normlen = unserialise_double(&ptr, end);
+    double delta = unserialise_double(&ptr, end);
+    if (rare(ptr != end))
+	throw Xapian::SerialisationError("Extra data in BM25PlusWeight::unserialise()");
+    return new BM25PlusWeight(k1, k2, k3, b, min_normlen, delta);
+}
+
+double
+BM25PlusWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount len,
+							Xapian::termcount) const
```
ojwb: There's too much indentation here - `Xapian::termcount` should be aligned with the `Xapian::termcount` at the start of the parameter list on the line above.

xapian-core/tests/api_weight.cc
```diff
@@ -134,6 +134,66 @@ DEFINE_TESTCASE(bm25weight5, backend) {
...
+    mset = enquire.get_mset(0, 10);
+    TEST_EQUAL(mset.size(), 5);
+    // Expect: neither wdf nor doclen affects weight.
+    TEST_EQUAL_DOUBLE(mset[0].get_weight(), mset[4].get_weight());
+
+    return true;
+}
+
+// Test to make sure k2 has no effect on weight. This is necessary because
+// k2 is associated with extra weight component and BM25+ has such no extra weight.
```
ojwb: The BM25+ formula doesn't use `k2`, so does it actually make sense to pass it as a parameter to this class at all?

xapian-core/include/xapian/weight.h
```diff
@@ -590,6 +590,101 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+    /** Construct a BM25PlusWeight.
+     *
+     *  @param delta  A parameter for pseudo tf value to control the scale
+     *		      of the tf lower bound. Delta(δ) can be tuned for e.g
+     *		      from 0.0 to 1.5 but BM25+ can still work effectively
+     *		      across collections with a fixed δ = 1.0. (default 1.0)
+     */
+    BM25PlusWeight(double k1, double k2, double k3, double b, double min_normlen, double delta)
+	: param_k1(k1), param_k2(k2), param_k3(k3), param_b(b),
+	  param_min_normlen(min_normlen), param_delta(delta)
```
ojwb: The documentation comment only covers parameter `delta`, and not any of the parameters which are in common with `BM25Weight` - to find out what they mean the user would have to guess to go and look at the documentation for that.  It would be better to just copy over the documentation for the common parameters (except that `k2` isn't actually used by BM25+).

xapian-core/include/xapian/weight.h
```diff
@@ -590,6 +590,101 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
     double get_maxextra() const;
 };
 
+/// Xapian::Weight subclass implementing the BM25+ probabilistic formula.
+class XAPIAN_VISIBILITY_DEFAULT BM25PlusWeight : public BM25Weight {
+
+  public:
+    /** Construct a BM25PlusWeight.
```
ojwb: Conventionally we don't have a blank line between the start of the class and `public:` like you have here in the Xapian codebase.

xapian-core/weight/bm25plusweight.cc
ivmarkp: @ojwb Thanks for the review. I have addressed all the issues highlighted by you in the recent commits.

xapian-bindings/python/doxy2swig.py
ivmarkp: Thanks. Is it okay to leave this change in this PR or remove it now?

ojwb: Probably the final step before merging will be to rebase to squash the "fixup" commits onto the commits they fixup - I'd suggest we rebase onto git master as part of that, which should automatically drop this commit as already applied (and similarly should drop the trailing whitespace change that's in another commit).

xapian-core/weight/bm25plusweight.cc
```diff
@@ -110,7 +109,7 @@ BM25PlusWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount len,
     // Parameter delta (δ) is a pseudo tf value to control the scale of the
     // tf lower bound. δ can be tuned for e.g from 0.0 to 1.5 but BM25+ can
     // still work effectively across collections with a fixed δ = 1.0
-    RETURN((termweight * (wdf_double / denom)) + param_delta);
+    RETURN((termweight * ((((param_k1 + 1) * wdf_double) / denom) + param_delta)));
 }
```
ojwb: There's now a `((` and `))` around the whole return expression, which just makes it harder to read the expression - this means exactly the same:

```
RETURN(termweight * ((((param_k1 + 1) * wdf_double) / denom) + param_delta));
```

But also `(x * y) / z` and `x * y / z` mean the same since `*` and `/` have the same precedence (http://en.cppreference.com/w/cpp/language/operator_precedence) and `+` has lower precedence, so you could actually lose another two pair of parentheses:

```
RETURN(termweight * ((param_k1 + 1) * wdf_double / denom + param_delta));
```

This step's more arguable - some people prefer explicit parentheses, especially when there are operators with different precedence involved.  But let's definitely lose the extra outer pair (and similarly below).

Also in a case like this where the initial implementation has a bug, it's a really good idea to add a regression testcase which will fail if the fix ever got undone.  Bugs have a tendency to recur - if the code got misimplemented in a particular way, it's natural that someone reworking it may make the same mistake again, and it's also good not to repeat the same mistake as it looks bad to users, management, etc - and adding regression testcases is a big help in preventing that.

xapian-core/include/xapian/weight.h
```diff
@@ -590,6 +590,118 @@ class XAPIAN_VISIBILITY_DEFAULT BM25Weight : public Weight {
...
+		       Xapian::termcount uniqterm) const;
+    double get_maxpart() const;
+
+    double get_sumextra(Xapian::termcount doclen,
+			Xapian::termcount uniqterms) const;
+    double get_maxextra() const;
+
+  protected:
+    /// Factor to multiply the document length by.
+    mutable Xapian::doclength len_factor;
```
ojwb: This member is never set, but does get used, which causes the testsuite to fail when run under valgrind (you really ought to have valgrind installed when doing Xapian development, as the testsuite hooks into it and detects such issues for you).

I think you need to keep the code from `bm25weight.cc` which initialises `len_factor`, and just adjust it to reflect that `param_k2` is effectively always zero for BM25+.

It would be really good to have some testcases which compare the weights for a simple case with expected weights worked out by hand from the original formulae - that would help to avoid problems like this (and should serve as a regression test for the other issue you fixed recently).

ivmarkp: I've been working on a different machine lately that didn't have valgrind installed until now; just installed it. Sorry, will be careful now.

xapian-core/tests/api_weight.cc
ivmarkp: I'm wondering if removing this test case would be fine as this is not useful after 212fbb0?

ojwb:
Thanks for the quick response - all looks good to me now.

@jaylett suggested on IRC that it was probably best to just squash this into one commit when merging to master, so I've done that and pushed it.

PR #106, opened 2016-06-08
letor: introduce code, refactor and clean-up #1

ayshtmr:
This PR does the following:

1. Introduce code mainly till features portion from v-hasu’s gsoc2014 branch
2. Refactor FeatureVector class by creating featurevector_internal.h and featurevector_internal.cc in api/
3. Delete non-functional letor_internal_refactored.cc
4. Minor refactoring of certain methods
5. Clean-up of dead code


xapian-letor/api/featurevector_internal.cc
```diff
@@ -41,26 +41,17 @@
...
-//     return c1.score < c2.score;
-// }
-
 map<string, map<string, int> >
 FeatureVector::Internal::load_relevance(const std::string & qrel_file)
 {
-    typedef map<string, int> Map1;      // docid and relevance judjement 0/1
-    typedef map<string, Map1> Map2;     // qid and map1
-    Map2 qrel;
+    map<string, map<string, int>> qrel1;     // < qid, <docid, relevance_judgement> >
```
jaylett: Is there a reason this has been renamed to `qrel1`?

xapian-letor/api/featurevector_internal.cc
```diff
@@ -41,26 +41,17 @@
...
 
     string inLine;
-    ifstream myfile(qrel_file.c_str(), ifstream::in);
+    ifstream myfile (qrel_file.c_str(), ifstream::in);
```
jaylett: I think we should avoid spaces in this situation.

xapian-letor/api/featuremanager_internal.cc
```diff
@@ -125,6 +125,9 @@ FeatureManager::Internal::load_relevance(const std::string & qrel_file)
 
     string inLine;
     ifstream myfile (qrel_file.c_str(), ifstream::in);
+    if(!myfile.good()){
+        cout << "No Qrel file found" << endl;
```
jaylett: This should probably raise an exception instead. It certainly can't proceed, can it?

xapian-letor/api/featuremanager.cc
```diff
@@ -69,9 +69,9 @@ FeatureManager::getlabel(map<string, map<string, int> > qrel2, const Document &d
 }
 
 Xapian::RankList
-FeatureManager::create_rank_list(const Xapian::MSet & mset, std::string & qid)
+FeatureManager::create_rank_list(const Xapian::MSet & mset, std::string & qid, int TrainorTest)
```
jaylett: Formal parameters should be in lower-case, really (`trainor_test`); also, I'm not sure from the name what this does.

xapian-letor/api/letor_internal.cc
```diff
@@ -262,15 +262,62 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
 
-    list<Xapian::RankList> l;
+    vector<Xapian::RankList> l;
 
     string str1;
     ifstream myfile1;
     myfile1.open(queryfile.c_str(), ios::in);
 
+    if(!myfile1.good()){
+        cout << "No Query file found"<<endl;
```
jaylett: Should raise an exception.

xapian-letor/api/letor_internal.cc
```diff
@@ -180,6 +180,12 @@ static void read_problem(const char *filename) {
 	}
 	++elements;
 	++prob.l;
+vector<Xapian::RankList>
```
jaylett: This commit looks funky, I think because the one that tidies up the old code around here comes after it. It looks okay in the overall diff, and we're going to collapse these all down once we're happy anyway, so it shouldn't be a problem.

xapian-letor/api/letor.cc
```diff
@@ -73,13 +72,13 @@ Letor::prepare_training_file(const string & query_file, const string & qrel_file
...
 
 void
-Letor::create_ranker(int ranker_type, int metric_type) {
-    switch(ranker_type) {
-        case 0: internal->ranker = * new SVMRanker;
-                break;
-        case 1: break;
-        default: ;//cout<<"Please specify proper ranker.";
-    }
+Letor::create_ranker(int ranker_type, int metric_type) { //TODO: update when adding rankers
```
jaylett: I'm pretty sure as we re-introduce multiple rankers, that we shouldn't have this function here at all. Client code should just create the ranker object is needs. Command line utilities would be better off having `--ranker=<name>` (which will result in code a little like this, but will generally be more clear). So it may be worth going through and eliminating this function entirely, fairly soon.

xapian-letor/include/xapian-letor/letor.h
```diff
@@ -61,7 +61,7 @@ class XAPIAN_VISIBILITY_DEFAULT Letor {
      *
      *  @return Letor score corresponding to each document in mset as map<docid, score> format.
      */
-    std::map<Xapian::docid, double> letor_score(const Xapian::MSet & mset);
+    std::vector<Xapian::docid> letor_rank(const Xapian::MSet & mset);
```
jaylett: With this change, the doccomment above is incorrect.

xapian-letor/api/featurevector_internal.cc
```diff
@@ -0,0 +1,129 @@
+/* featurevector.cc: The file responsible for transforming the document into the feature space.
```
jaylett: The filename is wrong in this comment :)

jaylett:
@ayshtmr I haven't had a chance to build this branch, but there are a few comments that you should address quickly before moving on. Ignore the travis build failure; it's trying to build without knowing what it's doing (I'm hoping to get that working properly in the next couple of weeks).

xapian-letor/api/letor.cc
ayshtmr: I agree. Marking this as a todo. Will get this done for both rankers and scorers, while integrating ListNET and NDCGScore.

xapian-letor/api/featurevector_internal.cc
ayshtmr: No. I probably didn't notice it. Will change it back.

xapian-letor/api/featuremanager.cc
ayshtmr: It's a check argument for training or testing. While training (TrainorTest == 1), if any doc in mset has no label associated (label == -1), then no featurevector is computed for it.
I'll change the name to lower-case (train_or_test). Even in camelcase it should have been TrainOrTest. Thanks for pointing it out.

xapian-letor/api/letor_internal.cc
ayshtmr: It indeed is funky. The reason I separated it from the commit that tidies up the code was because of the whitespace issues you pointed out in my mail on the mailing list. (https://github.com/ayshtmr/xapian/commit/2abaf6b8d64b1624699c3f490b6dc441e1c6cdf3)
It was probably an issue with my editor configuration, but I couldn't understand what was happening even after indentation setting were set fine. Hence, I separated the commits so that it looks fine in the overall diffs. I know it's not a good practise, though.

xapian-letor/api/featuremanager_internal.cc
ayshtmr: Will update.

ayshtmr:
@jaylett Fixed the issues. Please review.

xapian-letor/api/featuremanager.cc
jaylett: Then shouldn't it be a `bool` argument? And if it's boolean, you don't need to call it `train_or_test`, you can just call it `train` (and explain the distinction in the docstring).

ayshtmr: Yes, it should be bool. Making the change.

xapian-letor/api/featuremanager_internal.cc
```diff
@@ -37,6 +37,7 @@
 using namespace Xapian;
 using namespace std;
 
+struct FileNotFound { };
```
jaylett: Note for future: I'm guessing we'll end up with a set of letor exceptions, with their own root, at which point this can move out to its own file and probably become a class. (Whether this inherits from the xapian-core root exception is something we'll have to discuss when we have some more experience working with the API.)

xapian-letor/api/featurevector_internal.cc
```diff
@@ -0,0 +1,119 @@
...
+FeatureVector::Internal::get_did() {
+    return did;
+}
+
+double
+FeatureVector::Internal::get_feature_value(int index) {
+    map<int,double>::const_iterator iter;
+    iter = fvals.find(index);
+    if(iter == fvals.end())
+    return 0;
```
jaylett: Some alignment issues in this function (and a missing space after `if`); worth checking the entire PR for similar remaining things, as I'll probably miss them even if I try to point them all out :)

xapian-letor/api/featuremanager_internal.cc
ayshtmr: Noted. This approach has already started to feel messy.

xapian-letor/api/featurevector_internal.cc
ayshtmr: Noted.

ayshtmr:
@jaylett Please hold merging of this PR for now. Apart from the additions you suggested, there have been some portions where I later realised the scope of improvement. I should probably be more careful before opening PRs.

jaylett:
@ayshtmr cool, I'll wait for you. It's not a problem to have PRs open for a while, providing you keep on introducing changes from master (so you don't get lots of merge conflict work later), and in the case of letor that probably doesn't even apply. Obviously getting stuff merged is the aim, but getting the right stuff merged is always better :-)

xapian-letor/api/featurevector_internal.h
```diff
@@ -0,0 +1,76 @@
...
+using namespace std;
+
+namespace Xapian{
+
+class FeatureVector::Internal : public Xapian::Internal::intrusive_base
+{
+    friend class FeatureVector;
+    double label;
+    double score;
+    std::map<int,double> fvals;
```
parthg: I would suggest to use logical names for the features than simple integers e.g. Enums. Usually a good way is to use different ranges for different types such as 100s for query (Q) related, 200s for document (D) related and 300s for the T and D related. That way it would be easier to add more features later.

xapian-letor/api/letor.cc
parthg: Though, it is a good idea to have defaults like NDCG with ListNET.

xapian-letor/api/letor_internal.cc
```diff
@@ -204,6 +210,68 @@ static void read_problem(const char *filename) {
...
+        std::map<int,double> fvals;//read features
+        for(int i = 1; i < 20; ++i){
+            train_file.ignore();
+            int feature_index;
+            double feature_value;
+            train_file >> feature_index;
+            train_file.ignore();
+            train_file >> feature_value;
+            //std::cout<<feature_index<<" "<<feature_value<<endl;
+            fvals[feature_index] = feature_value;
```
parthg: If we go with the Enum style feature references, which I strongly believe soon we should, it should be resolved with the reverse map/dictionary of features with ids.

xapian-letor/api/letor_internal.cc
```diff
@@ -262,15 +262,62 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+    		break;
+    	    string temp;
+    	    temp.append(title);
+    	    temp.append(t);
+    	    temp.append(" ");
+    	    temp.append(qq);
+    	    qq = temp;
+    	}
+
+    	//cout << "Processing Query: " << qq << "\n";
```
parthg: Such lines should be removed.

xapian-letor/api/letor_internal.cc
```diff
@@ -262,15 +262,62 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+        cout << "No Query file found"<<endl;
+    }
+
 
     while (!myfile1.eof()) {           //reading all the queries line by line from the query file
 
+    	getline(myfile1, str1);
+
+    	if (str1.empty()) {
+            //cout<< "str1 empty";
```
parthg: Such intermediate debugs should be removed.

xapian-letor/api/featuremanager_internal.cc
```diff
@@ -89,16 +90,19 @@ FeatureManager::Internal::create_rank_list(const Xapian::MSet & mset, std::strin
         double weight = i.get_weight();
 
         map<int,double> fVals = transform(doc, weight);
```
parthg: `transform()` is the method which will benefit the most from `feature_type` introduction and also the method names `calculate_fx()` therein.

xapian-letor/api/featurevector_internal.h
jaylett: Another approach to this is to make the `FeatureVector` into an actual vector of feature values, since (once calculated for each document) it doesn't actually matter how they were calculated, just that the vectors are the same in the system. That would allow for more flexibility by users, who could configure the system (from training through to use) with the feature generators they actually want to use. (With a default set that is our 'recommended' approach.) One of the big advantages there is that we wouldn't have to assign numbers universally, so we wouldn't have to worry about where to put experimental features, for instance.

Alternatively, if we do need ids for them, we could use URIs and have a registry which converts them to numbers (somehow). That again would be within the system.

Hopefully as we progress across the summer it'll become clear which route is more appropriate. If we do continue using identifiers that are numeric at a low level, then definitely using an `enum` type would be an improvement.

xapian-letor/api/ranklist.cc
```diff
@@ -1,101 +1,83 @@
...
-	    for (norm_inner = norm_outer->second.begin();norm_inner!=norm_outer->second.end();++norm_inner) {
-		if (max!=0)      // sometimes value for whole feature is 0 and hence it may cause 'divide-by-zero'
-		    *norm_inner /= max;
-		k++;
-	    }
-	}
 
+    std::vector<FeatureVector> local_fvv = this->fvv;
+
+    // find the max value for each feature gpr all the FeatureVectors in the RankList rl.
```
parthg: gpr?

xapian-letor/api/ranklist.cc
```diff
@@ -1,101 +1,83 @@
...
+    int num_features = 19;
```
parthg: Let's not hard-code it.

xapian-letor/api/ranklist.cc
```diff
@@ -1,101 +1,83 @@
...
+    double temp = 0.0;
+    double max[num_features];
+
+    for(int i=0; i<19; ++i)
```
parthg: Again, get features size from somewhere rather than hard-coding it.

xapian-letor/api/ranklist.cc
```diff
@@ -1,101 +1,83 @@
...
+    max[i] = 0.0;
+
+    int num_fv = local_fvv.size();
+    for(int i=0; i < num_fv; ++i) {
+        for(int j=0; j<19; ++j) {
```
parthg: Again, total feature size.

xapian-letor/api/ranklist.cc
```diff
@@ -1,101 +1,83 @@
...
+        }
     }
-    return norm;
+
+    /* We have the maximum value of each feature overall.
+       Now we need to normalize each feature value of a featureVector by dividing it by the corresponding max of the feature value
     */
+
+    for(int i=0; i < num_fv; ++i) {
+        for(int j=0; j<19; ++j) {
```
parthg: Again, total feature size.

xapian-letor/api/featurevector_internal.h
parthg: Right, either way would also use to allow users to specify a subset of features they want to work with.

xapian-letor/api/letor.cc
jaylett: @parthg certainly the CLI programs should have sensible defaults; the API will either (depending on what feels right) do something similar (as we do for instance with weighting schemes in core, which almost certainly will be appropriate for scorers), or you'll just instantiate whichever you need (which may make sense for rankers, depending on exactly what the API looks like).

Are you saying that the scorer default should be dependent on the ranker? That's obviously not a problem, and it may help settle the best direction for the API.

ayshtmr:
@jaylett As you have been saying, I think merging this PR would make more sense than to let it continue to grow. Following this, I'll be opening two PRs, one which performs some refactoring, documentation addition, cleanup and added support for exception class over this PR, and one that adds tests.

jaylett:
@ayshtmr can you squash & rebase this onto master, please? It's not doing so automatically, so I suspect some of your fixups are colliding when out of order.

ayshtmr:
@jaylett Done.

PR #107, opened 2016-06-13
Clustering base classes and similarity measure

richhiey1996:
This PR contains the classes which will be required by the clusterer to implement KMeans

### DIFF
--- a/xapian-core/cluster/Makefile.mk
+++ b/xapian-core/cluster/Makefile.mk
@@ -1,0 +1,7 @@
+EXTRA_DIST +=\
+	cluster/Makefile
+
+lib_src +=\
+	cluster/cluster.cc\
+	cluster/euclidian_sim.cc\
+	cluster/round_robin.cc
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -1,0 +1,223 @@
+/** @file cluster.cc
+ *  @brief Cluster API
+ */
+/* Copyright (C) 2010 Richard Boulton
+ * Copyright (C) 2016 Richhiey Thomas
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
+#include <config.h>
+
+#include "xapian/cluster.h"
+
+#include <xapian/error.h>
+#include <api/termlist.h>
+#include <debuglog.h>
+
+#include <vector>
+#include <map>
+
+using namespace Xapian;
+using namespace std;
+
+MSetDocumentSource::MSetDocumentSource(const MSet &mset_)
+    :mset(mset_), maxitems(mset_.size()), index(0)
+{
+    LOGCALL_CTOR(API,"MSetDocumentSource", mset_);
+    mset.fetch();
+}
+
+MSetDocumentSource::MSetDocumentSource(const MSet & mset_, doccount maxitems_)
+    :mset(mset_), maxitems(maxitems_), index(0)
+{
+    LOGCALL_CTOR(API, "MSetDocumentSource", mset | maxitems);
+    if (maxitems > mset.size())
+	maxitems = mset.size();
+
+    if (maxitems > 0)
+	mset.fetch(mset.begin(), mset[maxitems - 1]);
+}
+
+Document
+MSetDocumentSource::next_document() {
+    LOGCALL(API, Document, "MSetDocumentSource::next_document()", "");
+    return (mset[index++].get_document());
+}
+
+bool
+MSetDocumentSource::at_end() const {
+    LOGCALL(API, bool, "MSetDocumentSource::at_end()", "");
+    return (index >= maxitems);
+}
+
+doccount
+MSetDocumentSource::size() {
+    LOGCALL(API, doccount, "MSetDocumentSource::size()", "");
+    return mset.size();
+}
+
+Similarity::~Similarity()
+{
+    LOGCALL_DTOR(API, "Similarity");
+}
+
+FreqSource::~FreqSource()
+{
+    LOGCALL_DTOR(API, "FreqSource");
+}
+
+doccount
+DummyFreqSource::get_termfreq(const string &) {
+    LOGCALL(API, doccount, "DummyFreqSource::get_termfreq()", "");
+    return 1;
+}
+
+doccount
+DummyFreqSource::get_doccount() {
+    LOGCALL(API, doccount, "DummyFreqSource::get_doccount()", "");
+    return 1;
+}
+
+void
+TermListGroup::add_document(const Document &document) {
+    LOGCALL_VOID(API, "TermListGroup::add_document()", document);
+
+    TermIterator titer(document.termlist_begin());
+    TermIterator end(document.termlist_end());
+
+    for (; titer != end; titer++) {
+	map<string, doccount>::iterator i;
+	i = termfreq.find(*titer);
+	if (i == termfreq.end())
+	    termfreq[*titer] = 1;
+	else
+	    i->second += 1;
+    }
+}
+
+void
+TermListGroup::add_documents(DocumentSource &docs) {
+    LOGCALL_VOID(API, "TermListGroup::add_documents()", docs);
+    while (!docs.at_end()) {
+	Document document = docs.next_document();
+	add_document(document);
+    }
+    docs_num = docs.size();
+}
+
+doccount
+TermListGroup::get_doccount() {
+    LOGCALL(API, doccount, "TermListGroup::get_doccount()", "");
+    return docs_num;
+}
+
+doccount
+TermListGroup::get_termfreq(const string &tname) {
+    LOGCALL(API, doccount, "TermListGroup::get_termfreq()", tname);
+    return termfreq[tname];
+}
+
+doccount
+ClusterSet::num_of_clusters() {
+    LOGCALL(API, doccount, "ClusterSet::num_of_clusters()", "");
+    return clusters.size();
+}
+
+doccount
+ClusterSet::cluster_size(clusterid cid) {
+    LOGCALL(API, doccount, "ClusterSet::cluster_size()", cid);
+    map<clusterid, vector<Document> >::iterator i;
+    i = clusters.find(cid);
+    if (i == clusters.end())
+	throw RangeError("The mentioned clusterid was out of range", 103);
+    return ((i->second).size());
+}
+
+DocumentSet
+ClusterSet::get_cluster(clusterid cid) {
+    LOGCALL(API, DocumentSet, "ClusterSet::get_cluster()", cid);
+    map<clusterid, vector<Document> >::iterator i;
+    i = clusters.find(cid);
+    if (i == clusters.end())
+	throw RangeError("The mentioned clusterid was out of range", 103);
+    vector<Document> docs;
+    docs = i->second;
+    DocumentSet docset;
+    for (vector<Document>::iterator k = docs.begin(); k != docs.end(); k++)
+	docset.add_document(*k);
+    return docset;
+}
+
+void
+ClusterSet::add_document(clusterid cid, Document &doc) {
+    LOGCALL_VOID(API, "ClusterSet::add_document()", cid | doc);
+    clusters[cid].push_back(doc);
+}
+
+DocumentSet
+ClusterSetIterator::get_cluster() {
+    LOGCALL(API, DocumentSet, "ClusterSetIterator::get_cluster()", "");
+    return cset.get_cluster(index);
+}
+
+ClusterSetIterator
+ClusterSet::begin() {
+    LOGCALL(API, ClusterSetIterator, "ClusterSet::begin()", "");
+    return ClusterSetIterator(*this, 0);
+}
+
+ClusterSetIterator
+ClusterSet::end() {
+    LOGCALL(API, ClusterSetIterator, "ClusterSet::end()", "");
+    return ClusterSetIterator(*this, num_of_clusters());
+}
+
+ClusterSetIterator
+ClusterSet::operator[](doccount i) {
+    return ClusterSetIterator(*this, i);
+}
+
+void
+DocumentSet::add_document(Document &doc) {
+    LOGCALL_VOID(API, "DocumentSet::add_document()", doc);
+    docs.push_back(doc);
+}
+
+Document
+DocumentSet::operator[](doccount i) {
+    Document d = docs[i];
+    return d;
+}
+
+DocumentSetIterator
+DocumentSet::begin() {
+    LOGCALL(API, DocumentSetIterator, "DocumentSet::begin()", "");
+    return DocumentSetIterator(*this, 0);
+}
+
+DocumentSetIterator
+DocumentSet::end() {
+    LOGCALL(API, DocumentSetIterator, "DocumentSet::end()", "");
+    return DocumentSetIterator(*this, size());
+}
+
+Document
+DocumentSetIterator::get_document() {
+    LOGCALL(API, Document, "DocumentSetIterator::get_document()", "");
+    Document d = docs[index];
+    return d;
+}
--- a/xapian-core/cluster/euclidian_sim.cc
+++ b/xapian-core/cluster/euclidian_sim.cc
@@ -1,0 +1,118 @@
+/** @file euclidian_sim.cc
+ *  @brief Document similarity calculation
+ */
+/* Copyright (C) 2010 Richard Boulton
+ * Copyright (C) 2016 Richhiey Thomas
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
+#include <config.h>
+
+#include "xapian/cluster.h"
+
+#include <debuglog.h>
+#include <api/termlist.h>
+
+#include <vector>
+#include <map>
+#include <cmath>
+
+using namespace std;
+using namespace Xapian;
+
+string
+EuclidianDistance::get_description() {
+    LOGCALL(API, string, "EuclidianDistance::get_description()", "");
+    return "Euclidian Distance metric";
+}
+
+double
+EuclidianDistance::similarity(TermListGroup tlg,
+			      TermIterator a_begin,
+			      TermIterator a_end,
+			      TermIterator b_begin,
+			      TermIterator b_end)
+{
+    LOGCALL(API, double, "EuclidianDistance::similarity()", tlg | a_begin | a_end | b_begin | b_end);
+    vector<pair<string, double> > docvec1;
+    vector<pair<string, double> > docvec2;
+    docvec1.clear();
+    docvec2.clear();
+
+    double doccount = tlg.get_doccount();
+    TermIterator titer = a_begin;
+
+    // Creating document vector for first document
+    for (; titer != a_end; titer++) {
+	double tf, wt;
+	tf = 1 + log(titer.get_wdf());
+	double idf;
+	double termfreq = tlg.get_termfreq(*titer);
+	idf = log(doccount/termfreq);
+	wt = tf*idf;
+	docvec1.push_back(make_pair(*titer, wt));
+    }
+
+    titer = b_begin;
+    // Creating document vector for the second document
+    for (; titer != b_end; titer++) {
+	double tf, wt;
+	tf = 1 + log(titer.get_wdf());
+	double idf;
+	double termfreq = tlg.get_termfreq(*titer);
+	idf = log(doccount/termfreq);
+	wt = tf*idf;
+	docvec2.push_back(make_pair(*titer, wt));
+    }
+
+    vector<pair<string, double> >::iterator ta;
+    vector<pair<string, double> >::iterator tb;
+    double result = 0;
+
+    // Finding euclidian distance between the above document vectors
+    for (ta = docvec1.begin(); ta != docvec1.end(); ta++) {
+	bool found = false;
+	double sum = 0;
+	for (tb = docvec2.begin(); tb != docvec2.end(); tb++) {
+	    if (tb->first == ta->first) {
+		found = true;
+		break;
+	    }
+	}
+
+	if (found)
+	    sum += (tb->second - ta->second)*(tb->second - ta->second);
+	else
+	    sum += (ta->second)*(ta->second);
+	result += sum;
+    }
+
+    for (ta = docvec2.begin(); ta != docvec2.end(); ta++) {
+	bool found = false;
+	double sum = 0;
+	for (tb = docvec1.begin(); tb != docvec1.end(); tb++) {
+	    if (tb->first == ta->first) {
+		found = true;
+		break;
+	    }
+	}
+	if (!found)
+	    sum += (ta->second)*(ta->second);
+	result += sum;
+    }
+    return result;
+}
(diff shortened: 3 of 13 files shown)


xapian-core/cluster/Makefile.mk
```diff
@@ -0,0 +1,6 @@
+EXTRA_DIST +=\
+	weight/Makefile
```
jaylett: You mean `cluster/Makefile` here!

xapian-core/include/xapian/cluster.h
```diff
@@ -0,0 +1,199 @@
+/** @file cluster.h
+ *  @brief Cluster API
+ */
+
+#ifndef XAPIAN_INCLUDED_CLUSTER_H
+#define XAPIAN_INCLUDED_CLUSTER_H
+
+#if !defined XAPIAN_IN_XAPIAN_H && !defined XAPIAN_LIB_BUILD
+# error "Never use <xapian/database.h> directly; include <xapian.h> instead."
```
jaylett: Error message should identify this as `xapian/cluster.h`.

xapian-core/cluster/Makefile.mk
richhiey1996: Aah yes, that was rectified! Saw it too late!

xapian-core/include/xapian/cluster.h
```diff
@@ -0,0 +1,199 @@
...
+};
+
+class ClusterSetIterator;
+
+// Class for storing the results returned by the Clusterer
+class XAPIAN_VISIBILITY_DEFAULT ClusterSet {
+	friend class ClusterSetIterator;
+  public:
+	// A map storing the clusterid and its corresponding Document cluster
+	std::map<clusterid, std::vector<Document> > clusters;
```
jaylett: Shouldn't this be a private member?

xapian-core/include/xapian/cluster.h
richhiey1996: I'm sorry I didn't realize that. I'll change that

xapian-core/cluster/cluster.cc
```diff
@@ -0,0 +1,179 @@
...
+	return ClusterSetIterator(*this, num_of_clusters());
+}
+
+ClusterSetIterator
+ClusterSet::operator[](doccount i) {
+	return ClusterSetIterator(*this, i);
+}
+
+
+/*
```
jaylett: You shouldn't commit commented code, because it's not part of the "active" changes in this commit. If you use `git add -p` to selectively add changes in a file, it becomes easier to skip the things you don't want in a commit. (For completely new files, `git add -N` is "intent to add" the file, which means you can then do `git add -p`.)

xapian-core/cluster/euclidian_sim.cc
```diff
@@ -0,0 +1,75 @@
+#include <config.h>
```
jaylett: Please include a file header for every new file you add.

xapian-core/cluster/cluster.cc
```diff
@@ -0,0 +1,179 @@
+/** @file cluster.cc
+ *  @brief Xapian::Cluster API class
+ */
+
```
jaylett: This should also have a copyright statement; given this seems to take code from the previous cluster branch, it should include the copyright owner from there also.

xapian-core/include/xapian/cluster.h
```diff
@@ -0,0 +1,199 @@
...
+inline bool operator!=(const ClusterSetIterator &a, const ClusterSetIterator &b) {
+	return !(a == b);
+}
+
+class XAPIAN_VISIBILITY_DEFAULT Clusterer {
+	//Core clustering functionality
+	virtual ClusterSet do_cluster(std::vector<docid> &docs, TermListGroup &tlg);
+};
+
+class XAPIAN_VISIBILITY_DEFAULT KMeans : public Clusterer {
```
jaylett: You aren't implementing `KMeans` in this commit (or even this PR), so I don't think this should be here at all. (Same goes for the base class above, since it's impossible to tell if this is the right API design until we're looking at code that exercises it.)

xapian-core/cluster/euclidian_sim.cc
richhiey1996: I used this file to separate the implementations from cluster.cc. I've included the "xapian/cluster.h" which declares the EuclidianDistance class. So should I create a new header file for this with the classes used for document similarity calculation?

xapian-core/cluster/cluster.cc
richhiey1996: Yes I'll keep this in mind. I directly add the files to be committed so thats why this happened.

xapian-core/tests/api_cluster.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "apitest.h"
+#include "testutils.h"
+
+#include <string.h>
+#include <cstring>
+
+using namespace std;
+
+DEFINE_TESTCASE(euclidian_distance, backend) {
+	Xapian::Database db(get_database("etext.txt"));
```
jaylett: Alignment of this line is off :-)

xapian-core/tests/api_cluster.cc
```diff
@@ -0,0 +1,40 @@
...
+
+    Xapian::QueryParser qp;
+    qp.set_database(db);
+    qp.set_stemmer(stemmer);
+
+    Xapian::Enquire enq(db);
+    enq.set_query(qp.parse_query(query));
+    Xapian::MSet matches = enq.get_mset(0, 30);
+
+	Xapian::MSetDocumentSource docs(matches);
```
jaylett: And these. If your editor is set to display tab characters as 4 columns, then you'll struggle to make this work. Tab characters should align to 8 columns, and your editor should be configured to a 4 column alignment indent.

xapian-core/tests/api_cluster.cc
```diff
@@ -0,0 +1,40 @@
...
+	Xapian::TermListGroup tlg;
+	tlg.add_documents(docs);
+	Xapian::Document doc1 = matches[0].get_document();
+	Xapian::Document doc2 = matches[1].get_document();
+	Xapian::EuclidianDistance d;
+	
+	int sim = d.similarity(tlg, doc1.termlist_begin(), doc1.termlist_end(), doc2.termlist_begin(), doc2.termlist_end());
+
+	TEST(sim != 0);
```
jaylett: If you use a different test database (say `apitest_allterms`) then you can calculate exactly what the similarity should be using euclidian distance.

You probably also want a test that the similarity _is_ 0 between two identical documents.

richhiey1996: I did test that locally and it worked but I wasn't too sure of how to write a test. So I just assumed that two different documents will have a similarity greater than zero. I'll take care of this in the next commit. :+1:

xapian-core/cluster/Makefile.mk
jaylett: Running `make dist` will tell you if these kinds of things work. You can always push fixup commits to a PR after opening it.

xapian-core/cluster/euclidian_sim.cc
jaylett: No, not a header file: there's no info and copyright header in _this_ file, such as there is in our other source files.

xapian-core/cluster/cluster.cc
jaylett: You can create a fixup commit to remove this code by deleting it in your workspace and doing `git add -p` then `git commit --fixup`.

jaylett:
I'd like to get this to a "minimal" system as quickly as possible. One way to do this is to implement a "round-robin" clusterer before we get to KMeans. That's just a clusterer class that you construct with a single parameter (number of clusters), and which will assign documents in order to the clusters. (So the first document goes in the first cluster, up to the Nth document in the Nth cluster, then the (N+1)th document in the first cluster and so on.)

It sounds like this is pointless, but it (a) gives us a bit of experience of the cluster API, (b) gets something we can do overall testing of more easily, (c) shouldn't take very long to write. You can write an overall test of the cluster API easily because you know exactly where each document will land beforehand, without having to do independent verification of the KMeans algorithm (or similar for other clustering algos in future).

It will also mean that, with KMeans, we definitely have two clustering algorithms before we tackle anything more complex, which should help shake out how the API will look.

xapian-core/cluster/euclidian_sim.cc
richhiey1996: Oh I'm sorry I misread!

richhiey1996:
Yes that sounds good. A round robin clusterer shouldn't be too hard to implement.
For now, I'll start fixing up mistakes in the present commits like indentation, missing file headers and other glitches and push a fixup commit.
After that I can start working on the simple clusterer.
Should that be fine?

jaylett:
@richhiey1996 sounds good! 👍

richhiey1996:
Also, I was able to test this implementation only on a 20 document cluster which I indexed manually because I wasn't sure how to set up a database for the manual testing purposes.
Are there any pre-compiled Xapian datasets that I can use? And how do I use them?

jaylett:
@richhiey1996 there aren't pre-compliled datasets, but you can easily generate a db to play around with by using something like omindex against a load of documents (of the various formats it supports).

richhiey1996:
That sounds good! I'll see what I can do there :+1:

xapian-core/cluster/cluster.cc
```diff
@@ -1,6 +1,24 @@
 /** @file cluster.cc
  *  @brief Xapian::Cluster API class
  */
+/* Copyright (C) 2016 Richhiey Thomas
+ * Copyright (C) 2016 Richard Boulton
```
jaylett: Strictly, this should be from the commit date on Richard's work, not 2016.

richhiey1996:
Sorry to have taken so long to implement this. I had fallen ill in between.
I have implemented the RoundRobin clusterer as mentioned and modified the API with things that I had missed out on before.
I've tested the API locally and it seems to work fine providing an easy way to iterate through the clusters (atleast according to me). I've also written a test 'round_robin1' which shows how the API should be used with the RoundRobin clusterer. I'm not sure whether the naming I have used for the classes are apt, so I would specially like to have some feedback on that.
Though I've written the tests, I'm still having some problem with running my tests but I'll make sure that is dealt with as soon as possible. I committed them to show how the API would look when used.

xapian-core/tests/testdata/apitest_cluster.txt
```diff
@@ -0,0 +1,7 @@
+A computer cluster consists of a set of loosely or tightly connected computers that work together so that, in many respects, they can be viewed as a single system. Unlike grid computers, computer clusters have each node set to perform the same task, controlled and scheduled by software.
```
jaylett: This is a bit weird — in a clustering system, having text about a totally different sort of cluster! It's probably not a problem, but isn't really an ideal choice of text.

xapian-core/tests/api_cluster.cc
```diff
@@ -51,14 +53,34 @@ DEFINE_TESTCASE(euclidian_distance2, backend) {
...
+	Xapian::MSetDocumentSource docs(matches);
+	Xapian::TermListGroup tlg;
+	tlg.add_documents(docs);
+	Xapian::MSetIterator i = matches.begin();
+	Xapian::Document doc1 = i.get_document();
+	i++;
+	Xapian::Document doc2 = i.get_document();
+	Xapian::EuclidianDistance d;
+	double sim = d.similarity(tlg, doc1.termlist_begin(), doc1.termlist_end(), doc2.termlist_begin(), doc2.termlist_end());
+	printf("%f",sim);
```
jaylett: Probably don't want to keep this `printf()` in once you've got it working?

xapian-core/cluster/cluster.cc
```diff
@@ -1,8 +1,8 @@
 /** @file cluster.cc
- *  @brief Xapian::Cluster API class
+ *  @brief Cluster API
  */
-/* Copyright (C) 2016 Richhiey Thomas
- * Copyright (C) 2016 Richard Boulton
+/* Copyright (C) 2010 Richard Boulton
+ * Copyright (C) 2016 Richhiey Thomas
```
jaylett: This would ideally have been two commits: one to fixup the previous work, and one (separately) to implement `RoundRobin`. Each commit should do one, complete, thing (or, in the case of a fixup, should combine with a previous commit to do one, complete, thing).

xapian-core/cluster/cluster.cc
```diff
@@ -149,24 +150,109 @@ ClusterSet::cluster_size(clusterid cid) {
...
+
+Document
+DocumentSetIterator::get_document() {
+	LOGCALL(API, Document, "DocumentSetIterator::get_document()", "");
+	Document d;
+	d = docs[index];
+	return d;
+}
+
+RoundRobin::~RoundRobin()
```
jaylett: I'd probably put the `RoundRobin` implementation in its own file, to keep it separate from the "mechanics" of the clustering system that are separate to any algorithm. (Otherwise, the next algorithm will also be in this file, and it'll become harder and harder to find the bit you need.)

xapian-core/cluster/euclidian_sim.cc
```diff
@@ -50,6 +50,8 @@ EuclidianDistance::similarity(TermListGroup tlg,
 	LOGCALL(API, double, "EuclidianDistance::similarity()", "");
 	vector<pair<string, double> > docvec1;
 	vector<pair<string, double> > docvec2;
+	docvec1.clear();
+	docvec2.clear();
```
jaylett: This isn't fixing the tests, it's fixing the implementation. Should this entire commit be a fixup for earlier work?

xapian-core/cluster/euclidian_sim.cc
```diff
@@ -77,29 +78,40 @@ EuclidianDistance::similarity(TermListGroup tlg,
...
 		double sum = 0;
 		for(tb = docvec2.begin(); tb != docvec2.end(); tb++) {
 			if(tb->first == ta->first) {
-				found=true;
+				found = true;
 				break;
 			}
 		}
-		
-		if(found) {
```
jaylett: We put spaces after `if` (also `for`, above, and similar), before the brackets. So:

```c++
if (found) {
```

xapian-core/cluster/euclidian_sim.cc
```diff
@@ -77,29 +78,40 @@ EuclidianDistance::similarity(TermListGroup tlg,
...
+
 	vector<pair<string, double> >::iterator ta;
 	vector<pair<string, double> >::iterator tb;
 	double result = 0;
-	
-	// Finding euclidian distance between the above document vectors	
+
+	// Finding euclidian distance between the above document vectors
 	for(ta = docvec1.begin(); ta != docvec1.end(); ta++) {
 		bool found = false;
```
jaylett: This alignment is too deep. Each successive block we move four characters further in, not eight. I think your editor is getting confused between the concept of the tab character (displayed by moving to the next eighth position) and alignment offset (which should be 4).

xapian-core/cluster/cluster.cc
```diff
@@ -0,0 +1,258 @@
...
+#include <debuglog.h>
+
+#include <vector>
+#include <map>
+
+using namespace Xapian;
+using namespace std;
+
+MSetDocumentSource::MSetDocumentSource(const MSet &mset_)
+: mset(mset_), maxitems(mset_.size()), index(0)
```
jaylett: We usually offset initialisers by 4 characters. (I say "usually", because I think some older bits of the code don't.) In my opinion it makes it slightly easier to scan through and see where the signatures and bodies are.

xapian-core/cluster/cluster.cc
```diff
@@ -0,0 +1,258 @@
...
+	LOGCALL_CTOR(API,"MSetDocumentSource", mset_);
+	mset.fetch();
+}
+
+MSetDocumentSource::MSetDocumentSource(const MSet & mset_, doccount maxitems_)
+: mset(mset_), maxitems(maxitems_), index(0)
+{
+    LOGCALL_CTOR(API, "MSetDocumentSource", mset | maxitems);
+    if (maxitems > mset.size()) {
+		maxitems = mset.size();
```
jaylett: Alignment is a little deep here.

jaylett:
I'm not sure I understand why internally everything works in terms of cluster ids rather than just clusters. It certainly doesn't make any difference to `RoundRobin`, and I can't seeing it being easier for more complex algorithms either.

However it doesn't seem to affect the bit of the API that users actually play with. I wonder if we should push that into a `protected` part of the API and make the relevant classes friends or something, so we aren't exposing parts of the API to regular users who don't actually care about them. (We don't have to resolve this now, it's just something to start thinking about as we build up more functionality and write more user code.)

xapian-core/tests/testdata/apitest_cluster.txt
richhiey1996: These were just a few paragraphs taken from wikipedia and they seemed to serve the purpose well enough. Will it be necessary to change it now? Or can I do that when we test more complex algorithms? Because currently, the tests seem to work as expected with this.

xapian-core/tests/api_cluster.cc
richhiey1996: Yes, forgot to remove that. Changed that.

xapian-core/cluster/cluster.cc
richhiey1996: I've added everything to a single commit and then added more commits to fix mistakes I made before. I'll make my commits more precise now on.

jaylett: Yeah, it takes a while to get into the habit of breaking things into distinct commits :-)

(`git add -p` and friends helps.)

richhiey1996:
About not exposing parts of the API's to the user, I agree with you and was going to touch upon that once this API was given a green flag.
I prefer using clusterid's as a way of identifying clusters because I find them easier to visualize. It's a similar to using docid's I suppose. I'm not sure I understand how it can be changed to work with directly with clusters. Won't it be a little confusing if we had to pass around vectors of documents compared to an unsigned int? I could be wrong, so could you please explain that?
Also, by my timeline, I had aimed to get done with an implementation of KMeans by the end of the mid term evaluation. So would it be wise for me to continue with that? Or should I put that on hold so we can finalize the API?

jaylett:
I don't think it's worth trying to tidy up the API now; better to keep moving forward, and learn more about what's useful from an API pov. It'd be good if you could tidy up some of the formatting issues, then we can merge this and move on.

richhiey1996:
Yes yes. I didn't include a few changes in this commit. I staged them for commit and then added a few more changes. I'll be done in a bit.

richhiey1996:
I guess I've tried to fix as many formatting issues as I could. Could you please have a look and tell me if I've missed out in certain places?

xapian-core/include/xapian/cluster.h
```diff
@@ -0,0 +1,325 @@
...
+     *  construction of termlists
+     */
+    doccount docs_num;
+
+    void add_document(const Document &doc);
+
+    // Add a number of documents from the DocumentSource
+    void add_documents(DocumentSource &docs);
+
+    Xapian::doccount get_termfreq(const std::string &tname);
```
jaylett: This has trailing whitespace. (I found this by looking at the output of `git diff`, where it's highlighted. That's also used by `git add -p`, which is one of the reasons I recommend adding things into the git index using it.)

jaylett:
@richhiey1996 that mostly looks good. `xapian-maintainer-tools/xapian-check-patch` is still warning a few things though:

```
$ git diff HEAD~8 | ./xapian-maintainer-tools/xapian-check-patch 
xapian-core/cluster/cluster.cc:105: added/changed line uses spaces for indentation rather than tab: +        if (i == termfreq.end())
xapian-core/cluster/cluster.cc:107: added/changed line uses spaces for indentation rather than tab: +        else
xapian-core/cluster/cluster.cc:108: added/changed line uses spaces for indentation rather than tab: +            i->second += 1;
xapian-core/cluster/cluster.cc:147: added/changed line has trailing whitespace: +    return ((i->second).size());	
xapian-core/cluster/euclidian_sim.cc:44: added/changed line has trailing whitespace: +EuclidianDistance::similarity(TermListGroup tlg, 
xapian-core/cluster/round_robin.cc:44: added/changed line has trailing whitespace: +    LOGCALL(API, ClusterSet, "RoundRobin::cluster()", mset | k);	
xapian-core/include/xapian/cluster.h:114: added/changed line uses spaces for indentation rather than tab: +        : docs(docs_), index(index_) {}
xapian-core/include/xapian/cluster.h:185: added/changed line has trailing whitespace: +    Xapian::doccount get_termfreq(const std::string &tname);        
Files patched:	13
Lines added:	855
Lines removed:	13
```

They should all be fairly quick to fix; half are trailing whitespace, the others are spaces instead of leading tabs for indentation which would be ideal for consistency to fix, but depends somewhat on your editor.

richhiey1996:
I've fixed all those errors pointed out by the maintainer tool. Should I squash all the commits into one?

jaylett:
@richhiey1996 you already have, which was a mistake because now I can't check the final changes you made. (You've also squashed without rewriting the commit message to make sense.)

richhiey1996:
In my final commit, I had just fixed the last few errors shown by xapian-check-patch. Is there anything I can do so that you can see the changes made by the final commit?
I apologize for squashing the commits. I should've waited till you approved of it.

jaylett:
> About not exposing parts of the API's to the user, I agree with you and was going to touch upon that once this API was given a green flag.

I still think it's a good idea to implement a second (real!) clusterer before worrying too much about the API. Doing the second one should make it clearer which bits need to be slightly more general.

However briefly these are my current thoughts, so you can have them in the back of your mind as your work on KMeans. You have:

```c++
ClusterSet RoundRobin::cluster(MSet &mset, unsigned int k);
double Similarity::similiarity(TermListGroup tlg, TermIterator a_begin, TermIterator a_end, TermIterator b_begin, TermIterator b_end);
```

1. `k` is probably fine here, although I wonder if there are clusterers which don't have to be told the number of clusters in advance, in which case moving it to the constructor might be better
2. `k` should be called `num_clusters` or something more descriptive
3. `cluster()` should take a `DocumentSource` instead of an `MSet`
  * otherwise all the `MSetDocumentSource` mechanics should be internal to clustering rather than part of the public API
  * it feels like something where being able to write your own might be useful in future
  * we can provide some helpers so you can still pass an `MSet` into `cluster()`
4. Obviously there'll be a Pull Up Class to do once you have two of them, but that's kind of why I don't want to get hung up on the API just yet -- let's keep moving and see what happens when we do that pull up
5. `Similarity` should probably take two documents, but I'm not sure what it should look like because of the `TermListGroup`, which is basically pre-calculating some useful bits
  * perhaps that could be an aspect of the `Similarity` class, or a subclass of it?
  * that would make `TermListGroup` private rather than part of the public API, which is probably desirable (I can't think of a use for it in client code)

> I prefer using clusterid's as a way of identifying clusters because I find them easier to visualize. It's similar to using docid's I suppose. I'm not sure I understand how it can be changed to work directly with clusters. Won't it be a little confusing if we had to pass around vectors of documents compared to an unsigned int? I could be wrong, so could you please explain that?

I'm less concerned about what happens internally than the public API. clusterids don't make a whole lot of sense for client code (you can just use the index into the `ClusterSet`).

More importantly, there's no `Cluster` object at the moment, which in future (say when we have an approach that can provide useful labels for the clusters automatically) would make more sense. I suspect that changing `ClusterSet` to use a `map<>` of `Cluster` objects rather than a `map<vector<Document> >` will provide a better API, but may also drive changes to how you do things internally.

jaylett:
@richhiey1996 can you rebase this onto upstream master, please? It should now run travis tests automatically (and they should pass!).

richhiey1996:
@jaylett I'm running into a slight problem. When I try to rebase the cluster branch with upstream master, git points out a few trailing whitespace errors and gives the merge conflict error.
I tried the same thing with the master branch and it gave the same errors. So I hard reset that to upstream master and then it was fine because I had no local changes out there.
But I don't see how to rebase the cluster branch onto upstream master. Could you help me out?

richhiey1996:
@jaylett I've rebased this onto upstream master and travis build seems to have completed successfully :)

richhiey1996:
> akk

richhiey1996:
> k is probably fine here, although I wonder if there are clusterers which don't have to be told the number of clusters in advance, in which case moving it to the constructor might be better
There are a few density based clusterers which don't exactly need the 'number' of clusters, but certain other parameters are still required. The parameters for every clusterer will vary, and thats why, from a users point of view, I think it's easier to keep varying function parameters rather than push it to a constructor. Correct me if I'm on the wrong path though.

> k should be called num_clusters or something more descriptive
Noted!

> cluster() should take a DocumentSource instead of an MSet
Yes, I remember bringing this up during the community bonding period too and agree on every point that you mentioned related to this. But as of now, I've used the MSet directly because we're only dealing with search results.

>  Similarity should probably take two documents, but I'm not sure what it should look like because of the TermListGroup, which is basically pre-calculating some useful bits
Well, it should ideally be taking two documents and even currently it is, but instead of the entire documents, just their termlists are being passed. TermListGroup is definitely a private part of the API, only used to calculate term frequencies.

> More importantly, there's no Cluster object at the moment, which in future (say when we have an approach that can provide useful labels for the clusters automatically) would make more sense. I suspect that changing ClusterSet to use a map<> of Cluster objects rather than a map<vector<Document> > will provide a better API, but may also drive changes to how you do things internally.
Absolutely! That makes much more sense going forward. I'll make sure thats changed before implementing the KMeans algorithm.

Thanks for all those suggestions. Really helped! I'll fix some parts of the API that will be needed before proceeding, and then start with KMeans.

jaylett:
I've committed this as `richhiey1996/cluster`, so future PRs can be into that branch. Once we have KMeans, I'd hope we'll have a complete (if minimal) system that we can merge to master :-)

PR #108, opened 2016-06-30
Add support for PL2+ Weighting Scheme.

ivmarkp:
@ojwb @jaylett The implementation of PL2+ weighting function is complete with all the initial tests covered so far -- builds successfully and all tests pass. Please have a look when you have some time. Next, I'll try to improve test coverage further and will update the PR soon.

### DIFF
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1083,7 +1083,7 @@ class XAPIAN_VISIBILITY_DEFAULT PL2Weight : public Weight {
 };
 
 /// Xapian::Weight subclass implementing the PL2+ probabilistic formula.
-class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public PL2Weight {
+class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public Weight {
   public:
     /** Construct a PL2PlusWeight.
      *
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1082,6 +1082,79 @@ class XAPIAN_VISIBILITY_DEFAULT PL2Weight : public Weight {
     double get_maxextra() const;
 };
 
+/// Xapian::Weight subclass implementing the PL2+ probabilistic formula.
+class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public PL2Weight {
+  public:
+    /** Construct a PL2PlusWeight.
+     *
+     *  @param c  A non-negative and non zero parameter controlling the extent
+     *		  of the normalization of the wdf to the document length. The
+     *		  default value of 1 is suitable for longer queries but it may
+     *		  need to be changed for shorter queries. For more information,
+     *		  please refer to Gianni Amati's PHD thesis titled
+     *		  Probabilistic Models for Information Retrieval based on
+     *		  Divergence from Randomness.
+     *  
+     *  @param delta  A parameter for pseudo tf value to control the scale
+     *                of the tf lower bound. Delta(δ) should be a positive
+     *                real number. It can be tuned for example from 0.1 to 1.5
+     *                in increments of 0.1 or so. Experiments have shown that
+     *                PL2+ works effectively across collections with a fixed δ = 0.8
+     *                (default 0.8)
+     */
+    PL2PlusWeight(double c, double delta);
+
+    PL2PlusWeight( )
+	: param_c(1.0), param_delta(0.8) {
+	need_stat(AVERAGE_LENGTH);
+	need_stat(DOC_LENGTH);
+	need_stat(DOC_LENGTH_MIN);
+	need_stat(DOC_LENGTH_MAX);
+	need_stat(COLLECTION_SIZE);
+	need_stat(COLLECTION_FREQ);
+	need_stat(WDF);
+	need_stat(WDF_MAX);
+	need_stat(WQF);
+    }
+
+    std::string name() const;
+
+    std::string serialise() const;
+    PL2PlusWeight * unserialise(const std::string & serialised) const;
+
+    double get_sumpart(Xapian::termcount wdf,
+		       Xapian::termcount doclen,
+		       Xapian::termcount uniqterms) const;
+    double get_maxpart() const;
+
+    double get_sumextra(Xapian::termcount doclen,
+			Xapian::termcount uniqterms) const;
+    double get_maxextra() const;
+  
+  protected:
+    /// The wdf normalization parameter in the formula.
+    double param_c;
+    
+    /// Additional parameter delta in the PL2+ weighting formula.
+    double param_delta;
+
+    /// The lower bound of the weight.
+    double lower_bound;
+
+    /// The upper bound on the weight.
+    double upper_bound;
+
+    /// Constants for a given term in a given query.
+    double P1, P2;
+
+    /// Set by init() to (param_c * get_average_length())
+    double cl;
+
+    PL2PlusWeight * clone() const;
+
+    void init(double factor);  
+};
+
 /** This class implements the DPH weighting scheme.
  *
  *  DPH is a representative scheme of the Divergence from Randomness Framework
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1094,7 +1094,7 @@ class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public Weight {
      *		  please refer to Gianni Amati's PHD thesis titled
      *		  Probabilistic Models for Information Retrieval based on
      *		  Divergence from Randomness.
-     *  
+     *
      *  @param delta  A parameter for pseudo tf value to control the scale
      *                of the tf lower bound. Delta(δ) should be a positive
      *                real number. It can be tuned for example from 0.1 to 1.5
@@ -1130,11 +1130,11 @@ class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public Weight {
     double get_sumextra(Xapian::termcount doclen,
 			Xapian::termcount uniqterms) const;
     double get_maxextra() const;
-  
+
   protected:
     /// The wdf normalization parameter in the formula.
     double param_c;
-    
+
     /// Additional parameter delta in the PL2+ weighting formula.
     double param_delta;
 
@@ -1152,7 +1152,7 @@ class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public Weight {
 
     PL2PlusWeight * clone() const;
 
-    void init(double factor);  
+    void init(double factor);
 };
 
 /** This class implements the DPH weighting scheme.
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1151,6 +1151,9 @@ class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public Weight {
     /// Set by init() to (param_c * get_average_length())
     double cl;
 
+    /// Weight contribution of delta term in the PL2+ function
+    double dw;
+
     PL2PlusWeight * clone() const;
 
     void init(double factor);
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1151,6 +1151,9 @@ class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public Weight {
     /// Set by init() to (param_c * get_average_length())
     double cl;
 
+    /// Set by init() to get_collection_freq()) / get_collection_size()
+    double mean;
+
     /// Weight contribution of delta term in the PL2+ function
     double dw;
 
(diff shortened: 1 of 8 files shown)


xapian-core/weight/pl2plusweight.cc
```diff
@@ -0,0 +1,153 @@
...
+{
+    if (wdf == 0) return 0.0;
+
+    double wdfn = wdf * log2(1 + cl / len);
+
+    double P = P1 + (wdfn + 0.5) * log2(wdfn) - P2 * wdfn;
+
+    double P_delta = P1 + (param_delta + 0.5) * log2(param_delta) - P2 * param_delta;
+
+    return get_wqf() * ((P / (wdfn + 1.0)) + (P_delta / (param_delta + 1.0))) - lower_bound;
```
ivmarkp: `get_wqf()` function is now called once per query term for calculating weight contribution within `get_sumpart(...)`.

xapian-core/include/xapian/weight.h
```diff
@@ -1082,6 +1082,80 @@ class XAPIAN_VISIBILITY_DEFAULT PL2Weight : public Weight {
...
+     *		  Divergence from Randomness.
+     *  
+     *  @param delta  A parameter for pseudo tf value to control the scale
+     *                of the tf lower bound. Delta(δ) should be a positive
+     *                real number or in other words, non-negative and non-zero.
+     *                It can be tuned for e.g from 0.1 to 1.5 in increments of
+     *                0.1 or so. Experiments have shown that PL2+ works effectivoly
+     *                across collections with a fixed δ = 0.8 (default 0.8)
+     */
+    PL2PlusWeight(double c, double delta);
```
samuelharden: I think the variable name c is too vague to be parameter

jaylett: It's just mirroring what we already have for PL2 though, and BM25 has some pretty impenetrable parameter names too.

xapian-core/tests/api_weight.cc
```diff
@@ -505,6 +505,86 @@ DEFINE_TESTCASE(pl2weight3, backend) {
...
+	    FAIL_TEST("Serialised PL2PlusWeight with junk appended unserialised to empty name!");
+	FAIL_TEST("Serialised PL2PlusWeight with junk appended unserialised OK");
+    } catch (const Xapian::SerialisationError &) {
+
+    }
+    return true;
+}
+
+// Test for invalid values of c.
+DEFINE_TESTCASE(pl2plusweight2, !backend) {
```
samuelharden: Either comment should be changed to say it test invalid argument exception and Default value or test should be slplit in two
Moreover, since we are writing new test we should give relevant names  to the test than pl2plusweight2 like pl2plusweight_test_invalid_parameters maybe  ?

ivmarkp: @samuelharden I wonder if grouping the test cases that test invalid argument exception together should serve our purpose here?
> Moreover, since we are writing new test we should give relevant names to the test

There seems to be a common practice of naming test cases as `<weighting-scheme><number>` for weighting schemes. So, I chose to stick with the convention for naming new test cases.

xapian-core/include/xapian/weight.h
```diff
@@ -1082,6 +1082,80 @@ class XAPIAN_VISIBILITY_DEFAULT PL2Weight : public Weight {
     double get_maxextra() const;
 };
 
+/// Xapian::Weight subclass implementing the PL2+ probabilistic formula.
+class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public PL2Weight {
+
+  public:
```
ojwb: As with BM25+, there's a superfluous blank line before `public:`.

xapian-core/include/xapian/weight.h
```diff
@@ -1082,6 +1082,80 @@ class XAPIAN_VISIBILITY_DEFAULT PL2Weight : public Weight {
...
+     *		  need to be changed for shorter queries. For more information,
+     *		  please refer to Gianni Amati's PHD thesis titled
+     *		  Probabilistic Models for Information Retrieval based on
+     *		  Divergence from Randomness.
+     *  
+     *  @param delta  A parameter for pseudo tf value to control the scale
+     *                of the tf lower bound. Delta(δ) should be a positive
+     *                real number or in other words, non-negative and non-zero.
+     *                It can be tuned for e.g from 0.1 to 1.5 in increments of
+     *                0.1 or so. Experiments have shown that PL2+ works effectivoly
```
ojwb: As with BM25+, "for e.g." would be better as just "e.g." or "for example" (I'd probably go for the latter myself).

"a positive real number or in other words, non-negative and non-zero" seems needlessly complicated - positive doesn't include zero, so clarification doesn't seem necessary.

xapian-core/include/xapian/weight.h
ojwb: The formula for `PL2` uses `c` for this parameter, so I think that's probably the best name for it here too.  The documentation comment for it can (and does) explain its purpose.  If we called it something like `length_normalisation_extent` we'd then also need to note that this is parameter c in the PL2 formula...

xapian-core/tests/api_nodb.cc
```diff
@@ -370,6 +370,18 @@ DEFINE_TESTCASE(weight1, !backend) {
...
+    TEST_EQUAL(pl2plusweight_dflt.serialise(), pl2plusweight.serialise());
+    wt = Xapian::PL2PlusWeight().unserialise(pl2plusweight.serialise());
+    TEST_EQUAL(pl2plusweight.serialise(), wt->serialise());
+    delete wt;
+
+    Xapian::PL2PlusWeight pl2plusweight2(2.0, 0.9);
+    TEST_NOT_EQUAL(pl2plusweight.serialise(), pl2plusweight2.serialise());
+
 
     Xapian::DPHWeight dphweight;
```
ojwb: There's an extra blank line above here.

xapian-core/weight/pl2plusweight.cc
ojwb: Just FYI, it shouldn't actually matter how many times you call `get_wqf()` - that method is inline in the class definition in the header and returns the already-fetched/calculated value from a member variable, and the compiler should be able to perform CSE on the inlined code.

Xapian knows whether a statistic is wanted thanks to the `need_stat()` machinery.

jaylett:
@ivmarkp I think this is good, now. It needs rebasing onto latest master, and some squashing — think about what the "ideal" commits are, and move to that using interactive rebase.

ivmarkp:
Travis build is failing with `configure error: Xapian requires GCC 4.8 or later`. Should gcc version be set to 4.8 in [.travis.yml](https://github.com/xapian/xapian/blob/master/.travis.yml)?

jaylett:
@ivmarkp no, sadly alone that doesn't work. For the time being, if it's passing on clang then that's enough. (Don't worry about updating `.travis.yml`, because that may well get updated separately as we try to resolve this, which will probably produce a merge conflict.)

ivmarkp:
Oh, okay. Although, yeah it's passing on clang.

> Don't worry about updating .travis.yml, because that may well get updated separately as we try to resolve this, which will probably produce a merge conflict

Okay, understood -- leaving this PR in its current state.

ojwb:
The automated checking tools reports some trailing whitespace:

```
git show 2ed0d35|xapian-maintainer-tools/xapian-check-patch 
xapian-core/include/xapian/weight.h:1097: added/changed line has trailing whitespace: +     *  
xapian-core/include/xapian/weight.h:1133: added/changed line has trailing whitespace: +  
xapian-core/include/xapian/weight.h:1137: added/changed line has trailing whitespace: +    
xapian-core/include/xapian/weight.h:1155: added/changed line has trailing whitespace: +    void init(double factor);  
xapian-core/tests/api_nodb.cc:373: added/changed line has trailing whitespace: +    
Files patched:	8
Lines added:	330
Lines removed:	8
```

xapian-core/include/xapian/weight.h
```diff
@@ -1082,6 +1082,79 @@ class XAPIAN_VISIBILITY_DEFAULT PL2Weight : public Weight {
     double get_maxextra() const;
 };
 
+/// Xapian::Weight subclass implementing the PL2+ probabilistic formula.
+class XAPIAN_VISIBILITY_DEFAULT PL2PlusWeight : public PL2Weight {
```
ojwb: I don't think we want to inherit from `PL2Weight` here - while conceptually there's a relationship between the weighting schemes, it isn't necessary to reflect that in the inheritance tree, as there isn't a situation in which we want to use a `PL2PlusWeight` reference or pointer where a `PL2Weight` reference or pointer is needed.  We always use a pointer or reference to `Xapian::Weight` where polymorphism is needed.

And in practical terms, inheriting like this means that a `PL2PlusWeight` object is about twice the size it needs to be as it has all the member variables from `PL2Weight` (which aren't used) plus its own, and both constructors will get called, and we will call `need_stat()` for the default `PL2Weight` parameters as well as for the actual `PL2PlusWeight` parameters, which may end up with us doing work to provide stats which are never used.

Looking at the BM25+ patch, I see that has the same issue - sorry for missing it there.

ivmarkp: I think we may as well make protected members (below) private? Although, I see no harm in keeping them as protected at present unless in future a subclass derives from `PL2PlusWeight`. Same for `BM25PlusWeight` subclass.

xapian-core/weight/pl2plusweight.cc
```diff
@@ -0,0 +1,153 @@
+/** @file pl2plusweight.cc
+ * @brief Xapian::PL2PlusWeight class - the PL2+ weighting scheme of the DFR framework.
+ */
+/* Copyright (C) 2016 Vivek Pal
```
ojwb: Again, you should preserve the copyright statements from the file you based this on.

You should also probably add copyright statements for your own contributions to files which you've made copyrightable changes to.  The dividing line of what's copyrightable is surprisingly complex but for example you've clearly made sufficient contributions to `include/xapian/weight.h`.

xapian-core/weight/pl2plusweight.cc
```diff
@@ -0,0 +1,153 @@
...
+    double P_min =
+	P1 + (wdfn_lower + 0.5) * log2(wdfn_lower) - P2 * wdfn_lower;
+    lower_bound = get_wqf() * P_min / (wdfn_upper + 1.0);
+
+    // Calculate the upper bound on the weight.
+    double P_max =
+	P1 + (wdfn_upper + 0.5) * log2(wdfn_upper) - P2 * wdfn_upper;
+    upper_bound = get_wqf() * P_max / (wdfn_lower + 1.0);
+
+    upper_bound -= lower_bound;
```
ojwb: The calculation of `upper_bound` is preserved unchanged from `pl2weight.cc`, which is incorrect as `get_maxpart()` is just `return upper_bound;` and we've changed the formula used in `get_sumpart()` - `get_maxpart()` is meant to return an upper bound on what `get_sumpart()` can return, and it's important that it's as tight a bound as we can provide.  Similarly, the calculation of `lower_bound` is unchanged.

The calculation of `lower_bound` and `upper_bound` needs to reflect the modifications made to the formula between PL2 and PL2+ - essentially the changes you make to `get_sumpart()` need to be reflected in the calculations here.

xapian-core/weight/pl2plusweight.cc
ojwb: `(P_delta / (param_delta + 1.0))` is constant for a given `PL2PlusWeight` object, so should be pre-calculated in `init()` and stored in a member variable - that will save a lot of logarithm calculations (which are typically quite expensive - potentially ~ 1000 CPU cycles each!)  You'll need to calculate this value in `init()` anyway in order to fix the calculations of `upper_bound` and `lower_bound`.

xapian-core/weight/pl2plusweight.cc
```diff
@@ -70,23 +72,38 @@ PL2PlusWeight::init(double)
...
     // Calculate the lower bound on the weight.
     double P_min =
 	P1 + (wdfn_lower + 0.5) * log2(wdfn_lower) - P2 * wdfn_lower;
-    lower_bound = get_wqf() * P_min / (wdfn_upper + 1.0);
+    // Calculate P_delta_min by setting param_delta to the
+    // lowest permitted value of 0.1.
+    double P_delta_min = P1 + (0.1 + 0.5) * log2(0.1) - P2 * 0.1;
+    // Calculate dw_min by setting param_delta to 1.5 based
+    // on the prescribed the value range for it in weight.h
+    double dw_min = P_delta_min / (1.5 + 1.0);
```
ojwb: At this point, the value for delta is known.  We only need to account for values which vary by document in calculating the bounds - i.e. `wdf`, `len`, and (unused by PL2, so not relevant in this case) the number of unique terms in the document.

`P_delta` doesn't depend on any of these unknowns, so it's a constant in this context (that is, a particular term in a given database).

ivmarkp: So we don't want `P_delta_min`, `P_delta_max` (and similarly `dw_min` and `dw_max`). Is that right?

ojwb: Indeed - `P_delta_min = P_delta_max = P_delta`, and similarly for `dw`, so we can just use `P_delta` and `dw` instead.

ojwb:
All looks good - merged (squashed down into one commit - not sure what's best here really, but that's what James suggested for the previous PR so I went with that again).

PR #111, opened 2016-07-15
Dir+ weighting function implementation.

ivmarkp:
This implementation is a bit different from the previous implementation of weighting functions, BM25+ and PL2+ because here the existing `LMWeight` class is extended with the additions of two new parameters, `param_delta` and `enable_dirplus` to support new Dir+ function alongside the implementation of existing Dir function. 

Also, the previously written test cases for lmweight are modified accordingly with new parameters so as to make sure it doesn't mess with the functioning of already available weighting functions (smoothings) in `lmweight.cc` -- all tests pass.

### DIFF
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -142,7 +142,8 @@ class XAPIAN_VISIBILITY_DEFAULT Weight {
 	TWO_STAGE_SMOOTHING = 1,
 	DIRICHLET_SMOOTHING = 2,
 	ABSOLUTE_DISCOUNT_SMOOTHING = 3,
-	JELINEK_MERCER_SMOOTHING = 4
+	JELINEK_MERCER_SMOOTHING = 4,
+	DIRICHLET_PLUS_SMOOTHING = 5
     } type_smoothing;
 
     class Internal;
@@ -1160,15 +1161,8 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
     /** The type of smoothing to use. */
     type_smoothing select_smoothing;
 
-    // Parameters for handling negative value of log, and for smoothing.
+    // Parameters for handling negative value for smoothing.
     double param_log, param_smoothing1, param_smoothing2;
-
-    // Additional parameter in Dir+ weighting formula.
-    double param_delta;
-
-    // Boolean parameter to control weight calculations when smoothing
-    // is set to Dirichlet.
-    bool enable_dirplus;
 
     // Collection weight.
     double weight_collection;
@@ -1203,28 +1197,19 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
      *					DIRCHLET(2000))
      *
      *  @param param_smoothing2_	A non-negative parameter which is used
-     *					only with TWO_STAGE_SMOOTHING as
-     *					parameter for Dirichlet's smoothing.
-     *					(default: 2000)
+     *					with TWO_STAGE_SMOOTHING as parameter for Dirichlet's
+     *					smoothing (default: 2000) and as parameter delta to
+     *					control the scale of the tf lower bound in the
+     *             			DIRICHLET_PLUS_SMOOTHING (default 0.05).
      *
-     *  @param param_delta_	A non-negative parameter for pseudo tf value to control the scale
-     *		      		of the tf lower bound in the Dir+ scoring funtion. Delta can be tuned
-     *				from 0.0 to 0.15 in increments of 0.01 but it is also advisable to
-     *				fix it to 0.05 for normal use. (default 0.05)
-     *
-     *  @param enable_dirplus_	A boolean parameter to control the weight calculations. It takes
-     *				effect only when smoothing is set to Dirichlet. By default, it is set to
-     *				zero so as to allow normal functioning of Dirichlet smoothing. (default 0)
      */
     // Unigram LM Constructor to specifically mention all parameters for handling negative log value and smoothing.
     explicit LMWeight(double param_log_ = 0.0,
 		      type_smoothing select_smoothing_ = TWO_STAGE_SMOOTHING,
 		      double param_smoothing1_ = 0.7,
-		      double param_smoothing2_ = 2000.0,
-		      double param_delta_ = 0.05,
-		      bool enable_dirplus_ = 0)
+		      double param_smoothing2_ = 2000.0)
 	: select_smoothing(select_smoothing_), param_log(param_log_), param_smoothing1(param_smoothing1_),
-	  param_smoothing2(param_smoothing2_), param_delta(param_delta_), enable_dirplus(enable_dirplus_)
+	  param_smoothing2(param_smoothing2_)
     {
 	need_stat(AVERAGE_LENGTH);
 	need_stat(DOC_LENGTH);
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1239,6 +1239,7 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
 	if (select_smoothing == ABSOLUTE_DISCOUNT_SMOOTHING)
 	    need_stat(UNIQUE_TERMS);
     }
+
     std::string name() const;
 
     std::string serialise() const;
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1200,7 +1200,7 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
      *					with TWO_STAGE_SMOOTHING as parameter for Dirichlet's
      *					smoothing (default: 2000) and as parameter delta to
      *					control the scale of the tf lower bound in the
-     *             			DIRICHLET_PLUS_SMOOTHING (default 0.05).
+     *					DIRICHLET_PLUS_SMOOTHING (default 0.05).
      *
      */
     // Unigram LM Constructor to specifically mention all parameters for handling negative log value and smoothing.
@@ -1216,7 +1216,7 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
 	    if (select_smoothing == TWO_STAGE_SMOOTHING)
 		param_smoothing2 = 2000.0;
 	    else
-	    	param_smoothing2 = 0.05;
+		param_smoothing2 = 0.05;
 	}
 	need_stat(AVERAGE_LENGTH);
 	need_stat(DOC_LENGTH);
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1230,7 +1230,7 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
 	if (select_smoothing == ABSOLUTE_DISCOUNT_SMOOTHING)
 	    need_stat(UNIQUE_TERMS);
 	if (select_smoothing == DIRICHLET_PLUS_SMOOTHING)
-	   need_stat(DOC_LENGTH_MIN);
+	    need_stat(DOC_LENGTH_MIN);
     }
 
     std::string name() const;
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -1163,6 +1163,13 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
     // Parameters for handling negative value of log, and for smoothing.
     double param_log, param_smoothing1, param_smoothing2;
 
+    // Additional parameter in Dir+ weighting formula.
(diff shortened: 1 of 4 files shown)


xapian-core/weight/lmweight.cc
```diff
@@ -159,8 +163,24 @@ LMWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount len,
...
+	if (enable_dirplus) {
+
+	/* In the Dir+ weighting formula, sumpart weight contribution is :-
+	 *
+	 * sum of log of (1 + (wdf/(param_smoothing1 * weight_collection))) and 
+	 * log of (1 + (delta/param_smoothing1 * weight_collection))). 
+	 
+	 * Since, sum of logs iss log of product so weight_sum is calculated as product
+	 * of terms in log in the Dir+ formula.	
+	 */
```
jaylett: The comment should be aligned with the following code, not the containing block.

xapian-core/weight/lmweight.cc
```diff
@@ -159,8 +163,24 @@ LMWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount len,
...
-		     (len_double + param_smoothing1);
+	// Enabling Dir+ weighting formula.
+	if (enable_dirplus) {
+
+	/* In the Dir+ weighting formula, sumpart weight contribution is :-
+	 *
+	 * sum of log of (1 + (wdf/(param_smoothing1 * weight_collection))) and 
+	 * log of (1 + (delta/param_smoothing1 * weight_collection))). 
+	 
+	 * Since, sum of logs iss log of product so weight_sum is calculated as product
```
jaylett: "is" instead of "iss"?

xapian-core/weight/lmweight.cc
```diff
@@ -182,12 +202,20 @@ LMWeight::get_maxpart() const
...
     // Calculating upper bound considering different smoothing option available to user.
     if (select_smoothing == JELINEK_MERCER_SMOOTHING) {
 	upper_bound = (param_smoothing1 * weight_collection) + (1 - param_smoothing1);
     } else if (select_smoothing == DIRICHLET_SMOOTHING) {
-	upper_bound = (get_doclength_upper_bound() + (param_smoothing1 * weight_collection)) / (get_doclength_upper_bound() + param_smoothing1);
+	if (enable_dirplus) {
+	    upper_bound = (1 + (wdf_max/(param_smoothing1 * weight_collection))) *
+			  (1 + (param_delta/(param_smoothing1 * weight_collection)));
+	}
+	else if (!enable_dirplus) {
```
jaylett: This can just be `else` (and should be collapsed with the previous line: `} else {`).

jaylett: Update: I see you've partly fixed this in a later commit :-)

xapian-core/weight/lmweight.cc
ivmarkp: Oh sorry, a typo.

ivmarkp:
> Note: This PR does not contain final code to be merged; opened for the purpose of code review and also, adding new tests for Dir+ is a pending work right now. Working on it at present.

Since new tests were added and all passing, I realise this PR now contains the final code that can be merged after the code review.

xapian-core/include/xapian/weight.h
```diff
@@ -1221,7 +1239,6 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
 	if (select_smoothing == ABSOLUTE_DISCOUNT_SMOOTHING)
 	    need_stat(UNIQUE_TERMS);
     }
-
     std::string name() const;
```
ojwb: Randomly removed newline here.

xapian-core/weight/lmweight.cc
```diff
@@ -200,16 +222,34 @@ LMWeight::get_maxpart() const
...
+    else
+	return 0;
 }
 
 double
 LMWeight::get_maxextra() const
 {
-    return 0;
+    if (select_smoothing == DIRICHLET_PLUS_SMOOTHING) {
+	double extra_weight = param_smoothing1 / (get_doclength_lower_bound() + param_smoothing1);
```
ivmarkp: @ojwb Test case `unigramlmweight7` is failing under valgrind with error message telling that an uninitialised variable is being used at line 248 :

```bash
Running test: unigramlmweight7...==22742== Conditional jump or move depends on uninitialised value(s)
==22742==    at 0x55E0D88: log (w_log.c:28)
==22742==    by 0x4FC6543: Xapian::LMWeight::get_maxextra() const (lmweight.cc:248)
```
I might be missing something here but I don't see any uninitialised variable in this line. Any pointers to help fix this failing test?

ojwb: You haven't requested the statistic used here (i.e. a lower bound on the document length), so that is presumably the uninitialised value - you need to add something like this to the constructor:

```
if (select_smoothing == DIRICHLET_PLUS_SMOOTHING)
   need_stat(DOC_LENGTH_MIN);
```

More generally, you can run `valgrind --track-origins=yes` to find the origin of uninitialised values (it's slower to do so, which is why it's not the default), e.g.:

```
./runtest valgrind  --track-origins=yes ./apitest -b glass unigramlmweight7
```

ojwb: (And note that it says "uninitialised value(s)" not "uninitialised variables"!)

xapian-core/include/xapian/weight.h
```diff
@@ -1160,16 +1161,9 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
     /** The type of smoothing to use. */
     type_smoothing select_smoothing;
 
-    // Parameters for handling negative value of log, and for smoothing.
+    // Parameters for handling negative value for smoothing.
```
ojwb: This particular change seems wrong...

xapian-core/include/xapian/weight.h
```diff
@@ -1203,28 +1197,19 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
...
-     *				zero so as to allow normal functioning of Dirichlet smoothing. (default 0)
      */
     // Unigram LM Constructor to specifically mention all parameters for handling negative log value and smoothing.
     explicit LMWeight(double param_log_ = 0.0,
 		      type_smoothing select_smoothing_ = TWO_STAGE_SMOOTHING,
 		      double param_smoothing1_ = 0.7,
-		      double param_smoothing2_ = 2000.0,
-		      double param_delta_ = 0.05,
-		      bool enable_dirplus_ = 0)
+		      double param_smoothing2_ = 2000.0)
```
ojwb: To make the default here actually match the documentation, we need something like:

```
	      double param_smoothing2_ = (select_smoothing_ == TWO_STAGE_SMOOTHING ? 2000.0 : 0.05))
```

xapian-core/weight/lmweight.cc
```diff
@@ -229,29 +225,30 @@ LMWeight::get_maxpart() const
...
-	double extra_weight = param_smoothing1 / (len_double + param_smoothing1);
-	return get_query_length() * log (extra_weight);
-    } else
+    if (select_smoothing == DIRICHLET_PLUS_SMOOTHING) {
+	double extra_weight = param_smoothing1 / (len + param_smoothing1);
+	return get_query_length() * log(extra_weight);
+    }
+    else
 	return 0;
 }
```
ojwb: The layout of the `else` clause here doesn't match our coding standards (it should be `} else` on one line.

But in a case like this where you've just handled a particular case with an `if` that ends in a `return` statement, and there's a default return following it, it's probably more readable just to remove the `else` and have the `return 0;` at the top level of the function:

```
{
    if (select_smoothing == DIRICHLET_PLUS_SMOOTHING) {
        double extra_weight = param_smoothing1 / (len + param_smoothing1);
        return get_query_length() * log(extra_weight);
    }

    return 0;
}
```

xapian-core/weight/lmweight.cc
```diff
@@ -163,23 +166,18 @@ LMWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount len,
...
-	    weight_sum = (1 + (wdf_double/(param_smoothing1 * weight_collection))) *
-			 (1 + (param_delta/(param_smoothing1 * weight_collection)));
-	} else {
-	    weight_sum = (wdf_double + (param_smoothing1 * weight_collection)) /
-			 (len_double + param_smoothing1);
-	}
+	weight_sum = (wdf_double + (param_smoothing1 * weight_collection)) /
+		    (len_double + param_smoothing1);
+    } else if (select_smoothing == DIRICHLET_PLUS_SMOOTHING) {
+    /* In the Dir+ weighting formula, sumpart weight contribution is :-
```
ojwb: The body of this `if` is under-indented.

xapian-core/weight/lmweight.cc
```diff
@@ -208,12 +206,10 @@ LMWeight::get_maxpart() const
...
-	if (enable_dirplus) {
-	    upper_bound = (1 + (wdf_max/(param_smoothing1 * weight_collection))) *
-			  (1 + (param_delta/(param_smoothing1 * weight_collection)));
-	} else {
 	    upper_bound = (get_doclength_upper_bound() + (param_smoothing1 * weight_collection)) / (get_doclength_upper_bound() + param_smoothing1);
-	}    
+    } else if (select_smoothing == DIRICHLET_PLUS_SMOOTHING) {
+	    upper_bound = (1 + (wdf_max/(param_smoothing1 * weight_collection))) *
+			  (1 + (param_smoothing2/(param_smoothing1 * weight_collection)));
     } else if (select_smoothing == ABSOLUTE_DISCOUNT_SMOOTHING) {
```
ojwb: While this one is over-indented.

ojwb: Also missing spaces around the `/` operators here and above.

xapian-core/include/xapian/weight.h
ojwb: Oh, seems that doesn't work then.

I guess we need to make the default value of `param_smoothing2_` something invalid (`-1.0` say) and then check for that and reset it when we initialise `param_smoothing2` instead.

ivmarkp: Okay, so after setting `param_smoothing2` to `-1.0` here initialisation can be done within the `LMWeight::init()` method alongside the places where `param_smoothing1` gets set for each of the four smoothings. Is this the right way to do what we want here?

ojwb: I think it's better to handle it just below here, where we initialise `param_smoothing2` - that means the `param_smoothing2` member always has a valid value.  If we handle it in `LMWeight::init()`, we no longer have the invariant `param_smoothing2 >= 0`.

I think the other similar things which currently get handled in `init()` would be better done here instead (and we should use `-1.0` as the default value for `param_smoothing1_` too - currently if you ask for ` DIRICHLET_SMOOTHING` and explicitly set `smoothing1` to `0.7`, that code can't tell that was explicit rather than the default value and changes it to `2000.0` which is nasty).  The `get_query_length()` check can't be done here though.  But let's leave those changes until after we've got Dir+ merged.

xapian-core/include/xapian/weight.h
```diff
@@ -1206,11 +1206,17 @@ class XAPIAN_VISIBILITY_DEFAULT LMWeight : public Weight {
...
+		      double param_smoothing1_ = -1.0,
+		      double param_smoothing2_ = -1.0)
 	: select_smoothing(select_smoothing_), param_log(param_log_), param_smoothing1(param_smoothing1_),
 	  param_smoothing2(param_smoothing2_)
     {
+	if (param_smoothing1 < 0) param_smoothing1 = 0.7;
+	if (param_smoothing2 < 0) {
+	    if (select_smoothing == TWO_STAGE_SMOOTHING)
+		param_smoothing2 = 2000.0;
+	    param_smoothing2 = 0.05;
```
ojwb: This will always end up setting `param_smoothing2 = 0.05;` - you probably wanted an `else` in there...

ivmarkp: Oops, sorry  missed it there.

ojwb:
The automated checks are flagging some whitespace issues:

```
xapian-core/include/xapian/weight.h:1387: added/changed line has space before tab: +     *             			DIRICHLET_PLUS_SMOOTHING (default 0.05).
xapian-core/include/xapian/weight.h:1403: added/changed line has space before tab: +	    	param_smoothing2 = 0.05;
xapian-core/weight/lmweight.cc:229: added/changed line has trailing whitespace: + * 
```

In the interests of this not dragging out into the weekend, I'll fix these and merge, but please do this check for yourself - it saves a lot of to-and-fro and means we can review patches more quickly, which means you can get feedback sooner.

ivmarkp:
@ojwb Sorry, I've fixed these whitespace issues and pushed

ojwb:
Thanks, merged as b2af2ce92f647d926fbb22f1320980f4c45763fd.

PR #112, opened 2016-07-15
Letor: ListNET

ayshtmr:
This PR integrated ListNet ranking algorithm with letor.

Note: This PR does not change the present api structure. PRs with updated api structure (as discussed on the mailing list) will follow after this.


ayshtmr:
This PR becomes outdated with upcoming API changes since it is based on RankList. Hence, no point merging this.

PR #113, opened 2016-07-23
Resize over-long terms instead of throwing an exception.

kevin-duraj-zefr:


jaylett:
@kevin-duraj-zefr I can imagine some user code might want to instead catch the exception, especially since that's the current behaviour. I wonder if another way of tackling this might be better: make a convenience function that truncates if needed, so you take care of things in the indexing process, before we hit `add_document()` / `replace_document()`.

xapian-core/backends/glass/glass_database.cc
```diff
@@ -1081,7 +1081,8 @@ GlassWritableDatabase::add_document_(Xapian::docid did,
 
 		string tname = *term;
 		if (tname.size() > MAX_SAFE_TERM_LENGTH)
-		    throw Xapian::InvalidArgumentError("Term too long (> " STRINGIZE(MAX_SAFE_TERM_LENGTH) "): " + tname);
+		    tname.resize(MAX_SAFE_TERM_LENGTH);		
+		    //throw Xapian::InvalidArgumentError("Term too long (> " STRINGIZE(MAX_SAFE_TERM_LENGTH) "): " + tname);
```
jaylett: Instead of commenting out lines of code, you should just remove them; the version control history makes it easy to recover them if we need.

ojwb:
I don't think we want to change the behaviour here - truncating loses information, and will cause nasty effects if the terms are unique ids (since you then get multiple documents with the same "unique" id).

AFAICS, this can only happen for terms added via `Document::add_term()`, `Document::add_boolean_term()` or `Document::add_posting()` (unless you specify an insanely long prefix to `TermGenerator`) - if an API change is warranted, I think a better approach would be to instead add variants of these which truncate.

But can someone give me an example of a situation in which silently truncating a long term is actually a good idea?

I'd think catching the exception and skipping the document is probably a better plan, in which case this is more of a documentation issue.

ojwb:
No response from the submitter for a month, and the discussion on the mailing list wasn't in favour of truncating, so closing this.

PR #115, opened 2016-08-01
Pivoted Normalization (Piv+) implementation

ivmarkp:
* Adds support for pivoted normalization in tfidfweight scheme
* Builds successfully and all `make check` tests pass.

### DIFF
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -477,7 +477,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     TfIdfWeight(const std::string &normalizations, double slope, double delta);
 
     TfIdfWeight()
-	: normalizations("ntn")
+	: normalizations("ntn"), param_slope(0.2), param_delta(1.0)
     {
 	need_stat(TERMFREQ);
 	need_stat(WDF);
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -465,11 +465,11 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *
      *     @li 'n': None wtn=tfn*idfn
      *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-slope+(slope*normlen))+delta*idfn) where
-     *         parameters slope and delta values should be non-negative & non-zero when
-     *         using pivoted normalization string. Specifically, "PPP" normalization
-     *         string should be used to use Piv+ normalization. In addtion, normalization
-     *         string such as "Pxx", "xPx" and "xxP" can also be used to have more options
-     *         of combinations for wdfn, idfn and wtn normalizations respectively.
+     *         parameters slope and delta values should be positive when using pivoted
+     *         normalization string. Specifically, "PPP" normalization string should
+     *         be used to use Piv+ normalization. In addition, different types of pivoted
+     *         normalization can also be used in combination with other normalisations
+     *         to have more options of combinations for wdfn, idfn and wtn normalizations.
      *
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -411,8 +411,8 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     /// The factor to multiply with the weight.
     double factor;
 
-    /// Parameters slope and delta in the Piv+ normalization weighting formula.
-    double param_slope, param_delta;
+    /// Parameters in the Piv+ normalization weighting formula.
+    double param_s, param_delta; 
 
     TfIdfWeight * clone() const;
 
@@ -472,7 +472,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
      */
-    explicit TfIdfWeight(const std::string &normalizations);
+    explicit TfIdfWeight(const std::string &normalizations, double param_s, double param_delta);
 
     TfIdfWeight(const std::string &normalizations, double slope, double delta);
 
@@ -483,6 +483,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 	need_stat(WDF);
 	need_stat(WDF_MAX);
 	need_stat(COLLECTION_SIZE);
+	need_stat(AVERAGE_LENGTH);
+	need_stat(DOC_LENGTH);
+	need_stat(WQF);
     }
 
     std::string name() const;
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -411,9 +411,10 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     /// The factor to multiply with the weight.
     double factor;
 
-    /// Parameters in the Piv+ normalization weighting formula.
-    double param_slope = 0;
-    double param_delta = 0;
+    /// Parameters slope and delta in the Piv+ normalization weighting formula.
+    /// Default values: 0.2 and 1.0 respectively.
+    double param_slope = 0.2;
+    double param_delta = 1.0;
 
     TfIdfWeight * clone() const;
 
@@ -475,7 +476,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      */
     explicit TfIdfWeight(const std::string &normalizations);
 
-    explicit TfIdfWeight(const std::string &normalizations, double slope, double delta);
+    TfIdfWeight(const std::string &normalizations, double slope, double delta);
 
     TfIdfWeight()
 	: normalizations("ntn")
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -412,7 +412,8 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     double factor;
 
     /// Parameters in the Piv+ normalization weighting formula.
-    double param_s, param_delta; 
+    double param_s = 0;
+    double param_delta = 0;
 
     TfIdfWeight * clone() const;
 
@@ -462,18 +463,20 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     supported:
      *
      *     @li 'n': None wtn=tfn*idfn
-     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-slope+(slope*normlen))+delta*idfn)
+     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-s+(s*normlen))+delta*idfn)
+     *              Note: Valid parameters s and delta values should be non-negative
+     *                    & non-zero when using pivoted normalization string("PPP")
      *
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
      */
-    explicit TfIdfWeight(const std::string &normalizations, double param_s, double param_delta);
+    explicit TfIdfWeight(const std::string &normalizations);
 
-    TfIdfWeight(const std::string &normalizations, double slope, double delta);
-
-    TfIdfWeight()
-	: normalizations("ntn"), param_s(0.2), param_delta(1.0)
+    explicit TfIdfWeight(double s, double delta)
+	: param_s(s), param_delta(delta)
     {
+	if (param_s < 0) param_s = 0;
+	if (param_delta < 0) param_delta = 0;
 	need_stat(TERMFREQ);
 	need_stat(WDF);
 	need_stat(WDF_MAX);
@@ -481,6 +484,15 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 	need_stat(AVERAGE_LENGTH);
 	need_stat(DOC_LENGTH);
 	need_stat(WQF);
+    }
+
+    TfIdfWeight()
+	: normalizations("ntn")
+    {
+	need_stat(TERMFREQ);
+	need_stat(WDF);
+	need_stat(WDF_MAX);
+	need_stat(COLLECTION_SIZE);
     }
 
     std::string name() const;
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -440,7 +440,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     @li 'b': Boolean    wdfn=1 if term in document else wdfn=0
      *     @li 's': Square     wdfn=wdf*wdf
      *     @li 'l': Logarithmic wdfn=1+log<sub>e</sub>(wdf)
-     *     @li 'P': Pivoted     wdfn=1+log(1+log(wdf))
+     *     @li 'P': Pivoted     wdfn=(1+log(1+log(wdf)))*(1/(1-slope+(slope*doclen/avg_len)))+delta
      *
      *     The Max-wdf and Augmented Max wdf normalizations haven't yet been
      *     implemented.
@@ -462,12 +462,6 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     supported:
      *
      *     @li 'n': None wtn=tfn*idfn
-     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-slope+(slope*normlen))+delta*idfn) where
-     *         parameters slope and delta values should be positive when using pivoted
-     *         normalization string. Specifically, "PPP" normalization string should
-     *         be used to use Piv+ normalization. In addition, different types of pivoted
-     *         normalization can also be used in combination with other normalisations
-     *         to have more options of combinations for wdfn, idfn and wtn normalizations.
      *
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -462,12 +462,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     supported:
      *
      *     @li 'n': None wtn=tfn*idfn
-     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-slope+(slope*normlen))+delta*idfn) where
-     *         parameters slope and delta values should be positive when using pivoted
-     *         normalization string. Specifically, "PPP" normalization string should
-     *         be used to use Piv+ normalization. In addition, different types of pivoted
-     *         normalization can also be used in combination with other normalisations
-     *         to have more options of combinations for wdfn, idfn and wtn normalizations.
+     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-slope+(slope*normlen))+delta*idfn)
      *
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
@@ -477,7 +472,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     TfIdfWeight(const std::string &normalizations, double slope, double delta);
 
     TfIdfWeight()
-	: normalizations("ntn"), param_slope(0.2), param_delta(1.0)
+	: normalizations("ntn"), param_s(0.2), param_delta(1.0)
     {
 	need_stat(TERMFREQ);
 	need_stat(WDF);
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -404,16 +404,16 @@ class XAPIAN_VISIBILITY_DEFAULT BoolWeight : public Weight {
 
 /// Xapian::Weight subclass implementing the tf-idf weighting scheme.
 class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
-    /* Three character string indicating the normalizations for tf(wdf), idf and
-       tfidf weight. */
-    std::string normalizations;
+    /// Parameters in the Piv+ normalization weighting formula.
+    double param_slope = 0;
+    double param_delta = 0;
 
     /// The factor to multiply with the weight.
     double factor;
 
-    /// Parameters in the Piv+ normalization weighting formula.
-    double param_s = 0;
-    double param_delta = 0;
+    /* Three character string indicating the normalizations for tf(wdf), idf and
+       tfidf weight. */
+    std::string normalizations;
 
     TfIdfWeight * clone() const;
 
@@ -463,20 +463,21 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      *     supported:
      *
      *     @li 'n': None wtn=tfn*idfn
-     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-s+(s*normlen))+delta*idfn)
-     *              Note: Valid parameters s and delta values should be non-negative
-     *                    & non-zero when using pivoted normalization string("PPP")
+     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-slope+(slope*normlen))+delta*idfn) where
+     *         parameters slope and delta values should be non-negative & non-zero when
+     *         using pivoted normalization string. Specifically, "PPP" normalization
+     *         string should be used to use Piv+ normalization. In addtion, normalization
+     *         string such as "Pxx", "xPx" and "xxP" can also be used to have more options
+     *         of combinations for wdfn, idfn and wtn normalizations respectively.
      *
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
      */
-    explicit TfIdfWeight(const std::string &normalizations);
+    explicit TfIdfWeight(double slope, double delta, const std::string &normalizations);
 
-    explicit TfIdfWeight(double s, double delta)
-	: param_s(s), param_delta(delta)
+    TfIdfWeight()
+	: param_slope(0.2), param_delta(1.0), normalizations("ntn")
     {
-	if (param_s < 0) param_s = 0;
-	if (param_delta < 0) param_delta = 0;
 	need_stat(TERMFREQ);
 	need_stat(WDF);
 	need_stat(WDF_MAX);
@@ -484,15 +485,6 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 	need_stat(AVERAGE_LENGTH);
 	need_stat(DOC_LENGTH);
 	need_stat(WQF);
-    }
-
-    TfIdfWeight()
-	: normalizations("ntn")
-    {
-	need_stat(TERMFREQ);
-	need_stat(WDF);
-	need_stat(WDF_MAX);
-	need_stat(COLLECTION_SIZE);
     }
 
     std::string name() const;
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -483,9 +483,6 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 	need_stat(WDF);
 	need_stat(WDF_MAX);
 	need_stat(COLLECTION_SIZE);
-	need_stat(AVERAGE_LENGTH);
-	need_stat(DOC_LENGTH);
-	need_stat(WQF);
     }
 
     std::string name() const;
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -412,9 +412,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     double factor;
 
     /// Parameters slope and delta in the Piv+ normalization weighting formula.
-    /// Default values: 0.2 and 1.0 respectively.
-    double param_slope = 0.2;
-    double param_delta = 1.0;
+    double param_slope, param_delta;
 
     TfIdfWeight * clone() const;
 
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -420,9 +420,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     /* When additional normalizations are implemented in the future, the additional statistics for them
        should be accessed by these functions. */
-    double get_wdfn(Xapian::termcount wdf, char c) const;
+    double get_wdfn(Xapian::termcount wdf, Xapian::termcount len, char c) const;
     double get_idfn(Xapian::doccount termfreq, char c) const;
-    double get_wtn(Xapian::termcount len, double wt, char c) const;
+    double get_wtn(double wt, char c) const;
 
   public:
     /** Construct a TfIdfWeight
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -404,16 +404,16 @@ class XAPIAN_VISIBILITY_DEFAULT BoolWeight : public Weight {
 
 /// Xapian::Weight subclass implementing the tf-idf weighting scheme.
 class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
-    /// Parameters in the Piv+ normalization weighting formula.
-    double param_slope = 0;
-    double param_delta = 0;
+    /* Three character string indicating the normalizations for tf(wdf), idf and
+       tfidf weight. */
+    std::string normalizations;
 
     /// The factor to multiply with the weight.
     double factor;
 
-    /* Three character string indicating the normalizations for tf(wdf), idf and
-       tfidf weight. */
-    std::string normalizations;
+    /// Parameters in the Piv+ normalization weighting formula.
+    double param_slope = 0;
+    double param_delta = 0;
 
     TfIdfWeight * clone() const;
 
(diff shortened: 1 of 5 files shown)


xapian-core/weight/tfidfweight.cc
```diff
@@ -70,26 +74,48 @@ TfIdfWeight::name() const
...
+	result += serialise_double(param_delta);
+	return result;
+    } else {
+	return normalizations;
+    }
 }
 
 TfIdfWeight *
-TfIdfWeight::unserialise(const string & s) const
+TfIdfWeight::unserialise(const string & str) const
```
jaylett: Was there a need to change the variable name? It makes it slightly harder to see what changes have been made (here it's not hugely important, but in a more complex function or change it might be).

xapian-core/weight/tfidfweight.cc
```diff
@@ -70,26 +74,48 @@ TfIdfWeight::name() const
...
+    }
 }
 
 double
-TfIdfWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount,
+TfIdfWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount doclen,
 			 Xapian::termcount) const
 {
     Xapian::doccount termfreq = 1;
+    double wqf_double = get_wqf();
```
jaylett: Should this not be just called `wqf`? Also, you probably don't want to pull it unless the normalisations are in play.

xapian-core/weight/tfidfweight.cc
```diff
@@ -153,18 +188,29 @@ TfIdfWeight::get_idfn(Xapian::doccount termfreq, char c) const
...
 }
 
 double
-TfIdfWeight::get_wtn(double wt, char c) const
+TfIdfWeight::get_wtn(Xapian::termcount doclen, double wt, char c) const
 {
-    (void)c;
-    AssertEq(c, 'n');
-    return wt;
+    double normlen;
```
jaylett: You should declare variables as locally as possible, which would be within the block for the `'P'` case.

xapian-core/include/xapian/weight.h
```diff
@@ -449,20 +454,38 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
...
+     *              Note: Valid parameters s and delta values should be non-negative
+     *                    & non-zero when using pivoted normalization string("PPP")
      *
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
      */
     explicit TfIdfWeight(const std::string &normalizations);
 
+    explicit TfIdfWeight(double s, double delta)
+	: param_s(s), param_delta(delta)
```
jaylett: What is `normalizations` in this case?

xapian-core/weight/tfidfweight.cc
```diff
@@ -70,26 +74,48 @@ TfIdfWeight::name() const
...
 }
 
 TfIdfWeight *
-TfIdfWeight::unserialise(const string & s) const
+TfIdfWeight::unserialise(const string & str) const
 {
-    if (s.length() != 3)
-	throw Xapian::SerialisationError("Extra data in TfIdfWeight::unserialise()");
-    return new TfIdfWeight(s);
+    if (param_s > 0 || param_delta > 0) {
```
jaylett: I'm not sure this is what you mean. I seem to need to construct a `TfIdfWeight` with positive params in order to unserialise one with positive params. And I need one with zero params to unserialise one with a normalisation string.

xapian-core/include/xapian/weight.h
```diff
@@ -449,20 +454,38 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
...
+     *     @li 'P': Pivoted idfn=log((N+1)/Termfreq)
      *
      * @li The third and the final character indicates the normalization for
      *     the document weight.  The following normalizations are currently
      *     supported:
      *
      *     @li 'n': None wtn=tfn*idfn
+     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-s+(s*normlen))+delta*idfn)
+     *              Note: Valid parameters s and delta values should be non-negative
+     *                    & non-zero when using pivoted normalization string("PPP")
```
jaylett: This suggests that `"PPP"` is the only valid use of pivoted normalisation. Is that the case? If it is, maybe we should be approaching configuring this class very differently.

xapian-core/weight/tfidfweight.cc
ivmarkp: I've used variable name `s` for parameter s in line 90:

`double s = unserialise_double(&ptr, end);`

Should I use some other name for it? May be using original name i.e. `slope` would be more appropriate.

xapian-core/weight/tfidfweight.cc
ivmarkp: Will change it with a fixup commit.

xapian-core/weight/tfidfweight.cc
ivmarkp: Noted. Thanks, will fix it.

xapian-core/include/xapian/weight.h
ivmarkp: Basically, this constructor facilitates changing values of parameters `s` and `delta`. For this particular constructor to be of use normalization string should be set to `"PPP"` using another constructor:

`explicit TfIdfWeight(const std::string &normalizations);`

xapian-core/weight/tfidfweight.cc
ivmarkp: We don't need to worry about setting parameters to zero as they have been purposely initialised to zero. And since we want only non-negative and non-zero values of parameters, this makes sure appropriate serialisation takes place when constructing `TfIdfWeight` either for normalization strings or passing parameter values. This [test](https://github.com/xapian/xapian/pull/115/files?diff=unified#diff-89288dcbe247bb4c55a8f4591ebda81bR672) demonstrates the idea.

And, this [test ](https://github.com/xapian/xapian/pull/115/files?diff=unified#diff-89288dcbe247bb4c55a8f4591ebda81bR688) is an example to see how we'll be using pivoted normalization where one constructor (at line 690) to select `"PPP"` normalization string and another in the next line to pass parameter values.

xapian-core/include/xapian/weight.h
ivmarkp: Yes,`"PPP"` normalization string should be used if we want to use Piv+ normalization but we can use `xPn` and `Pxn` to provide more combination options for tf and idf normalization to users.

xapian-core/weight/tfidfweight.cc
jaylett: `slope` would be better; single-character variable names should generally be avoided. (`i`, `j` for indices is okay, but for everything else being more explicit is generally more readable.)

xapian-core/include/xapian/weight.h
jaylett: You can't do that. You only get to call one constructor.

xapian-core/weight/tfidfweight.cc
jaylett: If I have a serialised `TfIdfWeight`, _without_ knowing whether it's using a normalisation string or the params (which, see previous comment about constructors, I don't think is the distinction you want), how do I know how to unserialise it? A serialised version needs to be able to be serialised without knowing how it was built.

xapian-core/tests/api_weight.cc
```diff
@@ -669,6 +669,35 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
...
+DEFINE_TESTCASE(tfidfweight5, backend) {
+    Xapian::Database db = get_database("apitest_simpledata");
+    Xapian::Enquire enquire(db);
+    Xapian::Query query("paragraph");
+    Xapian::MSet mset;
+
+    // Check for "PPP" normalization string.
+    enquire.set_query(query);
+    enquire.set_weighting_scheme(Xapian::TfIdfWeight("PPP"));
+    Xapian::TfIdfWeight(0.2, 1.0);
```
jaylett: What do you think this does?

xapian-core/include/xapian/weight.h
jaylett: Okay, so this comment needs to be more carefully-worded. Is `xxP` with the `x`s not `P` allowed? I'm not quite clear what the range of allowable configurations is at this point.

xapian-core/include/xapian/weight.h
ivmarkp: I can't update parameters values in that case. May be should I write a new method that updates parameter values?

xapian-core/weight/tfidfweight.cc
ivmarkp: >  A serialised version needs to be able to be serialised without knowing how it was built.
Oh, now I understand. It wouldn't be obvious for someone who doesn't know how serialisation was done to use how to unseriunserialise. Sorry, it looked so obvious to me. Do you any ideas about a good way of doing this? I can only think of writing a new method and use it to update parameter values at this point.

xapian-core/include/xapian/weight.h
ivmarkp: `xxP` can also be used if someone wants different `wtn` normalization. Sorry, missed that one. I'll make this comment more clear.

xapian-core/tests/api_weight.cc
ivmarkp: Create `TfIdfWeight` with updated parameter `s` and `delta` with values `0.2` and `1.0` respectively

xapian-core/include/xapian/weight.h
jaylett: Can't you just provide an overloaded constructor that takes `const std::string &normalizations, double _param_s, double _param_delta`?

xapian-core/tests/api_weight.cc
jaylett: No, it creates one `TfIdfWeight` using `"PPP"` as the normalisations, then another with `0.2, 1.0` as the two params (and then throws the second one away because you don't assign it to anything).

If you break line 690 down, you can assign the `TfIdfWeight` to a variable so you can play with it directly — however you can't set the parameters using the current API, so this isn't helpful. But I do think that a single constructor is the right approach.

xapian-core/weight/tfidfweight.cc
jaylett: Did you find [this document](https://xapian.org/docs/serialisation.html)? It explains what serialisation is used for, which might help.

For serialisation/unserialisation, you need a way of passing all the parameters you might need. I'd suggest putting the two doubles first, then using the rest as the normalisations string. (I don't believe it's a problem that this is different to previously, because I _think_ this is just a wire protocol, ie used at runtime. You could always check for a serialised string of length 3 and use that as normalisations, to support the old way if it's needed.)

xapian-core/include/xapian/weight.h
jaylett: Ah, okay. I think we definitely are going to need an advanced document in the getting started guide, similar to the one for tuning `LMWeight`, which explains all these different variants, points out that `"PPP"` is Piv+, and (hopefully) gives some idea of when you might use the different ones.

xapian-core/include/xapian/weight.h
ivmarkp: I think I tried that earlier ([commit](https://github.com/ivmarkp/xapian/commit/e31a46e5a5ee1bdd3931de25e25722e46df056fe)) but tests were failing with serialisation errors. But as you've mentioned below:

> For serialisation/unserialisation, you need a way of passing all the parameters you might 
> need. I'd suggest putting the two doubles first, then using the rest as the normalisations 
> string.

I think the problem was with putting normalization strings before the two parameters everywhere. I'll try the way you suggest and will update this PR. Thanks for clarifying doubts.

xapian-core/include/xapian/weight.h
```diff
@@ -471,10 +471,12 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
      */
-    explicit TfIdfWeight(double slope, double delta, const std::string &normalizations);
+    explicit TfIdfWeight(const std::string &normalizations);
+
+    explicit TfIdfWeight(const std::string &normalizations, double slope, double delta);
```
jaylett: You shouldn't need explicit with three formal parameters to a constructor.

xapian-core/include/xapian/weight.h
```diff
@@ -471,10 +471,12 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
...
 
     TfIdfWeight()
-	: param_slope(0.2), param_delta(1.0), normalizations("ntn")
+	: normalizations("ntn")
```
jaylett: If I construct `TfIdfWeight("PPP")`, should slope and delta be `0.0` or `0.2` and `1.0` respectively? What's more useful?

ivmarkp: Slope and delta should be `0.2` and `1.0` respectively by default as proposed in the paper. Updated the default values from the current initialised `0.0` to `0.2` and `1.0`. 

So, in case of `TfIdfWeight("PPP")` default values will be there for slope and delta or if someone wants to use `0.0` then constructing `TfIdfWeight("PPP", 0.0, 0.0)` should be the ideal choice.

jaylett: That feels like a more natural API to me — great!

xapian-core/include/xapian/weight.h
```diff
@@ -449,25 +455,37 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
...
      *     @li 's': Squared idfn=log(N/Termfreq)^2
+     *     @li 'P': Pivoted idfn=log((N+1)/Termfreq)
      *
      * @li The third and the final character indicates the normalization for
      *     the document weight.  The following normalizations are currently
      *     supported:
      *
      *     @li 'n': None wtn=tfn*idfn
+     *     @li 'P': Pivoted wtn=wqf(tfn*idfn*(1-slope+(slope*normlen))+delta*idfn) where
+     *         parameters slope and delta values should be non-negative & non-zero when
```
ojwb: Let's just say "positive"!

xapian-core/include/xapian/weight.h
```diff
@@ -449,25 +455,37 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
...
+     *         using pivoted normalization string. Specifically, "PPP" normalization
+     *         string should be used to use Piv+ normalization. In addtion, normalization
```
ojwb: typo: "addtion"

xapian-core/include/xapian/weight.h
```diff
@@ -449,25 +455,37 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
...
+     *         string such as "Pxx", "xPx" and "xxP" can also be used to have more options
+     *         of combinations for wdfn, idfn and wtn normalizations respectively.
```
ojwb: I don't think it's clear enough that the `x` characters here are mean to be placeholders rather than literal - perhaps better to instead say something like "The different types of pivoted normalisation can also be used in combination with other normalisations"?

xapian-core/weight/tfidfweight.cc
```diff
@@ -49,10 +50,31 @@ TfIdfWeight::TfIdfWeight(const std::string &normals)
...
+	need_stat(TERMFREQ);
+	need_stat(COLLECTION_SIZE);
+    }
+    need_stat(WDF);
+    need_stat(WDF_MAX);
+    if (param_slope != 0 || param_delta != 0) {
+	need_stat(AVERAGE_LENGTH);
+	need_stat(DOC_LENGTH);
+	need_stat(WQF);
+    }
```
ojwb: Shouldn't this also be conditional on the relevant `P` normalisations actually being specified?

Otherwise we end up fetching stats that will never be used.

xapian-core/weight/tfidfweight.cc
```diff
@@ -49,10 +50,31 @@ TfIdfWeight::TfIdfWeight(const std::string &normals)
...
+}
+
 TfIdfWeight *
 TfIdfWeight::clone() const
 {
-    return new TfIdfWeight(normalizations);
+	return new TfIdfWeight(normalizations, param_slope, param_delta);
 }
```
ojwb: Overindented...

xapian-core/include/xapian/weight.h
```diff
@@ -409,6 +409,11 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     /// The factor to multiply with the weight.
     double factor;
 
+    /// Parameters slope and delta in the Piv+ normalization weighting formula.
+    /// Default values: 0.2 and 1.0 respectively.
+    double param_slope = 0.2;
+    double param_delta = 1.0;
+
     TfIdfWeight * clone() const;
```
ojwb: We've so far avoided C++11 features in the public API headers (because many commonly used compiler versions need C++11 enabling explicitly, and if we use C++11 features, projects using Xapian will need to make their build systems detect what (if any) option is needed.

Given this one is easy to avoid, I think we should - just initialise in the default constructor instead.  It's probably clearer to keep the initialisations together anyway.

ojwb: Actually, with the default constructor they won't be used at all, so there's no need to initialise - it's the constructor just taking a `const std::string&` which needs them initialising.

xapian-core/weight/tfidfweight.cc
```diff
@@ -49,10 +50,31 @@ TfIdfWeight::TfIdfWeight(const std::string &normals)
     need_stat(WDF_MAX);
 }
```
ojwb: You don't request the extra stats needed by one of the `P` normalisations if this constructor is used.

xapian-core/weight/tfidfweight.cc
```diff
@@ -70,26 +92,41 @@ TfIdfWeight::name() const
...
+TfIdfWeight::get_sumpart(Xapian::termcount wdf, Xapian::termcount doclen,
 			 Xapian::termcount) const
 {
     Xapian::doccount termfreq = 1;
     if (normalizations[1] != 'n') termfreq = get_termfreq();
     double wt = get_wdfn(wdf, normalizations[0]) *
 		get_idfn(termfreq, normalizations[1]);
-    return get_wtn(wt, normalizations[2]) * factor;
+    if (normalizations[2] == 'P' && normalizations[1] == 'P') {
+	double wqf = get_wqf();
```
ojwb: Why is this conditional of the second and third normalisations both being `P`?  The documentation you wrote says that xPx and xxP can both be used, but in those terms here you check for xPP .

xapian-core/weight/tfidfweight.cc
```diff
@@ -70,26 +92,41 @@ TfIdfWeight::name() const
...
+	wt = get_wtn(doclen, wt, normalizations[2]) + param_delta * get_idfn(termfreq, normalizations[1]);
+	return wqf * get_wtn(get_doclength_lower_bound(), wt, normalizations[2]) * factor;
+    }
+    else
+	return get_wtn(doclen, wt, normalizations[2]) * factor;
 }
```
ojwb: Another naughty `else` here (and below).

xapian-core/weight/tfidfweight.cc
```diff
@@ -153,18 +199,29 @@ TfIdfWeight::get_idfn(Xapian::doccount termfreq, char c) const
...
 {
-    (void)c;
-    AssertEq(c, 'n');
-    return wt;
+    switch (c) {
+	case 'P':
+	{
+	    double normlen;
+	    normlen = doclen / get_average_length();
+	    double norm_factor =  1 / (1 - param_slope + (param_slope * normlen));
```
ojwb: Xapian convention is to "cuddle" braces, so that should be:

```
        case 'P': {
```

The rationale is that vertical screen space is limited, and a layout which uses fewer lines without trading off readability means you can view more code in the same height window.

Also it's cleaner to initialise a variable at the point of declaration where you can, like so:

```
    double normlen = doclen / get_average_length();
```

xapian-core/include/xapian/weight.h
ivmarkp: Yes, that sounds more clear. Thanks, making a change here.

xapian-core/weight/tfidfweight.cc
ivmarkp: It is based on the assumption that if `param_slope` and `param_delta` are set externally then the intention must have be to use one of the pivoted normalisation so it seemed fine to leave `P` normalisation check out here.

xapian-core/include/xapian/weight.h
ivmarkp: The constructor just taking a `const std::string&` takes non-pivoted normalizations that don't require `param_slope` and `param_delta` and so does default constructor. Making a way for initialisation of `param_slope` and `param_delta` requires changing the parameter list of this constructor which would break existing tests using this constructor e.g. `TfIdfWeight("ntn")`.

I think we can do the initialisation in overloaded constructor at line [477](https://github.com/xapian/xapian/pull/115/files#diff-8a2d77838da5b7da44ddb21e3d97acb1R477) as:

```c++
TfIdfWeight(const std::string &normalizations, double slope = 0.2, double delta = 1.0);

File: xapian-core/weight/tfidfweight.cc (hunk 14 above)
Comment (ojwb) [inline xapian-core/weight/tfidfweight.cc]: Still unfixed.

Comment (ivmarkp) [inline xapian-core/weight/tfidfweight.cc]: Sorry, I was planning to make changes at one go after getting your input on some comments below. Anyways, fixing these formatting issues now.

File: xapian-core/include/xapian/weight.h (hunk 15 above)
Comment (ojwb) [inline xapian-core/include/xapian/weight.h]: As things stand, you initialise the new parameters to non-zero values in the class definition:

```
    double param_slope = 0.2;
    double param_delta = 1.0;
```

So actually I could write: `TfIdfWeight weight("PPP");` and it would work, except that the stats aren't requested correctly in this case.

> Making a way for initialisation of `param_slope` and `param_delta` requires changing the parameter list of this constructor which would break existing tests using this constructor e.g. `TfIdfWeight("ntn")`.

No, you're already initialising them (as above), though as I already noted I don't think you should do it like that, as that's a C++11 feature, which we don't want in the public headers currently - we need to instead initialise in the constructor definition (in `tfidfweight.cc`) like so:

```
TfIdfWeight::TfIdfWeight(const std::string &normals)
    : normalizations(normals), param_slope(0.2), param_slope(1.0)
{
```

The default constructor doesn't need these parameters initialising, as the normalisations is fixed as `"ntn"` in that case.

Are those default parameter values actually sensible choices for a typical user?

If not, then I think we need to reject attempts to use the `P` normalisations without explicitly specifying these parameters.

If they are, then we need to request the stats correctly, and based on whether the normalisation(s) which used them are `P`, not whether the parameters are set (because they are set by default).

File: xapian-core/weight/tfidfweight.cc (hunk 13 above)
Comment (ojwb) [inline xapian-core/weight/tfidfweight.cc]: That's not a good assumption - people may make the normalisations configurable, and then it's probably simpler to just call the most general constructor and always supply slope and delta values, rather than conditionalise which constructor to call based on the contents of the normalisation string.

Fundamentally, the checks for which stats we request should exactly match the conditions under which we use those stats - that's more robust and we don't have to worry about what happens in the cases where one condition is met and the other isn't.  We check for `P` in the normalisation string before we use them, so we need to check for that here too.

File: xapian-core/include/xapian/weight.h (hunk 15 above)
Comment (ivmarkp) [inline xapian-core/include/xapian/weight.h]: @ojwb I've made the changes you suggested and all tests pass under `make check` and gdb but valgrind reports some errors -- mainly two errors spread over a no. of statements. 
First one is again: 
```
Conditional jump or move depends on uninitialised value(s)
at 0x4F937C5: MultiMatch::get_mset(unsigned int, unsigned int, unsigned int, Xapian::MSet&, Xapian::Weight::Internal&, Xapian::MatchDecider const*, Xapian::KeyMaker const*) (multimatch.cc:802)
==1046==    by 0x4E92250: Xapian::Enquire::Internal::get_mset(unsigned int, unsigned int, unsigned int, Xapian::RSet const*, Xapian::MatchDecider const*) const (omenquire.cc:548)
==1046==    by 0x4E92573: Xapian::Enquire::get_mset(unsigned int, unsigned int, unsigned int, Xapian::RSet const*, Xapian::MatchDecider const*) const (omenquire.cc:906)
==1046==    by 0x5B9118: test_tfidfweight4() (api_weight.cc:682)
```
and similar one by `api_weight.cc:691`. Also, another error which I think occurs due to something's wrong with `TfIdfWeight::clone()`:
```
Uninitialised value was created by a heap allocation
==1046==    at 0x4C2E0EF: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1046==    by 0x4FC7F32: Xapian::TfIdfWeight::clone() const (tfidfweight.cc:82)
==1046==    by 0x4F8D789: LocalSubMatch::open_post_list(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int, double, bool, QueryOptimiser*, bool) (localsubmatch.cc:300)
==1046==    by 0x4F8C7D9: LocalSubMatch::get_postlist(MultiMatch*, unsigned int*) (localsubmatch.cc:204)
==1046==    by 0x4F92AF0: MultiMatch::get_mset(unsigned int, unsigned int, unsigned int, Xapian::MSet&, Xapian::Weight::Internal&, Xapian::MatchDecider const*, Xapian::KeyMaker const*) (multimatch.cc:419)
==1046==    by 0x4E92250: Xapian::Enquire::Internal::get_mset(unsigned int, unsigned int, unsigned int, Xapian::RSet const*, Xapian::MatchDecider const*) const (omenquire.cc:548)
==1046==    by 0x4E92573: Xapian::Enquire::get_mset(unsigned int, unsigned int, unsigned int, Xapian::RSet const*, Xapian::MatchDecider const*) const (omenquire.cc:906)
==1046==    by 0x5B9975: test_tfidfweight4() (api_weight.cc:691)
```
Could you please tell me what I might be missing here?

Comment (ojwb) [inline xapian-core/include/xapian/weight.h]: It's probably most useful if I interactively help you work through how to interpret these.  There's a bit of a knack to working out what the problem is, and if you understand the processes to go through then you're more likely to be able to work it out yourself next time.

I'd suggest we do that on IRC.  I'm about to disconnect for the evening, but I'm mostly around NZ daytime currently - just ping me on #xapian.

Comment (ivmarkp) [inline xapian-core/include/xapian/weight.h]: @ojwb Sorry about the confusion. I think I was confusing delta used here with the one used with BM25+ and thus set its default value to be 1.0 but for slope, the value 0.2 came from table 10 and conclusions from the text in the two paragraphs below it. Overall there's no particular default value for piv+ delta, not even in the summary section (7.6). 

We actually want the slope to assume values between 0.01 to 0.25 and delta between 0.0 to 1.5 (or 0.1 to 1.5 so that both params assume only positive values) according to the paper.

Should we try to reject `P` normalisations when params values are not specified as you suggested above? We can achieve this by removing `P` normalisation character from the invalid normalisation string check in `TfIdfWeight::TfIdfWeight(const std::string &normals)` as:

```c++
if (normalizations.length() != 3 ||
	!strchr("nbsl", normalizations[0]) ||
	!strchr("ntpfs", normalizations[1]) ||
	!strchr("n", normalizations[2]))
	throw Xapian::InvalidArgumentError("Normalization string is invalid");
``` 

and add a check for the situation when `para_slope` and `param_delta` values are set zero or negative which assigns a positive value to them (not sure which values to assign) in the definition of `TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)`.

ojwb: My reading of the paper is that they are saying they tuned the parameters within the ranges they give in each case, but that doesn't mean that any values outside those ranges are invalid (just that the authors think the optimal values are likely to be within those ranges, for reasons they don't seem to explain).

Here they describe the tuning for BM25 parameter's `b` and `k1`:
> In both BM25+ and BM25, we train b and k1 using cross validation, where b is tuned from 0.1 to 0.9 in increments of 0.1, and k1 is tuned from 0.2 to 4.0 in increments of 0.2.

But BM25 only requires `0<=b<=1` and `k1>=0`.

Some values of the parameters added in the + variants may actually be invalid (I don't see anything explicitly saying it, but it seems in each case δ should be positive, or at least non-negative) - I think we should reject any values which are really invalid, but still allow values outside the ranges Lv and Zhai tuned within as these may still be interesting.

Checking the paper, I agree that they don't really suggest default values for Piv+ (table 10 suggests values for `s` between 0.01 and 0.19 depending on the collection and nature of the query, which is a sufficiently large range that picking a single good default is hard).

Rejecting `P` for `normalizations[2]` in the default constructor so the user has to specify slope and delta seems a good plan - I'm not sure we need to reject `P` for the other two normalisations though - they don't seem to use slope or delta.

I think in the "full" constructor that `slope <= 0` or `delta <= 0` should be rejected with an exception - that's what we do for most of the weighting schemes.  `BM25Weight` just clamps parameters to be within the valid range because it always has and people could be relying on that happening, but for the schemes implemented more recently we decided rejecting invalid values with an exception was more helpful to users.  And as you note, if we force them to something valid we have to pick a value to force them to, which would just bring us back to effectively picking a good default.

ivmarkp: Thanks for your detailed clarification; I've made the changes accordingly.

ivmarkp:
`dlhweight1` test case is failing in travis check which is weird because I did not make any changes in `api_weight.cc`. All tests pass locally; not sure what caused that test case to fail here though. :/

xapian-core/weight/tfidfweight.cc
```diff
@@ -63,16 +59,21 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
...
+    if (param_delta <= 0)
+	throw Xapian::InvalidArgumentError("Parameter delta is invalid.");
     if (normalizations[1] != 'n') {
 	need_stat(TERMFREQ);
 	need_stat(COLLECTION_SIZE);
     }
     need_stat(WDF);
     need_stat(WDF_MAX);
-    if (param_slope != 0 || param_delta != 0 || normalizations[2] == 'P') {
+    if ((param_slope > 0 || param_delta > 0) && normalizations[2] == 'P') {
```
ojwb: Now you're rejecting values `<= 0` in the constructor, `param_slope > 0` and `param_delta > 0` are now what's called "invariants" of this class.  If you want to check invariants, generally you use assertions, e.g.:

```
AssertRel(param_slope, >, 0);
AssertRel(param_delta, >, 0);
```

I'm not sure I'd bother here as you rejected the invalid values just above in the same function.  But more generally (e.g. when you're testing an invariant in different methods to where it was established) it can be useful as a defence against someone changing the code later and violating the invariant.

So you only need the final check here now.

ojwb:
The dlhweight1 test failure is presumably my fault - I removed the flawed lower bounding approach from `DLHWeight` and `DPHWeight` recently.  I'll investigate.

ojwb:
Fixed dlhweight1 in 0bec6b3919f0055176dd852088afae3f5da6e580.  I thought I'd tested it before committing, but it was failing on my machine too, so particularly sorry about that.

ivmarkp:
@ojwb Thanks for the fix.

xapian-core/weight/tfidfweight.cc
ivmarkp: Okay, got it; dropping `param_slope > 0` and `param_delta > 0` checks and leaving just the final check here.

ojwb:
Don't worry about changing it here now, but for future reference `fixup! 1371b37` doesn't work - the text after `fixup!` needs to match the start of the commit **message** for the commit you are fixing up (the reason to use the message instead of the hash is that it's more robust to the stack of commits being rebased or the commit messages being changed or subsets being squashed.  The hash is over various things include the contents of the commit, the commit message and the hashes of its parent(s), author and probably a few other things, so it changes when any of these are changed.

The simplest way to actually make a fixup commit is to add the files as usual (`git add -p` or whatever) and then just do `git commit --fixup 1371b37` - this will commit with a suitable message based on the commit message of commit 1371b37.

ivmarkp:
Noted. Will use the usual `git commit --fixup` command from now on.

ojwb:
OK, I've squashed and pushed these changes (and a missing space I didn't spot before in cbc6034eae343e9663dc524c880f8f59575b2680).

However, I'm wondering though if the split of the Piv+ formula into the three normalisations isn't the most natural one given the existing SMART framework.

The standard pattern is to apply the SMART normalisations as:

```
get_wtn(get_wdfn(wdf) * get_idfn(term_freq))
```

where each of the three functions is "pluggable" and specified by a different letter of the three letter code.

The way you split up Piv+ was to add a `doclen` parameter to `get_wtn()` and then as a special case change the pattern the three normalisations are combined using to be:

```
(get_wdfn(wdf) * get_wtn(doclen, 1.0) + delta) * get_idfn(term_freq)
```

This special casing seems inelegant, as does passing `1.0` for the weight to `get_wtn()`.

I wonder if we'd be better to split the normalisation up such that the square brackets in (14) in the paper are `get_wdfn(wdf, len, 'P')`, the `log` is `idf_norm(termfreq, 'P')` and the weight normalisation is just the identity (i.e. 'n').  So the code for Piv+ would then be `PPn`.

In this approach, `get_wdfn()` gains a `doclength` parameter (instead of `get_wtn()` doing so), which is something I think it would also need to do to support the `L` normalisation listed here:

https://en.wikipedia.org/wiki/SMART_Information_Retrieval_System

In Xapian terms, `tf(t,d)` is `wdf`, and `ave({t in d})(tf(t,d))` is `doclen/unique_terms` (since in Xapian `doclen(d)` is defined as `sum({t in d})(tf(t,d))`, and `unique_terms` is the size of the set `{t in d}`).

I've merged the current state to git master anyway, as it seems unhelpful to throw in a spanner like this late on, and I'd rather not block merging of the evaluation PRs on something like this.  But it would be good to think about, and resolve unless there's a good reason for the current design.

ivmarkp:
> I wonder if we'd be better to split the normalisation up such that the square brackets in (14) in the paper are get_wdfn(wdf, len, 'P'), the log is idf_norm(termfreq, 'P') and the weight normalisation is just the identity (i.e. 'n'). So the code for Piv+ would then be PPn.

I think this is a better approach indeed. I've made the changes. All tests pass, although I dropped the `xxP` test case entirely as it seemed irrelevant with `P` not being present anymore as `normalizations[2]` character.

Also, while we are at it I'm planning open another PR implementing the `L` wdf normalisation support next. It seems simple given that we have the required stats readily available. Is there any other detail I'm missing out?

ivmarkp:
Rebased commits in this PR on top of latest master to resolve merge conflicts. Travis check should now be functional on some of the recent commits.

ojwb:
OK - squashed and merged as 334c09c6737425639741f4a0c7fed6959ae27ae7 (the fixup commits didn't all apply cleanly, so just squashing to one commit seemed simplest).

> I think this is a better approach indeed. I've made the changes. All tests pass, although I dropped the xxP test case entirely as it seemed irrelevant with P not being present anymore as `normalizations[2]` character.

Yes, removing that testcase makes sense with this change.

> Also, while we are at it I'm planning to open another PR implementing the L wdf normalisation support next. It seems simple given that we have the required stats readily available. Is there any other detail I'm missing out?

That'd be great - I can't see any particular obstacles to doing it.

ivmarkp:
@ojwb Great, thanks.

PR #116, opened 2016-08-02
letor: Remove RankList

ayshtmr:
This PR is majorly centered around eliminating `RankList` from letor.

`RankList` is simply a class that bundles a `string qid` with corresponding `vector<FeatureVector>`. The use of `RankList` only makes sense while preparing the training file, when qrel and query files are to be parsed. In rest of letor, all methods can be performed independently of `qid`; `vector<FeatureVector>` is self-sufficient. Hence, following changes are done:
- Remove `RankList` and replace it by `vector<FeatureVector>` in all methods.
- Reposition some methods like `getlabel(*)` in FeatureManager into letor_internal.cc, to be able to work directly in `prepare_training_file(*)` method.

Apart from `RankList` changes, this PR also does the following:

- Update includedir path in configure.ac for Xapian 1.4.0 release
- In `Ranker class`: Add documentation, update doccomments, and made some functions as pure virtual functions, that will be overriden by corresponding subclass functions.


xapian-letor/configure.ac
```diff
@@ -36,7 +36,7 @@ AC_SUBST(LIBRARY_VERSION_INFO)
 
 dnl Where xapian-letor.h, etc go.  In development release append "/xapian-1.3".
 dnl incdir=$includedir
-incdir=$includedir/xapian-1.3
+incdir=$includedir
```
jaylett: I suspect the comment is wrong, and we actually want `$includedir/xapian`.

xapian-letor/include/xapian-letor/featuremanager.h
```diff
@@ -67,34 +66,18 @@ class XAPIAN_VISIBILITY_DEFAULT FeatureManager {
...
-    /** Loads up a qrel file.
-     *  @return a map: < qid <docid, relevance_judgement> >
-     */
-    map<string, map<string,int> > load_relevance(const std::string & qrel_file);
-
-    /** Creates instance of FeatureVector class that stores features_array, relevance label and corresponding document id
-     *  @return a FeatureVector instance containing features_array, relevance label and corresponding document id
-     */
-    Xapian::FeatureVector create_feature_vector(map<int,double> fvals, int &label, Xapian::docid & did);
+    std::vector<Xapian::FeatureVector> create_vec_fvec(const Xapian::MSet & mset);
```
jaylett: `create_feature_vectors` (with an 's') might be a clearer name?

xapian-letor/api/ranker.cc
```diff
@@ -59,6 +63,19 @@ Ranker::Ranker(int metric_type) { //TODO: update this when adding scorers
 
 }
 
+Ranker::Ranker(int metric_type, double learn_rate, int num_iterations) {
+    MAXPATHLEN = 200;
+    // switch(metric_type) {
```
jaylett: Let's not commit this commented code.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -49,28 +49,50 @@ class XAPIAN_VISIBILITY_DEFAULT Ranker { //TODO: Update documentation
...
 
-    virtual void train_model();
+    /// Method to train the model. Overrided in ranker subclass.
+    virtual void train_model()=0;
+
+    /// Method to save model as an external file. Overrided in ranker subclass.
+    virtual void save_model_to_file()=0;
 
-    virtual void save_model_to_file();
+    /// Method to save model as an external file. Overrided in ranker subclass. Overrided in ranker subclass.
```
jaylett: This comment seems wrong for this method?

xapian-letor/api/ranker.cc
```diff
@@ -37,11 +37,15 @@
...
 
-struct scoreComparer {
-    bool operator()(const pair<Xapian::docid, int>& first_pair, const pair<Xapian::docid, int>& second_pair) const {
-        return first_pair.second > second_pair.second;
-    }
-};
+bool
+Ranker::ScoreComparer::operator()(const FeatureVector & firstfv, const FeatureVector& secondfv) const {
+    return firstfv.get_score() > secondfv.get_score();
+}
```
jaylett: Is there a reason these would need to be overridden? I'd have thought just providing them as utility functions would be as helpful (and should take a little less memory).

xapian-letor/api/featurevector_internal.cc
```diff
@@ -117,3 +117,31 @@ FeatureVector::Internal::get_nonzero_num(){
...
+
+     /* We have the maximum value of each feature overall.
+        Now we need to normalize each feature value of a featureVector
+        by dividing it by the corresponding max of the feature value
+    */
+    for(int j=0; j<19; ++j) {
+        temp = fvals.find(j)->second;
+        temp /= max[j];
+        fvals.insert(pair<int,double>(j,temp));
+        temp = 0.9;
```
jaylett: This shouldn't be needed?

xapian-letor/api/featurevector_internal.cc
```diff
@@ -117,3 +117,31 @@ FeatureVector::Internal::get_nonzero_num(){
...
+    }
+
+     /* We have the maximum value of each feature overall.
+        Now we need to normalize each feature value of a featureVector
+        by dividing it by the corresponding max of the feature value
+    */
+    for(int j=0; j<19; ++j) {
+        temp = fvals.find(j)->second;
+        temp /= max[j];
+        fvals.insert(pair<int,double>(j,temp));
```
jaylett: AIUI, `std::map::insert()` won't overwrite an existing key. You may need to use `fvals[j] = temp` to get the desired effect.

xapian-letor/api/featurevector_internal.cc
```diff
@@ -117,3 +117,31 @@ FeatureVector::Internal::get_nonzero_num(){
...
 }
+
+void
+FeatureVector::Internal::normalise() {
+
+    int num_features = 19;
+    double temp = 0.0;
+    double max[num_features];
+
+    for(int i=0; i<19; ++i)
```
jaylett: Indentation seems missing here, but it's probably clearest to wrap as an actual block.

xapian-letor/api/featurevector_internal.cc
```diff
@@ -117,3 +117,31 @@ FeatureVector::Internal::get_nonzero_num(){
     //cout << "nonzero: "<< nonzero <<endl;
     return nonzero;
 }
+
+void
+FeatureVector::Internal::normalise() {
+
+    int num_features = 19;
```
jaylett: Magic number! Should be declared once, or (better?) detected from the size of the map.

xapian-letor/api/featurevector_internal.cc
```diff
@@ -117,3 +117,31 @@ FeatureVector::Internal::get_nonzero_num(){
     //cout << "nonzero: "<< nonzero <<endl;
     return nonzero;
 }
+
+void
+FeatureVector::Internal::normalise() {
```
jaylett: Looking at this a little further, I'm not convinced this function does what is needed. Is it normalising the feature vector, so its length is 1, or should it be normalising each feature's values across the range of feature vectors. Because this code seems to do neither. Perhaps some tests would help clarify this?

jaylett:
This doesn't build for me because of the unused `Scorer * scorer` variable in `Xapian::Ranker`. Are you extracting this from a wider set of changes? I'm not sure how you could get this to build without overriding CXXFLAGS (which you really should not do).

I wonder if we should take the plunge at this point and remove `xapian-letor/.nobootstrap`, remove `xapian-letor` from the `bootstrap` defaults, and put it explicitly into `.travis.yml` so letor PRs get built (and, in future, tested) automatically.

xapian-letor/configure.ac
ayshtmr: Yes, it should be `$includedir/xapian`. -core's configure.ac has the same comment and `incdir` path. That should also be fixed. Will push a change for both of them.

xapian-letor/include/xapian-letor/featuremanager.h
ayshtmr: I put the name as `create_vec_fvec` instead of `create_feature_vectors` since FeatureManager has a closely named method called `create_feature_vector` (which will actually be removed in the Features PR). But still, `create_vec_fvec` isn't very intuitive. Will make the change.

xapian-letor/api/ranker.cc
ayshtmr: Sorry for being careless about that. Will remove.

xapian-letor/api/featurevector_internal.cc
ayshtmr: Thanks for pointing this out. So stupid of me, there wasn't any normalisation taking place up until now.

xapian-letor/api/featurevector_internal.cc
ayshtmr: I plan to get rid of these hard-coded values of '19's in the next `Features` PR. Can be merged for now.

xapian-letor/api/featurevector_internal.cc
ayshtmr: Its intended use was to normalise each feature's values across the set of feature vectors contained in a `RankList` element. I did not realise this while removing `RankList` and blindly converted it to work on individual FeatureVectors. 

I will re-implement it as a utility function that normalises the complete `vector<FeatureVector>` returned by `create_vec_fvec` (or by `create_feature_vectors` method, after renaming).

Thanks for pointing this out.

ayshtmr:
Yes, I pulled these changes out from a wider set of changes. Surprising how this build fine for me although I haven't overrided CXXFLAGS.

I will remove `Scorer * scorer` from `Xapian::Ranker`, since Scorers will get a class of their own anyway. They aren't used in ranking or training anywhere.

ayshtmr:
@jaylett All the letor classes are written under `namespace Xapian` at present. Should I move them all under `namespace Xapian::Letor` in the next PR?

jaylett:
At any convenient point, yes please :-)
On 3 Aug 2016 9:01 p.m., Ayush Tomar <notifications@github.com> wrote:@jaylett All the letor classes are written under namespace Xapian at present. Should I move them all under namespace Xapian::Letor in the next PR?

ayshtmr:
@jaylett I couldn't fully understand when you said "remove `xapian-letor/.nobootstrap`, remove `xapian-letor` from the `bootstrap` defaults, and put it explicitly into `.travis.yml` so letor PRs get built (and, in future, tested) automatically."

Shouldn't just removing `xapian-letor/.nobootstrap` get travis to build letor automatically? Please correct me why those extra steps are needed.

jaylett:
Because we don't want the default for all users to be building ltr since it has extra dependencies -- libsvm.
J
On 3 Aug 2016 9:40 p.m., Ayush Tomar <notifications@github.com> wrote:@jaylett I couldn't fully understand when you said "remove xapian-letor/.nobootstrap, remove xapian-letor from the bootstrap defaults, and put it explicitly into .travis.yml so letor PRs get built (and, in future, tested) automatically."

Shouldn't just removing xapian-letor/.nobootstrap get travis to build letor automatically? Please correct me why those extra steps are needed.

xapian-letor/api/ranker.cc
ayshtmr: I'll be using this later as a comparator in ranker implementations, to sort `vector<FeatureVector`> by scores.

jaylett: That's still not a reason to make them into functors. They could just be non-static functions, couldn't they? The only reason to make them into functors is to allow other versions to be swapped in, and I'm not sure why they might be.

ayshtmr: As I understand it, `std::sort` does not allow non-static comparison function. But yes, functors weren't quite needed here.

Pushed them as `static` functions in `ranker.h`. Please review if its done correctly this time or not.

jaylett:
You should add `libsvm-dev` to the packages installed in `.travis.yml`.

jaylett:
👍 

Probably best to squash this down to a more reasonable number of commits. I think it's ready otherwise?

ayshtmr:
Yes, I think its ready. Squashing it down to 4-5 commits.

ayshtmr:
I think it looks reasonable now.

jaylett:
@ayshtmr great, thanks. I'll try to have a final look over this & merge as soon as possible :-)

jaylett:
@ayshtmr I've rebased this onto latest master (here: https://github.com/jaylett/xapian/tree/letor-and-build-updates), but we're now using GCC 4.8 as a minimum, and it fails on Travis (not in letor). I don't want to merge until I can get a full GCC build to work somewhere. Either Olly will be able to figure out quickly what's going on, or one of us will be able to demonstrate that it's a Travis-specific issue. This will hold merging up for a little, but hopefully not for too long.

(It's probably worth rebasing subsequent work on top of my branch, because I moved some commits around and obviously there's the more recent work in there as well.)

jaylett:
Merged in be88933.

PR #117, opened 2016-08-03
Maximise query term coverage for snippets

rsto:
This pull request adds a new snippet relevance mechanism to MSet's snippet method. It attempts to find the snippet that covers as many query terms as possible. The commit message contains an example that compares the current snippet generator with the proposed new feature.

The existing snippet generator remains default and the new feature is only activated by setting the newly introduced SNIPPET_TERMCOVER flag.

The added unit tests test for both regression and subtleties introduced by the new feature.

### DIFF
--- a/xapian-core/common/debuglog.h
+++ b/xapian-core/common/debuglog.h
@@ -81,6 +81,9 @@ enum debuglog_categories {
     /// Query stuff.
     DEBUGLOG_CATEGORY_QUERY = ('Y' - '@'),
 
+    /// Termgenerator.
+    DEBUGLOG_CATEGORY_TERMGEN = ('T' - '@'),
+
     /// Messages which are always logged.
     DEBUGLOG_CATEGORY_ALWAYS = 31
 };
--- a/xapian-core/include/xapian/mset.h
+++ b/xapian-core/include/xapian/mset.h
@@ -97,7 +97,18 @@ class XAPIAN_VISIBILITY_DEFAULT MSet {
 
     enum {
 	SNIPPET_BACKGROUND_MODEL = 1,
-	SNIPPET_EXHAUSTIVE = 2
+	SNIPPET_EXHAUSTIVE = 2,
+      /** Attempt to cover all search terms.
+       *
+       * If enabled, snippet() attempts to find a snippet that covers most of
+       * the search terms. The weight of each snippet term contributes only once
+       * to a snippet's cumulated relevance.
+       *
+       * This flag is only useful in search for "loose terms" as occuring in
+       * ORed and ANDed queries. It is ignored for queries containing phrases
+       * or wildcards.
+       */
+	SNIPPET_TERMCOVER = 4
     };
 
     /** Generate a snippet.
--- a/xapian-core/queryparser/termgenerator_internal.cc
+++ b/xapian-core/queryparser/termgenerator_internal.cc
@@ -40,8 +40,10 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <sstream> // for debugging
 
 #include "cjk-tokenizer.h"
+#include "debuglog.h"
 
 using namespace std;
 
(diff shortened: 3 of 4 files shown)


ojwb:
@rsto Are you OK with the patch licensing requirements?

https://trac.xapian.org/browser/trunk/xapian-core/HACKING#L1364

rsto:
yes, all OK

ojwb:
I've done some profiling of this, using a test chert database of an index of `/usr/share/doc` on my dev box created with `omindex` just under a year ago, and set to store a large amount of text (I don't recall exactly how large).  I profiled using valgrind's cachegrind tool, since snippet generation is clearly a CPU-bound operation, and cachegrind gives an exactly repeatable number, which means no worrying about whether a difference is just noise (though very small differences might be due to cache modelling errors).  I used kcachegrind to get a cycle estimate.

The time overhead when `SNIPPET_TERMCOVER` is **NOT** in use seems to be on the order of 2%, which doesn't thrill me - if I had to guess I'd suspect the virtual method calls, as there will be a lot of them.  When it is in use, it's something like 50% slower, which I guess is OK, but I'd really like to be able to make the smarter behaviour the default - the only reason it isn't is I failed to see a viable approach to achieve that originally.

However, it occurred to me that if we store a pointer to the relevance, and halve it after we add a term to the end of pipe, then double it before we remove a term from the front of the pipe, we get a cheap implementation of exponential decay for repeated occurrences in the same snippet (in exchange for some rounding-off when a term occurs many times in a row).  Exponential decay is not the same as "term cover", but perhaps is actually better - a second occurrence of the term in the snippet probably does add something, just not as much as the first occurrence (especially if you are generating large snippets).

So I first did a bit of profiling of the existing code, and made a few tweaks to get a ~45% combined reduction in run time for my testcase.  I've pushed those changes to the repo.

Then I implemented my idea on top of that.  With `GNU GPL` as a test query, 100/100 results include both words in a 200 byte snippet vs 67/100 before, and the runtime overhead (vs the version in the repo) is under 1% - that seems too good to be true, but even if there's something wrong in the current implementation, it suggests the approach has promise.

I need to stop for today and the patch needs much more thorough testing, plus the testsuite sorting out for it, but if you (or anyone else) wants to give it a go, I've uploaded it to:

https://oligarchy.co.uk/xapian/patches/snippet-exponential-relevance-decay.patch

rsto:
Thanks. Sorry for the delay, I've been traveling the last weeks.

I have applied the patch to Xapian's latest master. I also rebased the relevant parts of the Cyrus fork to make use of our regression and integration tests. The branch is at [xapian/cyrus_snippet_decay](https://github.com/rsto/xapian/tree/cyrus_snippet_decay).

 The results look promising! One difference is that using the default term weights broke a few deliberately pathological test cases. E.g. with our current termcover algorithm the following input

    rubbish rubbish example rubbish rubbish

and search terms ``rubbish examples`` produced the snippet

    example rubbish

but now it produces

    rubbish rubbish

If I am not mistaken, that's because the term "rubbish" occurs so often, that its termweight exceeds the one of "example" significantly. Once I switched to boolean weights our regression tests in Cyrus worked flawlessly with the exponential decay snippet generator.

That being said, I haven't done any load testing, nor used real-world documents. Probably @brong can run this in his development environment to check with the FastMail interface? A custom Cyrus branch is at [cyrus-imapd/snippet_decay](https://github.com/rsto/cyrus-imapd/tree/snippet_decay).

ojwb:
Thanks for taking a look.

The weights used to pick the snippet are from the weighting formula used for ranking the search results, which are based on frequencies in the db - in this tests that's indexed from `tests/testdata/apitest_simpledata.txt` (and the glass DB is cached as `.glass/db=apitest_simpledata`).  It depends on the weighting formula in use, but typically really frequent terms will actually get lower weights, as such terms don't have much useful discriminating power.

In this case, `Zexampl` isn't actually in that database - while it is good to check that case is handled sanely, it's not good to build the majority of our snippet tests around such an unusual case.

Looking at this in more detail, the relevance used for each term is `termweight(term) + max_tw`, where `max_tw` which is the largest `termweight` for any query term, so I think in general you should always get a `term2` preferred to a second occurrence of `term1`, except in this unusual case where `termweight(term2) == 0` (which happens when the term doesn't exist, and can also happen for extreme cases with some weighting formulae, for example a term which indexes all or nearly all documents might get zero weighting).

Perhaps we ought to do `max_tw *= 1.01` or similar to avoid this - if I make that tweak then I get `...<b>example</b> <b>rubbish</b>...` for that case without needing to use `BoolWeight`.  However, perhaps a term which gets no weight really is less interesting than a second occurrence of the highest weighted term.

For now I've applied my patch and merged that testcase to master and pushed them.  I'll go through and adjust most of the snippet tests to use terms that are actually in the database shortly.

rsto:
Thanks for merging the relevant parts of this PR. I've just opened https://github.com/xapian/xapian/pull/130 with an additional change that we use for the Cyrus mail server.

ojwb:
Hang on, need to deal with the zero weight case still...

ojwb:
I think we want to always prefer a fresh term to a second occurrence of a term already in the snippet - if for no other reason than it seems unnatural to have different behaviour when the fresh term has termweight 0 versus a very small but positive value.  So I've addressed this in 964f5d5bcdf9090849b2fbeeb5fccec6db1a34a9 by scaling up `max_tw` a little.  I went for `1+1/64` rather than `1.01` as that probably plays better with the exponential decay stuff.

PR #119, opened 2016-08-07
K-Means clusterer

richhiey1996:
This is an implementation of the K-Means clustering algorithm. The API has changed significantly because the previous API wasn't able to satisfy certain operations that were required for K-Means. I'm currently working on optimizing it in terms of time and space complexity. I'm also working on evaluation metrics so that we can provide the user with ways to check the quality of the clusters produced by the clusterer.

### DIFF
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -26,10 +26,14 @@
 
 #include <xapian/error.h>
 #include <api/termlist.h>
-#include <debuglog.h>
+#include "debuglog.h"
+#include "omassert.h"
 
 #include <vector>
 #include <map>
+#include <set>
+#include <cmath>
+#include <cstdlib>
 
 using namespace Xapian;
 using namespace std;
@@ -110,7 +114,7 @@ TermListGroup::add_document(const Document &document) {
 }
 
 void
-TermListGroup::add_documents(DocumentSource &docs) {
+TermListGroup::add_documents(MSetDocumentSource docs) {
     LOGCALL_VOID(API, "TermListGroup::add_documents()", docs);
     while (!docs.at_end()) {
 	Document document = docs.next_document();
@@ -132,67 +136,48 @@ TermListGroup::get_termfreq(const string &tname) {
 }
 
 doccount
-ClusterSet::num_of_clusters() {
-    LOGCALL(API, doccount, "ClusterSet::num_of_clusters()", "");
+ClusterSet::size() {
+    LOGCALL(API, doccount, "ClusterSet::size()", "");
     return clusters.size();
 }
 
 doccount
 ClusterSet::cluster_size(clusterid cid) {
     LOGCALL(API, doccount, "ClusterSet::cluster_size()", cid);
-    map<clusterid, vector<Document> >::iterator i;
-    i = clusters.find(cid);
-    if (i == clusters.end())
+    unsigned int size = clusters.size();
+    if (cid >= size)
 	throw RangeError("The mentioned clusterid was out of range", 103);
-    return ((i->second).size());
+    return clusters[cid].size();
 }
 
-DocumentSet
+Cluster
 ClusterSet::get_cluster(clusterid cid) {
-    LOGCALL(API, DocumentSet, "ClusterSet::get_cluster()", cid);
-    map<clusterid, vector<Document> >::iterator i;
-    i = clusters.find(cid);
-    if (i == clusters.end())
+    LOGCALL(API, Cluster, "ClusterSet::get_cluster()", cid);
+    unsigned int size = clusters.size();
+    if (cid >= size)
 	throw RangeError("The mentioned clusterid was out of range", 103);
-    vector<Document> docs;
-    docs = i->second;
-    DocumentSet docset;
-    for (vector<Document>::iterator k = docs.begin(); k != docs.end(); k++)
-	docset.add_document(*k);
-    return docset;
+    return clusters[cid];
 }
 
 void
-ClusterSet::add_document(clusterid cid, Document &doc) {
-    LOGCALL_VOID(API, "ClusterSet::add_document()", cid | doc);
-    clusters[cid].push_back(doc);
+ClusterSet::add_cluster(Cluster c) {
+    LOGCALL_VOID(API, "ClusterSet::add_cluster()", c);
+    clusters.push_back(c);
 }
 
-DocumentSet
+Cluster
 ClusterSetIterator::get_cluster() {
-    LOGCALL(API, DocumentSet, "ClusterSetIterator::get_cluster()", "");
+    LOGCALL(API, DocumentSet, "ClusterSetIterator::get_cluster()", index);
     return cset.get_cluster(index);
 }
 
-ClusterSetIterator
-ClusterSet::begin() {
-    LOGCALL(API, ClusterSetIterator, "ClusterSet::begin()", "");
-    return ClusterSetIterator(*this, 0);
-}
-
-ClusterSetIterator
-ClusterSet::end() {
-    LOGCALL(API, ClusterSetIterator, "ClusterSet::end()", "");
-    return ClusterSetIterator(*this, num_of_clusters());
-}
-
-ClusterSetIterator
+Cluster
 ClusterSet::operator[](doccount i) {
-    return ClusterSetIterator(*this, i);
+    return clusters[i];
 }
 
 void
-DocumentSet::add_document(Document &doc) {
+DocumentSet::add_document(Document doc) {
     LOGCALL_VOID(API, "DocumentSet::add_document()", doc);
     docs.push_back(doc);
 }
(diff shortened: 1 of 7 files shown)


jaylett:
@richhiey1996 that first commit already exists to some extent in the tree, and you seem to be based on distant history rather than the clustering branch. Can you rebase onto that clustering branch, and figure out whether your first commit is actually the same as the last one on the branch?

xapian-core/cluster/euclidian_sim.cc
```diff
@@ -40,86 +40,10 @@ EuclidianDistance::get_description() {
     return "Euclidian Distance metric";
 }
 
+/** TODO
+ *  Implement the euclidian distance metric for the current API
+ */
```
jaylett: Rather than do this, maybe just remove the euclidian distance metric completely? This also feels like this commit (ec28b37) is actually a fixup to the previous one, and should be noted as such to ensure it gets collapsed before merging.

xapian-core/include/xapian/cluster.h
```diff
@@ -180,64 +182,153 @@ class XAPIAN_VISIBILITY_DEFAULT TermListGroup : public FreqSource {
...
+    double similarity(Point a, Point b);
+
+    std::string get_description();
+};
+
+// Class for calculating the cosine distance between two documents
+class XAPIAN_VISIBILITY_DEFAULT CosineDistance : public Similarity {
+    TermListGroup tlg;
+  public:
+    CosineDistance(TermListGroup tlg_) : tlg(tlg_) {}
```
jaylett: `tlg_` doesn't seem to be used in `similarity()`. Is it needed at all?

xapian-core/include/xapian/cluster.h
```diff
@@ -316,10 +407,30 @@ class XAPIAN_VISIBILITY_DEFAULT RoundRobin {
...
 
     // Returns the description of the clusterer being used
     std::string get_description();
 };
-}
+
+/** Kmeans clusterer:
+ *  This clusterer implements the K-Means clustering algorithm
+ */
+class XAPIAN_VISIBILITY_DEFAULT KMeans {
```
jaylett: I might have preferred to have the API refactor in one commit, and the KMeans implementation in another — it would have been easier to see what was going on with the refactor by only looking at the round robin version.

xapian-core/include/xapian/cluster.h
```diff
@@ -316,10 +407,30 @@ class XAPIAN_VISIBILITY_DEFAULT RoundRobin {
...
+    std::vector<Point> docs;
+    //std::map<int, docid> doc_list;
+    std::vector<Point> centroids;
+  public:
+    void initialize_centroids(const unsigned int k, ClusterSet &cset);
```
jaylett: Public methods should have docstrings.

xapian-core/include/xapian/cluster.h
```diff
@@ -180,64 +182,153 @@ class XAPIAN_VISIBILITY_DEFAULT TermListGroup : public FreqSource {
...
 
     Xapian::doccount get_termfreq(const std::string &tname);
 
     Xapian::doccount get_doccount();
 };
 
+/** Structure to store term and corresponding wdf.
+ *  This is used with PointTermIterator to return wdf
+ */
+struct XAPIAN_VISIBILITY_DEFAULT Wdf {
```
jaylett: Is this used by clients of Xapian, or just within Xapian? If the latter, it shouldn't be in a public header file. (In which case you might be able to use a `std::pair<>` instead.) I'm not entirely sure that the `Point` stuff needs to be exposed, either.

xapian-core/include/xapian/cluster.h
```diff
@@ -316,10 +407,30 @@ class XAPIAN_VISIBILITY_DEFAULT RoundRobin {
...
+    std::vector<Point> docs;
+    //std::map<int, docid> doc_list;
+    std::vector<Point> centroids;
+  public:
+    void initialize_centroids(const unsigned int k, ClusterSet &cset);
+
+    void initialize_points(MSetDocumentSource docs, TermListGroup tlg);
+
+    // Implements KMeans clustering
+    ClusterSet cluster(MSet mset, unsigned int k);
```
jaylett: Does `k` here have the match `k` in `initialize_centroids()`? I'd have thought this should 

Also, you really want `KMeans` and `RoundRobin` to be subclasses of the same (abstract) base, so that they can be genuinely interchangeable.

xapian-core/cluster/cluster.cc
```diff
@@ -110,7 +114,7 @@ TermListGroup::add_document(const Document &document) {
 }
 
 void
-TermListGroup::add_documents(DocumentSource &docs) {
+TermListGroup::add_documents(MSetDocumentSource docs) {
```
jaylett: The point of using `DocumentSource` was to allow using this without driving it from an MSet. If you think that's too much abstraction for the moment, then get rid of the abstraction completely by passing an MSet directly. Otherwise, you want to keep the abstract base class as the way you're passing things around.

xapian-core/cluster/euclidian_sim.cc
richhiey1996: I did plan to implement the euclidian distance in the upcoming commits, so just kept this as a dummy for now. I think having two distance measures can help us in some ways.

xapian-core/include/xapian/cluster.h
richhiey1996: Thanks for pointing this out! I forgot to change the constructor when I changed the way cosine  similarity function was implemented.

xapian-core/include/xapian/cluster.h
richhiey1996: Sorry about that. Will break the commits into small pieces henceforth.

xapian-core/cluster/euclidian_sim.cc
jaylett: I agree about having multiple distance measures, but what I mean is that if it isn't working, the entire class shouldn't be present. Each commit should represent a good state for the code, wherever possible. This isn't, because it contains options that people can compile against but don't make any sense — removing the `EuclidianDistance` class (for the time being) should solve this.

The second part of my comment is still relevant.

xapian-core/include/xapian/cluster.h
jaylett: It's a difficult thing to get right, and there's no hard and fast rules. It's not about small pieces so much as self-contained pieces. The refactor goes from one good state to another (with a different API), so it's reasonable to have a commit that contains only that. Hope that's helpful, but let me know if I've just confusing things :)

xapian-core/include/xapian/cluster.h
richhiey1996: > Does k here have the match k in initialize_centroids()? I'd have thought this should 
I'm sorry I didn't get you here.
And I initially thought that every Clusterer implemented can have completely different properties so the only thing common between them would be a 'cluster' function. I'll add in that change.

xapian-core/cluster/cluster.cc
richhiey1996: Sure. As you mentioned, I just changed that to get rid of the abstraction a while.

xapian-core/include/xapian/cluster.h
jaylett: If it's a _property_ of a `Clusterer` then I suggest you pass it in the constructor, not to the `cluster()` method. There are clustering algorithms that don't need to know in advance how many clusters they're generating, so that's probably a better approach.

richhiey1996:
I've again modified the API to add a Centroid class which inherits from an abstract class named PointType. This helped me solve the problem of differentiating between clusters and points.
I've run this implementation on a set of BBC news articles (around 1000 of them), and the average ADVDC values returned by the clusters are around 0.6 - 0.75 (Cosine distance) in most cases, which I'm guessing is normal. (Please correct me here if I'm wrong). The clusters also returned best results with respect to ADVDC when k=6, and since there were news articles in 6 domains, it looks like it worked fine.
One of the main problems I'm facing currently is that, this implementation works fine till n = 300 ( n is number of documents) but beyond that, the amount of time taken increases and becomes undesirable.
So I'm looking to cut down on unnecessary computations in various parts. Would be great if I could get ideas on improving the code to be more efficient in time.
Apart from optimization, I'm looking into quickly implementing a kmeans++ initialization before moving onto PSO.
Once I'm done with optimizing a little more, I'll run these with FIRE datasets and post my findings on the mailing list.

jaylett:
If you're looking at optimisation you should just look at that. Use a profiler to find where time is being spent (do runs with increasing numbers of documents, for instance).
On 12 Aug 2016 10:06 p.m., richhiey1996 <notifications@github.com> wrote:I've again modified the API to add a Centroid class which inherits from an abstract class named PointType. This helped me solve the problem of differentiating between clusters and points.
I've run this implementation on a set of BBC news articles (around 1000 of them), and the average ADVDC values returned by the clusters are around 0.6 - 0.75 (Cosine distance) in most cases, which I'm guessing is normal. (Please correct me here if I'm wrong). The clusters also returned best results with respect to ADVDC when k=6, and since there were news articles in 6 domains, it looks like it worked fine.
One of the main problems I'm facing currently is that, this implementation works fine till n = 300 ( n is number of documents) but beyond that, the amount of time taken increases and becomes undesirable.
So I'm looking to cut down on unnecessary computations in various parts. Would be great if I could get ideas on improving the code to be more efficient in time.
Apart from optimization, I'm looking into quickly implementing a kmeans++ initialization before moving onto PSO.
Once I'm done with optimizing a little more, I'll run these with FIRE datasets and post my findings on the mailing list.

richhiey1996:
I've optimized the implementation a little. It seems to work much better than before now, although a there will is more work required to make it faster.
I've also implemented a KMeans++ initialization which works well sometimes, and not so well in a few situations where cluster sizes go out of proportion.
I've run this KMeans on various datasets so I'll be posting the results on the mailing list in a few hours, so everyone can give in suggestions for the same.

xapian-core/include/xapian/types.h
```diff
@@ -30,6 +30,13 @@
 
 namespace Xapian {
 
+/**A unique identifier for a cluster of documents
+ *
+ * clusterid of 0 is invalid, providing an "out of range" value which can
+   be used to mean "not a valid cluster"
+ */
+typedef unsigned XAPIAN_DOCID_BASE_TYPE clusterid;
```
jaylett: `XAPIAN_DOCID_BASE_TYPE` isn't appropriate for this. Do you need a signed type? It's probably best to use `int` or `unsigned int`.

xapian-core/include/xapian/cluster.h
```diff
@@ -0,0 +1,489 @@
...
+     *  by checking for change in centroid of the clusters
+     */
+    bool converge(std::vector<Centroid> &previous, std::vector<Centroid> &current);
+
+    /** Initialization of centroids using a certain specified method
+     *  Current methods that are supported :
+     *  random - Random Initialization
+     *  kmeanspp - KMeans++ Initialization
+     */
+    void initialize_centroids(ClusterSet &cset);
```
jaylett: Are these four methods really public? They feel internal to the implementation of `cluster()`.

xapian-core/include/xapian/types.h
richhiey1996: Yes you're right. I had added that with a different idea in mind when I started off. Will change that right away.

jaylett:
A reminder that this is unmergable as it stands because it seems to include the last commit that was already merged into the clustering branch. It should be rebased cleanly on top of that existing branch.

Also, you should work on pruning the public API down to the bare minimum (hide as much of the detail of `Point` and `Centroid` and so forth away from the end user). You may want to use PIMPL to achieve this. (Xapian uses PIMPL fairly extensively.)

Finally, as noted previously, you should include a superclass for clusterers so that client code can construct one of whatever type and pass it around. (This because important when there are multiple 'real' clusterers available.)

xapian-core/cluster/kmeans.cc
```diff
@@ -61,22 +70,85 @@ KMeans::initialize_centroids(ClusterSet &cset) {
...
+}
+
+void
+KMeans::initialize_points(MSetDocumentSource source, TermListGroup &tlg) {
+    LOGCALL_VOID(API, "KMeans::initialize_points()", source | tlg);
+    while (!source.at_end()) {
 	Document temp = source.next_document();
 	Point p;
 	p.initialize(tlg, temp);
+	TermIterator it = p.termlist_begin();
```
jaylett: What does this line do?

xapian-core/cluster/kmeans.cc
```diff
@@ -61,22 +70,85 @@ KMeans::initialize_centroids(ClusterSet &cset) {
 }
 
 void
-KMeans::initialize_points(MSetDocumentSource source, TermListGroup tlg) {
-    while(!source.at_end()) {
+KMeans::initialize_kmeanspp(ClusterSet &cset) {
```
jaylett: Would it perhaps be better to have a separate function or class to create the initial set of clusters using KMeans++, and a way of initialising `KMeans` with an existing `ClusterSet`? That feels like it's going to flow better into having a PSO implementation, and also allow experimentation with other ways of initialising the clusters in future.

jaylett:
The commit "optimisations for KMeans clusterer" is doing a lot more than that, including introducing KMeans++. Please break it apart into commits that each do one thing.

xapian-core/cluster/kmeans.cc
jaylett: Another pre-clustering option we might want to explore (in the future, not within GSoC) is [canopy clustering](https://en.wikipedia.org/wiki/Canopy_clustering_algorithm), and in deciding on the API for using KMeans++ it may be worth thinking about what happens when you have KMeans++, PSO and Canopy as different pre-clusterers.

richhiey1996:
Closing this PR and opened a new one.

PR #120, opened 2016-08-11
letor: API changes

ayshtmr:
This PR proposes final API changes to `xapian-letor`

1. `FeatureVector` changes:
    * Change type of `fvals` from `map<int,double>` to `vector<double>`
    * Enable variable `fcount`
    * Clean-up of unnecessary methods and remove `featurevector_internal.cc`
2. Add `Feature` class and subclasses:
    * `Feature` is an abstract base class. Each subclass calculates a single feature.
    * `Feature::Internal` class does the work of supplying `Feature` subclasses with document statistics (`tf`, `coll_tf` etc). 
    * `Feature` subclass implementations and `Feature::Internal` class defined in `xapian-letor/feature`
3. Add `FeatureList` class:
    * `FeatureList` does the work of creating `FeatureVectors`
    * It contains `vector<Feature*>` - calls each Feature object to return `fval` values
4. Update `Letor` class methods according to changes mentioned above
5. Add filename parameters in `Letor` and `Ranker` methods
    * Training file and model to be saved as user supplied filenames
6. Eliminate `FeatureManager`:
    * `FeatureManager::create_feature_vectors(*)` is replaced by `FeatureList::create_feature_vectors(*)`
    * `FeatureManager's` role of supplying certain document stats is replaced by `Feature::Internal` class methods
7. Eliminate `Features` (`letor_features.h` & `letor_features.cc`):
    * `Feature` subclasses do the work of calculating feature values
8. Documentation updates


xapian-letor/api/featurelist.cc
```diff
@@ -0,0 +1,144 @@
...
+
+        for (std::vector<Feature*>::iterator it = feature.begin() ; it != feature.end(); ++it) {
+
+            (*it)->set_database(letor_db);
+            (*it)->set_query(letor_query);
+            (*it)->set_doc(doc);
+
+            fVals.push_back((*it)->get_fval('t')); // for title
+            fVals.push_back((*it)->get_fval('b')); // for body
+            fVals.push_back((*it)->get_fval('w')); // for whole
```
jaylett: I wonder if a `Feature` should act as a generator that can produce all three of these, rather than having to be explicitly called. Otherwise we need changes here to support a feature that doesn't vary by field (perhaps something based on other document metadata rather than terms).

I also think for clarity (and general C++-ness) we should use constants rather than `'t'`, `'b'` and `'w'`, even if we're switching on things internally to the feature classes. (`TITLE_FIELD`, `BODY_FIELD`, `WHOLE_DOCUMENT` or similar, perhaps?)

xapian-letor/api/letor_internal.cc
```diff
@@ -123,16 +113,18 @@ Letor::Internal::load_list_fvecs(const char *filename) { //directly use train.tx
...
 
         qid = qid.substr(qid.find(":")+1,qid.length());
+        train_file.ignore();
 
-        std::map<int,double> fvals;//read features
-        for(int i = 1; i < 20; ++i){
-            train_file.ignore();
+        std::vector<double> fvals;
+
+        while (train_file.peek() != 35) { // read till '#docid' is found (35 is ascii for '#')
```
jaylett: Pretty sure `'#'` (`char`) will upcast to `int` happily for this comparson. The comment is still needed to point out we're looking for `#docid`, though.

xapian-letor/api/letor_internal.cc
```diff
@@ -123,16 +113,18 @@ Letor::Internal::load_list_fvecs(const char *filename) { //directly use train.tx
...
             int feature_index;
             double feature_value;
             train_file >> feature_index;
             train_file.ignore();
             train_file >> feature_value;
-            fvals[feature_index] = feature_value;
+            fvals.push_back(feature_value);
+            train_file.ignore();
```
jaylett: I think this parsing needs a little more commenting. You're reading a double, skipping a character without validation (space?), reading an int, skipping another character…?

xapian-letor/api/featurelist.cc
```diff
@@ -0,0 +1,144 @@
...
+
+        fvec.push_back(fv);
+
+    }
+    normalise(fvec);
+    return fvec;
+}
+
+Xapian::FeatureVector
+FeatureList::create_feature_vector(std::vector<double> & fvals, Xapian::docid & did)
```
jaylett: This seems to be a constructor for `FeatureVector` that's implemented as a method of another class. Any reason not to move it across?

xapian-letor/api/featurelist.cc
ayshtmr: Probably what we can do here is return a `vector<double> fval` from each `Feature` instead of a single value, and append it to `fVals`. This vector would contain all three `t,b,w` feature values.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -0,0 +1,71 @@
...
+CollTfCollLenFeature::get_fval(char ch) {
+
+    Query query = Feature::internal->feature_query;
+    map<string, long int> coll_tf = Feature::internal->collection_termfreq();
+    map<string, long int> coll_len = Feature::internal->collection_length();
+
+    double value = 0;
+
+    if (ch == 't') {
+    for (Xapian::TermIterator qt = query.get_terms_begin();
```
jaylett: Indentation is weird throughout this bit, making it hard to read (at least on github).

xapian-letor/api/letor_internal.cc
ayshtmr: I tried with '#' instead of 35 but suprsingly it didn't work. Will check once again, maybe I was doing something wrong at that time.

xapian-letor/api/letor_internal.cc
ayshtmr: I'll add required comments to it.

jaylett:
This is looking great :-)

Something about `Feature::Internal` is niggling at me, but I can't figure out what. It may simply be that those helper functions actually want to be somewhere else (so, for instance, we don't need to recalculate query-dependent but doc-independent values for every different feature but can cache that on…`FeatureList`, maybe?). Not a worry right now, though.

Probably disable G++ builds in travis for the time being (just remove that one line from `env.matrix`). I need to figure out how to get travis to build with gcc 4.8, or how to get clang to work with its own stdlib instead of gcc 4.9. (I was unable to look further at this today, because gnu.org was offline for most of it!)

xapian-letor/api/letor_internal.cc
jaylett: 👍

xapian-letor/api/featurelist.cc
ayshtmr: You're right. It makes more sense as a constructor. Will move it across.

xapian-letor/api/letor_internal.cc
jaylett: Huh. I ran a quick check here in what I thought was an equivalent case, and it worked okay. Weird.

xapian-letor/api/featurelist.cc
jaylett: Yep, and call it `get_fvals()` (or just `get_values()`?).

jaylett:
Hmm, as we're beginning to get somewhere with this…I can't find the implementation of `SVMRanker`. Which is weird. Am I being stupid?

xapian-letor/feature/colltfcolllenfeature.cc
ayshtmr: Will fix these up

ayshtmr:
Rankers have not been integrated yet. I have them lined up once we close this one.

jaylett:
@ayshtmr ah, cool. It's just weird having the svmranker C++ file be basically the same as the header file :-)

ayshtmr:
We could cache doc-independent values on `FeatureList`, but that would in some way "pollute" the class with Feature specific metrics. I put them all on `Feature::Internal` so that we could preserve the generality of `FeatureList`.

ayshtmr:
Yeah, I should've removed svmranker files in the clean up PR earlier.

jaylett:
I'm not suggesting we move anything now from `Feature::Internal`. In some ways it's similar to how `Weight` works, but that's always felt a little odd to me too :-)

xapian-letor/api/letor_internal.cc
ayshtmr: Its probably verbose to excess now

jaylett: :-)
On 12 Aug 2016 9:00 p.m., Ayush Tomar <notifications@github.com> wrote:In xapian-letor/api/letor_internal.cc:
>              int feature_index;
>              double feature_value;
>              train_file >> feature_index;
>              train_file.ignore();
>              train_file >> feature_value;
> -            fvals[feature_index] = feature_value;
> +            fvals.push_back(feature_value);
> +            train_file.ignore();

Its probably verbose to excess now

jaylett:
Are there tests we can write for anything here? Normalisation, maybe some of the `Feature` classes, that kind of thing?

Can you autosquash these commits? They look ready otherwise :-)

ayshtmr:
I planned all the tests as a separate PR after ListNet (which is next) since some of the methods in `Letor` class require `Ranker` instances, which are yet to be integrated. Is it fine to do so?

jaylett:
Yes, that sounds fine.

ayshtmr:
@jaylett If you think its ready, could you please merge this?

jaylett:
Can you run it through `xapian-maintainer-tools/xapian-check-patch`? There are a lot of space warnings, and some about Doxygen commands.

ayshtmr:
Should a push individual fixup commits or a single commit to fix all of them together?

jaylett:
Do a single fixup commit for it.

jaylett:
Sorry, a single commit that fixes it up.

jaylett:
I'm just manually running the tests on G++ 4.9 since we can't get travis to work at the moment. Should be ready to merge once you're done with the fixes :)

jaylett:
Merged in 9637c0e.

PR #121, opened 2016-08-16
BM25+ fixes

ojwb:
There are essentially three changes here, but I've bundled two of them together in a commit since they involve recomputing the hand-computed weights so the tests pass, and it seemed pointless extra work to do that twice.

- `BM25PlusWeight` now inherits from `Weight`
- `(get_collection_size() + 1) / tf` performs an integer division as `get_collection_size()` and `tf` are both integer types - we need to ensure at least one argument of the division is a `double` to get a floating point division
- Eliminate the fudge to avoid negative term weights, inherited from our BM25 implementation


ivmarkp:
@ojwb Thanks for this PR. Changes look good. I'll run an evaluation on it and update the evaluation results of BM25+ and also update xapian-evaluation module to reflect recent changes in `BM25PLusWeight`.

PR #122, opened 2016-08-16
Kmeans Clusterer

richhiey1996:
Due to git history problems in the previous PR, I'm closing that PR and starting a new one. I tried rebasing the old branch with upstream cluster, but for some reason it was getting all the previous commits. So I created a new branch and made it level with the merged cluster branch, and then cherry-picked the commits.


jaylett:
You should be able to get the tests passing for this; I'll review once this PR is green.

richhiey1996:
I had a later version of g++ and gcc, so it had hidden quite a lot of errors from me. I'm currently working on that.

jaylett:
@richhiey1996 what version? 6 is a fairly major change (certainly from 4.x), so that wouldn't surprise me. However what's worrying is if your compiler is missing things like the equality/assignment mistake, which is a pretty common compiler warning…

richhiey1996:
It was on 5.4. Luckily not that far. I'm currently getting an undefined reference to GeoEncode::encode(double, double, string&) and to Xapian::Internal::str()

`./.libs/libxapian.so: undefined reference to `GeoEncode::encode(double, double, std::string&)'
./.libs/libxapian.so: undefined reference to `Xapian::Internal::str(long)'
./.libs/libxapian.so: undefined reference to `Xapian::Internal::str(double)'
./.libs/libxapian.so: undefined reference to `errno_to_string(int, std::string&)'
./.libs/libxapian.so: undefined reference to `Xapian::Internal::str(unsigned int)'
./.libs/libxapian.so: undefined reference to `Xapian::Internal::str(long long)'
./.libs/libxapian.so: undefined reference to `Xapian::Internal::str(unsigned long long)'
./.libs/libxapian.so: undefined reference to `Xapian::Internal::str(int)'
`
Would you know what could be causing this?

richhiey1996:
I've moved to 4.7 and didn't any equality / assignment mistake as you mentioned. So I'm guessing that works fine. It was pretty easy of shadowing though.

jaylett:
The undefined references mean you aren't linking everything together in the library. At a guess, your `cluster/Makefile.mk` has errors in it, so `Makefile.am` (which includes it) isn't getting fully processed. Look at the generated `Makefile.in` to see if there's a problem in the definition of `lib_src`.

richhiey1996:
It seems to be some error in my local configuration. I'll check on that. I have fixed the shadow errors and another error that was coming up, and removed the "api_cluster.h" file. I'll be writing new tests now. For now, the build seems to have passed

richhiey1996:
Should I now concentrate on going ahead and tidying up and fixing mistakes in the API?

jaylett:
You should concentrate on getting this PR into a mergeable state. Things that immediately come to mind:

 * `xapian-check-patch` shows a number of whitespace errors; check over them but also anything else from a stylistic or indentation point of view
 * you've removed the tests you wrote for the previous work. So now none of your work is tested. You need to return the tests for `RoundRobin`, and introduce tests for `KMeans`.
 * ensure the public API is fully documented so people could start using it if it were merged
 * beyond this, all class members should have docstrings so that people working with the internals know what's going on and what the purpose of everything is
 * `RoundRobin` isn't a subclass of `Clusterer`. I'm not going to bother mentioning this again, but this needs to be fixed before this PR can be merged.
 * there are 13 commits in this PR, but not all of them should exist as separate commits. Each commit should make a single, complete change. Each commit should be well-described, with a good commit message. It's fine to have tiny commits that do just one thing. What you might want to do is to reset your branch back to the upstream cluster branch, and build new commits out of the changes you have. Refactor the API to prepare for KMeans in one (adapting RoundRobin, and updating its tests as you go). Introduce KMeans in another, with new tests. Perhaps move things out from the public API in a third, although you can probably do that either as part of the refactor or introducing KMeans.

If there are "mistakes in the API", then you need to fix them before rebuilding commits. I'm not sure what you mean by this, though. If you're talking about _improvements_ to the API, then those can wait.

Say when you're happy that things are in a good state for me (and/or others) to review. I can start doing this before you have docstrings in place, but you should look at the other things above, and have a high confidence that I'm not going to find any lurking problems. (Commits in a good structure are vastly easier to review than a number of commits that don't tell a clean story of the changes.)

richhiey1996:
Sure. I'll make sure the PR is in good shape after I've completed all this.

I'm a little confused about the 6th point that you mentioned. When you say make new commits out of the one's you have currently, do you mean I should rebase to upstream cluster and then cherry pick and squash the existing commits?

jaylett:
You can rebuild the branch with better-chosen commits any way you want, but I'd suggest (as I said):

 * resetting to the upstream cluster branch (ie throwing away the commits you have on this branch)
 * creating new commits carefully from the end point (ie you can choose what the right commits are once you have all the code written), using `git add -p`

Cherry picking and squashing won't get it done, because it's highly unlikely that you'll be able to simply combine existing commits to get the final set you want. While it's possible to use interactive rebase to split apart commits, it's often too much work compared to just resetting (which leaves your commits as your working directory) and building new commits on the upstream.

richhiey1996:
@jaylett I feel the API is in decent condition now. I've fixed the problems you mentioned above, added documentation and tests. I've also removed the PIMPL implementation of Cluster to ease things for a bit, so that once this is merged, I can concentrate on tidying up the API.

jaylett:
The commits are still at the wrong granularity in a number of places. If you have the word 'and' in a commit summary, you're probably doing too much.

For instance, "Tests for Clusterers and new Round Robin impl". Ideally, the KMeans tests would have been with the KMeans implementation, and moving RR and its tests to the new API would either be in its own commit, or in the API refactor commit.

You also make Cluster into a PIMPL class, then remove it again. Since across the PR this is no change, there shouldn't be commits for it.

I think you want to have:

 * refactor API to support more clusterers (probably include updating RoundRobin & its test for this); this is quite a small change
 * introduce KMeans, with its supporting classes (the `Point` classes aren't needed for `RoundRobin`) and tests
 * any distance metric changes that aren't part of the above
 * KMeans++ method for initialising the centroids, possibly including the API change to the `KMeans` class (although that might be more hassle than it's worth)

You may need a final fixes commit to tidy things up at the end.

richhiey1996:
Yes I should have added the KMeans tests with its implementation and should have added the RoundRobin with the API refactor itself. I just felt since these are independent changes, I'd add them differently.
I'm always unsure about breaking things down into useful commits. Sorry for the trouble here.

jaylett:
Two things that may help in thinking about commits:

1. The commits in a pull request tell a story. It should _not_ be the story of how you did the work; in six months time no one will care what order things were done in, or that you changed you mind about an API, or using `std::map<>` over `std::vector<>` partway through. It should be the story of the new features that the PR introduces. You should be able to read the commit messages and be able to put together in your head what they add up to.

2. A good rule of thumb is: each commit should build, pass tests, and be usable. It may not be particularly useful (eg the point where you had an API and the `RoundRobin` clusterer wasn't terribly useful to anyone, since RR clustering is unlikely to be used by anyone in real life); but it's possible to use.

It follows from this that if you change a public API, you should change the implementations (and the tests) in the same commit. Otherwise the implementation will stop being usable in that commit, until a subsequent one restores it.

Also, if you add a new implementation of an API (eg `KMeans`) then you should add tests for it in the same commit. (Otherwise, how do you know that this commit is usable? The tests, amongst other things, should demonstrate that a user sitting down and programming with the Xapian library can actually do what the API docs say they can.)

richhiey1996:
@jaylett Great! That's the reason I was making mistakes with it. I saw each commit as an independent change made so I thought I could add the tests in different commits, and I wasn't too bothered about whether the tests pass for every commit I pushed. It's clear now.
Also, I've made the changes the way you mentioned and added them. Let me know if these are fine or if there are any more changes which would be required. :)

xapian-core/cluster/cluster.cc
```diff
@@ -27,9 +27,15 @@
...
 #include <api/termlist.h>
 #include <debuglog.h>
+#include <omassert.h>
 
+#include "points.h"
+#include "similarity.h"
+
+#include <cmath>
+#include <string>
+#include <tr1/unordered_map>
```
ojwb: We require C++11, so that can (and should) just be `<unordered_map>`.

xapian-core/cluster/cluster.cc
```diff
@@ -132,92 +141,357 @@ TermListGroup::get_termfreq(const string &tname) {
...
 
 doccount
 ClusterSet::cluster_size(clusterid cid) {
     LOGCALL(API, doccount, "ClusterSet::cluster_size()", cid);
-    map<clusterid, vector<Document> >::iterator i;
-    i = clusters.find(cid);
-    if (i == clusters.end())
+    unsigned int s = clusters.size();
+    if (cid >= s)
 	throw RangeError("The mentioned clusterid was out of range", 103);
```
ojwb: Why are you passing `103` here?  If I read the overloads right, it'll be interpreted as an `errno` value, but what a literal `103` means depends on the platform (and for Linux, it even varies between architectures in some cases).  For me, `103` is `ECONNABORTED` - "Software caused connection abort", which seems to make no sense.

I don't think you want to be injecting an `errno` value for a `RangeError` anyway, but more generally if you do want to refer to an error value, use the `EXXX` constant for it.

xapian-core/cluster/cluster.cc
```diff
@@ -132,92 +141,357 @@ TermListGroup::get_termfreq(const string &tname) {
 }
 
 doccount
-ClusterSet::num_of_clusters() {
-    LOGCALL(API, doccount, "ClusterSet::num_of_clusters()", "");
+ClusterSet::size() {
```
ojwb: This method (and many others) should be `const` - they don't touch the internal state of the object, and so should be usable on a `const` object - e.g. `cluster_set.size()` should be callable in a function which takes a `const ClusterSet& cluster_set` parameter.

xapian-core/cluster/cluster.cc
```diff
@@ -54,41 +60,46 @@ MSetDocumentSource::MSetDocumentSource(const MSet & mset_, doccount maxitems_)
 
 Document
 MSetDocumentSource::next_document() {
```
ojwb: Xapian convention is that the opening `{` should go on a new line here.  Generally we "cuddle" braces in the interests of conserving vertical space so more code fits on the screen.  The exception here is mostly just because it ends up cleaner with long parameter lists to start a new line.  And the "fit more on the screen" argument is more important within a function (or a class, struct, etc).

xapian-core/cluster/cluster.cc
```diff
@@ -110,18 +121,16 @@ TermListGroup::add_document(const Document &document) {
 }
 
 void
-TermListGroup::add_documents(DocumentSource &docs) {
+TermListGroup::add_documents(MSetDocumentSource docs) {
```
ojwb: Passing by const reference would probably be better than passing by value.

xapian-core/include/xapian/cluster.h
```diff
@@ -24,302 +24,392 @@
...
 
-    // Returns the description of the clusterer being used
+    /// This method implements the RoundRobin clustering
+    ClusterSet cluster(MSet &mset);
+
+    /// This method returns the description of the clusterer
     std::string get_description();
 };
-}
+};
```
ojwb: You don't need a `;` after `namespace { /*...*/ }` (`class X { /*...*/ }` needs one for irritating historical reasons - it's modelled on `struct X { /* ... */ };` in C).  I have a feeling it causes (or at least caused) a warning with one of the less common compilers.  I'm not sure if that's still the case (requiring C++11 has ruled out a few older compilers) but it's probably better avoided.  Certainly don't go around adding them, as you do here!

xapian-core/tests/testdata/apitest_cluster.txt
```diff
@@ -3,3 +3,13 @@ A computer cluster consists of a set of loosely or tightly connected computers t
 The components of a cluster are usually connected to each other through fast local area networks ("LAN"), with each node (computer used as a server) running its own instance of an operating cluster system. In most circumstances, all of the nodes use the same hardware and the same operating system, cluster although in some setups (i.e. using Open Source Cluster Application Resources (OSCAR)), different operating systems can be used on each computer, clusterand/or different hardware.
 
 They are usually deployed to improve performance and availability over that of a single computer, while typically being much more cost-effective than single computers of comparable speed or availability.
+
```
ojwb: This is taken from wikipedia, right?  Something like the first paragraph of each entry here it looks like:

https://en.wikipedia.org/wiki/Cluster

That's not an inherent problem, but the Wikipedia licence requires both attribution and making it clear what the licence is - see "Attribution" and "Notice" here:

https://en.wikipedia.org/wiki/Wikipedia:Text_of_Creative_Commons_Attribution-ShareAlike_3.0_Unported_License

If you only excerpted a small amount of text, one could argue that's not required, but this seems like a sizeable amount of text to me.

xapian-core/cluster/cluster.cc
richhiey1996: Oh yes. I'll change that right away.

xapian-core/cluster/cluster.cc
richhiey1996: Yes that was a mistake on my part! When I wrote it, I didn't know where to find what different errorno values would be interpreted as. I hadn't added the errono value, but I got an error saying less arguments, so I just added one. Will change that.

xapian-core/cluster/cluster.cc
richhiey1996: Thanks for pointing this out.

xapian-core/include/xapian/cluster.h
richhiey1996: Oh yes, didn't know these could cause warnings. Will remove them.

xapian-core/tests/testdata/apitest_cluster.txt
richhiey1996: I had no clue about the license requirements. I'd rather write up a few lines which wouldn't require any license's and will help me write test data according to the clusterers being tested. Thanks for pointing this out.

xapian-core/cluster/cluster.cc
richhiey1996: The way I've thought of the convention while writing code was to put opening braces on the same line when its a function and on a different line when its a constructor. Have I misinterpreted things here?
Should I be necessarily adding this brace on a next line? Because that will require me to change other functions that I had written in a similar way.

xapian-core/cluster/cluster.cc
richhiey1996: Problem here is, that the parameter 'MSetDocumentSource docs' is modified inside add_documents( .. ), so I can't cast it as a const because that leads to an error. And I can't even pass this by reference because I do not want this object to change its state outside this function. That's why I pass it by value.
It won't make a very huge performance difference because these functions are more like a pre-processing thing.

jaylett: Do we need `MSetDocumentSource` at all? We don't have any other types of `DocumentSource` at the moment, so we could just pass an `MSet` or a pair of `MSetIterator`s in here.

richhiey1996: It would be beneficial to provide a documents to the clusterer in different ways through the document source. I can't think of any other way for now, except a vector of Documents. But it would be beneficial later on.
I could just make the parameter accept a DocumentSource rather, so we can pass in subclasses of DocumentSource later on.

jaylett: The key words here are "later on". And yes, you certainly want to pass the superclass if you want to support that. However for now I'd just pass an `MSet` and be done with it. We can add something that takes a suitable `Document` iterator in future.

richhiey1996: Okay great! I'll add that in right away. I guess that removes two classes from the public API, so it does help 👍

richhiey1996: @jaylett I've removed the DocumentSource classes and am passing MSet to go through the documents where required. Could you look into that?

xapian-core/cluster/cluster.cc
richhiey1996: @ojwb Yes I've made the functions which don't touch the internal state of the object 'const'. Would be great if you could have a look into it and let me know whether I've missed out on anything :)

xapian-core/cluster/cluster.cc
```diff
@@ -121,10 +85,11 @@ TermListGroup::add_document(const Document &document) {
 }
 
 void
-TermListGroup::add_documents(MSetDocumentSource docs) {
+TermListGroup::add_documents(const MSet &docs) {
     LOGCALL_VOID(API, "TermListGroup::add_documents()", docs);
-    while (!docs.at_end())
-	add_document(docs.next_document());
+    MSetIterator it = docs.begin();
```
jaylett: You can put this initialiser inside the `for`. This would be more natural C++. See for example the code in the user guide. (Unless you build it yourself, it's probably easiest to look at the source code to get the C++ examples.)

xapian-core/cluster/kmeans.cc
```diff
@@ -130,13 +130,13 @@ KMeans::initialize_kmeanspp(ClusterSet &cset) {
...
 
 void
-KMeans::initialize_points(MSetDocumentSource source, TermListGroup &tlg) {
+KMeans::initialize_points(const MSet &source, TermListGroup &tlg) {
     LOGCALL_VOID(API, "KMeans::initialize_points()", source | tlg);
-    while (!source.at_end()) {
-	Document temp = source.next_document();
+    MSetIterator it = source.begin();
+    for(; it != source.end(); it++) {
+	Document temp = it.get_document();
```
jaylett: You don't need a temporary variable here.

xapian-core/cluster/cluster.cc
richhiey1996: Yes will do that. I seem to initialize iterators in a similar way in other parts of the code though. Would I need to change that too? Or was this only in relation to MSet because its part of Xapian's public API?

xapian-core/cluster/kmeans.cc
richhiey1996: Yes I had removed them in most other parts. Didn't see this. Thanks for pointing out!

xapian-core/cluster/cluster.cc
jaylett: It's a C++ pattern, so you're better off using it everywhere.

richhiey1996:
@jaylett I've used the C++ pattern you mentioned in most places and just avoided it with unordered_map because code becomes really hard to read. Also fixed a few indentation mistakes that I could see to make the code more readable.

jaylett:
The main thing is that you've been adding changes without using fixup commits, so you'll have to go through again and rebase this to make sense out of things. (And more: your patterns and indentation commit has a functional change in `initialize_centroids()`, where you seem to have previously missed part of the introduction of KMeans++.) It's possible you'll be able to just reset the last few commits and recreate appropriate fixup (ie `git commit --fixup <previous-sha>`) commits out of them without having to touch the four commits you ended up with before this last round of review.

richhiey1996:
@jaylett That would mean I would have one fixup commit which combines the fixes for the refactored API that were added through 'small fixes for refactored API' and the latest commit, keeping the others commits in between the same, right? Since the other two commits are more like changes, rather than fixing up something.

jaylett:
@richhiey1996 I think we probably want to rewrite the previously-merged commit at this point. (Which introduced `DocumentSource`, which you're now removing; and also which introduced the test data you're now replacing. The latter is important because we need to avoid having the wiki data with unspecified license in master when we merge all this.)

That may mean that the first commit (refactor API) needs to be squash with the previous one already on the branch (along with removing `DocumentSource` and replacing the test data; although the first of those will need to be split since it includes changes to KMeans as well — which then can be squashed with the KMeans commit).

richhiey1996:
@jaylett would it be fine if I rewrite the refactor API commit and remove the DocumentSource classes there? It would be complicated to rewrite the previously merged commit according to me. The wiki data can be removed though, which is the more important thing that needs to be done.
It would probably be better to 
rebase to the merged branch -> remove the DocumentSource classes -> squash with previously merged commit -> rewrite the other commits and remove unlicensed wiki data so that its not there when merging.
Sorry about the late response. I was on very low internet connectivity for the past 3 weeks.

jaylett:
I think in practice at this point we're going to squash a fair amount at the end during rebase, so refactor whichever commits make sense to you.

PR #123, opened 2016-08-17
letor: Integrate ListNET and NDCGScore

ayshtmr:
This PR does the following:
1. Integrate `ListNETRanker`.
2. Integrate `NDCGScore`
3. `Letor` class updates for `ListNetRanker` and `NDCGScore`
4. Update `questletor.cc` and add `trainletor.cc` in `/bin`
5. Update `letor.rst`

This PR will put xapian-letor in a useable state. Exception handling and automated tests will follow next.


jaylett:
@ayshtmr can you rebase this onto latest master? I've jiggled the travis configuration, and it should build both clang & gcc again now.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -82,7 +76,58 @@ class XAPIAN_VISIBILITY_DEFAULT Ranker {
...
+
+    /** Method to save model as an external file.
+     *  @param output_filename      Filename by which model is to be stored.
+     */
+    void save_model_to_file(const char* output_filename);
+
+    /// Method to load model as an external file.
+    void load_model_from_file(const char* model_filename);
+
+    /// Method to re-rank a std::vector<Xapian::FeatureVector> by using the model.
```
jaylett: While this comment is strictly true, it might be better as "Method to re-rank a list of FeatureVectors (each representing a Xapian::Document) by using the model" or similar? However more to think about for now, as I think this will need more detailed help (since it's one of the core methods for LTR).

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+#include <vector>
+
+using namespace std;
+using namespace Xapian;
+
+ListNETRanker::ListNETRanker() {
+    // default values
+    learning_rate = 0.0001;
+    iterations = 15;
+    cout << "Initializing ListNETRanker with default parameters" << endl;
```
jaylett: You can use the `LOGCALL` macros for this (they're in `common/debuglog.h`) although it may involve linking in another object file. But it's probably a good idea to do a pass and convert all of these at some point.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+#include <sstream>
+#include <vector>
+
+using namespace std;
+using namespace Xapian;
+
+ListNETRanker::ListNETRanker() {
+    // default values
+    learning_rate = 0.0001;
+    iterations = 15;
```
jaylett: Is it not possible to just have these as parameter defaults to the other constructor?

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+    cout << "Initializing ListNETRanker with default parameters" << endl;
+}
+
+ListNETRanker::ListNETRanker(double learn_rate, int num_iterations) {
+    learning_rate = learn_rate;
+    iterations = num_iterations;
```
jaylett: These can be done in an initialiser list, and so possibly just put it into the header.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+
+	for (int i = 0; i < int(fvals.size()); ++i){
+
+		inner_product += parameters[i] * fvals[i];
+	}
+
+	return inner_product;
+}
+
+// From Theorem (6) in paper.
```
jaylett: I don't suppose we know which paper? It'd be good to provide a more complete reference ;-)

(The Cao et al paper doesn't seem to have a theorem 6.)

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+static vector< vector<double> >
```
jaylett: I wonder if this should be a locally-defined type, to make it clearer what's going on. Pulling `[0]` and `[1]` out of it is pretty opaque…

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+
+// Equation (3) in paper.
+static vector<double>
+calculateGradient(vector<FeatureVector> feature_vectors, vector< vector<double> > prob) {
+
+	vector<double> gradient(feature_vectors[0].get_fcount(),0);
+	int list_length = feature_vectors.size();
+
+	vector<double> prob_y(prob[0]);
+	vector<double> prob_z(prob[1]);
```
jaylett: Are these intermediate variables needed? I think it's just copying a `vector<double>` for no reason.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+	int num = new_parameters.size();
+
+	for (int i = 0; i < num; i++){
+		new_parameters[i] -= gradient[i] * learning_rate;
+	}
+
+}
+
+static void
+batchLearning(const std::vector<FeatureVector> & fvv, vector<double> & new_parameters, double learning_rate) {
```
jaylett: Given how short this is, I'd be inclined to inline it in the loop within `train_model()` below?

xapian-letor/api/letor.cc
```diff
@@ -36,8 +36,22 @@ namespace Xapian {
...
 
+Letor::Letor(const Xapian::Database & db, const Xapian::Query & query, Xapian::Ranker * ranker) {
+    internal = new Letor::Internal();
+    internal->letor_db = db;
+    internal->letor_query = query;
+    internal->ranker = ranker;
+}
+
+Letor::~Letor() {
+    delete internal->ranker;
```
jaylett: If you're going to do this, it should be documented in the API to avoid people getting confused as to why the `Ranker` they made has disappeared. (Unless it's possible to use the ref counting stuff here?)

xapian-letor/include/xapian-letor/letor.h
```diff
@@ -45,6 +46,20 @@ class XAPIAN_VISIBILITY_DEFAULT Letor {
     /// Default constructor.
     Letor();
 
+    /** Constructor to initialise Letor object during training.
+     *  Xapian::Query(ies) are supplied via the 'query' file.
+     * @param db     Xapian::Database to use for LTR
+     * @param ranker Ranker to be used. By default it is ListNETRanker
+     */
+    Letor(const Xapian::Database & db, Xapian::Ranker * ranker = new ListNETRanker());
```
jaylett: I'd construct the `ListNETRanker()` in the constructor, if `ranker == 0`.

xapian-letor/bin/questletor.cc
```diff
@@ -42,16 +43,8 @@
...
-#define PROG_DESC "Xapian command line search tool with Lerning to Rank Facility"
-
-typedef std::pair<Xapian::docid, double> MyPair;
-
-struct MyTestCompare {
-    bool operator()(const MyPair& firstPair, const MyPair& secondPair) const {
-	return firstPair.second < secondPair.second;
-    }
-};
+#define PROG_NAME "letor-rank"
```
jaylett: If its help text is going to say this, it should probably be renamed to that. (Or pull the `PROG_NAME` out of `argv[0]`?)

xapian-letor/bin/questletor.cc
```diff
@@ -70,28 +63,30 @@ static void show_usage() {
...
+"  --boolean-prefix=PFX:TERMPFX  Add a boolean prefix\n"
+"  --help          display this help and exit\n"
+"  --version       output version information and exit\n";
 }
 
 int
 main(int argc, char **argv)
 try {
-    const char * opts = "d:m:s:p:b:hv";
+    const char * opts = "a:b:c:d:e:f:g:h";
```
jaylett: The short options for this were better before. `-d` for `--db` and so forth.

xapian-letor/bin/trainletor.cc
```diff
@@ -0,0 +1,146 @@
...
+#include <string>
+#include <map>
+#include <set>
+#include <math.h>
+
+#include "gnu_getopt.h"
+
+using namespace std;
+
+#define PROG_NAME "letor-train"
```
jaylett: These days we prefix all our binaries with `xapian-` to avoid collisions. Perhaps just `xapian-train` is sufficient here? (And rename the file, binary &c.)

xapian-letor/bin/trainletor.cc
```diff
@@ -0,0 +1,146 @@
...
+"  --train=PATH  path to file by which training file is to be saved\n"
+"  --model=PATH  path to file by training file is to be saved\n"
+"  --help          display this help and exit\n"
+"  --version       output version information and exit\n";
+}
+
+int
+main(int argc, char **argv)
+try {
+    const char * opts = "a:b:c:d:e:f:g:h";
```
jaylett: These shouldn't just be alphabetically-ordered; they should be reasonable short options matching the long ones below.

xapian-letor/api/letor.cc
```diff
@@ -64,8 +64,8 @@ Letor::set_query(const Xapian::Query & query) {
 }
 
 std::vector<Xapian::docid>
-Letor::letor_rank(const Xapian::MSet & mset, Xapian::FeatureList & flist, const char* model_filename) {
-    return internal->letor_rank(mset, flist, model_filename);
+Letor::letor_rank(const Xapian::MSet & mset, const char* model_filename, Xapian::FeatureList & flist) {
```
jaylett: I think we shouldn't load the model during `letor_rank()`. Once we have a REPL on top of this (say, python), I can see people experimenting with queries on top of the same learned model, in which case we'd want to bring up a suitable ranker, load its model, and then run lots of queries without loading the model again. (Loading the model probably isn't hugely slow, but it'll add up, and future rankers might need to load more data, and possibly unpack it or upload onto a CUDA card or something.)

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+
+		batchLearning(fvv, new_parameters, learning_rate);
+	}
+
+    parameters = new_parameters;
+
+}
+
+void
+ListNETRanker::save_model_to_file(const char* output_filename) {
```
jaylett: We should document the file format used for ListNET model files here (or point to it in other documentation).

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+    train_file.close();
+
+    loaded_parameters.pop_back();
+    parameters = loaded_parameters;
+
+    cout<< "ListNET model loaded from: \"" << model_filename << "\""  << endl;
+}
+
+std::vector<FeatureVector>
+ListNETRanker::rank(std::vector<FeatureVector> & fvv) {
```
jaylett: `fvv` can (and probably should) be `const` here?

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,264 @@
...
+
+    std::vector<FeatureVector> testfvv = fvv;
+    int testfvvsize = testfvv.size();
+
+    cout << "Performing ranking using ListNET model..." << endl;
+
+    std::vector<double> new_parameters = parameters;
```
jaylett: You aren't modifying `parameters`, so is there a reason not to just use it directly?

xapian-letor/scorer/scorer.cc
```diff
@@ -0,0 +1,31 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
+#include "xapian-letor/scorer.h"
+
+using namespace std;
+using namespace Xapian;
+
+Scorer::Scorer() {
```
jaylett: Can you just fold these into the header file?

xapian-letor/scorer/ndcg_score.cc
```diff
@@ -0,0 +1,71 @@
+/** @file ndcg_score.cc
+ *  @brief Implementation of NDCGScore
+ */
+/* Copyright (C) 2016 Ayush Tomar
```
jaylett: Is this a fresh implementation of NDCG?

jaylett:
Most of my comments are suggested refactors to improve the incoming code you've integrated. Overall this is feeling really great to see all the pieces slotting into place :-)

xapian-letor/include/xapian-letor/ranker.h
ayshtmr: I'll make the change. I think the entire documentation in LTR needs to be revisited once, specially `letor.rst`, which I think is written in a somewhat casual manner right now.

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: Will have a look at `LOGCALL`

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: silly me

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: It seems the paper I referred to was the "long version" of the ICML paper. Will make the change.
https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tr-2007-40.pdf

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: Not needed per se. Had them for the sake of clarity.

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: will make the change

xapian-letor/api/letor.cc
ayshtmr: You're right. Better to use a reference counted pointer.

xapian-letor/bin/questletor.cc
ayshtmr: okay

xapian-letor/bin/questletor.cc
ayshtmr: I got confused as to who should get the `-m` option, `msetsize` or `model`, and what option to give to the one that is left. Hence decided to get rid of short options alltogether. Maybe its a messy approach?

xapian-letor/bin/trainletor.cc
ayshtmr: will make the change

xapian-letor/bin/trainletor.cc
ayshtmr: same as what I did in `questletor.cc`, couldn't figure out what to do in case of conflicts. Does it matter if short options are not reasonable, given that i have a choice of only a single character? (maybe i'm missing something here)

xapian-letor/api/letor.cc
ayshtmr: I'm internall calling `Ranker::load_model_from_file(const char* model_filename)` in this method. We could have a similar function in `Letor` as well, so that `letor_rank(*)` doesn't have to do this. Another reason why `Letor` class should be eliminated and be put inside `Ranker`.

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: Will add the doccomments. Its probably better to have seperate topic documentation pages which mentions these type of notes for each ranker and scorer.

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: will update

xapian-letor/ranker/listnet_ranker.cc
ayshtmr: Will update

xapian-letor/scorer/scorer.cc
ayshtmr: okay

xapian-letor/scorer/ndcg_score.cc
ayshtmr: I missed v-hasu on this one. 
It uses a different version of DCG calculation, but its not entirely fresh.

xapian-letor/include/xapian-letor/ranker.h
jaylett: 👍 

I wonder if the right thing with the existing documentation is to convert it into a section in the user guide, and keep documentation in the source code to API docs. (Olly's issue with `INSTALL` could be addressed at the same time.)

xapian-letor/bin/questletor.cc
jaylett: I think it's better to have short options. I'd normally say that whichever is more likely to be used should get `-m`, so it's more guessable. However here I wonder if we can use `-f` for "model file", since `-f` is so often used to specify a file. (And `--model-file`, perhaps.)

xapian-letor/bin/trainletor.cc
jaylett: It's probably harder to get familiar with short options (which is what I tend to use with most CLI programs) if they don't "connect" with their use at all.

This might be easier if we split creating the training file out from training the model. Then we could have:

```
$ xapian-prepare-trainingfile [options] <queryfile> <qrelfile> <trainingfile>
$ xapian-train-model [options] <trainingfile> <modelfile>
```

or something. Since most of the options are required for each stage, they may well be better as arguments instead.

xapian-letor/api/letor.cc
jaylett: [This is getting quite future speculative, so we don't have to worry about it yet. However:]

Ideally I think we'd just let people call it on `Ranker` directly. The only problem is for the default ranker as we currently do things. However I'm wondering if we should actually be storing the model as database metadata instead of as a file (so it'll get replicated, for instance). Then we could have a default metadata key name for each ranker type, which would take care of the default, or you could specify a different key (in the ranker constructor, probably) if you wanted to have different models trained with the same algorithm. So:

```
Ranker ranker = ListNET(); // Uses default, loads model from key letor.models.listnet or similar
Ranker ranker2 = ListNET("my.listnet.model"); // Loads a different model
```

xapian-letor/ranker/listnet_ranker.cc
jaylett: I think file formats are actually internal developer documentation, so we don't need them in the user guide. I'd leavea them in doc comments at least for now. (They might go in the nascent developer guide at some point, I guess, along with documentation on the database backend file formats, for instance.)

xapian-letor/scorer/ndcg_score.cc
jaylett: Sorry, by fresh I was just confirming that you wrote it yourself. I wasn't expecting it to be a novel twist on NDCG itself. It may be worth pointing to a good public definition of NDCG though, now I think about it.

ayshtmr:
@jaylett I have address all of the issues that you highlighted, except for `LOGCALL` macros. I couldn't understand its use-case from the documentation. Could you please explain in what cases should I be using it and how?

jaylett:
For examples of how to use the `LOGCALL` macros, see `xapian-core/api/valueiterator.cc`. Use while debugging is covered under "Debugging Messages" in `xapian-core/HACKINGS`. Include `common/debuglog.h` to get them, and there's documentation of the macros there (skip to about line 368).

jaylett:
This looks good :)

Once it's gone green, can you squash as appropriate please?

ayshtmr:
Sure!

ayshtmr:
An incorrectly named fixup commit gave me some trouble in autosquashing commits. Anyway, I think its ready to merge now.

jaylett:
This hasn't squashed properly…you add `trainletor.cc`, then you remove references in it, so you end up with `trainletor.cc` still there. You should be able to eliminate both of those commits entirely (use interactive rebase). Also, if you run the diff through `xapian-check-patch`, there are a small number of outstanding whitespace issues.

I think you can happily squash removing `questletor` & adding `xapian-rank`, as it's mostly a rename. (And include the earlier commit that changes `questletor.cc`.)

Also, unless you can collapse them with other commits, I'd squash all the docs updates together in one go and list what's changed. (It'd be better to have the docs updated in the same commit as changes the code, but that may not be practical to identify with all the changes.)

ayshtmr:
@jaylett If you think this is ready, can you please merge? (I have the exception handling PR ready, which enhances over this one).

xapian-letor/bin/trainletor.cc
ojwb: It's also very frustrating when short options aren't consistent, especially within a codebase.  Consistency makes complex systems much easier to learn and to remember.

For example, in Xapian we use `-d` as the short option for `--db` consistently.  Imagine how annoying it would be if one program used `-d` for something else (especially if it was short for something like `--delete-database`!)

It's similarly annoying when long options aren't consistent (e.g. `--db` in some places, `--database` in others).

So where an option is already present in another program, use the same long and short names.  And try to avoid picking new short names for a program that are used elsewhere for an option which might make sense to add to this program.

Also, there's no requirement to have a short option for everything.  If an option is dangerous or very rarely used, it may make sense to not allocate a short option for it.

Actually, I notice `quest` has `-v` for `--version` which we generally use for `--verbose`, and `quest` could sensibly have a `--verbose` option one day.  Please don't copy that mistake to here!

ayshtmr:
letor's `Makefile.am` does not include `common/Makefile.mk`, instead, `noinst_HEADERS` are included in the following way:
```
noinst_HEADERS +=\
	common/safeunistd.h\
	common/str.h\
	common/stringutils.h
```
Its probably better to remove this all and include `common/Makefile.mk` instead.

ayshtmr:
Please ignore my previous comment. I did not realise that `-core` already includes `common/Makefile.mk`. `noinst_HEADERS` are still unnecessary though, and need to be removed.

xapian-letor/bin/trainletor.cc
ayshtmr: Will eliminate the short version `-v` for `--version` and update

ayshtmr: I still need to assign a character for switch case to correspond to long option `version`. It is fine to give it any junk character, say `'*'`? e.g.
```
static const struct option long_opts[] = {
	    { "db",     required_argument, 0, 'd' },
	    { "help",   no_argument, 0, 'h' },
	    { "version",    no_argument, 0, '*' },
	    { NULL,     0, 0, 0}
	};
```

ojwb: It's better to just create some symbolic constants:

https://github.com/xapian/xapian/blob/master/xapian-core/bin/xapian-compact.cc#L37L39
https://github.com/xapian/xapian/blob/master/xapian-core/bin/xapian-compact.cc#L113L117

ojwb:
Ah yes, `common/Makefile.mk` contains all the files that core uses, but letor doesn't need most of those, and referring to the unused ones is probably a bad idea - they're likely to end up shipped in the letor tarball.

Probably simplest to just handle the common files you do want in `xapian-letor/Makefile.am` for now.  We're hoping to come up with a better plan for the sharing stuff than the current `common/` in the not too distant future anyway.

jaylett:
Except for a final squash, I'm happy with this. (Sorry about the confusion over the `main()` braces.) If @ojwb is happy, I can get this merged.

ayshtmr:
Should I leave the last commit [Update noinst_HEADERS](https://github.com/xapian/xapian/pull/123/commits/15d471495cf478f5d25d1df898c9ce5e34c87716) as it is, or should it be squashed with [Delete questletor.cc & add xapian-rank.cc](https://github.com/xapian/xapian/pull/123/commits/fe30fe92b5a76a7a80c00bfd16e9d085bb5cd0a4) ?

jaylett:
It should be squashed with whatever needs it to build.

ayshtmr:
done

xapian-letor/.gitignore
```diff
@@ -8,10 +8,14 @@ autom4te.cache
...
-bin/questletor.1
+bin/xapian-rank
+bin/xapian-rank.1
 bin/xapian-letor-update
 bin/xapian-letor-update.1
+bin/xapian-prepare-trainingfile
+bin/xapian-prepare-trainingfile.1
+bin/xapian-train
+bin/xapian-train.1
 config.guess
```
ojwb: When there's a sorted list and you add or modify an entry, it's helpful to preserved the sorted nature (so `xapian-rank` would be just before `xapian-train`).

xapian-letor/Makefile.am
```diff
@@ -61,13 +61,16 @@ include api/Makefile.mk
...
+include scorer/Makefile.mk
 
 libxapianletor_la_LIBADD += $(LIBSVM_LIBS)
 
 noinst_HEADERS +=\
 	common/safeunistd.h\
 	common/str.h\
-	common/stringutils.h
+	common/stringutils.h\
+	common/gnu_getopt.h
```
ojwb: Again, this was a sorted list before.

xapian-letor/api/featurelist.cc
```diff
@@ -31,8 +32,7 @@ namespace Xapian {
 
 FeatureList::FeatureList() {
 
-    std::cout << "Initialising DEFAULT list of features" << endl;
-
+    LOGCALL_CTOR(API, "FeatureList", NO_ARGS);
     feature.push_back(new TfFeature());
```
ojwb: Blank lines at the start of blocks are better avoided too.  Again, probably just avoid introducing more, and we can do a mass clean-up commit at some point.

xapian-letor/api/letor.cc
```diff
@@ -34,58 +35,91 @@ using namespace std;
...
-Letor::Letor() : internal(new Letor::Internal) { }
+Letor::Letor(const Xapian::Database & db, Xapian::Ranker * ranker) {
+    LOGCALL_CTOR(API, "Letor", db | ranker);
+    internal = new Letor::Internal();
+    internal->letor_db = db;
+    if (ranker == 0) {
+	internal->ranker = new ListNETRanker();
+    }
+    else {
+	internal->ranker = ranker;
```
ojwb: Should be `} else {`

xapian-letor/api/letor.cc
```diff
@@ -34,58 +35,91 @@ using namespace std;
...
-Letor::~Letor() { }
+Letor::Letor(const Xapian::Database & db, const Xapian::Query & query, Xapian::Ranker * ranker) {
+    LOGCALL_CTOR(API, "Letor", db | query | ranker);
+    internal = new Letor::Internal();
+    internal->letor_db = db;
+    internal->letor_query = query;
+    if (ranker == 0) {
+	internal->ranker = new ListNETRanker();
+    }
+    else {
```
ojwb: Ditto.

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
 	for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
 
 	    Xapian::Document doc = i.get_document();
-	    fvv_mset[k++].set_label(getlabel(doc, qid));
+	    int label = getlabel(doc, qid);
+	    if (label != -1) { // only add FeatureVector which is found in the qrel file
+		fvv_mset[k].set_label(label);
+		fvv_qrel.push_back(fvv_mset[k]);
+	    }
+	    k += 1;
```
ojwb: `++k` is preferable here (see `HACKING`).

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+    parser.set_stemming_strategy(Xapian::QueryParser::STEM_SOME);
+    parser.set_stopper(&mystopper);
+
+    this->qrel = load_relevance(qrel_file);
+
+    string str1;
+    ifstream queryfile;
+    queryfile.open(query_file.c_str(), ios::in);
+
+    if(!queryfile.good()){
```
ojwb: Space before `(` and after `)` on an `if`.

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+    while (!queryfile.eof()) {           //reading all the queries line by line from the query file
+
+	getline(queryfile, str1);
+
+	if (str1.empty()) {
+	    break;
+	}
+
+	string qid = str1.substr(0, (int)str1.find(" "));
+	string querystr = str1.substr((int)str1.find("'")+1, (str1.length() - ((int)str1.find("'") + 2)));
```
ojwb: Spaces missing around the first `+`.

The parsing here is rather crude and lacking in error checking (there's nothing to require the `'` to be after the space for example, and it'll happily continue even if there's no '`' in the line).  It also searches for the same `'` twice, though perhaps the compiler can spot that it doesn't need to repeat that search.  I guess these can get dealt with later though.

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+
+	string qq = querystr;
+	istringstream iss(querystr);
+	string title = "title:";
+	while (iss) {
+	    string t;
+	    iss >> t;
+	    if (t.empty())
+	    break;
```
ojwb: Underindented.

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+	int k = 0;
+	for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
+
+	    Xapian::Document doc = i.get_document();
+	    int label = getlabel(doc, qid);
+	    if (label != -1) { // only add FeatureVector which is found in the qrel file
+		rankedfvv[k].set_label(label);
+		rankedfvv_qrel.push_back(rankedfvv[k]);
+	    }
+	    k += 1;
```
ojwb: `++k`

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+
+	Xapian::MSet mset = enquire.get_mset(0, msetsize);
+
+	std::vector<FeatureVector> fvv_mset = flist.create_feature_vectors(mset, query, letor_db);
+	std::vector<FeatureVector> rankedfvv = ranker->rank(fvv_mset);
+	std::vector<FeatureVector> rankedfvv_qrel;
+
+	int k = 0;
+	for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
+
```
ojwb: Blank line.

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+	    }
+	    k += 1;
+	}
+
+	double score = scorer->score(rankedfvv);
+	cout << "Ranking score for qid:" << qid << " = " << score << endl;
+
+	total_score += score;
+	num_queries += 1;
+
```
ojwb: `++`, blank line.

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+    }
+    queryfile.close();
+    total_score = total_score/num_queries;
```
ojwb: Spaces around `/`.

xapian-letor/api/letor_internal.cc
```diff
@@ -315,19 +313,125 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
+    cout << "Average ranking score = " << total_score << endl;
+
+}
```
ojwb: Blank line (I'm going to give up pointing out every one of these you've added...)

xapian-letor/bin/Makefile.mk
```diff
@@ -15,16 +11,39 @@ bin_xapian_letor_update_SOURCES =\
...
+bin_xapian_train_LDADD = libxapianletor.la
+
 if !MAINTAINER_NO_DOCS
 dist_man_MANS +=\
-	bin/questletor.1\
-	bin/xapian-letor-update.1
+	bin/xapian-letor-update.1\
+	bin/xapian-rank.1\
+	bin/xapian-prepare-trainingfile.1\
+	bin/xapian-train.1
```
ojwb: No longer sorted.

xapian-letor/bin/Makefile.mk
```diff
@@ -15,16 +11,39 @@ bin_xapian_letor_update_SOURCES =\
...
+
+bin/xapian-rank.1: bin/xapian-rank$(EXEEXT) makemanpage
+	./makemanpage bin/xapian-rank $(srcdir)/bin/xapian-rank.cc bin/xapian-rank.1
+
+bin/xapian-prepare-trainingfile.1: bin/xapian-prepare-trainingfile$(EXEEXT) makemanpage
+	./makemanpage bin/xapian-prepare-trainingfile $(srcdir)/bin/xapian-prepare-trainingfile.cc bin/xapian-prepare-trainingfile.1
+
+bin/xapian-train.1: bin/xapian-train$(EXEEXT) makemanpage
+	./makemanpage bin/xapian-train $(srcdir)/bin/xapian-train.cc bin/xapian-train.1
 endif
```
ojwb: No longer sorted (target order).

xapian-letor/bin/questletor.cc
```diff
@@ -166,7 +166,7 @@ try {
     parser.set_stemming_strategy(Xapian::QueryParser::STEM_SOME);
     parser.set_stopper(&mystopper);
```
ojwb: Don't we want to just delete this file?  It's been replaced by `xapian-rank.cc` hasn't it?

xapian-letor/bin/trainletor.cc
```diff
@@ -0,0 +1,146 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <iostream>
+#include <fstream>
+#include <sstream>
+#include <string>
+#include <map>
+#include <set>
+#include <math.h>
```
ojwb: `<cmath>` is preferred, unless there's a good reason for needing `<math.h>`, like something platform extension not in ISO C++11.  Also `<cstdio>` rather than `<stdio.h>` above, and it's unlikely you want both `<cstdlib>` and `<stdlib.h>`.

Also sorting this list would make it easier for future coders.

xapian-letor/bin/trainletor.cc
```diff
@@ -0,0 +1,146 @@
...
+    static const struct option long_opts[] = {
+    { "db",     required_argument, 0, 'a' },
+    { "query",     required_argument, 0, 'b' },
+    { "qrel",     required_argument, 0, 'c' },
+    { "msize",  required_argument, 0, 'd' },
+    { "train",     required_argument, 0, 'e' },
+    { "model",     required_argument, 0, 'f' },
+    { "help",   no_argument, 0, 'g' },
+    { "version",    no_argument, 0, 'h' }
+    };
```
ojwb: The contents of the struct isn't indented enough, and there's a seeming random offset to the second column - we generally try to line up the second column with tabs here for readability - e.g. see `xapian-core/bin/xapian-compact.cc`.

You're also missing a final terminating element `{NULL, 0, 0, 0}` - I guess you happen to get lucky and there's some zeroed memory just after this structure when you compile, or else perhaps an unknown long option will cause a crash (for any known long option, it'll presumably stop looking once it finds it).

xapian-letor/bin/trainletor.cc
```diff
@@ -0,0 +1,146 @@
...
+"Options:\n"
+"  --db=DIRECTORY  database to search (multiple databases may be specified)\n"
+"  --query=PATH  path to query file\n"
+"  --qrel=PATH  path to qrel file\n"
+"  --msize=MSIZE   maximum number of matches to return\n"
+"  --train=PATH  path to file by which training file is to be saved\n"
+"  --model=PATH  path to file by training file is to be saved\n"
+"  --help          display this help and exit\n"
+"  --version       output version information and exit\n";
+}
```
ojwb: Conventionally the short options are documented above, where they exist, and the descriptions should be aligned (with **spaces** since that means the alignment works even if the terminal/pager has different tab stop settings) - e.g.:

```
"  -d, --db=DIRECTORY  database to search (multiple databases may be specified)\n"
"       --version               output version information and exit\n"
```

xapian-letor/bin/trainletor.cc
```diff
@@ -0,0 +1,146 @@
+/** @file trainletor.cc
+ * @brief Command line tool to prepare Letor training file and train the model
```
ojwb: I started to review this, but I'm starting to suspect it's been renamed to something else and the old version not removed (though I didn't spot an exact replacement).

@ayshtmr Could you do a pass through and make sure any files which have been renamed (or just removed) have actually been removed from git too (with `git rm`).  That'll save both of us time and effort - there's no point me giving feedback on dead code which you then have to wade through and try to work out if it applies to a relocated version somewhere.

xapian-letor/tests/generate-api_generated
```diff
@@ -24,15 +24,16 @@ my %uncopyableclasses = (
...
 
 my %no_get_description = (
     'Letor' => '',
     'FeatureVector' => '',
     'Feature' => '',
-    'FeatureList' => ''
+    'FeatureList' => '',
+    'ListNETRanker' => '',
+    'NDCGScore' => ''
 );
```
ojwb: This list wasn't sorted before admittedly, but really should be.

xapian-letor/scorer/ndcg_score.cc
```diff
@@ -0,0 +1,76 @@
...
+NDCGScore::NDCGScore() {
+    LOGCALL_CTOR(API, "NDCGScore", NO_ARGS);
+}
+
+NDCGScore::~NDCGScore() {
+    LOGCALL_DTOR(API, "NDCGScore");
+}
+
+static double
+get_dcg(const std::vector<double> labels) {
```
ojwb: This currently passes `labels` by value - I think you're missing an `&`, i.e.:

```
get_dcg(const std::vector<double>& labels)
```

xapian-letor/scorer/ndcg_score.cc
```diff
@@ -0,0 +1,76 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
```
ojwb: Every `.cc` source file (but NB **not** `.h` files) should `#include <config.h>` as the first header just after the copyright boilerplate.  This means that any `HAVE_XXX` preprocessor defines set as a result of `configure` checks will be available (both to this source file and to any headers that get included).

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    prob.push_back(prob_y);
+    prob.push_back(prob_z);
+
+    return prob;
+}
+
+// Equation (6) in paper Cao et al. "Learning to rank: from pairwise approach to listwise approach."
+static vector<double>
+calculateGradient(vector<FeatureVector> feature_vectors, prob_distrib_vector prob) {
+    LOGCALL_STATIC_VOID(API, "calculateGradient", feature_vectors | prob_distrib_vector);
```
ojwb: The last part here should be the parameter **names** not their types.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    vector< vector<double> > prob;
+    prob.push_back(prob_y);
+    prob.push_back(prob_z);
+
+    return prob;
+}
+
+// Equation (6) in paper Cao et al. "Learning to rank: from pairwise approach to listwise approach."
+static vector<double>
+calculateGradient(vector<FeatureVector> feature_vectors, prob_distrib_vector prob) {
```
ojwb: Both are these are `std::vector`, so you really want to pass them by const reference not value - currently this will have to copy both the passed vectors.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    LOGCALL_STATIC_VOID(API, "calculateGradient", feature_vectors | prob_distrib_vector);
+
+    vector<double> gradient(feature_vectors[0].get_fcount(),0);
```
ojwb: Space after `,`

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    int list_length = feature_vectors.size();
+
+    // Hold ground truth probability distribution
+    vector<double> prob_y(prob[0]);
+    // Hold prediction score probability distribution
+    vector<double> prob_z(prob[1]);
```
ojwb: There's no need to copy these vectors out of `prob` - just use a const reference, e.g.:

```
const vector<double>& prob_y(prob[0]);
```

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+
+    for (int i = 0; i < list_length; i++) {
+	vector<double> fvals = feature_vectors[i].get_fvals();
+
+	for (int k = 0; k < int(fvals.size()); k++){
```
ojwb: It would be cleaner to make the `k` the same type as `fvals.size()` - i.e. `fvals::size_type`.  Then you don't need to cast things to `int` to avoid warnings, and you don't even need to consider what happens if the cast truncates the value.

Similarly for `i` and `list_length`.

Also, missing space before `{`.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+	    gradient[k] += second_term;
+
+	}
+    }
+
+    return gradient;
+}
+
+static void
+updateParameters(vector<double> & new_parameters, vector<double> gradient, double learning_rate) {
```
ojwb: Pass `gradient` by const reference here (basically, check everywhere that complex types like `vector`, `map`, etc are passed by value).

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    LOGCALL_STATIC_VOID(API, "updateParameters", new_parameters | gradient | learning_rate);
+	int num = new_parameters.size();
```
ojwb: Indentation (and type)

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+
+    // get the training data
+    std::vector<Xapian::FeatureVector> fvv = get_traindata();
+    int fvv_len = fvv.size();
+
+    int feature_cnt = -1;
+    if (fvv_len != 0) {
+	feature_cnt = fvv[0].get_fcount();
+    }
+    else {
```
ojwb: `} else {`

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+	new_parameters[i] -= gradient[i] * learning_rate;
+    }
+
+}
+
+void
+ListNETRanker::train_model() {
+    LOGCALL_VOID(API, "ListNETRanker::train_model", NO_ARGS);
+
+    std::cout << "ListNET model begin to train..." << endl;
```
ojwb: A library shouldn't spew stuff to stdout.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    int fvv_len = fvv.size();
+
+    int feature_cnt = -1;
+    if (fvv_len != 0) {
+	feature_cnt = fvv[0].get_fcount();
+    }
+    else {
+	std::cout << "The training data is NULL!" << endl;
+	exit(1); // TODO: Throw exception
+    }
```
ojwb: And it certainly shouldn't call `exit()`.

Resolving this TODO would be good - even just `throw Xapian::InvalidArgumentError("The training data is empty")`. would be an improvement (`NULL` seems the wrong word...)

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+
+	// initialize Probability distributions of y and z
+	prob_distrib_vector prob = initializeProbability(fvv, new_parameters);
+	// compute gradient
+	vector<double> gradient = calculateGradient(fvv, prob);
+	// update parameters: w = w - gradient * learningRate
+	updateParameters(new_parameters, gradient, learning_rate);
+    }
+
+    parameters = new_parameters;
```
ojwb: Better to swap than assign here:

```
    swap(parameters, new_parameters);
```

A copy is O(new_parameters.size()), while swapping a vector can typically be achieved by just exchange a pointer and a couple of integers.  Note that this works here because `new_parameters` is about to go out of scope, so we don't need it to keep the same value.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    int parameters_size = parameters.size();
+
+    std::ostringstream oss;
+    oss.precision (std::numeric_limits<double>::digits10);
+
+    for (int i = 0; i < parameters_size; ++i) {
+	oss << parameters[i] << endl;
+    }
+
+    parameters_file << oss.str();
```
ojwb: Why not eliminate `oss` entirely, and just set the precision and write the parameter values straight to `parameters_file`?

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    parameters_file.close();
+
+    cout<< "ListNET model saved to \"" << output_filename << "\"" << endl;
```
ojwb: Please get rid of all use of `iostream` in the library code - if you need these for debug, you can use the logging functionality in `debuglog.h`.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    while (train_file.peek() != EOF) {
+
+	double parameter;
+	train_file >> parameter;
+	loaded_parameters.push_back(parameter);
+    }
+
+    train_file.close();
+
+    loaded_parameters.pop_back();
```
ojwb: Why do you throw away the last parameter read?

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    parameters = loaded_parameters;
```
ojwb: The `swap()` idiom can be used again here.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+    for (int i = 0; i <testfvvsize; ++i) {
+
+	double listnet_score = 0;
+
+	std::vector<double> fvals = testfvv[i].get_fvals();
+	int fvalsize = fvals.size();
+
+	if (fvalsize != parameters_size) {
+	    std::cout << "Number of fvals don't match the number of ListNet parameters" << endl;
+	    exit(1); // TODO: should throw an exception
```
ojwb: -> exception

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	for(int j = 0; j < fvalsize; ++j) {
```
ojwb: Space after `for`.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+	    listnet_score += fvals[j]* parameters[j];
+	}
```
ojwb: Space before `*`.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,256 @@
...
+
+	testfvv[i].set_score(listnet_score);
+    }
+
+    std::sort(testfvv.begin(),testfvv.end(),&Ranker::scorecomparer);
```
ojwb: Spaces after commas

xapian-letor/bin/trainletor.cc
ayshtmr: Sorry for this. Will remove both `questletor.cc` and `trainletor.cc`.

ayshtmr:
Added exception handling, const correctness fixes and layout fixes.

ayshtmr:
@jaylett @ojwb `dlhweight1` tests are failing. I wonder whats causing these, I haven't touched anything on `core`.

ojwb:
> `dlhweight1` tests are failing. I wonder whats causing these, I haven't touched anything on core.

Sorry, my fault - I fixed that 3 days ago, so should be good now, but for the purposes of this PR you can just ignore it - as you say you didn't touch core.

ayshtmr:
@ojwb I think this PR is more or less complete. You can do a final review.

xapian-letor/api/letor.cc
```diff
@@ -21,71 +22,115 @@
...
+	internal->ranker = ranker;
+    }
+}
 
-Letor::~Letor() { }
+Letor::Letor(const Xapian::Database & db, const Xapian::Query & query, Xapian::Ranker * ranker)
+{
+    LOGCALL_CTOR(API, "Letor", db | query | ranker);
+    if (query.empty()) {
+	throw Xapian::InvalidArgumentError("Can't initialise with an empty query string");
```
ojwb: It's actually an empty query **object**.

xapian-letor/api/letor_internal.cc
```diff
@@ -271,23 +231,149 @@ Letor::Internal::prepare_training_file(const string & queryfile, const string &
...
 	    break;
 	}
+	++num_queries;
 
-	string qid = str1.substr(0, (int)str1.find(" "));
-	string querystr = str1.substr((int)str1.find("'")+1, (str1.length() - ((int)str1.find("'") + 2)));
+	char * str;
+	char * x1;
+	x1 = const_cast<char*>(str1.c_str());
+	str = strtok(x1, " ");
```
ojwb: This call to `strtok()` will modify the string pointed to by `x1`, which I'm fairly sure is undefined behaviour.

It's usually better to use methods of `std::string` to manipulate a `std::string` rather than the string functions C++ inherits from C, but that's especially true of `strtok()` which modifies the string it works on and isn't even reentrant (`strtok_r()` is reentrant, but is specified by POSIX rather than the ISO C standard, so is less portable).

You can split a string into words at spaces using methods of `std::string` like this:

```
        size_t j = 0;
        while (j < s.size()) {
            size_t i = s.find_first_not_of(' ', j);
            if (i == string::npos) break;
            j = s.find_first_of(' ', i);
            token.push_back(s.substr(i, j - i));
        }
```

(The second argument of `substr()` can be larger than the length of the string, which happens if there's a word at the end of the string, since then `j` is `string::npos`).

You can also specify several characters instead of a single one, so to split at space or tab:

```
s.find_first_not_of(" \t", j);
```

etc.

xapian-letor/api/letor_internal.h
```diff
@@ -23,38 +24,38 @@
 #define XAPIAN_INCLUDED_LETOR_INTERNAL_H
 
 #include "xapian-letor/letor.h"
-#include "xapian-letor/ranker.h"
-
-#include <map>
```
ojwb: `std::map` is used in this header, so it's better to `#include <map>` explicitly rather than relying on it happening to get included by another header included from here.  That way each header is more self-contained, and eliminating an apparently unnecessary include from one header doesn't trigger an avalanche of required changes to other files.

xapian-letor/bin/xapian-prepare-trainingfile.cc
```diff
@@ -0,0 +1,119 @@
...
+    "Options:\n"
+    "  -d, --db=DIRECTORY  database to search (multiple databases may be specified)\n"
+    "  -m, --msize=MSIZE   maximum number of matches to return\n"
+    "      --help          display this help and exit\n"
+    "      --version       output version information and exit\n";
+}
+
+int
+main(int argc, char **argv)
+try {
```
ojwb: The body of this `try` block should be indented one level...

xapian-letor/bin/xapian-rank.cc
```diff
@@ -0,0 +1,229 @@
...
+    "                      (pass 'none' to disable stemming)\n"
+    "  -p, --prefix=PFX:TERMPFX  Add a prefix\n"
+    "  -b, --boolean-prefix=PFX:TERMPFX  Add a boolean prefix\n"
+    "      --help          display this help and exit\n"
+    "      --version       output version information and exit\n";
+}
+
+int
+main(int argc, char **argv)
+try {
```
ojwb: Here too.

xapian-letor/bin/xapian-rank.cc
```diff
@@ -0,0 +1,229 @@
...
+    exit(1);
+}
+
+cout << "Docids before re-ranking by LTR model:" << endl;
+int rank = 0;
+for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
+
+    Xapian::Document doc = i.get_document();
+    Xapian::docid did = doc.get_docid();
+    cout<<"Rank " << ++rank << ": " << int(did) << endl;
```
ojwb: Missing spaces around the first `<<`.

xapian-letor/bin/xapian-rank.cc
```diff
@@ -0,0 +1,229 @@
...
+Xapian::Ranker * ranker = new Xapian::ListNETRanker();
+Xapian::Letor ltr(db, query, ranker);
+
+// Get vector of re-ranked docids
+std::vector<Xapian::docid> ranked_docids = ltr.letor_rank(mset, model_path);
+
+cout << "Docids after re-ranking by LTR model:" << endl;
+rank = 0;
+for (int i = 0; i < int(ranked_docids.size()); i++)
+    cout<<"Rank " << ++rank <<": "<< ranked_docids[i] << endl;
```
ojwb: Whitespace around `<<` (twice).

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -28,61 +30,116 @@
 #include <xapian/visibility.h>
 
 #include "featurevector.h"
+#include "letor_error.h"
 
 #include <list>
+#include <iostream>
```
ojwb: `<iostream>` shouldn't be needed in the library code (and you've added it to the wrong place in the sorted list too).

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -0,0 +1,192 @@
...
+    // get the training data
+    std::vector<Xapian::FeatureVector> fvv = get_traindata();
+    size_t fvv_len = fvv.size();
+
+    int feature_cnt = -1;
+    if (fvv_len != 0) {
+	feature_cnt = fvv[0].get_fcount();
+    } else {
+        throw LetorInternalError("Training data is empty. Check training file.");
+    }
```
ojwb: Should be indented with a tab, not 8 spaces.

xapian-letor/scorer/ndcg_score.cc
```diff
@@ -0,0 +1,73 @@
...
+    LOGCALL_DTOR(API, "NDCGScore");
+}
+
+static double
+get_dcg(const std::vector<double> &labels)
+{
+    LOGCALL_STATIC(API, double, "get_dcg", labels);
+    double dcg = 0;
+    for (int i = 0; i < int(labels.size()); i++){
+	dcg += (pow(2,labels[i]) - 1)/(log(i+2)/log(2));
```
ojwb: Missing spaces around operators and after the comma.

Also, the `log2()` function is available so there's no need to write `log(x) / log(2)`.  Xapian has `common/log2.h` which provides it even if the standard library doesn't (it ought to with C++11, but the libraries tend to lag a bit on some platforms, so we've left the compatibility code in place for this release cycle).

xapian-letor/scorer/ndcg_score.cc
```diff
@@ -0,0 +1,73 @@
...
+    }
+    //DCG score of original ranking
+    double DCG = get_dcg(labels);
+    //DCG score of ideal ranking
+    sort(labels.begin(), labels.begin() + labels.size(), std::greater<double>());
+    double iDCG = get_dcg(labels);
+
+    if (iDCG == 0) // Don't divide by 0
+	return 0;
+    return DCG/iDCG;
```
ojwb: Spaces...

xapian-letor/bin/xapian-prepare-trainingfile.cc
ayshtmr: I remember us having a discussion about not adding an indentation level here since the entire `main()` block is contained in `try`. Couldn't find out where it was discussed. Please correct me if I'm wrong.

ojwb: I think you may be mixing up two things.

There doesn't need to be an extra block around the `try` here, but the body of the `try` should still be indented.

There was a case I mentioned earlier where we typically don't indent, but that was for the lines which are just strings in the help text in `show_usage()` above (in other programs we don't indent them really just because they're formatted to fit in 80 columns, and without an indent the text looks much more like it will once output - in particular if your editor window is 80 columns wide, you can see where the lines are too long easily) - e.g. look at `xapian-core/bin/xapian-delve.cc`.

ayshtmr: I get it now. Making the changes.

ayshtmr:
@ojwb Forgot to add a note. Pushed the errors that were highlighted. You can review.

xapian-letor/bin/xapian-rank.cc
```diff
@@ -202,7 +202,7 @@ for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
 
     Xapian::Document doc = i.get_document();
     Xapian::docid did = doc.get_docid();
-    cout<<"Rank " << ++rank << ": " << int(did) << endl;
+    cout<< "Rank " << ++rank << ": " << did << endl;
```
ojwb: There's a missing space between `cout` and `<<` here and below.

xapian-letor/scorer/ndcg_score.cc
```diff
@@ -49,7 +49,7 @@ get_dcg(const std::vector<double> &labels)
     LOGCALL_STATIC(API, double, "get_dcg", labels);
     double dcg = 0;
     for (int i = 0; i < int(labels.size()); i++){
-	dcg += (pow(2,labels[i]) - 1)/(log(i+2)/log(2));
+	dcg += (pow(2, labels[i]) - 1) / log2(i+2);
```
ojwb: Spaces around operators: `i+2` -> `i + 2`

ayshtmr:
@ojwb There's an incorrectly named fixup commit, so autosquash fails. If you think this PR is ready to merge, I'll reword that commit and perform the squash.

ojwb:
> There's an incorrectly named fixup commit, so autosquash fails. If you think this PR is ready to merge, I'll reword that commit and perform the squash.

Please do.

ojwb:
@ayshtmr Sorry, I missed that you'd done that - github doesn't seem to notify in this case.

PR #124, opened 2016-08-22
letor: Disable backend build options and update test harness

ayshtmr:
This PR does the following:

1. Remove backend build options from `configure.ac` and `tests/Makefile.am`
2. Add the new test harness which does the following:
    - `BackendManager` class with primarily a `get_database()` method, which returns an instance of      `Xapian::Database`, of default type set by `-core`. `get_writable_database` methods have been stripped off, since `xapian-letor` tests only need to operate on read-only databases.
   - Rest of the harness just copies from `-core`, (with relevant changes made to `TestRunner` class).
3. Update `apitest.h` according to new harness


xapian-letor/tests/Makefile.am
```diff
@@ -19,78 +19,9 @@ libxapian_la = libxapian.la
...
-	$(TESTS_ENVIRONMENT) ./apitest$(EXEEXT) -b singlefile_glass
-endif
-
-if BUILD_BACKEND_CHERT
-check-multi-chert: apitest$(EXEEXT)
-	$(TESTS_ENVIRONMENT) ./apitest$(EXEEXT) -b multi_chert
-check-chert: apitest$(EXEEXT)
-	$(TESTS_ENVIRONMENT) ./apitest$(EXEEXT) -b chert
-endif
+up: cd .. && $(MAKE)
```
jaylett: This doesn't do what you want :-)

ayshtmr: Removed the `up` option.

jaylett: Yes, but now you've changed the behaviour of `make` in the `tests` directory. Previously it built everything, because `up` was the first rule in the file. Now you've removed it, so `make` will pick something else to do.

ayshtmr: hmm.. I couldn't imagine a use case for anyone using phony option for `make up`. Obviously, I don't know enough when I say this.

ayshtmr: I'm adding back that line again as this:
```
up:
	cd .. && $(MAKE)
```

jaylett: Generally it's a good idea not to remove functionality that's already present (even if it was copied from another module).

ojwb: A phony target is one where the target name isn't a file to create, but rather effectively just a named list of commands to run.  We want `make up` should always run `cd .. && $(MAKE)`, but if there happened to be a file called `up` in the current directory `make` would see that and decide there was nothing to do.

By declaring a target as phony you tell `make` that it should always run the rule (at least for implementations such as GNU make which understand `.PHONY` - those which don't understand it will just treat `.PHONY` as another target, so you're no worse off that without the `.PHONY` list at least).

While it's pretty unlikely someone would create a file called `up`, some target names are more prone to this problem (`core` comes to mind!)

It also means `make` doesn't have to search for a match for the target in its list of implicit rules (the POSIX ones like `.c.o` and GNU make's `%.o` pattern rules), and presumably also avoids a `stat("up", &sb)` syscall to check if `up` exists and what its timestamp is - a small one-off overhead here, but might mount up for the complex tree of targets which automake creates.

I must admit I'd pretty much forgotten about `make up`, but I used to use it a lot and I likely will again now I've been reminded about it.  It's certainly the sort of thing that's good to have it supported consistently if we're going to have it.

xapian-letor/tests/harness/backendmanager.cc
```diff
@@ -0,0 +1,156 @@
...
+    // create a directory if not present
+    struct stat sbuf;
+    int result = stat(dirname.c_str(), &sbuf);
+    if (result < 0) {
+    if (errno != ENOENT)
+	throw Xapian::DatabaseOpeningError("Can't stat directory");
+    if (mkdir(dirname.c_str(), 0700) < 0)
+	throw Xapian::DatabaseOpeningError("Can't create directory");
+	return true; // Successfully created a directory.
+    }
```
ojwb: The body of the `if` above is under-indented.

xapian-letor/tests/harness/backendmanager.cc
```diff
@@ -0,0 +1,156 @@
...
+    string dbdir = parent_dir + "/db";
+    for (vector<string>::const_iterator i = files.begin();
+	 i != files.end(); ++i) {
+	dbdir += '_';
+	dbdir += *i;
+    }
+    // If the database is readonly, we can reuse it if it exists.
+    if (create_dir_if_needed(dbdir)) {
+    // Directory was created, so do the indexing.
+	Xapian::WritableDatabase db(dbdir, Xapian::DB_CREATE);
```
ojwb: The comment above should be indented to match the block it is in.

ojwb:
Great - rebased onto master and merged.

I did a quick diff of these files against the versions in xapian-core they were based on which uncovered a couple of remaining indentation issues, but I just fixed those myself in the interests of actually getting this work merged - see 23af2bf7a77aefddf4db5e989f1915c130e8a7e2.  The second one is much worse as the indentation is misleading in that case.

ayshtmr:
Thanks @ojwb

PR #125, opened 2016-10-04
Remove Letor class & store model as db metadata

ayshtmr:
This PR does the following:
1. Remove `Letor` class and merge it with `Ranker` class.
2. Store model as database metadata instead of as an external file.
3. Corresponding changes to `/bin`/ files.


xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -136,46 +132,56 @@ ListNETRanker::train_model() {
...
+    if (key.empty()) {
+	key = "LisntNET.model.default";
+    }
+    ostringstream oss;
     for (size_t i = 0; i <  parameters.size(); ++i)
-	parameters_file << setprecision(numeric_limits<double>::digits10) << parameters[i] << endl;
-    parameters_file.close();
+	oss << setprecision(numeric_limits<double>::digits10) << parameters[i] << endl;
+    string model_data = oss.str();
+    letor_db.set_metadata(key, model_data);
```
ojwb: I'd suggest using `serialise_double()` here - see `common/serialise-double.h` (and `.cc`).

This would reliably store the values without any loss of precision (I'm not entirely convinced this code always will), but should result in a smaller string for a given model (`serialise_double()` will encode any value in at most 10 bytes, and many in less than that; an ASCII string of the decimal representation will be much longer on average, as it only encodes 10 possible values per byte).

I'd expect `serialise_double()` to be faster too, though that's not a big concern here.

xapian-letor/ranker/ranker.cc
```diff
@@ -22,13 +22,280 @@
 
 #include <config.h>
 
+#include "xapian-letor/featurelist.h"
+#include "xapian-letor/featurevector.h"
+#include "xapian-letor/letor_error.h"
 #include "xapian-letor/ranker.h"
```
ojwb: Our convention is to include the header corresponding to the source file first in that source file (well, second really - `<config.h>` comes first).

The reason for this is that it means that every header (unless it has no corresponding source file) gets included once without anything else having been included first, so we know that it must be self-contained, rather than happening to work because some other header is always included first.

So this ought to be:

```
#include <config.h>

#include "xapian-letor/ranker.h"
 
#include "xapian-letor/featurelist.h"
#include "xapian-letor/featurevector.h"
#include "xapian-letor/letor_error.h"
#include "xapian-letor/scorer.h"
```

xapian-letor/ranker/ranker.cc
```diff
@@ -22,13 +22,280 @@
...
+    "i", "in", "is", "it",
+    "of", "on", "or",
+    "that", "the", "this", "to",
+    "was", "what", "when", "where", "which", "who", "why", "will", "with"
+};
+
+static vector<FeatureVector>
+load_list_fvecs(const string & filename)
+{
+    fstream train_file (filename, ios::in);
```
ojwb: There shouldn't be a space before the `(` here.

xapian-letor/ranker/ranker.cc
```diff
@@ -22,13 +22,280 @@
...
+	    if (token.size() != 4 || token[1] != "Q0") {
+		throw LetorParseError("Could not parse Qrel file at line:" + Xapian::Internal::str(qrel_count));
+	    }
+	    // Exception if relevance label is not a number
+	    int label;
+	    char * end;
+	    label = int(strtol(token[3].c_str(), &end, 10));
+	    if (*end) {
+		throw LetorParseError("Could not parse relevance label in Qrel file at line:"
+				       + Xapian::Internal::str(qrel_count));
```
ojwb: Just use `str()` (the `Xapian::Internal::` qualification isn't needed, and is just an implementation detail really, so could change in the future).

xapian-letor/ranker/ranker.cc
```diff
@@ -22,13 +22,280 @@
...
+		token.push_back(inLine.substr(i, j - i));
+	    }
+	    // Exceptions for parse errors
+	    if (token.size() != 4 || token[1] != "Q0") {
+		throw LetorParseError("Could not parse Qrel file at line:" + Xapian::Internal::str(qrel_count));
+	    }
+	    // Exception if relevance label is not a number
+	    int label;
+	    char * end;
+	    label = int(strtol(token[3].c_str(), &end, 10));
```
ojwb: Better to initialise when you declare where that's easy to do:

```
char * end;
int label = int(strtol(token[3].c_str(), &end, 10));
```

xapian-letor/ranker/ranker.cc
```diff
@@ -22,13 +22,280 @@
...
+
+    qrel = load_relevance(qrel_file);
+
+    ofstream train_file;
+    train_file.open(filename);
+
+    string str1;
+    ifstream myfile1;
+    myfile1.open(queryfile.c_str(), ios::in);
+    if(!myfile1.good()) {
```
ojwb: Missing space between `if` and `(`.

xapian-letor/ranker/ranker.cc
```diff
@@ -22,13 +22,280 @@
...
+	    iss >> t;
+	    if (t.empty())
+		break;
+	    string temp;
+	    temp.append(title);
+	    temp.append(t);
+	    temp.append(" ");
+	    temp.append(qq);
+	    qq = temp;
+	}
```
ojwb: Is this attempting to add `title:` before every word in the query string?

If so, it would be much better to just specify the default prefix as being the title and parse `querystr`.  The code here's going to get things horribly wrong in non-trivial cases - for example, a quoted phrase.

xapian-letor/ranker/ranker.cc
```diff
@@ -53,16 +341,147 @@ Ranker::labelcomparer(const FeatureVector & firstfv, const FeatureVector& second
...
+    // Set db
+    Xapian::Database letor_db(db_path);
+    // Load ranker model
+    load_model_from_metadata(model_key);
+    // Set scorer
+    Xapian::Internal::intrusive_ptr<Scorer> scorer;
+    if (scorer_type == "NDCGScore") {
+	scorer = new NDCGScore();
+    }
+    else {
```
ojwb: `} else {`

xapian-letor/ranker/ranker.cc
```diff
@@ -53,16 +341,147 @@ Ranker::labelcomparer(const FeatureVector & firstfv, const FeatureVector& second
...
+    parser.set_stemmer(stemmer);
+    parser.set_stemming_strategy(Xapian::QueryParser::STEM_SOME);
+    parser.set_stopper(&mystopper);
+
+    qrel = load_relevance(qrel_file);
+
+    string str1;
+    ifstream queryfile;
+    queryfile.open(query_file.c_str(), ios::in);
+    if(!queryfile.good()) {
```
ojwb: `if (`

xapian-letor/ranker/ranker.cc
```diff
@@ -53,16 +341,147 @@ Ranker::labelcomparer(const FeatureVector & firstfv, const FeatureVector& second
...
+	    iss >> t;
+	    if (t.empty())
+		break;
+	    string temp;
+	    temp.append(title);
+	    temp.append(t);
+	    temp.append(" ");
+	    temp.append(qq);
+	    qq = temp;
+	}
```
ojwb: This code again?  Why are there two copies?

xapian-letor/bin/xapian-train.cc
```diff
@@ -88,22 +88,23 @@ try {
...
     if (!have_database) {
 	cout << "No database specified so not running the query." << endl;
 	exit(0);
     }
 
-    // Initialise Letor object with db, query and ListNETRanker
-    // If not explicitly passed as done below, the default ranker is used.
+    // Initialise Ranker object.
     // See Ranker documentation for available Ranker options.
     Xapian::Ranker * ranker = new Xapian::ListNETRanker();
```
ayshtmr: Yes, there should be `delete ranker`. `Letor` class contained a smart pointer to `Ranker` but it is not the case now. 
Yes, this is dynamically allocated.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -23,62 +23,152 @@
...
 
     /// Virtual destructor since we have virtual methods.
     virtual ~Ranker();
 
-    /// Returns a vector of RankLists i.e. the training data
-    std::vector<Xapian::FeatureVector> get_traindata() const;
+    /** Specify path to Xapian::Database to be used for Ranker methods.
+     *  @param  dbpath      Path to Xapian::Database to be used.
+     */
+    void set_database_path(const std::string & dbpath);
```
ayshtmr: Yes, I need a `WritableDatabase` to be able to save model as db metadata, and `Database` for everything else.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -136,46 +132,56 @@ ListNETRanker::train_model() {
...
-ListNETRanker::save_model_to_file(const char* output_filename) {
-    LOGCALL_VOID(API, "ListNETRanker::save_model_to_file", output_filename);
-    ofstream parameters_file;
-    parameters_file.open(output_filename);
+ListNETRanker::save_model_to_metadata(const string & model_key) {
+    LOGCALL_VOID(API, "ListNETRanker::save_model_to_file", model_key);
+    Xapian::WritableDatabase letor_db(get_database_path());
+    string key = model_key;
+    if (key.empty()) {
+	key = "LisntNET.model.default";
```
ayshtmr: Yep, will fix.

xapian-letor/ranker/ranker.cc
ayshtmr: This one is part of `score(*)` function and the previous one was in `prepare_training_file(*)`. Both of them do very similar things, but the `score(*)` function here is a `Ranker` member function and encloses a `Scorer` instance to be able to check the quality of ranking.

xapian-letor/ranker/ranker.cc
ayshtmr: @ojwb You are correct to point out that it won't work for non-trivial cases, and I should use a default prefix. However, what I want to do over here is to parse `querystr` with `title` as default prefix, and with no default prefix (i.e. normal parsing) all in one go, since even if I parse them independently, there doesn't seem to be a way to combine two `Xapian::Query`.

ayshtmr: I can generate two `MSet`s, one for `Query` with default prefix set and one for without, but in that case, there will be extra `FeatureVector`s written to training file, since each feature calculates value 'with title', 'body only' and 'both body and title'. So, 'both body and title' feature value will be written off twice for the same query string in query file.

jaylett: You can combine query objects using `Xapian::Query(op, q1, q2)`; which operator makes sense will depend a little on how the relevant features are actually defined. (`::OP_AND` or `::OP_OR` are probably good candidates, depending…)

ayshtmr:
Thanks. I was looking at the documentation for Xapian::Query in apidoc.
Could find the method in sourcedoc.

jaylett:
Current API docs aren't listing undocumented methods, which I think is a victim of recent doxygen versions. If you change `HIDE_UNDOC_MEMBERS` to `NO` in `xapian-core/docs/doxygen_api.conf.in` then you can get everything listed. I haven't applied this to trunk yet, because there are some questions about whether this is the right choice.

ayshtmr:
Pushed the changes.

xapian-letor/ranker/ranker.cc
ojwb: The other way to achieve this is to just specify both prefixes to the `QueryParser`:

```
qp.add_prefix(string(), string());
qp.add_prefix(string(), string("S"));
```

If you do that and parse `hello world` you get:

```
Query(((Zhello@1 OR ZShello@1) AND (Zworld@2 OR ZSworld@2)))
```

whereas if you parse twice with a different default prefix each time and then use `OP_OR` to combine the results you get:

```
Query(((Zhello@1 AND Zworld@2) OR (ZShello@1 AND ZSworld@2)))
```

So which approach to take depends on whether you want the match to be entirely within one field, or if it's fine for different words to match in different fields.

xapian-letor/ranker/ranker.cc
ojwb: That doesn't explain why there's a large piece of repeated code - the code to parse the query could be in a separate function called from each.

xapian-letor/ranker/ranker.cc
ayshtmr: With `OP_OR` set as default op for `QueryParser` it will get parsed as 
``
Query(((Zhello@1 OR Zworld@2) OR (ZShello@1 OR ZSworld@2)))
``
which is what I think is required here

xapian-letor/ranker/ranker.cc
ayshtmr: Pushed the change in latest commit. You can review.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -151,7 +153,7 @@ ListNETRanker::save_model_to_metadata(const string & model_key) {
     }
     ostringstream oss;
     for (size_t i = 0; i <  parameters.size(); ++i)
-	oss << setprecision(numeric_limits<double>::digits10) << parameters[i] << endl;
+	oss << serialise_double(parameters[i]) << endl;
```
ojwb: You really just need to concatenate the output of `serialise_double()` into a string.  No need to the `ostringstream` (which is terribly slow IME) and no need to put in newlines.

```
string model_data;
for (size_t i = 0; i < parameters.size(); ++i)
    model_data += serialise_double(parameters[i];
```

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -151,7 +153,7 @@ ListNETRanker::save_model_to_metadata(const string & model_key) {
     }
     ostringstream oss;
     for (size_t i = 0; i <  parameters.size(); ++i)
```
ojwb: There are two spaces after the `<` here.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -173,7 +175,9 @@ ListNETRanker::load_model_from_metadata(const string & model_key) {
     istringstream model_str(model_data);
     string line;
     while (getline(model_str, line)) {
-	double parameter = stod(line);
+	const char *ptr = line.data();
+	const char *end = ptr + line.size();
+	double parameter = unserialise_double(&ptr, end);
 	loaded_parameters.push_back(parameter);
     }
```
ojwb: And here, no need to read line-by-line or use `istringstream` at all - just unserialise directly from `model_str`:

```
const char *ptr = model_str.data();
const char *end = ptr + model_str.size();
while (ptr != end)
    loaded_parameters.push_back(unserialise_double(&ptr, end));
```

xapian-letor/include/xapian-letor/ranker.h
ojwb: Hmm, I wonder if there's a better way.  I'm OK with this for now anyway.

ojwb:
Thanks for all the updates - looking good to me aside from the serialise_double-relate comments.

ayshtmr:
@ojwb Pushed the serialise-double changes.

ojwb:
Thanks - auto-squashed and merged to git master.

PR #126, opened 2016-11-06
python bindings: sphinx-build is an executable

dunn:
sphinx-build can be run directly (it is a Python executable with a
shebang), so it is sometimes replaced by a Bash wrapper script.  Trying
to run it as a Python script will break in those situations.

See e.g. https://github.com/Homebrew/homebrew-core/issues/6660#issuecomment-258699832

### DIFF
--- a/xapian-bindings/python/Makefile.am
+++ b/xapian-bindings/python/Makefile.am
@@ -163,4 +163,4 @@ install-data-local:
 all-local: $(sphinxdocs)
 
 $(sphinxdocs): xapian/__init__.py xapian/_xapian$(PYTHON2_SO) docs/conf.py $(RST_DOCS) $(dist_exampledata_DATA)
-	PYTHONPATH=..:$$PYTHONPATH $(PYTHON2) $(SPHINX_BUILD) -b html -d doctrees -c docs $(srcdir)/docs docs/html
+	PYTHONPATH=..:$$PYTHONPATH $(SPHINX_BUILD) -b html -d doctrees -c docs $(srcdir)/docs docs/html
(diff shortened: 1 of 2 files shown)


jaylett:
I believe we originally did this because of problems with running (say) sphinx under python2 against the python3 bindings (which includes python3 code). However it's possible that the generated code as it currently stands doesn't have that problem (the tests passing does suggest that), in which case we can drop the indirection. (We may be able to tidy up some of the configure mechanism at the same time; I have vague memories of that too.) Requires a bit more testing on platforms not covered by Travis, I'd suggest. (For homebrew you could happily carry the diff over the 1.4.1 release.)

ojwb:
There's no testing of the **content** of the sphinx-generated docs, so CI passing really only means that sphinx didn't throw an error generating them.  We need to compare the output from python2 sphinx and python3 sphinx in each case to ensure that this change doesn't introduce issues.

If the patch here isn't suitable, I guess we can do something like this to ensure we get the Sphinx for the appropriate Python version:

```
$(PYTHON2) -c 'import sphinx,sys;sys.exit(sphinx.main(sys.argv))'
```

We do something similar for the Perl bindings instead of running `pod2html` directly, though the reasons aren't exactly the same.

dunn:
> I believe we originally did this because of problems with running (say) sphinx under python2 against the python3 bindings (which includes python3 code).

Yeah, that makes sense.

Homebrew is unwilling to carry patches that aren't eventually going to be merged upstream, so can we expect some sort of patch like this to be merged, or will Sphinx perhaps be dropped altogether per https://trac.xapian.org/ticket/740 ?

dunn:
@ojwb Sorry, commented before I refreshed the page.

Either solution sounds good! I don't mind if this PR is replaced with something better.

ojwb:
@dunn We'll definitely do something so it works when `sphinx-build` isn't a Python script - hopefully in 1.4.2.

ojwb:
I checked on a machine where `/usr/bin/sphinx-build` is Python 2 sphinx and the change here results in incorrect Python 3 docs.  The `import sphinx,sys;sys.exit(sphinx.main(sys.argv))'` approach produces docs that only have one minor difference in the generated `searchindex.js` (not sure why, maybe just some non-determinism in Sphinx?  It's not an important difference anyway).

dunn:
Thanks!

PR #127, opened 2016-11-27
Letor: Add SVMRanker

ayshtmr:
This PR adds `SVMRanker` to `xapian-letor`.

### DIFF
--- a/xapian-letor/ranker/svmranker.cc
+++ b/xapian-letor/ranker/svmranker.cc
@@ -1,0 +1,205 @@
+/** @file svmranker.cc
+ *  @brief Implementation of Ranking-SVM
+ */
+/* Copyright (C) 2012 Parth Gupta
+ * Copyright (C) 2016 Ayush Tomar
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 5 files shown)


xapian-letor/ranker/svmranker.cc
```diff
@@ -0,0 +1,205 @@
...
+    std::ostringstream ss;
+    ss << f.rdbuf();
+    // Save model string
+    this->model_data = ss.str();
+    // Delete the temporary model file
+    std::remove(tempname);
+}
+
+void
+SVMRanker::save_model_to_metadata(const string & model_key)
```
jaylett: Since writing to the database can raise all sorts of exceptions, we should either deal with the more obvious ones or explain in a comment why we don't have to worry.

ayshtmr:
@ojwb @jaylett clang build failed with `queryparser/termgenerator_internal.cc:748:22: error: no member named 'emplace'...`. gcc build went fine.

xapian-letor/ranker/svmranker.cc
ayshtmr: I couldn't understand what you mean here. Could you please explain in detail?

ayshtmr:
@jaylett Thanks for a prompt review. Yes, writing tests should be the next step now.

xapian-letor/ranker/svmranker.cc
jaylett: This is a protected method, so users aren't going to see it. Therefore it gets called by things that users called, which means (for instance) that it may be less likely that there's another writer to the database at the same time. If this is the case, then a comment here explaining why we're not bothering to handle exceptions would be appropriate. If not, I guess we should have something in the documentation for whatever public API call ends up calling this method saying that various exceptions might be raised, and in particular that you shouldn't have another writer on the same database while training.

ayshtmr: That makes sense. Thanks.

ayshtmr:
Just to remind - I think this PR is complete and can be merged.

ojwb:
Thanks - rebased, squashed and merged to git master.

PR #128, opened 2016-12-15
L wdf normalisation support in TfIdfWeight

ivmarkp:
This PR adds support for the L wdf normalisation in TfIdfWeight. 
* Code builds successfully.
* All tests pass.

### DIFF
--- a/xapian-core/weight/tfidfweight.cc
+++ b/xapian-core/weight/tfidfweight.cc
@@ -148,8 +148,7 @@ TfIdfWeight::get_maxpart() const
 {
     Xapian::termcount wdf_max = get_wdf_upper_bound();
     Xapian::termcount len_min = get_doclength_lower_bound();
-    Xapian::termcount len_max = get_doclength_upper_bound();
-    double wdfn = get_wdfn(wdf_max, len_min, len_max, normalizations[0]);
+    double wdfn = get_wdfn(wdf_max, len_min, len_min, normalizations[0]);
     return get_wtn(wdfn * idfn, normalizations[2]) * wqf_factor;
 }
 
--- a/xapian-core/weight/tfidfweight.cc
+++ b/xapian-core/weight/tfidfweight.cc
@@ -83,13 +83,11 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
 	need_stat(DOC_LENGTH);
 	need_stat(DOC_LENGTH_MIN);
     }
-    else if (normalizations[0] == 'L') {
+    if (normalizations[0] == 'L') {
 	need_stat(DOC_LENGTH);
 	need_stat(DOC_LENGTH_MIN);
 	need_stat(DOC_LENGTH_MAX);
 	need_stat(UNIQUE_TERMS);
-	if (normalizations[1] == 'P')
-	    need_stat(AVERAGE_LENGTH);
     }
 }
 
--- a/xapian-core/weight/tfidfweight.cc
+++ b/xapian-core/weight/tfidfweight.cc
@@ -51,10 +51,12 @@ TfIdfWeight::TfIdfWeight(const std::string &normals)
     need_stat(WDF);
     need_stat(WDF_MAX);
     need_stat(WQF);
-    need_stat(DOC_LENGTH);
-    need_stat(DOC_LENGTH_MIN);
-    need_stat(DOC_LENGTH_MAX);
-    need_stat(UNIQUE_TERMS);
+    if (normalizations[0] == 'L') {
+	need_stat(DOC_LENGTH);
+	need_stat(DOC_LENGTH_MIN);
+	need_stat(DOC_LENGTH_MAX);
+	need_stat(UNIQUE_TERMS);
+    }
 }
 
 TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
@@ -76,12 +78,18 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
     need_stat(WDF);
     need_stat(WDF_MAX);
     need_stat(WQF);
-    need_stat(DOC_LENGTH);
-    need_stat(DOC_LENGTH_MIN);
-    need_stat(DOC_LENGTH_MAX);
-    need_stat(UNIQUE_TERMS);
     if (normalizations[0] == 'P' || normalizations[1] == 'P') {
 	need_stat(AVERAGE_LENGTH);
+	need_stat(DOC_LENGTH);
+	need_stat(DOC_LENGTH_MIN);
+    }
+    else if (normalizations[0] == 'L') {
+	need_stat(DOC_LENGTH);
+	need_stat(DOC_LENGTH_MIN);
+	need_stat(DOC_LENGTH_MAX);
+	need_stat(UNIQUE_TERMS);
+	if (normalizations[1] == 'P')
+	    need_stat(AVERAGE_LENGTH);
     }
 }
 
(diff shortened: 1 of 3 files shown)


ivmarkp:
`smoketest.py`failing... not sure why. I didn't touch that part of the code. Although, all `xapian-core` tests pass. Anyway, trying to figure out what went wrong.

ojwb:
Python tests fixed in git master by 9cd1712209aa32ad53dfa0ab834820331314c3ec.

xapian-core/weight/tfidfweight.cc
```diff
@@ -51,13 +51,17 @@ TfIdfWeight::TfIdfWeight(const std::string &normals)
     need_stat(WDF);
     need_stat(WDF_MAX);
     need_stat(WQF);
+    need_stat(DOC_LENGTH);
+    need_stat(DOC_LENGTH_MIN);
+    need_stat(DOC_LENGTH_MAX);
+    need_stat(UNIQUE_TERMS);
 }
```
ivmarkp: Yes, making the changes. Should've been more careful with this part.

xapian-core/weight/tfidfweight.cc
```diff
@@ -136,7 +142,8 @@ TfIdfWeight::get_maxpart() const
 {
     Xapian::termcount wdf_max = get_wdf_upper_bound();
     Xapian::termcount len_min = get_doclength_lower_bound();
-    double wdfn = get_wdfn(wdf_max, len_min, normalizations[0]);
+    Xapian::termcount len_max = get_doclength_upper_bound();
+    double wdfn = get_wdfn(wdf_max, len_min, len_max, normalizations[0]);
     return get_wtn(wdfn * idfn, normalizations[2]) * wqf_factor;
```
ivmarkp: I understand what you're saying. I'm wondering if there is a simpler way of pulling `uniqterms` stat here?

As of now I can't see how to pull `uniqterms` stat in this `double get_maxpart()` method since it lacks the required parameters. So, we'd need to change its declaration in `weight.h` to provide `Xapian::termcount uniqterms` parameters which would further require changing its definition in all the weighting schemes subclasses.

ivmarkp:
Thanks Olly, all tests now pass on gcc build job. However, clang compilation is failing; probably due to older compiler version (not sure though).

xapian-core/weight/tfidfweight.cc
ojwb: `get_maxpart()` needs to return an upper bound over all documents in the collection, but `uniqterms` varies per document.  Note how we use the **bounds** on `doclen`, we don't pass in `doclen` itself because there is no appropriate single `doclen` value - the same applies to `uniqterms`.

We don't track bounds on `uniqterms` currently, but I don't think we can do better than evaluating at `doclen == uniqterms` unless `doclen_min >= uniqterms_max`.

(Actually, it's not quite true that `uniqterms <= doclen` since a term in a document can have wdf=0 but such terms would still count in the unique count - this is an esoteric corner case, and probably best dealt with by setting `doclen = max(doclen, uniqterms)` (or something equivalent) when calculating the `L` norm value).

xapian-core/weight/tfidfweight.cc
```diff
@@ -76,12 +78,18 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
...
+	need_stat(DOC_LENGTH);
+	need_stat(DOC_LENGTH_MIN);
+    }
+    else if (normalizations[0] == 'L') {
+	need_stat(DOC_LENGTH);
+	need_stat(DOC_LENGTH_MIN);
+	need_stat(DOC_LENGTH_MAX);
+	need_stat(UNIQUE_TERMS);
+	if (normalizations[1] == 'P')
+	    need_stat(AVERAGE_LENGTH);
```
ojwb: I would just make this a separate `if (normalizations[0] == 'L') {` and drop the duplicated handling for `normalizations[1] == 'P'` from it.  That way the knowledge about `(normalizations[1] == 'P')` isn't duplicated (https://en.wikipedia.org/wiki/Don't_repeat_yourself).

The implementation of `need_stat(flag);` is carefully written to just inline `stats_needed = stat_flags(stats_needed | flag);` so a contiguous block of such calls with constant values should optimise down to a single bitwise or with a constant.  So this code is really:

```
if (normalizations[0] == 'P' || normalizations[1] == 'P') {
    stats_needed |= SOME_CONSTANT1;
} else if (normalizations[0] == 'L') {
    stats_needed |= SOME_CONSTANT2;
    if (normalizations[1] == 'P')
        stats_needed |= SOME_CONSTANT3;
}
```

And my suggested version would be:

```
if (normalizations[0] == 'P' || normalizations[1] == 'P') {
    stats_needed |= SOME_CONSTANT1;
}
if (normalizations[0] == 'L') {
    stats_needed |= SOME_CONSTANT2;
}
```

ivmarkp: With this structure of conditionals, I was actually trying to avoid making duplicate `need_stat()` calls specifically in case of `LPx` as `L` and `P` require two common stats i.e. `DOC_LENGTH` and `DOC_LENGTH_MIN`.

I think having two independent `if` conditionals would mean making a couple extra duplicate calls for the common stats in case of `LPx`.

xapian-core/weight/tfidfweight.cc
ivmarkp: Sorry, I don't quite get it. I'm unclear on what values should be passed instead of the current `len_min` and `len_max` parameter values?

xapian-core/weight/tfidfweight.cc
ojwb: But as I noted above, `need_stat()` just does a bitwise or, it doesn't go and fetch or calculate the statistic. So those "extra duplicate calls" only means the bit for `AVERAGE_LENGTH` is set in `SOME_CONSTANT1` and `SOME_CONSTANT2`.

Once the compiler has inlined and optimised the code, from an efficiency point of view what matters is not the number of calls to `need_stat()` in the source code, but how many separate runs of such calls.

ojwb:
> However, clang compilation is failing; probably due to older compiler version (not sure though).

Yeah, the clang it's using is a rather old version (older than would bother to support) and doesn't have full C++11 library support - I've not had a chance to look at how to fix that.  Just ignore that error for now.

xapian-core/weight/tfidfweight.cc
ivmarkp: @ojwb I'd appreciate if you could clarify a bit more based on my previous comment? Thanks!

ojwb: I think you just want to pass `len_min` for both.  For the 'P' case we want to minimise `doclen`, so we have to pass `len_min` for `doclen`.  Only 'L' currently looks at `uniqterms` and for that we just need to pass the same value for `doclen` and `uniqterms` - for 'L'  `get_wdfn()` only divides one by the other, so it only matters that they are equal (and non-zero).

(More generally, I doubt there will always be single a set of parameters we can just call `get_wdfn()` with to get an upper bound for any plausible wdf norm, but for the set of wdf norms implemented so far it seems there is).

ojwb:
Thanks, rebased and pushed to master.

ojwb:
Oh, this didn't get attended to:

> (Actually, it's not quite true that uniqterms <= doclen since a term in a document can have wdf=0 but such terms would still count in the unique count - this is an esoteric corner case, and probably best dealt with by setting doclen = max(doclen, uniqterms) (or something equivalent) when calculating the L norm value).

Having thought about this some more, I think this is a bug in the current implementation of `get_unique_terms()`, which just returns the length of the termlist (which includes boolean terms too).  It really should only count terms with `wdf > 0`.  To do that properly would be much slower, so for now I've pushed 1b4ab656c277488e4f7b0208443d75aad4b8faa3 which simply clamps `get_unique_terms()` to be <= the document length.

PR #130, opened 2017-01-09
Add EMPTY_NOMATCH flag to return empty string for non-matching snippets

rsto:
Currently, if MSet::snippet() can't find any matching snippet term it returns the first n bytes of the input text. This patch adds an optional flag to the  snippet generator to return the empty string if no match was found. The current behaviour remains default.

### DIFF
--- a/xapian-core/tests/api_snippets.cc
+++ b/xapian-core/tests/api_snippets.cc
@@ -274,3 +274,42 @@ DEFINE_TESTCASE(snippet_termcover2, backend) {
 
     return true;
 }
+
+/// Test snippet EMPTY_WITHOUT_MATCH flag
+DEFINE_TESTCASE(snippet_empty, backend) {
+
+    Xapian::Stem stem("en");
+
+    Xapian::Enquire enquire(get_database("apitest_simpledata"));
+    enquire.set_query(Xapian::Query(Xapian::Query::OP_OR,
+                Xapian::Query("rubbish"),
+                Xapian::Query("Zexampl")));
+
+    Xapian::MSet mset = enquire.get_mset(0, 0);
+
+    // A non-matching text
+    const char *input = "A string without a match.";
+    size_t len = strlen(input);
+
+    // By default, snippet() returns len bytes of input without markup
+    unsigned flags = 0;
+    TEST_STRINGS_EQUAL(mset.snippet(input, len, stem, 0), input);
+
+    // force snippet() to return the empty string if no term got matched
+    flags |= Xapian::MSet::SNIPPET_EMPTY_WITHOUT_MATCH;
+    TEST_STRINGS_EQUAL(mset.snippet(input, len, stem, flags), "");
+
+    // A text with a match
+    input = "A rubbish example text";
+    len = strlen(input);
+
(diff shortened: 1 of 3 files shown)


xapian-core/tests/api_snippets.cc
```diff
@@ -260,3 +260,45 @@ DEFINE_TESTCASE(snippet_termcover1, backend) {
...
     return true;
 }
+
+/// Test snippet nomatch flag
+DEFINE_TESTCASE(snippet_nomatch, backend) {
+
+    Xapian::Stem stem("en");
+    const char *input;
+    size_t len;
+    unsigned flags;
```
ojwb: Declaring variables at the point of initialisation is better C++ practice.

xapian-core/tests/api_snippets.cc
```diff
@@ -260,3 +260,45 @@ DEFINE_TESTCASE(snippet_termcover1, backend) {
...
+                Xapian::Query("Zexampl")));
+
+    Xapian::MSet mset = enquire.get_mset(0, 0);
+
+    // A non-matching text
+    input = "A string without a match.";
+    len = strlen(input);
+
+    flags = 0;
+    // Regression: snippet() returns len bytes of input without markup
```
ojwb: How is that a regression?

rsto:
Thanks for the feedback. I uploaded an updated version.

Do you know how I can make it pass Travis checks? Travis fails again while building xapian-letor [here](https://travis-ci.org/xapian/xapian/jobs/191582880#L3553), and this doesn't break on my local machine.

ojwb:
I've pushed the commit linked above which should solve the travis failure.

[I typed the below at the time of my previous comments, but it seems I failed to actually click the "Comment" button]

> Currently, if MSet::snippet() can't find any matching snippet term it returns the first n bytes of the input text.

Actually, it'll pick (up to) n "interesting" bytes with flag `SNIPPET_BACKGROUND_MODEL`, which is on by default.

I agree that having a flag for returning no snippet for this case would be useful, but the `EMPTY_NOMATCH` is inconsistently named as "nomatch" isn't one word.  `EMPTY_NO_MATCH` would be better, though I wonder if there's a name which would make the purpose clearer.

rsto:
I've reworded  EMPTY_NOMATCH to SNIPPET_EMPTY_WITHOUT_MATCH, which hopefully makes the purpose of this flag more clear.

I am still encountering Travis build issues and could fix one of them, but doxygen crashes for the clang image. I've opened a separate pull request for this: https://github.com/xapian/xapian/pull/132

ojwb:
Oh yes, that seems clearer without being excessively long.

I just clicked the button to make it restart the OS X build, but I guess it'll just use the config on the branch and so fail again.  Assuming it fails again, can you rebase the branch on current master?

rsto:
Done. It passes Travis now.

ojwb:
Great that the CI works again.  Thanks for your work on that.

Rebased onto current master.  I fixed up some indentation issues in 959fe2f09fdb - maybe you need to configure your editor to tab fill indentation with tab width 8?

PR #131, opened 2017-01-14
Fix compilation on Darwin

tychobrailleur:
OSX doesn't have the `exp10` function.

### DIFF
--- a/xapian-core/api/omenquire.cc
+++ b/xapian-core/api/omenquire.cc
@@ -44,6 +44,8 @@
 #include "str.h"
 #include "weight/weightinternal.h"
 
+#include "exp10.h"
+
 #include <algorithm>
 #include "autoptr.h"
 #include <cfloat>
@@ -255,7 +257,7 @@ MSet::get_matches_estimated() const
 	return e;
     }
 
-    Xapian::doccount r = Xapian::doccount(exp10(int(log10(D))) + 0.5);
+    Xapian::doccount r = Xapian::doccount(EXP10(int(log10(D))) + 0.5);
     while (r > e) r /= 10;
 
     Xapian::doccount R = e / r * r;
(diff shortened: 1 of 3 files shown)


jaylett:
Darwin does however have `__exp10()`, which "may be computed more efficiently than `pow(10,x)`". If this is important, we could find the appropriate function at configure time. (`__exp10()` is in `math.h` on Darwin.)

tychobrailleur:
@jaylett Here is my clumsy attempt at selecting the correct one.  I have no autoconf experience, so this may be somewhat misguided.

jaylett:
@tychobrailleur thanks for the rapid feedback — I'll leave @ojwb to assess whether this is the right approach from an autoconf pov (it isn't the way I'd have done it, but that doesn't necessarily mean it isn't reasonable!).

ojwb:
Thanks for raising this - seems `exp10()` is a GNU libc extension - sorry, I just assumed it was one of the "missing" math.h functions added by C99 and so available with C++11.

This use of `exp10()` isn't performance sensitive (it will generally be called at most once per search, unless you call `MSet::get_matches_estimated()` multiple times) and if we wanted to optimise this code, then `exp10(int(log10(x))` is probably better done in integer arithmetic, e.g. along the lines of: https://graphics.stanford.edu/~seander/bithacks.html#IntegerLog10Obvious or one of the other tricks in that document.

But it's probably worth probing for since a number of common platforms provide an implementation - it'll mean future code needing to calculate `exp10()` can benefit, and a general `pow()` function is quite a complex beast which can take a surprisingly long time - e.g. the Linux man page for `pow` tersely notes "On 64-bits, pow() may be more than 10,000 times slower for some (rare) inputs than for other nearby inputs".

The more natural autoconf check would be:

```
AC_CHECK_FUNCS([exp10 __exp10], [break])
```

And then check for `defined AC_HAVE_EXP10` and `defined AC_HAVE___EXP10` which `AC_CHECK_FUNCS` defines for you.

GCC also provides `__builtin_exp10()`, but clang doesn't appear to.  I've pushed a commit which checks for all three.

tychobrailleur:
Perfect!  Thanks @ojwb for looking into this.

PR #132, opened 2017-01-16
Fix Travis builds for clang configuration

rsto:
The libsvm3 package has been upgraded from 3.21 to 3.22 for the clang build image, so /tmp/xapian-libsvm-fixed-include/libsvm needs to point to the new version (https://travis-ci.org/xapian/xapian/jobs/192285478#L2315).

Fixing this, I see Travis fail with a doxygen segmentation fault (e.g. https://travis-ci.org/xapian/xapian/jobs/192290918#L5247), so this PR might still require more commits to make Travis pass clang builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,8 @@ matrix:
     - os: osx
       before_install:
         - brew update
-        - brew install doxygen help2man graphviz pngcrush libmagic pcre libsvm lua mono python3
+        - brew install doxygen --HEAD
+        - brew install help2man graphviz pngcrush libmagic pcre libsvm lua mono python3
         - pip install sphinx docutils
         - pip3 install sphinx
         - mkdir -p /tmp/xapian-libsvm-fixed-include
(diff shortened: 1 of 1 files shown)


rsto:
Unfortunately, I'm at loss here. For debugging I tried to downgrade brew's doxygen version to 1.8.12 using this Gist (https://gist.github.com/gcatlin/1847248), but didn't get far since the "versions" command is unknown to the brew command in Travis. Not having a local system with homebrew doesn't help either.

Is there a sane way to debug this, e.g. is there a Travis system with shell access for this kind of issues?

rsto:
Thanks to @tychobrailleur for pointing me at the workaround for doxygen.

ojwb:
Oddly https://travis-ci.org/xapian/xapian/jobs/192290918#L5247 links to a log which doesn't have 5247 lines, but that log does show a segfault generating XML from doxygen.  Looks like it's with 1.8.13, but Debian's 1.8.13-4 works - it has an extra patch for https://bugzilla.gnome.org/show_bug.cgi?id=776791 which is presumably the issue we're hitting here.

So it looks like stock doxygen 1.8.13 is a problem with xapian-bindings (but only in maintainer mode).

Thanks for patch, will merge shortly.

ojwb:
Merged, thanks.

I also tweaked it to avoid hard-coding the libsvm version, though I can't fully test that.

PR #134, opened 2017-01-22
Implemented FileSizeRangeProcessor Class

shivanshuiitg:
Hi, I have implemented FileSizeRangeProcessor to handle the ranges for file size. Following files have been modified:-
	1) xapian-core/api/valuerangeproc.cc
	2) xapian-core/include/xapian/queryparser.h

It will handle ranges like, "size:10K..1M".
I am working on the tests and its use in omega.

### DIFF
--- a/xapian-applications/omega/query.cc
+++ b/xapian-applications/omega/query.cc
@@ -112,8 +112,6 @@ static Xapian::Query query;
 Xapian::Query::op default_op = Xapian::Query::OP_AND; // default matching mode
 
 static Xapian::QueryParser qp;
-//static Xapian::NumberRangeProcessor * size_rp = NULL;
-// Using FileSizeRangeProcessor instead of NumberRangeProcessor
 static Xapian::FileSizeRangeProcessor * size_rp = NULL;
 static Xapian::Stem *stemmer = NULL;
 
@@ -321,9 +319,7 @@ set_probabilistic(const string &oldp)
     qp.set_stopper(new MyStopper());
     qp.set_default_op(default_op);
     qp.set_database(db);
-    // FIXME: provide a custom RP which handles size:10..20K, etc.
     if (!size_rp){
-    //size_rp = new Xapian::NumberRangeProcessor(VALUE_SIZE, "size:");
     size_rp = new Xapian::FileSizeRangeProcessor(VALUE_SIZE, "size:");
 	}
     qp.add_rangeprocessor(size_rp);
--- a/xapian-applications/omega/query.cc
+++ b/xapian-applications/omega/query.cc
@@ -112,7 +112,9 @@ static Xapian::Query query;
 Xapian::Query::op default_op = Xapian::Query::OP_AND; // default matching mode
 
 static Xapian::QueryParser qp;
-static Xapian::NumberRangeProcessor * size_rp = NULL;
+//static Xapian::NumberRangeProcessor * size_rp = NULL;
+// Using FileSizeRangeProcessor instead of NumberRangeProcessor
+static Xapian::FileSizeRangeProcessor * size_rp = NULL;
 static Xapian::Stem *stemmer = NULL;
 
 static string eval_file(const string &fmtfile);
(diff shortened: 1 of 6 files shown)


ojwb:
The CI (Continuous Integration) tests are failing - if you look at the logs (e.g. https://travis-ci.org/xapian/xapian/jobs/194232409) then it's because there's a new API class which needs to be added to the list of generated Java classes in `xapian-bindings/java/Makefile.am`, and similarly for csharp.

shivanshuiitg:
FileSizeRangeProcessor added to the list of generated Java classes in xapian-bindings/java/Makefile.am and csharp classes in xapian-bindings/csharp/Makefile.am

jaylett:
@shivanshuiitg we're still trying to get this hooked up correctly in travis, but one of the tests looks for common formatting problems, and [is throwing up some for your code](https://travis-ci.org/xapian/xapian/jobs/194369395).

There are some [further hints on coding style](https://github.com/xapian/xapian/blob/master/xapian-core/HACKING#L1216) in `xapian-core/HACKING`, which you should look through anyway because it gives useful information on things like the test suite and debugging.

There's also some useful information in our [developer guide](https://xapian-developer-guide.readthedocs.io/en/latest/), although since this is very early days for it, it hasn't yet started to get all the information from `HACKING` and elsewhere. However it does talk about what makes a good commit message (and a good commit), and I'd like to see a better one here. You've included useful information in the PR comment, but that generally does not get preserved in the git history, and in any case should be in the comment for the commit that makes the change, since that makes tools like `git blame` and `git bisect` easier to use.

Your second commit should also have been a "fixup" commit of the previous one (which you can make in this case using `git commit --fixup a051a91`), which does two helpful things: it indicates to reviewers and collaborators that you intend it to improve the previous commit rather than stand alone as its own change; and it makes it very easy to merge these commits together before we finally commit, using `git rebase`. There's a useful link on this in the developer guide.

jaylett:
You still haven't got all the problems that CI throws up, and there are others as well that it doesn't check for (in particular I spotted lack of spaces around the parenthesised expression for `if`). Also, this again should have been a fixup commit. (I suggest you rebase the whole lot to two commits at this point, one for the xapian-core change and one to introduce it into omega. Then you can add tests and documentation as fixups to those two commits.)

You can run the CI check yourself by doing (in this case) `git diff 7a4d4f5f597aba1c625014220856fe5bc786bcc0 | ./xapian-maintainer-tools/xapian-check-patch`. However it's a good idea to also go over the diff yourself manually, checking for things like changes in indentation, and looking for places in which your code doesn't look like the other code around it.

ojwb:
> You can run the CI check yourself by doing (in this case) git diff 7a4d4f5f597aba1c625014220856fe5bc786bcc0 | ./xapian-maintainer-tools/xapian-check-patch. 

Here's a handy tip to avoid having to look up the commit hash to diff against - you can diff against where the current branch branched from `master` like so (note **three** dots):

```
git diff master... | ./xapian-maintainer-tools/xapian-check-patch
```

This is essentially what the CI config in `.travis.yml` does when it runs this script, except it also tries to work out what branch the PR was is based on.  Generally xapian PRs will be against master though.

It's a good idea to run it by hand - that way you can find errors more quickly than by pushing and waiting.  Also, the script currently only reports one issue for each source line to try to avoid a deluge of messages, but that means that fixing the list you see in CI may not result in a clean run.

> However it's a good idea to also go over the diff yourself manually, checking for things like changes in indentation, and looking for places in which your code doesn't look like the other code around it.

Indeed - the script spots many common issues, but it isn't a complete check.  You can't just pound away at the keyboard without a thought for the expected coding standards and then expect to run a magic script, fix the reported issues and have a clean patch.

One particularly notable omission is that it doesn't check indentation of each line relative to the previous line yet.

xapian-applications/omega/query.cc
```diff
@@ -112,6 +112,11 @@ static Xapian::Query query;
 Xapian::Query::op default_op = Xapian::Query::OP_AND; // default matching mode
 
 static Xapian::QueryParser qp;
+<<<<<<< HEAD
```
shivanshuiitg: This issue is because I forgot to save the file after making change.

jaylett:
Work moved to #135.

PR #135, opened 2017-01-25
Implement FileSizeRangeProcessor class

shivanshuiitg:
Hi, I have implemented FileSizeRangeProcessor for handling custom range in Omega. 
I was somehow confused with structure of previous PR "Implemented FileSizeRangeProcessor Class". So, I will be closing that PR and rest all work will be done on this PR.

It will handle ranges like 10..20K

### DIFF
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -268,6 +268,12 @@ testcase '1.000000.' B=Hexample.org B=Hexample.net
 testcase '2.000000.' B=Hexample.org B=XFOObar
 testcase '3.000000.' B=Hexample.org B=XFOObar B=XFOObar
 
+# size-range search test for omega
+printf '$querydescription' > "$TEST_TEMPLATE"
+qtestcase '0 * VALUE_RANGE 2 10240 20480' P='size:10..20K'
+qtestcase '0 * VALUE_RANGE 2 10240 30721' P='size:10K..30K'
+qtestcase '0 * VALUE_RANGE 2 10240 40960' P='size:10..40'
+
 printf '$filters' > "$TEST_TEMPLATE"
 testcase 'Hexample.net~Hexample.org~.~~' B=Hexample.org B=Hexample.net
 testcase '!Gmisc.test~.~~' N=Gmisc.test
--- a/xapian-core/api/valuerangeproc.cc
+++ b/xapian-core/api/valuerangeproc.cc
@@ -470,7 +470,7 @@ Xapian::Query
 FileSizeRangeProcessor::operator()(const string& b, const string& e){
 	// Here b and e will be like "100K" and "1M"
 	double size_b, size_e;
-	char unit_b, unit_e;
+	char unit_b = '\0', unit_e = '\0';
 	string temp_b = b, temp_e = e;
 
 	if (!b.empty()) {
@@ -516,11 +516,11 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e){
 	} else {
 		size_e = 0.0;
 	}
-	if (unit_e.empty()) {
+	if (unit_e == '\0') {
 		// 1M..5 is not valid. So, if e does not have unit, then invalid
 		goto not_our_range;
 	}
-	if (unit_b.empty()) {
+	if (unit_b == '\0') {
 		unit_b = unit_e;
 	}
 
--- a/xapian-core/api/valuerangeproc.cc
+++ b/xapian-core/api/valuerangeproc.cc
@@ -439,7 +439,7 @@ not_our_range:
     return Xapian::Query(Xapian::Query::OP_INVALID);
 }
 
-static const int size_map[26] = {
+static const unsigned int size_map[26] = {
     0,
     1, // B
     0, 0, 0, 0,
@@ -451,23 +451,23 @@ static const int size_map[26] = {
     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
 };
 
-static int
-get_bytes(char ch)
+static unsigned int
+get_size_multiplier_from_suffix(const char ch)
 {
 	// It takes character as input and return number of bytes
 	// B - 1 byte
 	// K - 1024 bytes
 	// M - 1024*1024 bytes
 	// G - 1024*1024*1024 bytes
-	if (ch >= 'A' && ch <= 'Z' && size_map[ch - 'A'] > 0) {
-    return size_map[ch - 'A'];
+	if (ch >= 'A' && ch <= 'Z') {
+		return size_map[ch - 'A'];
 	} else {
-	return 0;
+		return 0;
 	}
 }
 
 Xapian::Query
-FileSizeRangeProcessor::operator()(const string& b, const string& e){
+FileSizeRangeProcessor::operator()(const string& b, const string& e) {
 	// Here b and e will be like "100K" and "1M"
 	double size_b, size_e;
 	char unit_b = '\0', unit_e = '\0';
@@ -480,7 +480,7 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e){
 			// If suffix of b is 'B','K','M','G'
 			unit_b = b_back;
 			temp_b.pop_back();
-		} else if (!isdigit(b_back)) {
+		} else if (!C_isdigit(b_back)) {
 			// If it is neither digit nor any of above character, then it is invalid
 			goto not_our_range;
 		}
@@ -502,7 +502,7 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e){
 			// If suffix of b is 'B','K','M','G'
 			unit_e = e_back;
 			temp_e.pop_back();
-		} else if(!isdigit(e_back)) {
+		} else if(!C_isdigit(e_back)) {
 			// If it is neither digit nor any of above character, then it is invalid
 		    goto not_our_range;
 		}
@@ -516,6 +516,11 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e){
 	} else {
 		size_e = 0.0;
 	}
+	if (unit_e == '\0' && unit_b == '\0'){
+		// If 10..20 then it means 10 to 20 bytes
+		unit_b = 'B';
+		unit_e = 'B';
+	}
 	if (unit_e == '\0') {
 		// 1M..5 is not valid. So, if e does not have unit, then invalid
 		goto not_our_range;
@@ -524,8 +529,8 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e){
 		unit_b = unit_e;
 	}
 
-	size_b = size_b * get_bytes(unit_b);
-	size_e = size_e * get_bytes(unit_e);
+	size_b = size_b * get_size_multiplier_from_suffix(unit_b);
+	size_e = size_e * get_size_multiplier_from_suffix(unit_e);
 
 	return RangeProcessor::operator()(
 	    b.empty() ? b : Xapian::sortable_serialise(size_b),
--- a/xapian-core/api/valuerangeproc.cc
+++ b/xapian-core/api/valuerangeproc.cc
@@ -466,6 +466,17 @@ get_size_multiplier_from_suffix(const char ch)
     }
 }
 
+static bool
+is_valid_size_character(const char ch)
+{
+    // Return true if ch is valid size character ('B','K','M','G')
+    if (ch =='B' || ch =='K' || ch =='M' || ch =='G') {
+	return true;
+    } else {
+	return false;
+    }
+}
+
 Xapian::Query
 FileSizeRangeProcessor::operator()(const string& b, const string& e) {
     // Here b and e will be like "100K" and "1M"
@@ -476,7 +487,7 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e) {
     if (!b.empty()) {
 	errno = 0;
 	char b_back = b.back();
-	if (b_back =='B' || b_back =='K' || b_back =='M' || b_back =='G') {
+	if (is_valid_size_character(b_back)) {
 	    // If suffix of b is 'B','K','M','G'
 	    unit_b = b_back;
 	    temp_b.pop_back();
@@ -498,11 +509,11 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e) {
     if (!e.empty()) {
 	errno = 0;
 	char e_back = e.back();
-	if (e_back =='B' || e_back =='K' || e_back =='M' || e_back =='G') {
+	if (is_valid_size_character(b_back)) {
 	    // If suffix of b is 'B','K','M','G'
 	    unit_e = e_back;
 	    temp_e.pop_back();
-	} else if(!C_isdigit(e_back)) {
+	} else if (!C_isdigit(e_back)) {
 	    // If it is neither digit nor any of above character, then it is invalid
 	    goto not_our_range;
 	}
@@ -514,9 +525,9 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e) {
 	    goto not_our_range;
 	}
     } else {
-		size_e = 0.0;
+	size_e = 0.0;
     }
-    if (unit_e == '\0' && unit_b == '\0'){
+    if (unit_e == '\0' && unit_b == '\0') {
 	// If 10..20 then it means 10 to 20 bytes
 	unit_b = 'B';
 	unit_e = 'B';
@@ -531,7 +542,7 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e) {
 
     size_b = size_b * get_size_multiplier_from_suffix(unit_b);
     size_e = size_e * get_size_multiplier_from_suffix(unit_e);
-
+    
     return RangeProcessor::operator()(
 	b.empty() ? b : Xapian::sortable_serialise(size_b),
 	e.empty() ? e : Xapian::sortable_serialise(size_e));
--- a/xapian-core/api/valuerangeproc.cc
+++ b/xapian-core/api/valuerangeproc.cc
@@ -362,35 +362,35 @@ DateRangeProcessor::operator()(const string& b, const string& e)
 
     {
 #ifdef SNPRINTF
-	char buf_b[9], buf_e[9];
-	if (!b.empty()) {
-	    SNPRINTF(buf_b, sizeof(buf_b), "%08d", b_y * 10000 + b_m * 100 + b_d);
-	} else {
-	    *buf_b = '\0';
-	}
-	if (!e.empty()) {
-	    SNPRINTF(buf_e, sizeof(buf_e), "%08d", e_y * 10000 + e_m * 100 + e_d);
-	} else {
-	    *buf_e = '\0';
-	}
+    char buf_b[9], buf_e[9];
+    if (!b.empty()) {
+	SNPRINTF(buf_b, sizeof(buf_b), "%08d", b_y * 10000 + b_m * 100 + b_d);
+    } else {
+	*buf_b = '\0';
+    }
+    if (!e.empty()) {
+	SNPRINTF(buf_e, sizeof(buf_e), "%08d", e_y * 10000 + e_m * 100 + e_d);
+    } else {
+	*buf_e = '\0';
+    }
 #else
-	char buf_b[100], buf_e[100];
-	buf_b[sizeof(buf_b) - 1] = '\0';
-	buf_e[sizeof(buf_e) - 1] = '\0';
-	if (!b.empty()) {
-	    sprintf(buf_b, "%08d", b_y * 10000 + b_m * 100 + b_d);
-	    if (buf_b[sizeof(buf_b) - 1]) abort(); // Buffer overrun!
-	} else {
-	    *buf_b = '\0';
-	}
-	if (!e.empty()) {
-	    sprintf(buf_e, "%08d", e_y * 10000 + e_m * 100 + e_d);
-	    if (buf_e[sizeof(buf_e) - 1]) abort(); // Buffer overrun!
-	} else {
-	    *buf_e = '\0';
-	}
+    char buf_b[100], buf_e[100];
+    buf_b[sizeof(buf_b) - 1] = '\0';
+    buf_e[sizeof(buf_e) - 1] = '\0';
+    if (!b.empty()) {
+	sprintf(buf_b, "%08d", b_y * 10000 + b_m * 100 + b_d);
+	if (buf_b[sizeof(buf_b) - 1]) abort(); // Buffer overrun!
+    } else {
+	*buf_b = '\0';
+    }
+    if (!e.empty()) {
+	sprintf(buf_e, "%08d", e_y * 10000 + e_m * 100 + e_d);
+	if (buf_e[sizeof(buf_e) - 1]) abort(); // Buffer overrun!
+    } else {
+	*buf_e = '\0';
+    }
 #endif
-	return RangeProcessor::operator()(buf_b, buf_e);
+    return RangeProcessor::operator()(buf_b, buf_e);
     }
 
 not_our_range:
@@ -432,8 +432,8 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
     }
 
     return RangeProcessor::operator()(
-	    b.empty() ? b : Xapian::sortable_serialise(num_b),
-	    e.empty() ? e : Xapian::sortable_serialise(num_e));
+	b.empty() ? b : Xapian::sortable_serialise(num_b),
+	e.empty() ? e : Xapian::sortable_serialise(num_e));
 
 not_our_range:
     return Xapian::Query(Xapian::Query::OP_INVALID);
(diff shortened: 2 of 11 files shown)


xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+static int
+get_bytes(char ch)
+{
+	// It takes character as input and return number of bytes
+	// B - 1 byte
+	// K - 1024 bytes
+	// M - 1024*1024 bytes
+	// G - 1024*1024*1024 bytes
+	if (ch >= 'A' && ch <= 'Z' && size_map[ch - 'A'] > 0) {
+    return size_map[ch - 'A'];
```
jaylett: The alignment of the if and else legs is wrong here, but in two different ways. One of them suggests that your tab width is set to 4 when it should be 8, the other suggests you aren't trying to indent the legs further, as:

```c
if (…) {
    return …;
}
```

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+
+static int
+get_bytes(char ch)
+{
+	// It takes character as input and return number of bytes
+	// B - 1 byte
+	// K - 1024 bytes
+	// M - 1024*1024 bytes
+	// G - 1024*1024*1024 bytes
+	if (ch >= 'A' && ch <= 'Z' && size_map[ch - 'A'] > 0) {
```
jaylett: You don't need the last part of the conditional, because if `size_map[ch - 'A'] == 0` then you will return `0` from the `else` leg, which is what it will do already.

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
     return Xapian::Query(Xapian::Query::OP_INVALID);
 }
 
+static const int size_map[26] = {
```
jaylett: You probably want `unsigned int` here, since negatives won't make sense.

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+	// G - 1024*1024*1024 bytes
+	if (ch >= 'A' && ch <= 'Z' && size_map[ch - 'A'] > 0) {
+    return size_map[ch - 'A'];
+	} else {
+	return 0;
+	}
+}
+
+Xapian::Query
+FileSizeRangeProcessor::operator()(const string& b, const string& e){
```
jaylett: There should be a space before the trailing `{`.

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+	// Here b and e will be like "100K" and "1M"
+	double size_b, size_e;
+	char unit_b, unit_e;
+	string temp_b = b, temp_e = e;
+
+	if (!b.empty()) {
+		errno = 0;
+		char b_back = b.back();
+		if (b_back =='B' || b_back =='K' || b_back =='M' || b_back =='G') {
```
jaylett: You should be able to do this without having the list of allowed suffices twice — once here and once, effectively, in the data structure that drives `get_bytes()`.

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+    1024 * 1024 * 1024, // G
+    0, 0, 0,
+    1024, // K
+    0,
+    1024 * 1024, // M
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+};
+
+static int
+get_bytes(char ch)
```
jaylett: This isn't a terrific name for this function; maybe `get_size_multiplier_from_suffix`? Or maybe that's too wordy.

Additionally, it should be `const char`, and `unsigned int` as the return type.

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+		size_e = strtod(startptr, &endptr);
+		if (endptr != startptr + temp_e.size() || errno) {
+			// Invalid characters in string || overflow or underflow.
+			goto not_our_range;
+		}
+	} else {
+		size_e = 0.0;
+	}
+	if (unit_e == '\0') {
+		// 1M..5 is not valid. So, if e does not have unit, then invalid
```
jaylett: What about `1..5`?

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+	if (ch >= 'A' && ch <= 'Z' && size_map[ch - 'A'] > 0) {
+    return size_map[ch - 'A'];
+	} else {
+	return 0;
+	}
+}
+
+Xapian::Query
+FileSizeRangeProcessor::operator()(const string& b, const string& e){
+	// Here b and e will be like "100K" and "1M"
```
jaylett: This function is over-indented. Indentation should be at 4, not 8. (Again, this may be because you have tabs set to 4 not 8.)

xapian-core/include/xapian/queryparser.h
```diff
@@ -234,6 +234,79 @@ class XAPIAN_VISIBILITY_DEFAULT RangeProcessor
     }
 };
 
+// Handles a file size range
```
jaylett: This comment should also note the form of the values, which is the same as for `NumberRangeProcessor`: they must be serialised appropriately.

xapian-core/include/xapian/queryparser.h
```diff
@@ -234,6 +234,79 @@ class XAPIAN_VISIBILITY_DEFAULT RangeProcessor
...
+     *             instead of a prefix.
+     *           * Xapian::RP_REPEATED - optionally allow @a str_
+     *             on both ends of the range - e.g. $1..$10 or
+     *             5m..50m.  By default a prefix is only checked for on
+     *             the start (e.g. date:1/1/1980..31/12/1989), and a
+     *             suffix only on the end (e.g. 2..12kg).
+     *
+     *  The string supplied in str_ is used by @a operator() to decide whether
+     *  the pair of strings supplied to it constitute a valid range.  If
+     *  prefix_ is true, the first value in a range must begin with str_ (and
```
jaylett: Looks like this was copied from `NumberRangeProcessor`, whose comment isn't great — `prefix_` doesn't exist for these classes.

More interestingly, do both of the `flags_` make sense here? `size:1K..10M` makes sense (`flags_=0`), but which others do? Should we allow the use of flags that don't make sense? We should at least discuss which ones are useful in the doccomment.

xapian-core/include/xapian/queryparser.h
```diff
@@ -234,6 +234,79 @@ class XAPIAN_VISIBILITY_DEFAULT RangeProcessor
...
+     *  the second value may optionally begin with str_);
+     *  if prefix_ is false, the second value in a range must end with str_
+     *  (and the first value may optionally end with str_).
+     *
+     *  If str_ is empty, the setting of prefix_ is irrelevant, and no special
+     *  strings are required at the start or end of the strings defining the
+     *  range.
+     *
+     *  The remainder of both strings defining the endpoints must be valid
+     *  floating point numbers.
```
jaylett: …with optional size suffices as documented below. Alternatively (and this might be better), move the size suffices into the class doccomment, and refer back to it from method doccomments.

xapian-core/include/xapian/queryparser.h
```diff
@@ -234,6 +234,79 @@ class XAPIAN_VISIBILITY_DEFAULT RangeProcessor
...
+     *  M - Mega Bytes
+     *  G - Giga Bytes
+     *  For example, if str_ is "size:",  Xapian::RP_SUFFIX is not specified,
+     *  and the range processor has been added to the queryparser, the
+     *  queryparser will accept "size:10K..1M".
+     *  Examples of valid ranges
+     *    size:10K..1M i.e. 10 KB to 1 MB
+     *  Examples of invalid ranges
+     *    size:10K..20
+     *    size:10..20
```
jaylett: I think Olly previously questioned why this isn't valid. It makes perfect sense, and additionally provides backwards compatibility for omega users.

xapian-core/tests/api_queryparser.cc
```diff
@@ -1725,6 +1725,86 @@ DEFINE_TESTCASE(qp_range4, !backend) {
...
+    { NULL, NULL }
+};
+
+// Test FileSizeRangeProcessor
+DEFINE_TESTCASE(qp_filesizerange1, !backend) {
+    Xapian::QueryParser qp;
+    Xapian::FileSizeRangeProcessor rp_filesize(1);
+    qp.add_rangeprocessor(&rp_date);
+    for (const test *p = test_value_filesizerange1_queries; p->query; ++p) {
+    string expect, parsed;
```
jaylett: These lines within the `for` loop should be indented further (and again in the next test).

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,100 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+	string temp_b = b, temp_e = e;
+
+	if (!b.empty()) {
+		errno = 0;
+		char b_back = b.back();
+		if (b_back =='B' || b_back =='K' || b_back =='M' || b_back =='G') {
+			// If suffix of b is 'B','K','M','G'
+			unit_b = b_back;
+			temp_b.pop_back();
+		} else if (!isdigit(b_back)) {
```
jaylett: As Olly previously said:

> Use the Xapian-specific `C_isdigit()` instead of `isdigit()` in Xapian code (also true for `isalpha()`, and other functions in the same family).
>
> The set of characters for which `isdigit()` returns a true value may vary if `setlocale()` has been called, which we want to avoid.
>
> Also `isdigit()` isn't defined for negative values, which is a problem on platforms where `char` is `signed char`,  Our `C_isdigit()` function works for both signed and unsigned char (the GNU libc implementation also does, but we can't rely on the system's libc extending the range of these functions for us, and it's cumbersome to have to write `isdigit(static_cast<unsigned char>(ch))` every time).

xapian-core/include/xapian/queryparser.h
```diff
@@ -234,6 +234,79 @@ class XAPIAN_VISIBILITY_DEFAULT RangeProcessor
...
+     *  FileSizeRangeProcessor will handle file size as:
+     *  B - Bytes
+     *  K - Kilo Bytes
+     *  M - Mega Bytes
+     *  G - Giga Bytes
+     *  For example, if str_ is "size:",  Xapian::RP_SUFFIX is not specified,
+     *  and the range processor has been added to the queryparser, the
+     *  queryparser will accept "size:10K..1M".
+     *  Examples of valid ranges
+     *    size:10K..1M i.e. 10 KB to 1 MB
```
jaylett: Generally, avoid using abbreviations like "i.e." or "e.g.". Here, I'd just say:

> size:10K..1M (10K to 10M)

xapian-core/include/xapian/queryparser.h
```diff
@@ -234,6 +234,79 @@ class XAPIAN_VISIBILITY_DEFAULT RangeProcessor
...
+     *    size:10..20
+     *    size:K10..K20
+     *    size:10..K20
+     *    size:K10..20
+     *
+     */
+    FileSizeRangeProcessor(Xapian::valueno slot_,
+		const std::string &str_ = std::string(),
+		unsigned flags_ = 0)
+    : RangeProcessor(slot_, str_, flags_) { }
```
jaylett: As Olly previously said:

> This constructor has a form which takes a single argument, so should be marked as `explicit` - we don't want an unsigned integer to be quietly converted to a `FileSizeRangeProcessor`.

xapian-core/include/xapian/queryparser.h
shivanshuiitg: I will make change. Should I keep 10..20 as 10 to 20 KB or 10 to 20 bytes ?

xapian-core/include/xapian/queryparser.h
ojwb: I don't see a sensible use for being able to specify a suffix - you could have some magic suffix indicate this is a size range, but (say) `1..100K^` just seems weird.

And repeating the prefix/suffix also doesn't seem useful here.  That's really meant for "units", e.g. `1..10kg` vs `1kg..10kg`; `$1.100` vs `$1..$100`, but I don't see that being helpful here.

However, I don't think we want to make it impossible to specify these as people may find uses (and there may be more flags in the future, and some of those may be useful here even if the current flags aren't).  So discussing in the doccomment seems best to me.

ojwb: > `prefix_` doesn't exist for these classes.

Oh yes, the boolean flag `prefix_` is now specified by `Xapian::RP_SUFFIX` being bitwise-or-ed into `flags_` (but with the opposite sense).

xapian-core/include/xapian/queryparser.h
ojwb: If it were in `KB` that would provide poor backwards compatibility as the meaning of such ranges would change by ~3 orders of magnitude.

Bytes seems most logical anyway.

xapian-core/include/xapian/queryparser.h
shivanshuiitg: Thanks @jaylett @ojwb  for pointing it out. 
I will remove the flags in FileSizeRangeProcessor class as it does not require RP_SUFFIX and RP_REPEATED.

Should I keep flag_ variable as it is & discuss in comments that it is not required OR just remove it from class ?

shivanshuiitg:
I have added a fixup commit, which resolves the issues mentioned in comments.

xapian-core/include/xapian/queryparser.h
ojwb: As I said above, I don't think we want to make it impossible to specify these.  Removing this parameter would make it impossible, so don't remove it.

shivanshuiitg: Ok ! So I have kept this parameter as it is and specified in comments that it has no use for FileSizeRangeProcessor class.

xapian-core/api/valuerangeproc.cc
```diff
@@ -451,23 +451,23 @@ static const int size_map[26] = {
...
 	// M - 1024*1024 bytes
 	// G - 1024*1024*1024 bytes
-	if (ch >= 'A' && ch <= 'Z' && size_map[ch - 'A'] > 0) {
-    return size_map[ch - 'A'];
+	if (ch >= 'A' && ch <= 'Z') {
+		return size_map[ch - 'A'];
 	} else {
-	return 0;
+		return 0;
 	}
```
ojwb: These two return statements seem to be over-indented - each indent level should be 4 columns, but the indent here is two tabs (16 columns) but the `if` is indented by one tab (8 columns).

xapian-core/api/valuerangeproc.cc
```diff
@@ -480,7 +480,7 @@ FileSizeRangeProcessor::operator()(const string& b, const string& e){
 			// If suffix of b is 'B','K','M','G'
```
ojwb: You've not addressed James' comment about this check.

xapian-core/tests/api_queryparser.cc
```diff
@@ -1743,19 +1743,19 @@ DEFINE_TESTCASE(qp_filesizerange1, !backend) {
...
     } catch (const Xapian::QueryParserError &e) {
-	parsed = e.get_msg();
+	   parsed = e.get_msg();
     } catch (const Xapian::Error &e) {
-	parsed = e.get_description();
+	   parsed = e.get_description();
     } catch (...) {
-	parsed = "Unknown exception!";
+	   parsed = "Unknown exception!";
     }
```
ojwb: These changes all seem incorrect - it looks to me like the indentation was correct before, and you've added 3 spaces for some reason.

xapian-core/include/xapian/queryparser.h
jaylett: I think we probably want to document what putting flags in would do, but point out that we don't see a need for them. (That way it might nudge anyone who finds a use to tell us about it, so we can document it for others.)

xapian-core/tests/api_queryparser.cc
jaylett: @shivanshuiitg I think you need to spend some time figuring out how to configure your editor for Xapian's indentation and alignment approach. If you get the config right, you shouldn't have to worry about indentation in your code very much, as it'll do most of the work for you. The amount of code you'd need to write over the summer is enough that this will save you a lot of time.

The main rules are: each level of indentation is by four positions, and indentation whitespace should be tabs (of eight positions) followed by spaces.

shivanshuiitg:
@ojwb  @jaylett  Thanks for reviewing my PR. From starting there was a issue of indentation with my Text-editor. I have figured out that culprit was my text editor's setting. Every time I set tab to 8 spaces, it was changing to 4 automatically(I still have no idea why?). Now, I have solved the issue by reinstalling the text editor and corrected the indentation. I will be resolving some issues in further commits. 

@jaylett I will restructure the commits to one commit per feature very soon.

Please check the code and let me know if there is still concerns with the indentation of code.

xapian-core/api/valuerangeproc.cc
shivanshuiitg: @jaylett Thanks for pointing it out. I will make required changes. :)

jaylett:
Before we merge, the "white space issue with code fixed." commit and its fixup (currently 66edb60 and af204ff) will need turning into appropriate fixups against the earlier commits. Those original commits still need better messages writing.

xapian-core/api/valuerangeproc.cc
```diff
@@ -466,6 +466,17 @@ get_size_multiplier_from_suffix(const char ch)
     }
 }
 
+static bool
+is_valid_size_character(const char ch)
+{
+    // Return true if ch is valid size character ('B','K','M','G')
+    if (ch =='B' || ch =='K' || ch =='M' || ch =='G') {
```
jaylett: This is not what I meant, because now you've just moved the check somewhere else. `get_size_multiplier_from_suffix()` will tell you if it's a valid suffix, so just use that and avoid this duplication.

xapian-applications/omega/omegatest
```diff
@@ -268,6 +268,12 @@ testcase '1.000000.' B=Hexample.org B=Hexample.net
 testcase '2.000000.' B=Hexample.org B=XFOObar
 testcase '3.000000.' B=Hexample.org B=XFOObar B=XFOObar
 
+# size-range search test for omega
+printf '$querydescription' > "$TEST_TEMPLATE"
+qtestcase '0 * VALUE_RANGE 2 10240 20480' P='size:10..20K'
+qtestcase '0 * VALUE_RANGE 2 10240 30721' P='size:10K..30K'
+qtestcase '0 * VALUE_RANGE 2 10240 40960' P='size:10..40'
```
ojwb: A note for when this gets merged - on master the `0 * ` prefix is no longer present here.

Once we're happy with the changes, I suggest we rebase on master, add a fixup commit for this, and then autosquash.

xapian-core/api/valuerangeproc.cc
```diff
@@ -362,35 +362,35 @@ DateRangeProcessor::operator()(const string& b, const string& e)
...
+    }
+    if (!e.empty()) {
+	sprintf(buf_e, "%08d", e_y * 10000 + e_m * 100 + e_d);
+	if (buf_e[sizeof(buf_e) - 1]) abort(); // Buffer overrun!
+    } else {
+	*buf_e = '\0';
+    }
 #endif
-	return RangeProcessor::operator()(buf_b, buf_e);
+    return RangeProcessor::operator()(buf_b, buf_e);
```
ojwb: All the above whitespace changes need reverting.

xapian-core/api/valuerangeproc.cc
```diff
@@ -432,8 +432,120 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
     }
 
     return RangeProcessor::operator()(
-	    b.empty() ? b : Xapian::sortable_serialise(num_b),
-	    e.empty() ? e : Xapian::sortable_serialise(num_e));
+	b.empty() ? b : Xapian::sortable_serialise(num_b),
+	e.empty() ? e : Xapian::sortable_serialise(num_e));
```
ojwb: The above whitespace changes need reverting.

xapian-core/api/valuerangeproc.cc
```diff
@@ -432,8 +432,120 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+    1024 * 1024 * 1024, // G
+    0, 0, 0,
+    1024, // K
+    0,
+    1024 * 1024, // M
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+};
+
+static unsigned int
+get_size_multiplier_from_suffix(const char ch)
```
ojwb: The `const` here doesn't really server a useful purpose.  Generally you only want to make a function parameter `const` if it's a pointer or reference.

xapian-core/api/valuerangeproc.cc
```diff
@@ -432,8 +432,120 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+{
+    // It takes character as input and return number of bytes
```
ojwb: The `It` sounds odd here.

Also, this barely goes beyond describing the function at a syntactic level.  One can see from the parameter list that it takes a character as input, and from the return type that it returns a number.  So all you've really told us is that the return value is measured in bytes (but actually it's not bytes but a scale factor - it could equally apply to bits as bytes).

What's more interesting to tell future developers (often including your future self) is the higher level semantics which aren't obvious from the function signature - what the function is for, e.g. the meaning of the output in relation to the input.  And handling of errors is important to document too.

It doesn't have to be an essay - indeed a good comments is often brief but to the point.  Something like this captures the important details: `Maps a file size suffix character to a scale factor.  Returns 0 for an unknown suffix character.`

If this was external API documentation, the list of recognised characters definitely ought to be in there too, but here that's already documented by the comments in `size_map[]` above - repeating that list here mostly means there are two versions which creates the risk of a future change updating one but not the other (https://en.wikipedia.org/wiki/Don't_repeat_yourself).  You could move the whole definition of `size_map[]` inside the function to make the connection clearer (and the code a little cleaner).

xapian-core/tests/api_queryparser.cc
```diff
@@ -1725,6 +1725,84 @@ DEFINE_TESTCASE(qp_range4, !backend) {
     return true;
 }
 
+static const test test_value_filesizerange1_queries[] = {
+    { "10K..20K", "0 * VALUE_RANGE 1 10240 20480" },
+    { "10..50M", "0 * VALUE_RANGE 1 10485760 52428800" },
```
ojwb: `0 * ` prefix gone on master here too.

xapian-core/tests/api_queryparser.cc
```diff
@@ -1725,6 +1725,84 @@ DEFINE_TESTCASE(qp_range4, !backend) {
...
+    { "10K..20", "Unknown range operation" },
+    { "10A..20A", "Unknown range operation" },
+    { "10..20H", "Unknown range operation" },
```
ojwb: You aren't testing any cases without suffices.

You aren't testing with empty range ends.

And you're only testing integer values.

xapian-core/tests/api_queryparser.cc
```diff
@@ -1725,6 +1725,84 @@ DEFINE_TESTCASE(qp_range4, !backend) {
...
+
+// Feature test FileSizeRangeProcessor with prefixes 
+DEFINE_TESTCASE(qp_filesizerange2, !backend) {
+    Xapian::QueryParser qp;
+    Xapian::FileSizeRangeProcessor rp_filesize_1(1, "size:");
+    Xapian::FileSizeRangeProcessor rp_filesize_2(2, "filesize:");
+    Xapian::FileSizeRangeProcessor rp_filesize_3(3, "file-size:");
+    qp.add_rangeprocessor(&rp_filesize_1);
+    qp.add_rangeprocessor(&rp_filesize_2);
+    qp.add_rangeprocessor(&rp_filesize_3);
```
ojwb: I think it makes sense to just merge the testcases from `qp_filesizerange1` into those tests here.  We'll get better coverage for the time spent running tests that way, as the difference between the single RP and multiple RP case isn't interesting.

xapian-core/tests/api_queryparser.cc
```diff
@@ -1725,6 +1725,84 @@ DEFINE_TESTCASE(qp_range4, !backend) {
...
+	tout << "Query: " << p->query << '\n';
+	TEST_STRINGS_EQUAL(parsed, expect);
+    }
+    return true;
+}
+
+static const test test_value_filesizerange2_queries[] = {
+    { "size:10..20K", "0 * VALUE_RANGE 1 10240 20480" },
+    { "size:10..50M", "0 * VALUE_RANGE 1 10485760 52428800" },
+    { "filesize:10B..1M", "0 * VALUE_RANGE 2 10 1048576" },
```
ojwb: The expected results here are wrong - it should be the result of `Xapian::sortable_serialise()` on the number, like  `NumberRangeProcessor` would give - e.g.:

```
{ "1..12", "VALUE_RANGE 2 \\xa0 \\xae" },
```

As I said already (on IRC I think) you can get the correct encoded values by using `NumberRangeProcessor` on the scaled end points (so for `size:10..20K` look at `number:10240..20480`).

jaylett:
@shivanshuiitg even if you're not going to have a chance to work on this further, can you give permission for your code to be [dual licensed as per our standards](https://trac.xapian.org/browser/git/xapian-core/HACKING#L1392)? All you need to do is to reply to this issue saying something like:

> I agree to dual-license all my contributions to Xapian under the GNU GPL version 2 and all later versions, and the MIT/X licence.

That way someone else in the community can pick this up and see if we can get it finished and merged. (Your contribution will of course be acknowledged as per the licenses above!)

ojwb:
There's been no response for over a year, so I think we have to assume this patch is abandoned. 

We don't have the required licensing agreement so this patch is not even something that somebody else can pick up and take forward.

shivanshuiitg:
Hi Olly,

Sorry for the late reply.

I agree to dual-license all my contributions to Xapian under the GNU GPL
version 2 and all later versions, and the MIT/X licence.

Thanks & Regards,
Shivanshu Chauhan

PR #136, opened 2017-01-29
Change use of atoi/atol to stoi/stol

nimitbhardwaj:
Referred from https://trac.xapian.org/wiki/ProjectIdeas#BiteSize Replace atoi() and atol() 

But instead of strtol or strtoul I used stoi and stol (part of stl string), as they raise exceptions on the wrong inputs but strtol/strtoul(the part of the stdlib.h) don't raise the exceptions

This is the change in the xapian-application/omega/ applications, all the required occurences of the atoi/atol are changed, and a try catch mechanism is added to extra so that if suppose in future there is error in converting from string to number, the invalid_exception is raised so that the proper position of the error is known, stoi and stol functions (the part of stl string) are similar to atoi/atol except that they throw error/exception whern we send string something like "ab123" and atoi gives value of 0, but on sending the string like "123ab" they both will give the valid output, in this case 123 http://www.cplusplus.com/reference/string/stoi/ , so according to me we should also change the occurrences of the atoi and atol in the cgiparam.cc file as if we get the valid output in the strings like "2[=" then both atoi and stoi can handle them, my next step is to remove the call of the member function c_str() at the places where I replaced atoi to stoi, but before I need verification, that I did this work correct,

### DIFF
--- a/xapian-applications/omega/date.cc
+++ b/xapian-applications/omega/date.cc
@@ -24,6 +24,8 @@
 
 #include <config.h>
 
+#include <iostream>
+
 #include "date.h"
 
 #include <vector>
@@ -130,9 +132,15 @@ parse_date(const string & date, int *y, int *m, int *d)
 {
     // FIXME: for now only support YYYYMMDD (e.g. 20011119)
     // and don't error check
-    *y = atoi(date.substr(0, 4).c_str());
-    *m = atoi(date.substr(4, 2).c_str());
-    *d = atoi(date.substr(6, 2).c_str());
+    try {
+    *y = stoi(date.substr(0, 4).c_str(), nullptr); //stoi can be used instead of atoi in this function
+    *m = stoi(date.substr(4, 2).c_str(), nullptr);
+    *d = stoi(date.substr(6, 2).c_str(), nullptr);
+    } catch(invalid_argument &e)
+    {
+    cerr << "error in file /xapian-application/omega/date.cc exception by stoi" << endl;
+    exit(EXIT_FAILURE);
+    }
 }
 
 Xapian::Query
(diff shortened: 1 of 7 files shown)


xapian-applications/omega/date.cc
```diff
@@ -130,9 +132,15 @@ parse_date(const string & date, int *y, int *m, int *d)
 {
     // FIXME: for now only support YYYYMMDD (e.g. 20011119)
     // and don't error check
```
nimitbhardwaj: The use of stoi can be the fix for the above FIXME as it will check only numbers if the string is encountered then it will throw exception

nimitbhardwaj:
Why those errors occured

ojwb:
The GCC build failed because you didn't add `#include <stdexcep>` to get the standard exception classes declared:

http://en.cppreference.com/w/cpp/error/invalid_argument

The `xapian-check-patch` build failed because your patch doesn't match the Xapian coding standards - almost all the reported issues look like they're due to not having your editor set to tab-fill indentation, e.g.:

```
xapian-applications/omega/omega.cc:422: error: added/changed line uses spaces for indentation rather than tab: +        try {
```

The `xapian-check-patch` script is in the source tree (under `xapian-maintainer-tools`) - I would strongly recommend running it locally to test changes and fix any issues it flags (it doesn't check for everything, but it does catch many common issues):

```
git diff master... | xapian-maintainer-tools/xapian-check-patch
```

If you editor has a "quick-fix" mode, you can probably feed the output into that - and have each flagged line opened automatically - e.g. for "vim" I use:

```
git diff master... | xapian-maintainer-tools/xapian-check-patch > tmp.qf
vim -q tmp.qf
```

ojwb:
Reviewing the changes in the patch, you've pretty much just added a boilerplate "chuck a message out on stderr and exit with non-zero status".  That's rarely a good response to an error, and is particularly poor in a CGI program (the user will get a truncated page or a cryptic error in their browser, and the error message will just disappear into a server log - for apache, `error.log` by default).

Another issue is that handling an exception is typically fairly expensive, so `stoi()` is not an appropriate choice for all situations - that's mostly why the project idea suggests using `strtol()`, etc.

There's a lot more to this task that it may appear.  Rather than just cutting and pasting essentially the same code wherever you find `atoi()`, I'd suggest slowing down and looking at each case by itself.  There's no need to address every use in one PR  - addressing (say) uses of `atoi()` in parameter parsing code is a useful change in itself, and what you learn from that would make it easier to produce another PR for uses elsewhere.

Think about what input values should be allowed, about how bad values would ideally be handled, and compare that to how they would currently be handled.  Consider how your change could affect users upgrading to a version containing your changes.  Think about what the error messages produced look like to the user.

nimitbhardwaj:
OK, I got that, I will fix this one as soon as I got the time, I am buzzy this week as there is an event going on in our college

ojwb:
The current patch needs a major reworking and there's been no progress in 3 months so I'm going to close this.  If you still wish to work on this, please reopen.

PR #137, opened 2017-01-29
Html semantics in omega templates

jronak:
BiteSized: Omega templates to use more modern web techniques

As suggested by @jaylett I have created a PR for review on using html5 semantics for the template of omega before proceeding to add styles. This PR for review and comments on how the structure can be improved. As of now, only query template has been changed. 

I have chosen this project to start off working on omega along with understanding the internals as I progress. 

Please provide your suggestions/advice,
Thank you

### DIFF
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -42,7 +42,8 @@ function write_date(time_t) {
 var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
-document.write("<span>Modified:<time datetime=\""+D+" "+T+"\">"+D+"</time></span>");
+var ISO=d.toISOString();
+document.write("<span>Modified:<time datetime=\""+ISO+"\"</time></span>");
 }
 // -->
 </script>
@@ -97,7 +98,7 @@ ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" heig
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
-<span>Modified:<time datetime="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
+<span>Modified:<time datetime="$html{$date{$field{modtime},%Y-%m-%dT%H:%M:%S}}">
 $html{$date{$field{modtime},%Y-%m-%d}}</time></span>
 </noscript>
 }
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -52,7 +52,6 @@ document.write("<span title=\""+D+" "+T+"\">Modified:<br><b>"+D+"</b></span>");
 <body bgcolor="#ffffff">
 <form name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
-<center>
 <input id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
 <script>
 if (!("autofocus" in document.createElement("input")))
@@ -82,7 +81,6 @@ $if{$not{$msizeexact},
   $nice{$add{$topdoc,1}}$if{$ne{$add{$topdoc,1},$last},-$nice{$last}} of exactly $nice{$msize} matches}
 }
 <hr>
-</center>
 $list{$map{$queryterms,$list{$html{$uniq{$unstem{$_}}},<b>,</b>/<b>,</b>}:&nbsp;$nice{$freq{$_}}},Term frequencies: ,$. ,}
 <br><small>Search took $time seconds</small>
 <table>
@@ -114,7 +112,7 @@ matching:
 <i>$list{$map{$terms,$html{$prettyterm{$_}}},</i>$. <i>,</i> and <i>}</i></small>${for lynx:}<p></td></tr>}
 </table>
 
-<br><center>
+<br>
 
 ${suppress next, prev, and page links if there's only one page}
 $if{$ne{$lastpage,1},
@@ -134,7 +132,7 @@ $if{$ne{$opt{y},$lastpage},...}
 $NEXT
 }
 }}
-</center><br>
+<br>
 $if{$dbname,<input type=hidden name=DB value="$html{$dbname}">}
 $if{$ne{$topdoc,0},<input type=hidden name=TOPDOC value=$topdoc>}
 $if{$ne{$hitsperpage,10},<input type=hidden name=HITSPERPAGE value=$hitsperpage>}
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -22,13 +22,13 @@ $def{S,<img src="/icons/omega/page-$2s.png" $opt{a} alt=$1>}
 $def{SPAGE,$if{$gt{$1,9},$if{$gt{$1,99},$S{$1,$div{$1,100}}}$S{$1,$mod{$div{$1,10},10}}}$S{$1,$mod{$1,10}}}
 }
 
-$def{PREV,$if{$ne{$topdoc,0},<input type=submit name="&lt;" value="Previous">}}
+$def{PREV,$if{$ne{$topdoc,0},<input class="nav-button" type=submit name="&lt;" value="Previous">}}
 
-$def{PAGE,<input type=submit name="[" value="$1">}
+$def{PAGE,<input class="nav-button" type=submit name="[" value="$1">}
 
-$def{SPAGE,<input type=submit name="[" value="$1" disabled=disabled>}
+$def{SPAGE,<input class="nav-button" type=submit name="[" value="$1" disabled=disabled>}
 
-$def{NEXT,$if{$ne{$last,$msize},<input type=submit name="&gt;" value="Next">}}
+$def{NEXT,$if{$ne{$last,$msize},<input class="nav-button" type=submit name="&gt;" value="Next">}}
 
 <head>
 <title>$if{$query,Omega Search: $html{$query},Omega Search}</title>
@@ -37,6 +37,7 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
 input[type=image] {
   border: 0px;
   }
+
 </style>
 <script type="text/javascript"><!--
 function f2(n) {return (n>9?"":"0")+n;}
@@ -53,51 +54,54 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 <form name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
 ${Search Control(Start) - shows search button, options and textbox}
-<fieldset>
-<input id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
+<fieldset class="query-search">
+<input class="query-search-textbox" d="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
 <script>
 if (!("autofocus" in document.createElement("input")))
  document.getElementById("omega-autofocus").focus();
 </script>
-<input type=submit value="Search">
+<input class="query-search-button" type=submit value="Search">
 </fieldset>
 ${Todo: Replace radios with search buttons}
 $if{,$include{inc/anyalldropbox},
-<fieldset>$include{inc/anyallradio}</fieldset>}
+<fieldset class="query-options">$include{inc/anyallradio}</fieldset>}
 ${Search Control(End)}
 ${Suggestions(Start) - shows suggested terms to improve query}
 $if{$opt{topterms},
- <div title="Suggested terms to add to your query"
+ <div class="query-suggested-container" title="Suggested terms to add to your query"
  style="text-align:left;background:#cfc;border:1px solid green;padding:2px;font:11px verdana$. arial$. helvetica$. sans-serif;">
- $map{$topterms,<span style="white-space:nowrap"><input type=checkbox name=X value="$html{$prettyterm{$_}}" onClick="C(this)">$html{$prettyterm{$_}}</span> }
+ $map{$topterms,<span class="query-suggested-element" style="white-space:nowrap"><input type=checkbox name=X value="$html{$prettyterm{$_}}" onClick="C(this)">$html{$prettyterm{$_}}</span> }
  <noscript><input type=hidden name=ADD value=1></noscript>
  </div>
 }
-$if{$suggestion,Did you mean: <a href="$html{$or{$env{SCRIPT_NAME},omega}?DB=$url{$dbname}&P=$url{$suggestion}$if{$ne{$hitsperpage,10},&HITSPERPAGE=$hitsperpage}$if{$fmt,&FMT=$url{$fmt}}$if{$cgi{COLLAPSE},&COLLAPSE=$url{$cgi{COLLAPSE}}}$map{$relevants,&R=$_}$if{$cgi{THRESHOLD},&THRESHOLD=$url{$cgi{THRESHOLD}}}$map{$cgilist{B},&B=$url{$_}}&SPELL=0}">$html{$suggestion}</a>}
+$if{$suggestion,Did you mean: <a class="query-suggestion" href="$html{$or{$env{SCRIPT_NAME},omega}?DB=$url{$dbname}&P=$url{$suggestion}$if{$ne{$hitsperpage,10},&HITSPERPAGE=$hitsperpage}$if{$fmt,&FMT=$url{$fmt}}$if{$cgi{COLLAPSE},&COLLAPSE=$url{$cgi{COLLAPSE}}}$map{$relevants,&R=$_}$if{$cgi{THRESHOLD},&THRESHOLD=$url{$cgi{THRESHOLD}}}$map{$cgilist{B},&B=$url{$_}}&SPELL=0}">$html{$suggestion}</a>}
 ${Suggestions(End)}
 ${Query metadata(start) - shows query performance or document size}
 $or{$html{$error},
+<p class="query-metadata">
 $if{$eq{$msize,0},
 $if{$query,No documents match your query,
 Searching $nice{$dbsize} documents
-},
+}</p>,
 $if{$not{$msizeexact},
  $nice{$add{$topdoc,1}}-$nice{$last} of about $nice{$msize} matches,
  $if{$and{$eq{$last,$msize},$eq{$topdoc,0}},
   All $nice{$msize} matches,
   $nice{$add{$topdoc,1}}$if{$ne{$add{$topdoc,1},$last},-$nice{$last}} of exactly $nice{$msize} matches}
 }
-$list{$map{$queryterms,$list{$html{$uniq{$unstem{$_}}},<b>,</b>/<b>,</b>}:&nbsp;$nice{$freq{$_}}},Term frequencies: ,$. ,}
+<br/>
+$list{$map{$queryterms,$list{$html{$uniq{$unstem{$_}}},,,}:&nbsp;$nice{$freq{$_}}},Term frequencies: ,$. ,}
 Search took $time seconds
+</p>
 ${Query metadata(End)}
 ${Results(start)}
 <table>
-$hitlist{<tr><td valign=top>
+$hitlist{<tr class="list-item-container"><td valign=top>
 ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" height=16 width=32>}
-<div title="$percentage%" style='float:left;width:60px;height:10px;border:1px solid black;margin-top:4px;'>
-<div style='width:$div{$mul{$percentage,6},10}px; height:10px; background-color: red;'>
+<div class="list-item-progress-container" title="$percentage%" style='float:left;width:60px;height:10px;border:1px solid black;margin-top:4px;'>
+<div class="list-item-progress" style='width:$div{$mul{$percentage,6},10}px; height:10px; background-color: red;'>
 </div></div>
-<div style='float:left;margin-top:2px;font-size:x-small;'>
+<div class="list-item-meta" style='float:left;margin-top:2px;font-size:x-small;'>
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
@@ -112,11 +116,13 @@ $if{$field{size},<span title="$html{$field{size}} bytes">Size: <b>$html{$filesiz
 </div>
 </div>
 </td>
-<td><header><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></header>
-$highlight{$snippet{$field{sample}},$terms}$if{$field{sample},...}
+<td><header class="list-item-header"><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></header>
+<span class="list-item-content" style="display: block;">
+$highlight{$snippet{$field{sample}},$terms}$if{$field{sample},...}</span>
 <a href="$html{$field{url}}">$html{$field{url}}</a>
+<span class="list-item-meta">
 matching:
-<i>$list{$map{$terms,$html{$prettyterm{$_}}},</i>$. <i>,</i> and <i>}</i>${for lynx:}<p></td></tr>}
+<i>$list{$map{$terms,$html{$prettyterm{$_}}},</i>$. <i>,</i> and <i>}</i>${for lynx:}</span></td></tr>}
 </table>
 ${Results(end)}
 
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -52,13 +52,18 @@ document.write("<span title=\""+D+" "+T+"\">Modified:<b>"+D+"</b></span>");
 <body bgcolor="#ffffff">
 <form name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
+${Search Control(Start) - shows search button, options and textbox}
+<fieldset>
 <input id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
 <script>
 if (!("autofocus" in document.createElement("input")))
  document.getElementById("omega-autofocus").focus();
 </script>
 <input type=submit value="Search">
-$if{,$include{inc/anyalldropbox},$include{inc/anyallradio}}
+</fieldset>
+${Todo: Replace radios with search buttons}
+$if{,$include{inc/anyalldropbox},
+<fieldset>$include{inc/anyallradio}</fieldset>}
 $if{$opt{topterms},
  <div title="Suggested terms to add to your query"
  style="text-align:left;background:#cfc;border:1px solid green;padding:2px;font:11px verdana$. arial$. helvetica$. sans-serif;">
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -46,7 +46,7 @@ function write_date(time_t) {
 var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
-document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time></span>");
+document.write("<span>Modified:<time datetime=\""+D+" "+T+"\">"+D+"</time></span>");
 }
 // -->
 </script>
@@ -99,7 +99,7 @@ ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" heig
 <div class="list-item-progress-container" title="$percentage%" >
 <div class="list-item-progress" style='width:$div{$mul{$percentage,6},10}px;'>
 </div></div>
-<div class="list-item-meta" style='float:left;margin-top:2px;font-size:x-small;'>
+<div class="list-item-meta">
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
@@ -107,8 +107,8 @@ $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 $html{$date{$field{modtime},%Y-%m-%d}}</time></span>
 </noscript>
 }
-$if{$field{language},Language: <b>$html{$field{language}}</b>}
-$if{$field{size},<br/><span title="$html{$field{size}} bytes">Size: <b>$html{$filesize{$field{size}}}</b></span>}
+$if{$field{language},Language: $html{$field{language}}}
+$if{$field{size},<br/><span title="$html{$field{size}} bytes">Size:$html{$filesize{$field{size}}}</span>}
 <div title="Mark this document as relevant">
 <input type=checkbox name=R value=$id$if{$relevant, checked}>
 </div>
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -52,7 +52,7 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 </script>
 </head>
 <body>
-<form name=P method=get
+<form id="query-form" name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
 ${Search Control - shows search button, options and textbox}
 <fieldset class="query-search">
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -43,7 +43,7 @@ var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
 var ISO=d.toISOString();
-document.write("<span>Modified:<time datetime=\""+ISO+"\"</time></span>");
+document.write("<span>Modified:<time datetime=\""+ISO+"\" title=\""+D+" "+T+"\">"+D+"</time></span>");
 }
 // -->
 </script>
@@ -98,7 +98,7 @@ ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" heig
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
-<span>Modified:<time datetime="$html{$date{$field{modtime},%Y-%m-%dT%H:%M:%S}}">
+<span>Modified:<time title="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}" datetime="$html{$date{$field{modtime},%Y-%m-%dT%H:%M:%S}}">
 $html{$date{$field{modtime},%Y-%m-%d}}</time></span>
 </noscript>
 }
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -61,8 +61,8 @@ if (!("autofocus" in document.createElement("input")))
 <input type=submit value="Search">
 </fieldset>
 ${Todo: Replace radios with search buttons}
-$if{,$include{inc/anyalldropbox},
-<fieldset class="options">$include{inc/anyallradio}</fieldset>}
+<fieldset class="options">$if{,$include{inc/anyalldropbox},
+$include{inc/anyallradio}}</fieldset>
 ${Suggestions - shows suggested terms to improve query}
 $if{$opt{topterms},
  <fieldset class="suggested-terms" title="Suggested terms to add to your query">
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -46,7 +46,7 @@ function write_date(time_t) {
 var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
-document.write("<span>Modified:<br/><time datetime=\""+D+" "+T+"\">"+D+"</time></span>");
+document.write("<span>Modified:<time datetime=\""+D+" "+T+"\">"+D+"</time></span>");
 }
 // -->
 </script>
@@ -103,12 +103,12 @@ ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" heig
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
-<span>Modified:<br/><time datetime="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
+<span>Modified:<time datetime="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
 $html{$date{$field{modtime},%Y-%m-%d}}</time></span>
 </noscript>
 }
 $if{$field{language},Language: $html{$field{language}}}
-$if{$field{size},<br/><span title="$html{$field{size}} bytes">Size:$html{$filesize{$field{size}}}</span>}
+$if{$field{size},<span title="$html{$field{size}} bytes">Size:$html{$filesize{$field{size}}}</span>}
 <div title="Mark this document as relevant">
 <input type=checkbox name=R value=$id$if{$relevant, checked}>
 </div>
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -56,7 +56,7 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
 ${Search Control - shows search button, options and textbox}
 <fieldset class="query-search">
-<input class="query-search-textbox" d="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
+<input class="query-search-textbox" d="omega-autofocus" type=search name=P value="$html{$query}" placeholder="Search..." size=65 autofocus>
 <script>
 if (!("autofocus" in document.createElement("input")))
  document.getElementById("omega-autofocus").focus();
--- a/xapian-applications/omega/templates/query
+++ b/xapian-applications/omega/templates/query
@@ -44,7 +44,7 @@ function write_date(time_t) {
 var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
-document.write("<span title=\""+D+" "+T+"\">Modified:<b>"+D+"</b></span>");
+document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time></span>");
 }
 // -->
 </script>
@@ -95,8 +95,8 @@ ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" heig
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
-<span title="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
-Modified:<b>$html{$date{$field{modtime},%Y-%m-%d}}</b></span>
+<span>Modified:<time datetime="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
+$html{$date{$field{modtime},%Y-%m-%d}}</time></span>
 </noscript>
 }
 $if{$field{language},Language: <b>$html{$field{language}}</b>}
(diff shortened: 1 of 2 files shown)


xapian-applications/omega/templates/query
```diff
@@ -81,23 +85,30 @@ $if{$not{$msizeexact},
...
+</section>
 <table>
 $hitlist{<tr><td valign=top>
 ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" height=16 width=32>}
+
 <div title="$percentage%" style='float:left;width:60px;height:10px;border:1px solid black;margin-top:4px;'>
 <div style='width:$div{$mul{$percentage,6},10}px; height:10px; background-color: red;'>
 </div></div>
 <div style='float:left;margin-top:2px;font-size:x-small;'>
+<time>
```
jaylett: This is great, but I think `time` should replace the `span` below, and change `title` to [`datetime`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) (or possibly include both, since I suspect `datetime` doesn't provide a hover). Now that the datetime format for `datetime` is different to how we're using `title` at the moment.

Note that with Javascript enabled something else happens (to show a localised time), and you'll want to update `write_date()` accordingly.

xapian-applications/omega/templates/query
```diff
@@ -81,23 +85,30 @@ $if{$not{$msizeexact},
...
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
 <span title="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
 Modified:<br><b>$html{$date{$field{modtime},%Y-%m-%d}}</b></span>
 </noscript><br>
```
jaylett: `br` is not semantic.

xapian-applications/omega/templates/query
```diff
@@ -49,17 +50,18 @@ document.write("<span title=\""+D+" "+T+"\">Modified:<br><b>"+D+"</b></span>");
 // -->
 </script>
 </head>
+
 <body bgcolor="#ffffff">
 <form name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
 <center>
```
jaylett: `center` is not semantic. There are probably others lurking that I won't catch.

xapian-applications/omega/templates/query
```diff
@@ -49,17 +50,18 @@ document.write("<span title=\""+D+" "+T+"\">Modified:<br><b>"+D+"</b></span>");
...
+<fieldset>
```
jaylett: This fieldset doesn't contain all the fields in the form, so it'd be good to have an `id` or similar for this to make it clear why you're grouping _these_ fields together.

xapian-applications/omega/templates/query
```diff
@@ -68,7 +70,9 @@ $if{$opt{topterms},
  <br><noscript><input type=hidden name=ADD value=1></noscript>
  </div>
 }
+</fieldset>
 <br>
+<section>
```
jaylett: > Do not use the <section> element as a generic container; this is what <div> is for, especially when the sectioning is only for styling purposes. A rule of thumb is that a section should logically appear in the outline of a document.

(from the [MDN description of <section>](https://developer.mozilla.org/en/docs/Web/HTML/Element/section))

I'm not sure this counts, but the way to think about it is: what are the names of the different parts of the page? Give them all headings that say what they're doing (when we add default CSS, we can throw them off the page using `position: absolute; left: -9999px`). That should make it clear when to use sections (and whether they need `id` or `class` to distinguish each of them).

xapian-applications/omega/templates/query
jronak: To center without using <center> tag we have to add CSS property to its parent, as there is no other tag which center's the content.

xapian-applications/omega/templates/query
jronak: I have missed id's on a lot of sections and fieldsets. It would be wise to call them query-fields

xapian-applications/omega/templates/query
```diff
@@ -149,7 +160,14 @@ ${Preserve any B filters used - this needs modifying if you modify this template
   to have B on radio buttons, SELECT, etc}
 $map{$cgilist{B},<input type=hidden name=B value="$html{$_}">}
 </form>
-<hr><div align=right><i><small><a href="https://xapian.org/">$html{$version}</a></small></i></div>
+<hr>
+<footer align=right>
```
jronak: Will undo

xapian-applications/omega/templates/query
```diff
@@ -149,7 +160,14 @@ ${Preserve any B filters used - this needs modifying if you modify this template
...
+<emp>
+<small>
```
jronak: Will fix it using CSS later. Shall add a class name to the footer

xapian-applications/omega/templates/query
```diff
@@ -149,7 +160,14 @@ ${Preserve any B filters used - this needs modifying if you modify this template
   to have B on radio buttons, SELECT, etc}
 $map{$cgilist{B},<input type=hidden name=B value="$html{$_}">}
 </form>
-<hr><div align=right><i><small><a href="https://xapian.org/">$html{$version}</a></small></i></div>
+<hr>
+<footer align=right>
+<emp>
```
jronak: It should be _em_ tag

xapian-applications/omega/templates/query
jronak: > In HTML5, the <hr> tag defines a thematic break. 
Should we replace it with something else

xapian-applications/omega/templates/query
jaylett: I think it's better to rip this stuff out, then come up with new styling when we add CSS (and if we have to add more hooks for selectors at that point, we can). The existing styling is pretty old school ;-)

xapian-applications/omega/templates/query
jaylett: html5 is a bit wishy-washy here; I'd only use `hr` in narrative text, like you get in novels within a chapter when some time has passed or something. Where we're moving between sections of the document, we should be able to style off existing elements.

(Note that my comment referred to a `br` not an `hr`.)

xapian-applications/omega/templates/query
jaylett: You still don't need it; emphasis only makes sense within running text (prose), which this is not. It's not being emphasised compared to anything around it, so just drop it.

xapian-applications/omega/templates/query
jronak: Names for different parts would be _query-Info_ and _response-info_

jronak:
@jaylett I have taken a note of all the things mentioned in the review at the moment. I will reset the head and start from scratch. Will add comments and todo's wherever necessary, as we have seen above that alot of them need to be styled using css. 

Thanks for your time and being very helpful.

xapian-applications/omega/templates/query
```diff
@@ -49,17 +50,18 @@ document.write("<span title=\""+D+" "+T+"\">Modified:<br><b>"+D+"</b></span>");
 // -->
 </script>
 </head>
+
 <body bgcolor="#ffffff">
```
ojwb: Probably that `bgcolor` should go, as that's really presentational (also it's not great to set the background colour but not the text colour and modern browsers default to white anyway).

jronak:
Updated PR, please review.

xapian-applications/omega/templates/query
```diff
@@ -58,7 +58,9 @@ if (!("autofocus" in document.createElement("input")))
  document.getElementById("omega-autofocus").focus();
 </script>
 <input type=submit value="Search">
-$if{,$include{inc/anyalldropbox},$include{inc/anyallradio}}
+${Todo: Replace radios with search buttons}
```
jaylett: Would buttons be the right approach here? Right now, you can toggle between the options. Would you want to replace a single search button with "search" and "search (AND)" or something?

xapian-applications/omega/templates/query
```diff
@@ -49,9 +49,10 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 // -->
 </script>
 </head>
-<body bgcolor="#ffffff">
```
jaylett: Each commit should do one thing; this one does two!

xapian-applications/omega/templates/query
```diff
@@ -61,6 +62,8 @@ if (!("autofocus" in document.createElement("input")))
 ${Todo: Replace radios with search buttons}
 $if{,$include{inc/anyalldropbox},
 <fieldset>$include{inc/anyallradio}</fieldset>}
+${Search Control(End)}
```
jaylett: I wouldn't bother putting "end" indicators in; there's no nesting so it's obvious when one has finished because the next one has started.

xapian-applications/omega/templates/query
```diff
@@ -58,7 +58,9 @@ if (!("autofocus" in document.createElement("input")))
  document.getElementById("omega-autofocus").focus();
 </script>
 <input type=submit value="Search">
-$if{,$include{inc/anyalldropbox},$include{inc/anyallradio}}
+${Todo: Replace radios with search buttons}
+$if{,$include{inc/anyalldropbox},
+<fieldset>$include{inc/anyallradio}</fieldset>}
```
jaylett: If `inc/anyallradio` gets a fieldset, `inc/anyalldropbox` probably should do also. (It's an example of another way of doing things, it won't get used automatically.)

This leaves other input elements outside fieldsets. The search box & submit button probably make sense to wrap as well. And topterms should almost certainly be a fieldset.

You should consider adding a legend to the fieldsets (we can style it away later if necessary).

(You should have put the classnames on this in the same commit. The change is that you're improving this part of the HTML, not that you're adding "fieldset" in.)

xapian-applications/omega/templates/query
```diff
@@ -44,7 +44,7 @@ function write_date(time_t) {
 var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
-document.write("<span title=\""+D+" "+T+"\">Modified:<b>"+D+"</b></span>");
+document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time></span>");
```
jaylett: If we're going to keep a `span` in there, it should have a class to distinguish it. Also, I suggested keeping the `title` but on `time` so the hover state works, and additionally I pointed out that the syntax for `datetime` is different to that used for the `title` attribute previously.

xapian-applications/omega/templates/query
```diff
@@ -22,13 +22,13 @@ $def{S,<img src="/icons/omega/page-$2s.png" $opt{a} alt=$1>}
...
 }
 
-$def{PREV,$if{$ne{$topdoc,0},<input type=submit name="&lt;" value="Previous">}}
+$def{PREV,$if{$ne{$topdoc,0},<input class="nav-button" type=submit name="&lt;" value="Previous">}}
 
-$def{PAGE,<input type=submit name="[" value="$1">}
+$def{PAGE,<input class="nav-button" type=submit name="[" value="$1">}
 
-$def{SPAGE,<input type=submit name="[" value="$1" disabled=disabled>}
+$def{SPAGE,<input class="nav-button" type=submit name="[" value="$1" disabled=disabled>}
```
jaylett: We've be better off turning the list of page controls into an actual list, or something. (Which would mean you'd end up with something like `<ul class='pagination'><li class='previous'>$PREV</li>…</ul>`.) Getting the details right is a bit fiddly, and the current usage isn't completely clear, so I'd start by adding comments where these are used, near the bottom of the file.

xapian-applications/omega/templates/query
```diff
@@ -52,50 +52,55 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 <body>
 <form name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
+<div class="query-container">
 ${Search Control(Start) - shows search button, options and textbox}
-<input id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
+<input class="query-textbox" id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
```
jaylett: You almost certainly don't need this for CSS styling. `input[type=search]` will probably do, as would `input[name=P]` or for that matter `input#omega-autofocus`.

It isn't needed for purely semantic reasons, as it should be obvious to anyone reading the code (at least the output HTML) what this input does.

xapian-applications/omega/templates/query
```diff
@@ -52,50 +52,55 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 <body>
 <form name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
+<div class="query-container">
```
jaylett: Either this should be a fieldset, or you should put a fieldset inside it (wrapping `input#omega-autofocus` and `input[type=submit]`). In the latter case, this should still be a fieldset, because you can nest them.

(Alternatively, put a fieldset round the two main search controls, and don't put anything around the whole lot.)

xapian-applications/omega/templates/query
```diff
@@ -52,50 +52,55 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
...
 ${Search Control(Start) - shows search button, options and textbox}
-<input id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
+<input class="query-textbox" id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
 <script>
 if (!("autofocus" in document.createElement("input")))
  document.getElementById("omega-autofocus").focus();
 </script>
-<input type=submit value="Search">
+<input class="query-search-button" type=submit value="Search">
```
jaylett: It's more common (and flexible) these days to use `<button type='submit'>Search</button>`. You almost certainly don't need a class name on this element.

xapian-applications/omega/templates/query
```diff
@@ -103,7 +103,7 @@ $if{$field{size},<span title="$html{$field{size}} bytes">Size: <b>$html{$filesiz
 </div>
 </div>
 </td>
-<td><b><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></b>
+<td><header><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></header>
```
jaylett: I didn't know that `td` counted as a sectioning root. Cool :-)

However: this shouldn't be a `table` at all, because it's semantically a list: it should be `ol`. (The table is just for layout.) Once you've done that, you can't use a `header` here.

xapian-applications/omega/templates/query
```diff
@@ -52,50 +52,55 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
...
 ${Todo: Replace radios with search buttons}
 $if{,$include{inc/anyalldropbox},
-<fieldset>$include{inc/anyallradio}</fieldset>}
+<fieldset class="query-options">$include{inc/anyallradio}</fieldset>}
+</div>
 ${Search Control(End)}
 ${Suggestions(Start) - shows suggested terms to improve query}
 $if{$opt{topterms},
- <div title="Suggested terms to add to your query"
+ <div class="query-suggested-container" title="Suggested terms to add to your query"
```
jaylett: Shouldn't this be a `fieldset`?

xapian-applications/omega/templates/query
```diff
@@ -52,50 +52,55 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
...
+ $map{$topterms,<span class="query-suggested-element" style="white-space:nowrap"><input type=checkbox name=X value="$html{$prettyterm{$_}}" onClick="C(this)">$html{$prettyterm{$_}}</span> }
  <noscript><input type=hidden name=ADD value=1></noscript>
  </div>
 }
-$if{$suggestion,Did you mean: <a href="$html{$or{$env{SCRIPT_NAME},omega}?DB=$url{$dbname}&P=$url{$suggestion}$if{$ne{$hitsperpage,10},&HITSPERPAGE=$hitsperpage}$if{$fmt,&FMT=$url{$fmt}}$if{$cgi{COLLAPSE},&COLLAPSE=$url{$cgi{COLLAPSE}}}$map{$relevants,&R=$_}$if{$cgi{THRESHOLD},&THRESHOLD=$url{$cgi{THRESHOLD}}}$map{$cgilist{B},&B=$url{$_}}&SPELL=0}">$html{$suggestion}</a>}
+$if{$suggestion,Did you mean: <a class="query-suggestion" href="$html{$or{$env{SCRIPT_NAME},omega}?DB=$url{$dbname}&P=$url{$suggestion}$if{$ne{$hitsperpage,10},&HITSPERPAGE=$hitsperpage}$if{$fmt,&FMT=$url{$fmt}}$if{$cgi{COLLAPSE},&COLLAPSE=$url{$cgi{COLLAPSE}}}$map{$relevants,&R=$_}$if{$cgi{THRESHOLD},&THRESHOLD=$url{$cgi{THRESHOLD}}}$map{$cgilist{B},&B=$url{$_}}&SPELL=0}">$html{$suggestion}</a>}
 ${Suggestions(End)}
 ${Query metadata(start) - shows query performance or document size}
 $or{$html{$error},
+<div class="query-metadata">
```
jaylett: Best thing we can do here is probably to make the metadata into a sentence, turn this into a `p`, and then providing styling hooks to make it pretty. (Alternatively, we could turn it into a `dl` of metadata. They're a pain to style, though.)

xapian-applications/omega/templates/query
```diff
@@ -52,50 +52,55 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
...
 $if{$eq{$msize,0},
 $if{$query,No documents match your query,
 Searching $nice{$dbsize} documents
-},
+}</div>,
```
jaylett: Too soon! The match counts are part of the metadata too.

xapian-applications/omega/templates/query
```diff
@@ -52,50 +52,55 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
...
 $if{$not{$msizeexact},
  $nice{$add{$topdoc,1}}-$nice{$last} of about $nice{$msize} matches,
  $if{$and{$eq{$last,$msize},$eq{$topdoc,0}},
   All $nice{$msize} matches,
   $nice{$add{$topdoc,1}}$if{$ne{$add{$topdoc,1},$last},-$nice{$last}} of exactly $nice{$msize} matches}
-}
-$list{$map{$queryterms,$list{$html{$uniq{$unstem{$_}}},<b>,</b>/<b>,</b>}:&nbsp;$nice{$freq{$_}}},Term frequencies: ,$. ,}
+}</div>
+<div class="query-performance">
```
jaylett: Still metadata.

xapian-applications/omega/templates/query
```diff
@@ -104,17 +109,19 @@ $html{$date{$field{modtime},%Y-%m-%d}}</time></span>
 </noscript>
 }
 $if{$field{language},Language: <b>$html{$field{language}}</b>}
-$if{$field{size},<span title="$html{$field{size}} bytes">Size: <b>$html{$filesize{$field{size}}}</b></span>}
+$if{$field{size},<div title="$html{$field{size}} bytes">Size: <b>$html{$filesize{$field{size}}}</b></div>}
```
jaylett: Span and div are not equivalent. This will probably become clearer when you eliminate the table structure.

xapian-applications/omega/templates/query
```diff
@@ -104,17 +109,19 @@ $html{$date{$field{modtime},%Y-%m-%d}}</time></span>
...
 </div>
 </div>
 </td>
-<td><header><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></header>
-$highlight{$snippet{$field{sample}},$terms}$if{$field{sample},...}
+<td><header class="list-item-header"><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></header>
+<span class="list-item-content" style="display: block;">
+$highlight{$snippet{$field{sample}},$terms}$if{$field{sample},...}</span>
 <a href="$html{$field{url}}">$html{$field{url}}</a>
+<span class="list-item-meta">
```
jaylett: Don't add this until we now we need it. Spans aren't particularly semantic.

xapian-applications/omega/templates/query
jronak: I have planned to add some neat styling even to the textbox, says something similar to google search textbox. 
True I could use the id, but  I wanted a generic class name as the css file could be shared with multiple templates or even multiple text boxes could be used.

xapian-applications/omega/templates/query
jronak: Again, planned to add hover effect on the buttons. So they do need a classname. 
Also the radio buttons option should be changed to buttons, I have added a todo in comment todo

xapian-applications/omega/templates/query
jronak: The <fieldset> tag is used to group related elements in a form. - W3

It just shows some text and its nothing related to form actually nor does a section make sense as it is not a main content of the page and it doesn't have a header. So I think it makes sense for it to stay as a div

xapian-applications/omega/templates/query
jronak: Haha dl would be a pretty bad. Yes we could combine the performance and document size line into one and add them under the same container. Let it stay as a div ?, as query performance has inner elements and it makes sense to put them under div.

xapian-applications/omega/templates/query
jronak: Yes I have thought of it as a separate props, query-metadata and query-performance.  I'll combine them into a single line or so

xapian-applications/omega/templates/query
jronak: True, it was added to reuse the list-item-meta, forgot to add the classs. Will revert it back to span

xapian-applications/omega/templates/query
jronak: Again adding some styles to metadata of the list

xapian-applications/omega/templates/query
jronak: There is a fieldset wrapping radio buttons. Why fieldset here as it doesn't have related elements ?, it just contains the actual form elements.

xapian-applications/omega/templates/query
jronak: My bad!. Will fixup this one.

xapian-applications/omega/templates/query
jronak: Okay will some comments. But James, how can the structure of navigation be improved, as turning this into list would be bad at least visually.

jronak:
Updated PR. I have added few CSS to the query container to style the search and background as floating elements. Please review.
xapian
xapian2

xapian-applications/omega/templates/query
jaylett: You can style a list to look like almost anything. Lots of pagination controls are lists.

xapian-applications/omega/templates/query
jaylett: Make that argument when you have CSS that justifies it, and keep things lean until then. Add classnames or ids when doing the semantic HTML pass only if they improve the semantics of the HTML.

xapian-applications/omega/templates/query
jaylett: `form#search button[type=submit]` would be fine. (I do think having something identifying the form is a good idea.)

xapian-applications/omega/templates/query
jaylett: It's utterly related to the form: it's a series of checkboxes for adding terms to add to the query (as the `div`'s title says).

xapian-applications/omega/templates/query
jaylett: `div` isn't semantic, and should be avoided if possible. What does the text look like here? Something like the following:

> Searching 71774 documents, 11-20 of about 200 matches

Then a break, and the query terms and term frequencies, then the search time. That's all close to narrative content, so just wrap it in `p`, which is otherwise identical to `div`.

ojwb:
Also, note the CI failure - the CSS file you've added has some whitespace irregularities:

https://travis-ci.org/xapian/xapian/jobs/197764336#L124

jronak:
Updated PR, fixed as suggested in the previous review. 

xapian3

jaylett:
@jronak I'm not going to be able to review this all right away, but I'll try to look over small batches of commits gradually as I have time.

In the meantime, can you please read the notes in our developer guide about commit messages? You should be using words like "Add" and "Remove", rather than "Added" and "Removed". You also aren't being consistent with capitalising the start of the commit summary, which would help people reading commits in future.

(You also don't need to link to the wiki for every commit, which may save you some time. We'll want to squash this work down to fewer commits, so don't worry about changing those parts for now.)

I think your commits are now a bit too small -- sorry, this is partly my fault for asking you to break them up. So for instance it'd be fine to have one commit which includes changes to the HTML and the CSS rules those support, for a part of the page (I'd probably have one commit for everything before the results, one for the results themselves, and one for everything after). Getting the right size is mostly a matter of experience, although there's also some taste involved, so for instance providing each is a complete change by itself it doesn't matter if you use three or five commits to add styling. However having too many little commits makes it as hard to figure out what's going on as one giant one!

xapian-applications/omega/templates/query
```diff
@@ -52,7 +52,7 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 </script>
 </head>
 <body>
-<form name=P method=get
+<form id="query-form" name=P method=get
```
jaylett: I'd probably just use `id=query`. The CSS selector `form#query` reads better to me than `form#query-form` :-)

xapian-applications/omega/templates/query
```diff
@@ -56,7 +56,7 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\"><b>"+D+"</b></time
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
 ${Search Control - shows search button, options and textbox}
 <fieldset class="query-search">
-<input class="query-search-textbox" d="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
+<input class="query-search-textbox" d="omega-autofocus" type=search name=P value="$html{$query}" placeholder="Search..." size=65 autofocus>
```
jaylett: You could (and I think should) use `&hellip;` instead of three literal dots in the placeholder.

xapian-applications/omega/templates/query
```diff
@@ -108,8 +108,8 @@ $html{$date{$field{modtime},%Y-%m-%d}}</time></span>
 </noscript>
 }
 $if{$field{language},Language: <b>$html{$field{language}}</b>}
-$if{$field{size},<span title="$html{$field{size}} bytes">Size: <b>$html{$filesize{$field{size}}}</b></span>}
-<div title="Mark this document as relevant" style="float:left;background-color:#cfc;border:1px solid green;margin-top:2px;">
+$if{$field{size},<br/><span title="$html{$field{size}} bytes">Size: <b>$html{$filesize{$field{size}}}</b></span>}
```
jaylett: You've introduced the `<br/>` here. Two things: under html5 it should just be `<br>`, and you also shouldn't need it (because you can do things like setting the subsequent element to `display: block;` to achieve the line break).

(We shouldn't attempt to make this "XHTML5" compatible, in my view. It's a pain, and not useful to many people.)

xapian-applications/omega/templates/query
```diff
@@ -96,8 +96,8 @@ ${Results}
 <table>
 $hitlist{<tr class="list-item-container"><td class="list-item-info">
 ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" height=16 width=32>}
-<div class="list-item-progress-container" title="$percentage%" style='float:left;width:60px;height:10px;border:1px solid black;margin-top:4px;'>
-<div class="list-item-progress" style='width:$div{$mul{$percentage,6},10}px; height:10px; background-color: red;'>
+<div class="list-item-progress-container" title="$percentage%" >
```
jaylett: You can remove the trailing space before `>` here.

jronak:
@jaylett Not a problem please take your time. I have updated the PR to include the following changes: 
Commit Messages: I have fixed the commit message, in fact, I followed the guideline for few commits then lost it.
Commit Size: As of now, I have squashed similar looking commits. As you said the experience will get me a better idea of the right size of a commit, hope to get there soon.
Styling: I have fixed the styling for the possible elements, but elements such as progress bar in the results row have its own class as of now.

ojwb:
> elements such as progress bar in the results row have its own class as of now

That's not a "progress bar" - it's a graphical indication of the result's relevance (or at least what the weighting scheme in use gives as its relevance).

jronak:
Oh thanks!, I hadn't read the weights in docs that time that's why I called it progress bar(UI)

jronak:
@jaylett Any possible suggestion or review ?

jaylett:
@jronak sorry, I haven't had time to review this yet. At this point not sure when I'm going to be able to get to it.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
 input[type=image] {
   border: 0px;
   }
```
jaylett: Should this existing rule be in the new `styles.css`?

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
+$include{styles.css}
 </style>
 <script type="text/javascript"><!--
 function f2(n) {return (n>9?"":"0")+n;}
 function write_date(time_t) {
 var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
-document.write("<span title=\""+D+" "+T+"\">Modified:<br><b>"+D+"</b></span>");
+document.write("<span>Modified:<br/><time datetime=\""+D+" "+T+"\">"+D+"</time></span>");
```
jaylett: If you want a line break inside the `span`, you can do that with something like `span > time { display: block; }`.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
 </head>
-<body bgcolor="#ffffff">
-<form name=P method=get
+<body>
+<form id="query-form" name=P method=get
 action="$html{$or{$env{SCRIPT_NAME},omega}}" target="_top">
-<center>
-<input id="omega-autofocus" type=search name=P value="$html{$query}" size=65 autofocus>
+${Search Control - shows search button, options and textbox}
+<fieldset class="query-search">
```
jaylett: It seems weird to me to use `query-search` as the class for all of `fieldset` and the two `input`s. You could alternatively just put `query-search` on the `fieldset`, and rewrite the CSS rules targeting the inputs as (for instance) `fieldset.query-search > input[type=search]`.

I'd probably just use `query` as the class. (I'd probably also use `search` as the `id` for the `form` itself, so you'd then be able to target things like `form#search fieldset.query > input[type=submit]`, which avoids the duplication of `form#query-form fieldset.query-search > input.query-search[type=submit]`.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
  document.getElementById("omega-autofocus").focus();
 </script>
-<input type=submit value="Search">
-<hr>
-$if{,$include{inc/anyalldropbox},$include{inc/anyallradio}}
+<input class="query-search" type=submit value="Search">
+</fieldset>
+${Todo: Replace radios with search buttons}
+$if{,$include{inc/anyalldropbox},
+<fieldset class="query-options">$include{inc/anyallradio}</fieldset>}
```
jaylett: Similar, this could be just `class=options`; below `class=suggested-terms` and so on.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
 if (!("autofocus" in document.createElement("input")))
  document.getElementById("omega-autofocus").focus();
 </script>
-<input type=submit value="Search">
-<hr>
-$if{,$include{inc/anyalldropbox},$include{inc/anyallradio}}
+<input class="query-search" type=submit value="Search">
+</fieldset>
+${Todo: Replace radios with search buttons}
+$if{,$include{inc/anyalldropbox},
```
jaylett: You want a `fieldset` round these no matter whether it's a dropbox (select) or radio buttons. I'd put the fieldset round the `$if{}`.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
+<fieldset class="query-options">$include{inc/anyallradio}</fieldset>}
+${Suggestions - shows suggested terms to improve query}
 $if{$opt{topterms},
- <div title="Suggested terms to add to your query"
+ <fieldset class="query-suggested-container" title="Suggested terms to add to your query"
  style="text-align:left;background:#cfc;border:1px solid green;padding:2px;font:11px verdana$. arial$. helvetica$. sans-serif;">
```
jaylett: These style rules can go into the CSS now.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
+}</p>,
 $if{$not{$msizeexact},
  $nice{$add{$topdoc,1}}-$nice{$last} of about $nice{$msize} matches,
  $if{$and{$eq{$last,$msize},$eq{$topdoc,0}},
   All $nice{$msize} matches,
   $nice{$add{$topdoc,1}}$if{$ne{$add{$topdoc,1},$last},-$nice{$last}} of exactly $nice{$msize} matches}
 }
-<hr>
-</center>
-$list{$map{$queryterms,$list{$html{$uniq{$unstem{$_}}},<b>,</b>/<b>,</b>}:&nbsp;$nice{$freq{$_}}},Term frequencies: ,$. ,}
```
jaylett: Even if you want to drop the `b` elements, there's a `/` in the middle which was an output character separating the terms. The documentation for that variant of `$list{}` says:

> "$list{LIST,List ,$. ,.}" = "List 1, 2, 3, 4."

What's happening here is that `$unstem{}` returns a list (because multiple words may stem to the same term), which we want to display as either **word** or **word/alternate** (or **word/alternate/another**, and so on). This then gets the frequency of that term bolted on, and (using the outer `$list{}`) ends up as something like:

> Term frequencies: **word**: 200, **word/alternate**: 1,024

(Dropping the bolding makes sense to me, but the `/` is still important.)

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
-<br><small>Search took $time seconds</small>
+<br/>
+$list{$map{$queryterms,$list{$html{$uniq{$unstem{$_}}},,,}:&nbsp;$nice{$freq{$_}}},Term frequencies: ,$. ,}
+Search took $time seconds
```
jaylett: You've dropped the `<br>` that was separating these two, which is the right thing to do, but now there's nothing to pick them apart, so we'll end up with the "Search took…" coming straight after the term frequencies. I'd probably think of this entire section (query metadata) as one paragraph, and turn it into narrative where possible, ie sentences separated by full stops.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
+</p>
+${Results}
 <table>
-$hitlist{<tr><td valign=top>
+$hitlist{<tr><td class="result-metadata">
 ${<img src="/icons/omega/score-$div{$percentage,10}.png" alt="$percentage%" height=16 width=32>}
-<div title="$percentage%" style='float:left;width:60px;height:10px;border:1px solid black;margin-top:4px;'>
-<div style='width:$div{$mul{$percentage,6},10}px; height:10px; background-color: red;'>
+<div class="result-metadata-progress-container" title="$percentage%" >
```
jaylett: Again, you don't need such long classnames; use more precise structural selectors instead: `tr.result > td.metadata > div.relevance-container` or similar.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
+<div class="result-metadata-progress" style='width:$div{$mul{$percentage,6},10}px;'>
```
jaylett: This is probably a separate piece of work, but we'd be better off coming up with a way of bounding the width of the container (say by putting CSS rules against the table column) and then just setting `width: $percentage%` here.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
 </div></div>
-<div style='float:left;margin-top:2px;font-size:x-small;'>
+<div class="result-metadata">
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
-<span title="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
-Modified:<br><b>$html{$date{$field{modtime},%Y-%m-%d}}</b></span>
-</noscript><br>
+<span>Modified:<br/><time datetime="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
```
jaylett: In this, and other situations where you're introduced `datetime`, you must:

 * provide the `datetime` attribute in the [correct format](https://www.w3.org/TR/html51/infrastructure.html#dates-and-times), which is for computers
 * retain the `title` on the container, since that's for humans

Again, you can avoid `br` elements here.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
+$html{$date{$field{modtime},%Y-%m-%d}}</time></span>
+</noscript>
 }
-$if{$field{language},Language: <b>$html{$field{language}}</b><br>}
-$if{$field{size},<span title="$html{$field{size}} bytes">Size: <b>$html{$filesize{$field{size}}}</b></span><br>}
-<div title="Mark this document as relevant" style="float:left;background-color:#cfc;border:1px solid green;margin-top:2px;">
+$if{$field{language},Language: $html{$field{language}}}
+$if{$field{size},<br/><span title="$html{$field{size}} bytes">Size:$html{$filesize{$field{size}}}</span>}
```
jaylett: This metadata list could do with some structure. Either `dl` or `ul` would be appropriate.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
 </div>
 </div>
 </td>
-<td><b><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></b><br>
-<small>$highlight{$snippet{$field{sample}},$terms}$if{$field{sample},...}</small><br>
-<a href="$html{$field{url}}">$html{$field{url}}</a><br>
-<small>
+<td class="result-data"><header class="result-data"><a href="$html{$field{url}}">$html{$or{$field{caption},$field{title},$prettyurl{$field{url}},Untitled}}</a></header>
+<span class="result-data" style="display: block;">
+$highlight{$snippet{$field{sample}},$terms}$if{$field{sample},...}</span>
```
jaylett: A `blockquote` would probably be good for surrounding the snippet.

xapian-applications/omega/templates/query
```diff
@@ -37,87 +37,96 @@ $if{$opt{topterms},$include{inc/toptermsjs}}
...
+<a class="result-data" href="$html{$field{url}}">$html{$field{url}}</a>
+<span class="result-data">
 matching:
-<i>$list{$map{$terms,$html{$prettyterm{$_}}},</i>$. <i>,</i> and <i>}</i></small>${for lynx:}<p></td></tr>}
+<i>$list{$map{$terms,$html{$prettyterm{$_}}},</i>$. <i>,</i> and <i>}</i>${for lynx:}</span></td></tr>}
```
jaylett: If you're going to drop the shim `p` element that improves layout for lynx, then you should drop the comment introducing it as well.

xapian-applications/omega/templates/query
```diff
@@ -132,9 +141,9 @@ $SPAGE{$thispage}
 $map{$range{$add{$thispage,1},$opt{y}},$PAGE{$_}}
 $if{$ne{$opt{y},$lastpage},...}
```
jaylett: This `...` is currently naked within the `ul`, which makes it harder to restyle subsequently. You can wrap it in an `li` with a suitable classname to distinguish from the pagination controls.

xapian-applications/omega/templates/query
```diff
@@ -149,7 +158,7 @@ ${Preserve any B filters used - this needs modifying if you modify this template
   to have B on radio buttons, SELECT, etc}
 $map{$cgilist{B},<input type=hidden name=B value="$html{$_}">}
 </form>
-<hr><div align=right><i><small><a href="https://xapian.org/">$html{$version}</a></small></i></div>
+<footer><i><a href="https://xapian.org/">$html{$version}</a></i></footer>
```
jaylett: You don't need the `i` element in there any more.

xapian-applications/omega/templates/styles.css
```diff
@@ -0,0 +1,128 @@
...
+}
+
+ul.pagination li input:hover {
+    background-color: #546E7A;
+}
+
+footer {
+    float: right;
+    margin: 2px;
+}
```
jaylett: It's always a good idea (unless the file format forbids it) to have a newline at the end of the file, because git works more smoothly that way when people edit around there subsequently.

xapian-applications/omega/templates/query
jronak: A description would look weird with the blockquote. Even though the text is from another source, we should use span.

xapian-applications/omega/templates/query
jronak: Fixed it. 
Is it normal to see repeated terms in terms frequency ?

xapian-applications/omega/templates/query
jronak: Could you elaborate, I quite didn't understand.

xapian-applications/omega/templates/query
jronak: I feel this would lead to adding more styles which could be unnecessary. Also, what is the significance of the checkbox in the metadata ?

xapian-applications/omega/templates/query
jaylett: No, my example was a little lazy. Should have been something like:

> Term frequencies: **word**: 200, **word2/alternate**: 1,024

xapian-applications/omega/templates/query
jaylett: If you look at the output HTML, what I believe you're getting is something like (with different whitespace):

```
<ul class="pagination">
  <li>PREV LINK</li>
  <li>PAGE 1 LINK</li>
  ...
  </li>THIS PAGE</li>
```

You aren't allowed to put naked text inside a `ul` like that (although browsers will generally figure out a way of rendering it), and it's impossible to style. (It should also be `&hellip;`, really.)

xapian-applications/omega/templates/query
jaylett: It's making the semantics explicit, which is good for various tools and assistive technology, and the styles are pretty minimal (`ul.whatever > li { display: inline; }` might even be sufficient).

The checkbox is for selecting documents for the relevance feedback mechanism. It isn't documented properly, unfortunately :-(. There's a [little on it in the API docs](https://xapian.org/docs/apidoc/html/classXapian_1_1Enquire.html#aab6fba434e5ddbbea2adcab85ba5ba4a).

xapian-applications/omega/templates/query
```diff
@@ -52,32 +48,31 @@ document.write("<span>Modified:<time datetime=\""+D+" "+T+"\">"+D+"</time></span
 </script>
 </head>
 <body>
-<form id="query-form" name=P method=get
+<form id="form" name=P method=get
```
jaylett: `form#form` ;-)

Is there a reason not to use the id 'search' here?

xapian-applications/omega/templates/styles.css
```diff
@@ -8,19 +8,27 @@ fieldset {
     text-align: center;
 }
 
-fieldset.query-search, .query-metadata,
-.query-suggested-container, .query-options, .query-suggestion {
+fieldset.search, fieldset.suggested-terms,
```
jaylett: Since `fieldset.suggested-terms` has its own block, it's probably easier to copy the couple of rules that aren't overridden down, just for clarity. (It should work fine as is, of course, but there's nothing more annoying than editing CSS and discovering that there's another rule being masked by the current one.)

xapian-applications/omega/templates/query
jaylett: You can make `blockquote` look however you like. It's also not inline content (although you get away with it because it follows a `header`, which will probably be default `display: block`), so it would be `div` not `span` even if you don't use `blockquote`; the fact that you're explicitly setting those `.result-data` elements to `display: block;` kind of underscores that.

And if you're using `div`, you should use `blockquote`, because:

> [The blockquote element represents a section that is quoted from another source.](https://w3c.github.io/html/grouping-content.html#the-blockquote-element)

It's common in "real" web systems, with more complex CSS, to start with a CSS reset or normalisation, which is a set of rules that turns everything pretty much flat and boring, but provides consistency across browsers (since, for instance, the indent used by browsers for `blockquote` doesn't have to be standard across all of them). In this case, `blockquote` would look identical to `div` anyway.

ojwb:
There's been no response for nearly a year, so I think we have to assume this patch is abandoned.

We don't appear to have the required licensing agreement (https://trac.xapian.org/browser/git/xapian-core/HACKING#L1376) so this patch is not even something that somebody else can pick up and take forward.

PR #138, opened 2017-03-05
KMeans Clusterer

richhiey1996:
@jaylett I have tired to rebuild the commits so that they are easier to review. I will be closing the previous PR since I have opened this.


xapian-core/tests/api_cluster.cc
```diff
@@ -66,6 +66,43 @@ DEFINE_TESTCASE(euclidian_distance2, backend) {
...
+    tlg.add_documents(matches);
+    Xapian::Document doc1 = matches[0].get_document();
+    Xapian::Document doc2 = matches[1].get_document();
+    Xapian::CosineDistance d;
+    Xapian::Point x1, x2;
+    x1.initialize(tlg, doc1);
+    x2.initialize(tlg, doc2);
+    double sim = d.similarity(x1, x2);
+    TEST_REL (sim,>=,0);
+    TEST_REL (sim,<=,1);
```
jaylett: You could just collapse the previous test into this one.

xapian-core/cluster/kmeans.cc
```diff
@@ -0,0 +1,146 @@
...
+    LOGCALL_VOID(API, "KMeans::initialize_random()", cset);
+    set<docid> centroid_docs;
+    int size = docs.size();
+    srand(time(NULL));
+    for (unsigned int i = 0; i < k;) {
+	int x = rand() % size;
+	if (centroid_docs.find(x) == centroid_docs.end()) {
+	    Point a = docs[x];
+	    Centroid centroid;
+	    centroid.set_to_point(a);
```
jaylett: What isn't this `Centroid centroid(a);`, with a suitable constructor? (Or indeed `Centroid centroid(docs[x]);` to avoid the temporary.)

xapian-core/cluster/kmeans.cc
```diff
@@ -0,0 +1,146 @@
...
+
+void
+KMeans::initialize_random(ClusterSet &cset) {
+    LOGCALL_VOID(API, "KMeans::initialize_random()", cset);
+    set<docid> centroid_docs;
+    int size = docs.size();
+    srand(time(NULL));
+    for (unsigned int i = 0; i < k;) {
+	int x = rand() % size;
+	if (centroid_docs.find(x) == centroid_docs.end()) {
```
jaylett: If this isn't true, will the `for` loop terminate?

xapian-core/cluster/kmeans.cc
```diff
@@ -0,0 +1,146 @@
...
+	    Point a = docs[x];
+	    Centroid centroid;
+	    centroid.set_to_point(a);
+	    centroids.push_back(centroid);
+	    centroid_docs.insert(x);
+	    Cluster cluster1(centroid);
+	    cset.add_cluster(cluster1);
```
jaylett: Again, no temporary needed here.

xapian-core/cluster/kmeans.cc
```diff
@@ -0,0 +1,146 @@
...
+
+bool
+KMeans::converge(vector<Centroid> &previous, vector<Centroid> &current) const {
+    LOGCALL(API, bool, "KMeans::converge()", previous | current);
+    CosineDistance cosine;
+    for (unsigned int i = 0; i < k; ++i) {
+	Centroid &prev = previous[i];
+	Centroid &curr = current[i];
+	double dist = cosine.similarity(prev, curr);
+	if (dist > 0.0001)
```
jaylett: This is a magic number in the code — there should be some justification for it, and it probably also should be a `#define` (assuming it shouldn't be a parameter for `KMeans`).

xapian-core/cluster/kmeans.cc
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return true;
+}
+
+ClusterSet
+KMeans::cluster(MSet &mset) {
+    LOGCALL(API, ClusterSet, "KMeans::cluster()", mset);
+    unsigned int size = mset.size();
+    mset.fetch();
+    if (size <= 0)
```
jaylett: Why not use `assert(size <= 0)`?

xapian-core/include/xapian/cluster.h
```diff
@@ -258,6 +258,28 @@ class XAPIAN_VISIBILITY_DEFAULT Point : public PointType {
...
+    void set_to_point(Point &x);
+
+    /// This method divides the weight of terms in the centroid by 'size'
+    void divide(int size);
+
+    /// This method clears the terms and corresponding values of the centroid
+    void clear();
+
+    /// This method recalculates the magnitude of the centroid
+    void recalc_magnitude();
```
jaylett: Should this really be public? Maybe you need to add some more detail to the class docstring to make it clear how the `Centroid` should be used. What sort of code creates it? What sort calls these different methods? If some are done only during (eg) Cluster recalculation, then it could be protected, with a suitable `friend` declaration.

xapian-core/tests/api_cluster.cc
```diff
@@ -35,6 +35,7 @@ using namespace std;
 DEFINE_TESTCASE(euclidian_distance1, backend) {
     Xapian::Enquire enquire(get_database("apitest_cluster"));
     enquire.set_query(Xapian::Query("cluster") );
+
```
jaylett: This line change isn't needed?

richhiey1996:
I think since I made the mistake of working with old work, it would be best for me to go back, sync with upstream master and rebuild these commits (breaking down the refactoring and kmeans commit as much as possible). There a few more changes that I introduced to optimize for speed, so I can add that in from there.

xapian-core/tests/api_cluster.cc
richhiey1996: Yes that can be done.

xapian-core/cluster/kmeans.cc
richhiey1996: Thanks for pointing this out.

xapian-core/cluster/kmeans.cc
richhiey1996: This is something I discovered by running it with different threshold values. The clusters do not change or tend to change very minimally once the distances between centroids gets below 0.0001. I'm not yet sure whether this is the most optimal value though.

jaylett: For now, make it a `#define`, and it can be investigated later. Turning it into a define means you can name it and explain what it's doing there; otherwise it's unclear how we arrived at that number. (So comment the define too!)

jaylett:
@richhiey1996 yes, that makes sense. As you're doing that, think carefully about [the story you're telling in the PR](http://www.annashipman.co.uk/jfdi/good-pull-requests.html).

xapian-core/cluster/kmeans.cc
richhiey1996: Great!

jaylett:
@richhiey1996 I don't know what you've done here, but it's not what you intended. I think you're trying to merge everything from master into the old branch as well. Given how long it's been, you're probably better off creating a new branch with the old branch's commits rebased to master plus the new ones from this, targeting master.

richhiey1996:
Yes I have created a new branch and started a new PR. I guess I should close this. Opening #149 in place of this.

PR #139, opened 2017-03-06
Fix broken snippetmisc1 test

rsto:
The snippetmisc1 unit test currently breaks on master when ran on backend 'glass'. This seems to have slipped through when the snippet test patch was initially merged.

### DIFF
--- a/xapian-core/tests/api_snippets.cc
+++ b/xapian-core/tests/api_snippets.cc
@@ -178,7 +178,7 @@ DEFINE_TESTCASE(snippetmisc1, generated) {
     Xapian::MSet mset = enquire.get_mset(0, 6);
     TEST_EQUAL(mset.size(), 3);
     TEST_STRINGS_EQUAL(mset.snippet(mset[0].get_document().get_data(), 40, stem),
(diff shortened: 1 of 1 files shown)


ojwb:
It's not failing for me on current master:

```
$ ./runtest ./apitest -v -b glass snippetmisc1
Running test './apitest -v -b glass snippetmisc1' under eatmydata and valgrind
Running test: snippetmisc1... ok
```

I could believe that there's some edge case rounding effect here that caused it to fail with some platforms or compilers or something, but I'm furthered confused that travis seems happy with your change, but also seems happy with other recent PRs without this patch.

ojwb:
OK, so the mystery of why tests seem to pass with and without this patch is solved - the travis config was only running bindings tests at the point this branched off master.  That was fixed in 319c609ade6ab8830e0b21b86d15b778c4314b2c which I've cherry picked to my test PR branch (#159).

ojwb:
So my tests in #159 with CI testing of xapian-core fixed show that tests pass without your patch and fail with, so this doesn't seem a useful change to me and I'm going to close it.

If you're still seeing problems, please can you resubmit a patch against current git master (where the CI tests should actually get run).

PR #140, opened 2017-03-10
Adding query introspection on leaf node.

nihal95:
Adding query introspection on leaf node to get wqf and pos. Also modifying queryintro1 test method to test leaf node.

### DIFF
--- a/xapian-core/api/query.cc
+++ b/xapian-core/api/query.cc
@@ -234,6 +234,12 @@ Query::get_leaf_wqf() const
 	return internal->get_wqf();
 }
 
+Xapian::termpos
+Query::get_leaf_pos() const
+{
+	return internal->get_pos();
+}
+
 string
 Query::get_description() const
 {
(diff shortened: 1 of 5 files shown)


xapian-core/api/query.cc
```diff
@@ -228,6 +228,12 @@ Query::get_subquery(size_t n) const
     return internal->get_subquery(n);
 }
 
+Xapian::termcount
+Query::get_leaf_wqf() const
+{
+	return internal->get_wqf();
```
ojwb: Here (and also in the next file) the code in the function is over-indented - it should be indented by 4 spaces.

Set your editor to display tabs with width 8, to indent by 4 columns, and to tab-fill the indents and you should avoid this problem.

ojwb:
Looking good.

One bit of admin I need to take care of before merging - are you OK with the patch licensing requirements?

https://trac.xapian.org/browser/git/xapian-core/HACKING#L1392

nihal95:
Hey olly, I hold the copyrights to the code and dual-license them under each of the following licences:
GPL version 2 and all later versions and MIT/X license. Is this fine?

ojwb:
Squashed into 4fb286495452f96c077fc31d85aaed95f1ab9ce5 and pushed.

Thanks!

PR #141, opened 2017-03-11
Changed the way Features request statistics (Ticket #733 on Trac)

AyushP123:
Link to the ticket -> https://trac.xapian.org/ticket/733

Intended Purpose for this Ticket:
1. Remove Internal which was making sub classing Feature Difficult.
2. Stats should be computed once for all the Features.
3. Reduction in the Memory Consumed by Features and FeatureList.

The following changes were made.
1. Feature::Internal was removed and Replaced by FeatureList::Internal
2. FeatureList now computes all the Stats beforehand for each document and stores them to Populate   each Feature with the necessary Stats. It then computes the Feature
3. Like Weight each Feature tells FeatureList what Statistics it needs beforehand.
4. All references to Internal were removed from Feature Subclasses.
5. @ojwb Although we discussed removing feature_query , feature_db and feature_doc from Feature I thought it was still better to keep them as we will have to keep passing them to the get_value function by reference which would still create additional pointers.

What I think was achieved:
1. Removed Internal from Feature and made it a part of FeatureList.
2. All the Stats are now being computed once for all the Features.

What I think got worse.
1. Memory allocation. Now all the stats are stored in all the Features ( which was being done before) and FeatureList.
2. Storing feature_q

Scope for Improvement:
1. In all the Features I am storing the Stats and feature_db,feature_doc,feature_query in dummy variables which I think should be avoided but I am not sure on extra space it is costing. 
2. @jaylett I think that suggestion of creating a FeatureList Object within the create_feature_vectors function , computing all the stats and storing them in the object along with feature_db,feature_query and feature_doc will help save a lot of space. However without feature_db , feature_query and feature_doc class Feature is just a class with two virtual functions. Also if we use this method then I don't think the Feature subclasses need to tell FeatureList which stats it needs as it can extract the required stats from the FeatureList Object that was passed to it.


jaylett:
@AyushP123 I think by initially doing as we discussed and concentrating on the API (ie the header files), but then going into implementation, you've ended up with commits that don't entirely make sense. Each commit should make one, entire change. So changing a class declaration in a header file is rarely a complete change, because it's unlikely to compile and work without changes to the implementation as well.

Have you read our developer's guide? There's a section on making commits that you should find useful. Another thing it covers is commit messages: yours aren't formatted according to git standards, which makes them harder to work with (including here in github; you'll see that it's difficult to tell what each of the non-whitespace commits does).

You should be very careful that each commit message describes why a change is made rather than just the changes. For instance, for 3075a5fc28a5a20e1ec7c47a7ad9314e7da7663e you say "Added a space after comma and added two comments in featurelist.cc", but we can tell this from the diff. Why did you make those changes?

In that particular case, it's because those changes were left out of a previous commit. In that case, you can use `git commit --fixup <sha>` to make a commit that is intended to "fixup" (ie improve) a previous one, referenced by sha. (This then hooks into interactive rebase so you can automatically have git merge these commits together later.)

Finally, commits in a PR should tell the story of the change you're trying to make. Each commit should do one thing. That doesn't mean it will be tiny, but if the summary line for your commit contains the word "and", it's probably not been broken down sufficiently. 5b728e7d459f936039a0e16da3d5cc59f148c3b9 is an example of this (although again,

This PR is consequently hard to review right now. I suggest at a minimum that you squash commits together so there aren't any whitespace fixups (look into interactive rebase for this). But you also need to think in terms of the functional changes you're making to Xapian (whether at an API level, or in terms of implementation), and create the commits out of that.

Since you're going to have to rebuild commits and force push the branch (which makes continuing a review difficult, because of limitations with github), I'm going to stop reviewing now and give you a chance to work through this.

PR #142, opened 2017-03-12
Using std::unordered_set with SimpleStopper

richhiey1996:
Since Xapian now uses c++11, we can use std::unordered_set instead of std::set for faster identification of stopwords

### DIFF
--- a/xapian-core/include/xapian/queryparser.h
+++ b/xapian-core/include/xapian/queryparser.h
@@ -33,8 +33,8 @@
 #include <xapian/termiterator.h>
 #include <xapian/visibility.h>
 
-#include <set>
 #include <string>
+#include <unordered_set>
 
 namespace Xapian {
 
@@ -79,7 +79,7 @@ class XAPIAN_VISIBILITY_DEFAULT Stopper
 
 /// Simple implementation of Stopper class - this will suit most users.
 class XAPIAN_VISIBILITY_DEFAULT SimpleStopper : public Stopper {
-    std::set<std::string> stop_words;
+    std::unordered_set<std::string> stop_words;
 
   public:
     /// Default constructor.
--- a/xapian-core/queryparser/queryparser.cc
+++ b/xapian-core/queryparser/queryparser.cc
@@ -47,7 +47,7 @@ string
 SimpleStopper::get_description() const
 {
     string desc("Xapian::SimpleStopper(");
-    set<string>::const_iterator i;
+    unordered_set<string>::const_iterator i;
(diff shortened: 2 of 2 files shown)


jaylett:
@richhiey1996 this should really have been branched off latest master, not off a much older commit from last year.

The commit message should explain _why_ we're doing this. Does it improve performance? Does it enable some other feature? Is there a test we can write, or some documentation that needs updating?

richhiey1996:
Since Xapian now requires C++11, it would be better for us to use an std::unordered_set which implements a hash table (facilitating O(1) lookups and insertions) compared to the std::set which implements an RB tree (O(log(n)) lookups and insertions).
I realized the SimpleStopper class worked slower than unordered_map for stopword removal when working on KMeans, and had a brief discussion with @ojwb on IRC. Olly suggested to try and change it if possible.
There are no tests or documents that need to be updated because it's just about using std::unordered_set instead of std::set. It surely does improve performance when we have a lot of lookups and insertions.

richhiey1996:
Again, I shall sync to updated master and push these changes.

jaylett:
@richhiey1996 1. put that explanation in the commit message! :) 2. ideally run some performance tests to give an indication of the kind of improvement we're looking at; this is helpful for other people (but less important than ensuring that the change is properly explained).

richhiey1996:
Yes, I have added this to the new commit message. I have synced to latest master and thus opening a new PR. Closing this in view of #143

ojwb:
BTW, there's no need to close a PR and open another to change the proposed commits - just sort out the branch locally and then `git push --force` to update an existing PR (note: this is a particular situation in which force-pushing is OK - it's unhelpful in many cases, especially for branches people are actively collaborating on).  Starting a new PR splits the discussion up and needlessly churns the PR numbers.

richhiey1996:
@ojwb I guess I was working on master that wasn't up to date with recent commits, so it would have been harder to rebase to latest master and do the changes, than to branch off from latest master and create a new PR.
I do understand the problem it creates with the workflow though. Thanks for pointing that out :+1:

ojwb:
@richhiey1996 You can re-push a fresh branch to the PR (if you don't reuse the same name you'll have a situation where the local and remote branch names don't match, but that's not an inherent problem).

But usually you'd just use `git rebase master` with the PR branch checked out to rebase it onto current master, then `git push --force` to update the PR with the rebased branch.  If master has changes which clash with commits in your PR you'll have to resolve them, but you'd also need to do that to apply the changes to a fresh branch created from the latest master.

PR #143, opened 2017-03-12
Use std::unordered_set instead of std::set to speed up Xapian::Simple…

richhiey1996:
Since Xapian now requires C++11, it would be better for us to use an
std::unordered_set which implements a hash table (facilitating O(1) lookups
and insertions) compared to the std::set which implements an RB tree (O(log(n))
lookups and insertions). This will improve performance when we have a lot
of insertions or/and lookups

### DIFF
--- a/xapian-core/include/xapian/queryparser.h
+++ b/xapian-core/include/xapian/queryparser.h
@@ -33,8 +33,8 @@
 #include <xapian/termiterator.h>
 #include <xapian/visibility.h>
 
-#include <set>
 #include <string>
+#include <unordered_set>
 
 namespace Xapian {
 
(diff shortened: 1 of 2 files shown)


ojwb:
Thanks, rebased onto master and pushed.

This isn't entirely invisible to the use as `SimpleStopper::get_description()` will now report the terms in an arbitrary order rather than sorted, but I think that's acceptable.

richhiey1996:
Yes that shouldn't really be a problem since the position of a term within its termlist never carried too much information. And keeping up with C++11 is the better thing to do :)

ojwb:
I'd say the more compelling reason is performance.  C++11 didn't deprecate `std::set` so "keeping up" isn't a good reason - `std::set` is valid C++11 and is still the appropriate choice for some situations.  If you need to iterate the set in order then `std::set` is probably going to be a better option than having to copy and sort the keys.  And even if you don't O(log n) may always beat O(n) for bounded n.

There are also less obvious cases - e.g. for a set of `std::string` where you do a lot of look-ups for long strings which aren't present, then `std::unordered_set` has to process the whole of the string being looked up to hash it (so it's `O(key.length())`) but the comparisons used to search the tree implementing `std::set` may only need to check the first few characters in the string to conclude that it isn't present.

richhiey1996:
> C++11 didn't deprecate std::set so "keeping up" isn't a good reason

That's actually right! The more important reason is performance as you rightly explained! :+1:

PR #145, opened 2017-03-14
Release/1.4

santokhsdg:


ojwb:
I can only assume this was accidentally submitted.

Anyway, we're not going to merge all of master into RELEASE/1.4 as it would break the ABI, so closing.

PR #146, opened 2017-03-17
Change the way Features request statistics ( Ticket #733 )

AyushP123:
Link to the Ticket -> https://trac.xapian.org/ticket/733

Improvements Suggested in the Ticket:

1. Eliminate Feature::Internal and replace it by FeatureList::Internal as Presence of Internal makes it difficult to subclass Feature.
2. Using Weight as a reference Model FeatureList::Internal should calculate the statistics once and should populate the Feature Subclasses with only the stats they require.

What I think was achieved in this PR:

1. Feature::Internal was eliminated and replaced by FeatureList::Internal.
 2. FeatureList computes Statistics once per document and then populates the Features with the required statistics.

Problems:

1. Additional Memory space required as now FeatureList holds a copy of stats in addition to Features.

Improvements:

1. Another approach. Create a FeatureList Object, compute statistics and populate the object with them, pass this object to get_value function of all the features. Advantage is same running time but a lot less memory as We can eliminate storing stats in Features along with feature_query , feature_db and feature_doc.

Why this approach then. Because it is more similar to Weight and plus the work done by a user writing a new feature will be slightly less as compared to the approach above.


AyushP123:
@ojwb @ayshtmr @jaylett  ..... we should start discussion on this PR

jaylett:
@AyushP123 before someone starts reviewing this, can you please check that:

1. Source code is correctly formatting as per our standards
2. You aren't copying objects around a lot (for instance, you may be able to use `const` references if you aren't changing the "copy")

AyushP123:
@jaylett, @ojwb, @ayshtmr : I have reviewed this PR and it conforms to xapian coding standards. I have not copied objects, altough I have copied the pointer to feature in FeatureList::Internal::populate_feature. I have tested it on a custom set of documents and avoiding the recalculation of statistics reduces the time to compute features by more than half. Please review this PR and let me know your suggestions.

xapian-letor/feature/tfdoclenfeature.cc
```diff
@@ -41,9 +40,9 @@ TfDoclenFeature::get_values() const
 {
     LOGCALL(API, std::vector<double>, "TfDoclenFeature::get_values", NO_ARGS);
 
-    Query query = Feature::internal->feature_query;
-    map<string, long int> tf = Feature::internal->termfreq();
-    map<string, long int> doc_len = Feature::internal->doc_length();
+    Query query = Feature::feature_query;
+    map<string, long int> tf = Feature::termfreq;
```
jaylett: 1. Do you need to scope this to `Feature::` to get it to compile?
2. You've copied this map from `FeatureList::Internal` object to `Feature` object, and then you're copying it again within this method. We can at least eliminate one of these copies, if not both by judicious use of references.

xapian-letor/api/Makefile.mk
```diff
@@ -1,9 +1,11 @@
 noinst_HEADERS +=\
-	api/featurevector_internal.h
+	api/featurevector_internal.h\
+	api/featurelist_internal.h
```
ojwb: These lists of source files should be sorted to make it easy to locate an entry, so you've added this in the wrong place.

xapian-letor/api/Makefile.mk
```diff
@@ -1,9 +1,11 @@
...
 
 EXTRA_DIST +=\
 	api/Makefile
 
 lib_src +=\
 	api/featurelist.cc\
-	api/featurevector.cc
+	api/featurevector.cc\
+	api/featurelist_internal.cc
```
ojwb: Again, list no longer in sorted order.

xapian-letor/include/xapian-letor/featurelist.h
```diff
@@ -42,6 +42,12 @@ class XAPIAN_VISIBILITY_DEFAULT FeatureList {
     /// Vector containing Feature pointer objects. Each will be used to return feature value.
     std::vector<Feature *> feature;
 
+  protected:
+    /// @internal Class representing the FeatureList internals.
+    class Internal;
+    /// @internal Reference counted internals.
+    Xapian::Internal::intrusive_ptr<Internal> internal;
+
```
ojwb: You've turned this into a PIMPL class, but left the `feature` member in place rather than move it into the internal class.

xapian-letor/api/featurelist.cc
```diff
@@ -40,12 +41,14 @@ FeatureList::FeatureList()
     feature.push_back(new CollTfCollLenFeature());
     feature.push_back(new TfIdfDoclenFeature());
     feature.push_back(new TfDoclenCollTfCollLenFeature());
+    internal = new FeatureList::Internal();
```
ojwb: It's cleaner to assign the internals using a member initialiser list (i.e. `    : internal(new FeatureList::Internal())` before the opening `{` of the constructor.

If you consistently initialise the internal pointer like that in all the constructors, then you can rely on it being set in other methods.  Similarly below.

xapian-letor/api/featurelist.cc
```diff
@@ -100,11 +103,14 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset, const Xapian::Que
     for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
 	Xapian::Document doc = i.get_document();
 	std::vector<double> fVals;
```
ojwb: Xapian coding style is not to use camel-case names like `fVals`.  Please use something like `fvals` or `f_vals` instead.

xapian-letor/api/featurelist.cc
```diff
@@ -100,11 +103,14 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset, const Xapian::Que
...
-
+	// All stats are computed and stored in the corresponding variables.
+	internal->compute_statistics(letor_query, letor_db, doc);
 	for (std::vector<Feature*>::const_iterator it = feature.begin(); it != feature.end(); ++it) {
 	    (*it)->set_database(letor_db);
 	    (*it)->set_query(letor_query);
 	    (*it)->set_doc(doc);
+	    // Populates the Feature with required stats
+	    internal->populate_feature(*it);
 	    vector<double> values = (*it)->get_values();
```
ojwb: It's better to use a const reference here as it helps the compiler avoid a copy (and copying a vector is not cheap):

```
const vector<double>& values = (*it)->get_values();
```

Assigning a temporary to a const reference like this is special in C++ and means it lives for the scope of the reference - see https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/ for example.  Note especially that a non-const reference doesn't work.

xapian-letor/api/featurelist.cc
```diff
@@ -100,11 +103,14 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset, const Xapian::Que
...
 	    fVals.insert(fVals.end(), values.begin(), values.end()); // Append feature values
```
ojwb: The trailing comment makes this line too long (Xapian code avoids lines longer than 80 columns).

In general, I'd suggest putting the comment on the line before anyway - trailing comments are much easier to miss when scanning down through the code.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+#include <cstring>
+#include "debuglog.h"
+
+using namespace std;
+using namespace Xapian;
+
+void
+FeatureList::Internal::set_database(const Xapian::Database & db)
+{
+    LOGCALL_VOID(API, "FeatureList::set_database", db);
```
ojwb: This is not `FeatureList::set_database`!

I'd just drop this ( - logging in both the API class method and the internal method it calls mostly just adds verbosity to the log, but if you are going to log here, use the correct method name and change the category to something appropriate (not `API`).

There are further instances of this below.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+    featurelist_db = db;
+}
+
+void
+FeatureList::Internal::set_query(const Xapian::Query & query)
+{
+    LOGCALL_VOID(API, "FeatureList::set_query", query);
+    if (query.empty()) {
+	throw Xapian::InvalidArgumentError("Can't initialise with an empty query string");
```
ojwb: The argument isn't a query string, it's a query object.

Is an empty query object really invalid here, or just not very useful?

We don't reject all cases of query objects which can't match anything, so I'm not sure it really helps to just reject those which simplify upon construction to give `Xapian::Query::MatchNothing`.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+void
+FeatureList::Internal::set_doc(const Xapian::Document & doc)
+{
+    LOGCALL_VOID(API, "FeatureList::set_doc", doc);
+    featurelist_doc = doc;
+}
+
+void
+FeatureList::Internal::compute_statistics(const Xapian::Query & letor_query, const Xapian::Database & letor_db,
+					  const Xapian::Document & letor_doc)
```
ojwb: The first line is over 80 columns - wrap at the first comma as well as the second...

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+    Xapian::TermIterator docterms = featurelist_doc.termlist_begin();
+    for (Xapian::TermIterator qt = featurelist_query.get_terms_begin(); qt != featurelist_query.get_terms_end(); ++qt) {
+	docterms.skip_to(*qt);
+	if (docterms != featurelist_doc.termlist_end() && *qt == *docterms) {
+	    tf[*qt] = docterms.get_wdf();
+	} else {
+	    tf[*qt] = 0;
+	}
+    }
+    termfreq = tf;
```
ojwb: This copies the whole map - better to use:

```
std::swap(termfreq, tf);
```

That will just exchange the two, which is typically copying a handful of pointers and integers.  When `tf` goes out of scope on the next line, the old `termfreq` will be destroyed.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+
+    Xapian::TermIterator docterms = featurelist_doc.termlist_begin();
+    for (Xapian::TermIterator qt = featurelist_query.get_terms_begin(); qt != featurelist_query.get_terms_end(); ++qt) {
+	docterms.skip_to(*qt);
+	if (docterms != featurelist_doc.termlist_end() && *qt == *docterms) {
+	    tf[*qt] = docterms.get_wdf();
+	} else {
+	    tf[*qt] = 0;
+	}
+    }
```
ojwb: This won't work correctly in general.

The `TermIterator` over a document returns the terms in sorted order, but `Query::get_terms_begin()` returns the terms in query position order.  When you get a term from the query which was before an earlier one, you'll fail to find it in the document.

I think you can just use `Query::get_unique_terms_begin()` and `Query::get_unique_terms_end()` instead.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+void
+FeatureList::Internal::compute_inverse_doc_freq()
+{
+    std::map<std::string, double> idf;
+
+    for (Xapian::TermIterator qt = featurelist_query.get_terms_begin(); qt != featurelist_query.get_terms_end(); ++qt) {
+	if (featurelist_db.term_exists(*qt)) {
+	    long int totaldocs = featurelist_db.get_doccount();
+	    long int df = featurelist_db.get_termfreq(*qt);
+	    idf[*qt] = log10(totaldocs / (1 + df));
```
ojwb: This is doing an integer division, which means unnecessary rounding (and if the term indexes all the documents will result in `log10(0)`.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+
+void
+FeatureList::Internal::compute_inverse_doc_freq()
+{
+    std::map<std::string, double> idf;
+
+    for (Xapian::TermIterator qt = featurelist_query.get_terms_begin(); qt != featurelist_query.get_terms_end(); ++qt) {
+	if (featurelist_db.term_exists(*qt)) {
+	    long int totaldocs = featurelist_db.get_doccount();
+	    long int df = featurelist_db.get_termfreq(*qt);
```
ojwb: It's better to use the correct types rather than `long int`.  If Xapian is built with 64-bit docid and termcounts, then the correct types may actually be able to hold **larger** values than `long int` can.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+	if ((*dt)[0] != 'S') {
+	    // We've reached the end of the S-prefixed terms.
+	    break;
+	}
+	temp_count += dt.get_wdf();
+    }
+    len["title"] = temp_count;
+    len["whole"] = featurelist_db.get_doclength(featurelist_doc.get_docid());
+    len["body"] = len["whole"] - len["title"];
+    doc_length = len;
```
ojwb: Same comments about using the correct types and `std::swap()` to avoid a copy of the map.  Also below...

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+    termfreq = tf;
+}
+
+void
+FeatureList::Internal::compute_inverse_doc_freq()
+{
+    std::map<std::string, double> idf;
+
+    for (Xapian::TermIterator qt = featurelist_query.get_terms_begin(); qt != featurelist_query.get_terms_end(); ++qt) {
+	if (featurelist_db.term_exists(*qt)) {
```
ojwb: Rather than calling `term_exists()` and then `get_termfreq() when it does, it would be more efficient to just call `get_termfreq()` and if it returns zero, you know the term doesn't exist.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,186 @@
...
+void
+FeatureList::Internal::compute_collection_termfreq()
+{
+    std::map<std::string, long int> tf;
+
+    for (Xapian::TermIterator qt = featurelist_query.get_terms_begin(); qt != featurelist_query.get_terms_end(); ++qt) {
+	if (featurelist_db.term_exists(*qt))
+	    tf[*qt] = featurelist_db.get_collection_freq(*qt);
+	else
+	    tf[*qt] = 0;
```
ojwb: This `term_exists()` check is unnecessary and inefficient - `get_collection_freq()` returns zero for terms not in the DB.

xapian-letor/api/featurelist_internal.h
```diff
@@ -23,30 +23,58 @@
...
+    std::map<std::string, long int> collection_termfreq;
 
     /** This method finds the frequency of the query terms in the specified documents. This method is a helping method and statistics gathered through
      *  this method are used in feature value calculation. It return the frequency of the terms of query in std::map<string, long int> form.
      */
-    std::map<std::string,long int> termfreq() const;
+    void compute_termfreq();
 
     /** This method calculates the inverse document frequency(idf) of query terms in the database. It returns the idf of each term in
      *  std::map<string, double> form.
```
ojwb: It now has a `void` return type.  Similar issue for other methods here.

xapian-letor/api/featurelist_internal.h
```diff
@@ -85,12 +113,34 @@ class Feature::Internal : public Xapian::Internal::intrusive_base {
...
+     *  query is supplied
+     */
+    void set_query(const Xapian::Query & query);
+
+    /// Specify the document to use for feature building. This will be used by the Internal class.
+    void set_doc(const Xapian::Document & doc);
+
+    public:
+
+     /** This Method computes all the statistics and stores them in their corresponding variables
```
ojwb: Randomly capitalised word (`Method`).

xapian-letor/api/featurelist_internal.cc
AyushP123: It should be invalid here as the query object with empty string will be used to calculate the features. This function is called by the ranker and using an empty string for ranking the documents would further complicate the things.

xapian-letor/api/featurelist_internal.h
AyushP123: you mean updating the docstring corresponding to the functions ??

xapian-letor/api/Makefile.mk
AyushP123: you mean featurelist_internal.h should be placed in xapian-letor/include/xapian-letor folder or should I just swap their positions ??

xapian-letor/feature/tfdoclenfeature.cc
AyushP123: I cannot eliminate the copying from FeatureList::Internal to Feature as it saves a lot of time which would be otherwise spent in recomputing the statistics in Feature. I can avoid the copy here but since its a const function, either I cannot avoid copying or I remove the const. Which would be a better option here ??. In my opinion having Feature::get_value() as a const function makes no sense.

xapian-letor/api/featurelist_internal.cc
jaylett: Or we can short-cut the entire feature vector generation and ranking system, which would be more consistent with the rest of Xapian. In particular, I believe this will be used during reranking, so it would require people to detect when they have zero items in their MSet and not call LTR, which is fiddly and should be unnecessary. A re-ranked empty MSet is just the empty MSet again.

In other words, you're trying to avoid complexity in the library at the cost of complexity in every library user's code. This is a poor tradeoff.

(I'm sure we've had a similar question before and did this, and we should resolve it in the same way here also.)

xapian-letor/api/featurelist_internal.h
```diff
@@ -0,0 +1,153 @@
...
+	TERM_FREQUENCY = 1,
+	/// Number of documents in the RSet.
+	INVERSE_DOCUMENT_FREQUENCY = 2,
+	/// Average length of documents in the collection.
+	DOCUMENT_LENGTH = 4,
+	/// How many documents the current term is in.
+	COLLECTION_LENGTH = 8,
+	/// How many documents in the RSet the current term is in.
+	COLLECTION_TERM_FREQ = 16,
+    }stat_flags;
```
jaylett: Space needed here.

xapian-letor/api/featurelist_internal.h
```diff
@@ -0,0 +1,153 @@
...
+     *  len["title"];
+     *  len["body"];
+     *  len["whole"];
+     *  @endcode
+     *
+     */
+    void compute_collection_length();
+
+    /** This method calculates the frequency of query terms in the whole database.
+     *  The information is stored in collection_termfreq in and used during the feature calculation methods.s
```
jaylett: Unwanted 's'.

xapian-letor/api/featurelist_internal.h
```diff
@@ -0,0 +1,153 @@
...
+
+    /// Vector containing Feature pointer objects. Each will be used to return feature value.
+    std::vector<Feature *> feature;
+
+    /// This method computes all the statistics and stores them in their corresponding variables
+    void compute_statistics(const Xapian::Query & query, const Xapian::Database & db, const Xapian::Document & doc);
+
+    /// Populates the statistics needed by a Feature
+    void populate_feature(Feature *feature_);
+
```
jaylett: Unwanted blank line. I specifically asked you to look for these, so I'll stop here until you've had a chance to complete your improvements based on the last round of review.

xapian-letor/api/featurelist_internal.cc
ojwb: > It should be invalid here as the query object with empty string

Stop thinking about query strings here.  A query object doesn't hold a query string - it's a tree of query objects representing terms, operations to combine them, etc (which is sometimes but not always obtained by parsing a query string).

A key point is that `query.empty()` tests for a **subset** of query objects which won't match anything - `Query()` aka `Query::MatchNothing`, and some forms which are reduced to this during construction - e.g. `Query() & Query("a")` happens to be currently, but these reductions are mostly opportunistic, and exactly what gets reduced has evolved with time, and is likely to continue to do so.  Meanwhile it's trivially easy to construct queries which don't get reduced to `Query()` but will never match anyway - e.g. `Query("a") & ~Query("a")`.

By checking `query.empty()` here, you reject a rather arbitrary subset of queries which can't match anything, and that may change between versions which is somewhat confusing to the user.  Why is `Query() & Query("a")` invalid, but `Query("a") & ~Query("a")` isn't?  And in a future version, why does the latter suddenly become invalid?

But there are still queries which can't match anything you can't catch here, and also queries which might match something but don't with the database being searched (e.g. `Query("foo")` if term "foo" isn't present), so with the current check you're going to have to deal with this case anyway aren't you?

I think as James suggests, the correct test is not `query.empty()` but `mset.empty()`.  If there are no results, there's nothing to rerank and you should be able to short-cut everything at a higher level.

xapian-letor/api/Makefile.mk
ojwb: Just swap the positions.

jaylett:
@AyushP123 you seem to have force pushed this branch in the middle of a review. Please, please do not do this. It makes it considerably harder to review, because now I basically have to start from the beginning.

xapian-letor/ranker/ranker.cc
```diff
@@ -354,6 +354,9 @@ Ranker::labelcomparer(const FeatureVector & firstfv, const FeatureVector& second
 std::vector<Xapian::docid>
 Ranker::rank(const Xapian::MSet & mset, const string & model_key, const Xapian::FeatureList & flist)
 {
+    if (mset.empty()) {
+	throw Xapian::InvalidArgumentError("Empty MSet. No documents could be retrieved with the given Query.\n");
```
jaylett: This would be a lot easier to get right in #156 where you could just `return` from the method is `mset.empty()`. Throwing an exception is the wrong thing, because it means that every client of `Ranker::rank()` has to catch it and do something useful with it — and the something useful in this case is to _ignore the exception_ and do any normal processing on an MSet (which may include checking if it's empty).

ojwb: Here we can just `return std::vector<Xapian::docid>()` - that's a valid reranking of an empty `MSet` (and indeed the only possible reranking).

As a general point though, exception messages should not include a newline at the end.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -41,33 +40,35 @@ CollTfCollLenFeature::get_values() const
...
 
-    for (Xapian::TermIterator qt = query.get_terms_begin(); qt != query.get_terms_end(); ++qt) {
+    for (Xapian::TermIterator qt = feature_query.get_unique_terms_begin();
+	 qt != feature_query.get_terms_end(); ++qt) {
 	if ((*qt).substr(0, 1) == "S" || (*qt).substr(1, 1) == "S")
-	    value += log10(1 + ((double)coll_len["title"] / (double)(1 + coll_tf[*qt])));
+	    value += log10(1 + ((double)collection_length.find("title")->second
+				/ (double)(1 + collection_termfreq.find(*qt)->second)));
 	else
 	    value += 0;
```
ojwb: This `else` branch seems to serve no useful purpose, please remove (and similarly below).

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -41,33 +40,35 @@ CollTfCollLenFeature::get_values() const
...
-
     vector<double> values;
     double value = 0;
 
-    for (Xapian::TermIterator qt = query.get_terms_begin(); qt != query.get_terms_end(); ++qt) {
+    for (Xapian::TermIterator qt = feature_query.get_unique_terms_begin();
+	 qt != feature_query.get_terms_end(); ++qt) {
 	if ((*qt).substr(0, 1) == "S" || (*qt).substr(1, 1) == "S")
-	    value += log10(1 + ((double)coll_len["title"] / (double)(1 + coll_tf[*qt])));
+	    value += log10(1 + ((double)collection_length.find("title")->second
```
ojwb: `collection_length.find("title")->second` is going to be the same for every iteration - stick in it a variable before the loop and that'll make this overly long expression shorter and easier to read, and may also be faster (I suspect the compiler won't be able to see that this lookup can be hoisted).

A similar comment applies to the loops below too.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -41,33 +40,35 @@ CollTfCollLenFeature::get_values() const
...
+				/ (double)(1 + collection_termfreq.find(*qt)->second)));
```
ojwb: If this `find()` doesn't find anything, then you dereference the end iterator which is undefined behaviour.

Now that should be "impossible" if I follow the code correctly, but that relies on a high level invariant which is not explicitly asserted (that all terms involved have an entry in `collection_termfreq`).  It would be all too easy for someone changing this code to stop that being the case - this is an example of what's sometimes called "code smell": https://en.wikipedia.org/wiki/Code_smell

My thought is that we should stop filling in 0 entries into these maps when the term isn't present in `FeatureList::Internal::compute_termfreq()` etc.  Then here do the `find()` just before the actual calculation and assume zero if the entry isn't found.  That avoids the potential for a crash, and saves some space in the maps.  Smaller maps also means faster lookup (because `std::map` lookup is O(log(N)) where N is the number of entries).

Same issue applies below too.

xapian-letor/feature/feature.cc
```diff
@@ -36,10 +39,17 @@ void
...
-    internal = o.internal;
+    feature_db = o.feature_db;
+    feature_query = o.feature_query;
+    feature_doc = o.feature_doc;
+    termfreq = o.termfreq;
+    inverse_doc_freq = o.inverse_doc_freq;
+    doc_length = o.doc_length;
+    collection_length = o.collection_length;
+    collection_termfreq = o.collection_termfreq;
 }
```
ojwb: Do `Feature` objects need to be copyable and/or assignable?  If not, I'd be tempted to not allow copying or assignment - they're both going to be costly operations and we tell users that API classes are cheap to copy and assign.

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -55,19 +104,39 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
     /// Virtual destructor because we have virtual methods.
     virtual ~Feature();
 
-    /// Specify the database to use for feature building. This will be used by the Internal class.
+    /// Specify the database to use for feature building. This will be used by FeatureList::Internal class.
```
ojwb: "by the FeatureList::Internal class" would read better.

This (and similar) doc comments should really be split.  The first line should be a short summary (the first sentence here is ideal), then there should be a blank (except for the comment `*`) line, and then more detail (the "This will be used ..." part can go there).

I'd say "This is used ..." though, as least if this describes the current situation.  Using the future tense suggests to the reader that it might not currently work that way, but is intended to after some planned code changes.

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -55,19 +104,39 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
...
+    /// Sets the doc_length that is going to be used for Feature building. This is used by FeatureList::Internal while populating Statistics.
+    void set_doc_length(const std::map<std::string, long int> & doc_len);
+
+    /// Sets the collection_length that is going to be used for Feature building. This is used by FeatureList::Internal while populating Statistics.
+    void set_collection_length(const std::map<std::string, long int> & collection_len);
+
+    /// Sets the collection_termfreq that is going to be used for Feature building. This is used by FeatureList::Internal while populating Statistics.
+    void set_collection_termfreq(const std::map<std::string, long int> & collection_tf);
+
+    /// Returns the Stats needed by a Sub class
```
ojwb: No need to capitalise "Stats" here.  And "subclass" not "Sub class".

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,177 @@
...
+    if (stats_needed & TERM_FREQUENCY)
+	feature_->set_termfreq(termfreq);
+    if (stats_needed & INVERSE_DOCUMENT_FREQUENCY)
+	feature_->set_inverse_doc_freq(inverse_doc_freq);
+    if (stats_needed & DOCUMENT_LENGTH)
+	feature_->set_doc_length(doc_length);
+    if (stats_needed & COLLECTION_LENGTH)
+	feature_->set_collection_length(collection_length);
+    if (stats_needed & COLLECTION_TERM_FREQ)
+	feature_->set_collection_termfreq(collection_termfreq);
```
AyushP123: it would create additional call to empty each stat once all the features have been computed for a document, query pair.

xapian-letor/feature/feature.cc
AyushP123: They are not copyable or assignable. The assignment operator and copy constructor are private. In the initial definition of the functions, internal was assigned as `internal = o.internal` for both the functions that is why I have assigned each variable to have a proper function definition. Although both methods are private and to be never called. Should I remove the assignment of each member variable in both the functions and just let there be log calls  ??

xapian-letor/api/featurelist_internal.cc
jaylett: @AyushP123 we have to empty them anyway between documents, don't we? For the document-specific ones, anyway.

xapian-letor/feature/feature.cc
jaylett: You don't need to provide an implementation for a private constructor or method that should never be called. (In C++11 you apparently can also do `= delete;` in the class declaration, but I don't know if there are any cross-compiler problems with that.)

xapian-letor/api/featurelist_internal.cc
AyushP123: No we don't empty them between the documents .... we just replace the old value with the new value

jaylett: Ah, I see. Even so, calling `.empty()` isn't a huge cost.

AyushP123: calling map.clear() for every stat will be a huge cost ..... we don't necessarily have to clear the stats after each document ... we could keep a flag `bool update ` which can be set to true after all the features for each document have been computed .... then we can check if the flag is true replace the old value with the new value ..... but I don't think we will be gaining any efficiency by doing this

jaylett: Replacing the old value with the new surely deconstructs the old value? Otherwise it's a memory leak. So I don't see how clearing the map can be significantly more expensive than what you're already doing. However, I'll defer to @ojwb on this.

AyushP123: yes @jaylett ... its a memory leak :smile:

AyushP123: No @jaylett ..... sorry its not a memory leak ..... the local map gets destroyed after the function call returns as @ojwb made it clear a while ago .... calling map.clear() will be O(N) cost for each map

AyushP123: https://stackoverflow.com/questions/15924488/delete-memory-of-stdmapint-string-completely

xapian-letor/feature/feature.cc
ojwb: Indeed - pre-C++11, the way to suppress the default copy constructor and/or assignment operator is just to declare (but **NOT** define) private versions.

It's not a good idea to also define them, because even private versions could still be called by methods of the class itself, or by friend classes/functions).

C++11 provides a cleaner way to do this by declaring these methods with `= delete` (the main benefit is that the compiler can give a better error message, because with the old-way it doesn't know if the definition is missing or deliberately omitted).  However, we're currently avoiding using C++11-specific features in the public API headers because at the moment it's still common to have to specify special compiler options to enable C++11 mode; if we use C++11-specific features in the API headers, that forces the build systems of code using Xapian to have to arrange for those compiler-specific options to be passed.  So for now, please stick with the old way (which still works with C++11 too).

xapian-letor/api/featurelist_internal.cc
ojwb: Clearing such a map will indeed involve O(N) work because the strings need to be released.  However, destroy the map will also be O(N) - this is not an additional cost, just a change in the timing of when it is incurred - at the moment this happens when the local map is destroyed after that `swap()`.

In fact by being lazier here we may be able to avoid that cost, because we avoid filling any of the maps which aren't going to be used.  And if we don't fill them then N is 0.

FWIW, I'm OK with leaving this as a follow-on change after we have this merged - the code as-is should work OK, but could be optimised.

ojwb: I think some of the confusion here is because I didn't really explain what the swap() trick achieves fully.  Sorry about that, and here goes...

The code you had before was essentially:

```
{
    map<string, int> tmp;
    // build up tmp

    // This assignment needs to:
    // (a) deallocate the current value of member, which is O(N) where N is current member.size()
    // (b) make a copy of tmp, which is O(M) where M is tmp.size()
    member = tmp;

    //  As we leave this block, tmp goes out of scope and destroying it is O(M)
}
```

With the swap trick we have:

```
{
    map<string, int> tmp;
    // build up tmp

    // This is cheap (O(1)) because it just needs to swap around pointers to the actual data:
    swap(member, tmp);

    //  As we leave this block, tmp goes out of scope and destroying it is O(N)
    // (because it holds the original contents of member)
}
```

So we still have to do the work to destroy the old value of member (O(N)) but with the swap we avoid having to copy tmp to member (O(M)) and then having to destroy the original (another O(M)).

AyushP123: The problem is that we will have to clear the member variable stats separately after all the features have been computed ..... that is going to be an extra O(N) work which we are not doing right now .... Ideally we should have a 5 bit mask and set each bit to 0 after all the features have been computed and the just recompute all the stats which we are doing now ..... since all the stats are required by the features, calculating them at the beginning or when the stat is computed for the first time wouldn't affect the time complexity at all ..... other than this I think i have incorporated all the remaining suggestions.

ojwb: There's clearly some part of this I'm not explaining to you well enough, but I'm not sure what part that is.

The complexity guarantees of of `std::map::clear()` and `std::map::~map()` are both linear:

http://en.cppreference.com/w/cpp/container/map/clear :
> Complexity
>
> Linear in the size of the container.

http://en.cppreference.com/w/cpp/container/map/%7Emap :
> Complexity
>
> Linear in the size of the container.

But if the container has already been cleared, then N=0 for the destructor (similarly `clear()` on an already empty container is with N=0).

So calling `clear()` does **NOT** incur extra work, it just does the work which has to be done at some point.  If we don't explicitly do it, it'll have to happen when the destructor gets called.  The only way to avoid doing that work is to avoid calling the destructor, which would require leaking the map and its contents, or causing the process to exit uncleanly (e.g. via `std::_Exit()` or an uncaught exception).

> since all the stats are required by the features, calculating them at the beginning or when the stat is computed for the first time wouldn't affect the time complexity at all

If this is true, then the optimisation won't help, but not for the reason you keep insisting.

But doesn't this depend on which features are being worked with?

xapian-letor/api/featurelist_internal.cc
```diff
@@ -0,0 +1,181 @@
...
+    dt.skip_to("S");
+    for ( ; dt != featurelist_doc.termlist_end(); ++dt) {
+	if ((*dt)[0] != 'S') {
+	    // We've reached the end of the S-prefixed terms.
+	    break;
+	}
+	title_len += dt.get_wdf();
+    }
+    len["title"] = title_len;
+    whole_len = featurelist_db.get_doclength(featurelist_doc.get_docid());
```
ojwb: It's better C++ style to declare a variable where you can initialise it, which you can for `whole_len`.

As the code is, someone might add or move some code which uses `whole_len` early in the function, and it'll still compile and run OK (but give the wrong results because it'll use `0` instead of the correct value, which hopefully the testsuite or the developer will notice, but it could easily result in a bug exposed to users).

If you instead declare and initialise `whole_len` here, then any attempt to use it before it's set to the correct value results in a compile-time error.

This same issue is repeated below - please find and address all cases.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -38,37 +38,74 @@ CollTfCollLenFeature::name() const
...
 
     vector<double> values;
     double value = 0;
+    double coll_len = 0;
+    map<string, Xapian::termcount>::const_iterator coll_len_iterator =
+	    collection_length.find("title");
+    if (coll_len_iterator != collection_length.end())
+	coll_len = (double)coll_len_iterator->second;
+    else
+	coll_len = 0;
```
ojwb: You already set `coll_len` to 0 above.  This pattern seems (a) redundant and (b) liable to trigger "value is assigned bu not used" compiler warnings.  It's repeated many times below too.

Given in some of those places you reuse the variable assigned in this way (which I'm not madly keen on, but let's ignore that for now) I would drop assignment in the declaration, so the consistent pattern is the if/else assignment.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -38,37 +38,74 @@ CollTfCollLenFeature::name() const
...
 CollTfCollLenFeature::get_values() const
 {
-    LOGCALL(API, std::vector<double>, "CollTfCollLenFeature::get_values", NO_ARGS);
+    LOGCALL(API, vector<double>, "CollTfCollLenFeature::get_values", NO_ARGS);
 
     vector<double> values;
     double value = 0;
+    double coll_len = 0;
+    map<string, Xapian::termcount>::const_iterator coll_len_iterator =
+	    collection_length.find("title");
```
ojwb: This seems a great place for C++11 `auto`!  Similarly for several cases below.

xapian-letor/feature/colltfcolllenfeature.cc
ojwb: You've not addressed this earlier feedback, despite claiming the branch was ready for merge.  There are many instances below too.

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -104,35 +106,50 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
     /// Virtual destructor because we have virtual methods.
     virtual ~Feature();
 
-    /// Specify the database to use for feature building. This will be used by FeatureList::Internal class.
+    /** Specify the database to use for feature building.
+     *  This will be used by FeatureList::Internal class.
+     */
```
ojwb: Thanks for reformatting these, but you've missed the part about adding an empty line between the brief description and the detailed description.  Here's an example of a doc comment with brief description, detailed description and parameter and return value descriptions showing the standard formatting we use:

https://trac.xapian.org/browser/git/xapian-core/include/xapian/queryparser.h#L197

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -146,10 +163,10 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
   private:
 
     /// Don't allow assignment.
-    void operator=(const Feature &);
+    void operator=(const Feature &) = delete;
 
     /// Don't allow copying.
-    Feature(const Feature & o);
+    Feature(const Feature & o) = delete;
```
ojwb: I clearly said not to use C++11 features in API headers.  Just keep these private and don't define them.

xapian-letor/ranker/ranker.cc
```diff
@@ -354,9 +354,6 @@ Ranker::labelcomparer(const FeatureVector & firstfv, const FeatureVector& second
 std::vector<Xapian::docid>
 Ranker::rank(const Xapian::MSet & mset, const string & model_key, const Xapian::FeatureList & flist)
 {
-    if (mset.empty()) {
-	throw Xapian::InvalidArgumentError("Empty MSet. No documents could be retrieved with the given Query.\n");
-    }
```
ojwb: I thought you said an empty `MSet` would cause problems, but you've just removed the check entirely rather than returning an empty vector here as I suggested.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -38,37 +38,74 @@ CollTfCollLenFeature::name() const
...
+	coll_len = (double)coll_len_iterator->second;
+    else
+	coll_len = 0;
 
     for (Xapian::TermIterator qt = feature_query.get_unique_terms_begin();
 	 qt != feature_query.get_terms_end(); ++qt) {
-	if ((*qt).substr(0, 1) == "S" || (*qt).substr(1, 1) == "S")
-	    value += log10(1 + ((double)collection_length.find("title")->second
-				/ (double)(1 + collection_termfreq.find(*qt)->second)));
+	if ((*qt).substr(0, 1) == "S" || (*qt).substr(1, 1) == "S") {
```
ojwb: The `substr(0, 1) == "S"` check I understand - that's a hard-coded check for this being the title field (though it'd be better written as `startswith(*qt, 'S')` - `startswith()` is a utility function in `"stringutils.h"` which can check for a string starting with a string or a character without creating the temporary string object which the current code does).

But `substr(1, 1) == "S"` is wrong (sorry, I missed this amongst all the stylistic issues).  That's checking if the second character of the term is `S`, but it's really wanting to check for terms with prefix `ZS` (stemmed terms from the title), so it ought to check `startswith(*qt, "ZS")`.  As-is, it will match those but also unwanted terms like `XSUMMARYfoo` (term "foo" with user prefix "XSUMMARY") and `KSI units` (keyword "SI units").

Given how many times this check is used, let's turn it into an inline helper:

```
inline bool
is_title_term(const string& term)
{
    return startswith(term, 'S') || startswith(term, "ZS");
}
```

That'll improve both readability and maintainability.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -38,37 +38,74 @@ CollTfCollLenFeature::name() const
...
+	    map<string, Xapian::termcount>::const_iterator coll_tf_iterator =
+		    collection_termfreq.find(*qt);
+	    if (coll_tf_iterator != collection_termfreq.end()) {
+		value += log10(1 + (coll_len /
+				    (double)(1 + coll_tf_iterator->second)));
+	    }
+	    else
```
ojwb: Xapian coding style is that you don't put a newline between `}` and `else`.

It's also better to put `{` and `}` around both or neither of the blocks.

This issue is repeated below.

xapian-letor/ranker/ranker.cc
AyushP123: I left it here because checking for an empty MSet in #156 ..... and resolving it there would be better because that would be the final function instead of this.

ojwb: This is a public API so passing an empty `Xapian::MSet` here needs to behave appropriately (not crash or otherwise invoke undefined behaviour, and end up returning a valid answer - in this case, that means returning an empty vector).

If I trace the execution, we call `flist.create_feature_vectors(mset, ...)` - that takes us to `FeatureList::create_feature_vectors()` in `xapian-letor/api/featurelist.cc` which creates an empty fvec, then loops over each entry in the `MSet` (which is a no-op when the `MSet` is empty), then calls `normalise(fvec)`.

If I look at `FeatureList::normalise()` then the first line of code is:

```
    int num_features = fvec[0].get_fcount();
```

But `fvec` is still empty, so `fvec[0]` is undefined behaviour.

xapian-letor/api/featurelist_internal.cc
AyushP123: sorry for bugging you on this .... I should have understood it the first time .... calling compute_stats will destroy the old stats .... its fair and simple .... I will implement this and push a fix

xapian-letor/api/featurelist.cc
```diff
@@ -103,13 +103,12 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
...
 	std::vector<double> fvals;
-	// All stats are computed and stored in the corresponding variables.
-	internal->compute_statistics(letor_query, letor_db, doc);
+	internal->set_data(letor_query, letor_db, doc);
 	for (Feature* it : internal->feature) {
 	    it->set_database(letor_db);
 	    it->set_query(letor_query);
 	    it->set_doc(doc);
-	    // Populates the Feature with required stats
+	    // Computes and Populates the Feature with required stats.
```
ojwb: "Computes and populates"

xapian-letor/api/featurelist_internal.cc
```diff
@@ -130,20 +124,23 @@ FeatureList::Internal::compute_collection_length()
     if (!featurelist_db.get_metadata("collection_len_title").empty() &&
 	    !featurelist_db.get_metadata("collection_len_body").empty() &&
 	    !featurelist_db.get_metadata("collection_len_whole").empty()) {
```
ojwb: The `!` should be aligned here (didn't I point this out already?)

xapian-letor/api/featurelist_internal.h
```diff
@@ -154,15 +172,16 @@ class FeatureList::Internal : public Xapian::Internal::intrusive_base {
...
-     *  corresponding variables
-     */
-    void compute_statistics(const Xapian::Query & query,
+    /// This method sets all the data members required for computing stats.
+    void set_data(const Xapian::Query & query,
 			    const Xapian::Database & db,
 			    const Xapian::Document & doc);
 
-    /// Populates the statistics needed by a Feature
+    /// Computes and Populates the stats needed by a Feature.
```
ojwb: "Computes and populates"

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -42,28 +42,26 @@ CollTfCollLenFeature::get_values() const
...
 				    (double)(1 + coll_tf_iterator->second)));
-	    }
-	    else
+	    } else {
 		value += log10(1 + coll_len);
-	}
-	else
+	    }
+	} else {
 	    value += 0;
```
ojwb: I keep saying to get rid of these pointless `value += 0;` statements!

They don't make much sense to the reader of the code, and the compiler generally can't optimise them away (because IEEE FP supports `-0`, and `(-0) + (+0)` is `+0`.  `+0` is what you'll get for a floating point constant for `0`, so the upshot is that adding `0` to a double changes it in the corner case where the input is `-0`, and `x + 0` can't be optimised to just `x` unless the compiler can prove that `x` can't be `-0`).

Checking recent GCC and clang, neither optimises away the addition for a simple test case:

```
$ cat test.cc
double f(double x) {
   x += 0;
   return x;
}
$ g++ -S -o - -O2 test.cc
	.file	"test.cc"
	.text
	.p2align 4,,15
	.globl	_Z1fd
	.type	_Z1fd, @function
_Z1fd:
.LFB0:
	.cfi_startproc
	addsd	.LC0(%rip), %xmm0
	ret
	.cfi_endproc
.LFE0:
	.size	_Z1fd, .-_Z1fd
	.section	.rodata.cst8,"aM",@progbits,8
	.align 8
.LC0:
	.long	0
	.long	0
	.ident	"GCC: (Debian 6.3.0-18) 6.3.0 20170516"
	.section	.note.GNU-stack,"",@progbits
```

`addsd	.LC0(%rip), %xmm0` loads constant `0.0` from `.LC0` and adds it.

```
$ clang++-4.0 -S -o - -O2 test.cc
	.text
	.file	"test.cc"
	.globl	_Z1fd
	.p2align	4, 0x90
	.type	_Z1fd,@function
_Z1fd:                                  # @_Z1fd
	.cfi_startproc
# BB#0:
	xorpd	%xmm1, %xmm1
	addsd	%xmm1, %xmm0
	retq
.Lfunc_end0:
	.size	_Z1fd, .Lfunc_end0-_Z1fd
	.cfi_endproc


	.ident	"clang version 4.0.1-+rc2-1 (tags/RELEASE_401/rc2)"
	.section	".note.GNU-stack","",@progbits
```

clang XORs a register with itself to generate the constant, but otherwise it's the same.

If you pass `-ffast-math` the compiler is allowed to use non-confirming assumptions when optimising, and the function becomes empty:

```
$ g++ -S -o - -ffast-math -O2 test.cc
	.file	"test.cc"
	.text
	.p2align 4,,15
	.globl	_Z1fd
	.type	_Z1fd, @function
_Z1fd:
.LFB0:
	.cfi_startproc
	rep ret
	.cfi_endproc
.LFE0:
	.size	_Z1fd, .-_Z1fd
	.ident	"GCC: (Debian 6.3.0-18) 6.3.0 20170516"
	.section	.note.GNU-stack,"",@progbits
```

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -99,6 +100,16 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
...
+    /** A helper function for feature->get_value()
+     *
+     *  Checks if the term belongs to the title or is stemmed from the title.
+     */
+    inline bool
+    is_title_term(const std::string& term) const
+    {
+	return startswith(term, 'S') || startswith(term, "ZS");
+    }
+
```
ojwb: This really doesn't want need to go in the public API header.  It doesn't need the context of the class at all so just declare it as an inline (non-class) function in the file which uses it.

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -163,10 +188,11 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
...
 
     /// Don't allow assignment.
-    void operator=(const Feature &) = delete;
+    void operator=(const Feature &);
 
     /// Don't allow copying.
-    Feature(const Feature & o) = delete;
+    Feature(const Feature & o);
+
 };
```
ojwb: You've added an unwanted blank line here.

xapian-letor/ranker/ranker.cc
```diff
@@ -354,6 +354,8 @@ Ranker::labelcomparer(const FeatureVector & firstfv, const FeatureVector& second
 std::vector<Xapian::docid>
 Ranker::rank(const Xapian::MSet & mset, const string & model_key, const Xapian::FeatureList & flist)
 {
+    if (mset.empty())
+	return std::vector<Xapian::docid>();
     LOGCALL(API, std::vector<Xapian::docid>, "Ranker::rank", mset | model_key | flist);
```
ojwb: Good, except the `LOGCALL` is meant to be the first line of the method.

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -55,19 +106,68 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
...
+    /** Sets the collection_termfreq that is going to be used
+     *  for Feature building.
+     *
+     *  This is used by FeatureList::Internal while populating Statistics.
+     */
+    void set_collection_termfreq(const std::map<std::string,
+				 Xapian::termcount> & collection_tf);
+
+    /// Returns the stats needed by a subclass
+    stat_flags get_stats(){
```
ojwb: Looking good to me now, except there should be a space between `)` and `{` here.

(Will add a rule to xapian-check-patch for that).

ojwb:
Auto-squashed, rebased and merged as 3d07e0f8b00b80ef2aac6ba2e0138a34e7513f8b.  Hurrah!

I hadn't looked at the commit message before, but it was rather, um ... freeform.  Rather than delaying the merge I reformatted it to the expected width and added a short summary line.

@AyushP123 please review the section in the developer's guide on writing good commit messages: https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages

PR #147, opened 2017-03-17
Implement word_in_list method using unordered_map

ivmarkp:
Build successful locally. All tests pass. Please review, thanks!

### DIFF
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -404,15 +404,15 @@ testcase "${tab}abc${tab}def" =1 abc=1 def=7
 
 # Feature tests for $highlight{}.
 printf '$highlight{$cgi{text},$cgi{list}}' > "$TEST_TEMPLATE"
-testcase 'A <b style="color:black;background-color:#ffff66">list</b> of <b style="color:black;background-color:#99ff99">words</b>' P=text list="list${tab}words" text="A list of words" open="<b>" close="</b>"
+testcase 'A <b style="color:black;background-color:#ffff66">list</b> of <b style="color:black;background-color:#99ff99">words</b>' list="list${tab}words" text="A list of words"
 
 printf '$highlight{$cgi{text},$cgi{list},$cgi{open}}' > "$TEST_TEMPLATE"
-testcase 'A list of <b>words</b>' P=text list="words" text="A list of words" open="<b>"
-testcase 'A list of <span>words</span>' P=text list="words" text="A list of words" open="<span>"
+testcase 'A list of <b>words</b>' list="words" text="A list of words" open="<b>"
+testcase 'A list of <span>words</span>' list="words" text="A list of words" open="<span>"
 
 printf '$highlight{$cgi{text},$cgi{list},$cgi{open},$cgi{close}}' > "$TEST_TEMPLATE"
-testcase 'A list of <b>words</b>' P=text list="words" text="A list of words" open="<b>" close="</b>"
-testcase 'A *list* of *words*' P=text list="words${tab}list" text="A list of words" open="*" close="*"
+testcase 'A list of <b>words</b>' list="words" text="A list of words" open="<b>" close="</b>"
+testcase 'A *list* of *words*' list="words${tab}list" text="A list of words" open="*" close="*"
 
 rm "$OMEGA_CONFIG_FILE" "$TEST_INDEXSCRIPT" "$TEST_TEMPLATE"
 rm -rf "$TEST_DB"
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -402,6 +402,18 @@ testcase "A${tab}AZ${tab}Z" A=1 A=2 A=3 AZ=1 AZ=2 Z=xxx
 testcase "${tab}abc" =1 abc=1
 testcase "${tab}abc${tab}def" =1 abc=1 def=7
 
+# Feature tests for $highlight{}.
+printf '$highlight{$cgi{text},$cgi{list}}' > "$TEST_TEMPLATE"
+testcase 'A <b style="color:black;background-color:#ffff66">list</b> of <b style="color:black;background-color:#99ff99">words</b>' P=text list="list${tab}words" text="A list of words" open="<b>" close="</b>"
+
+printf '$highlight{$cgi{text},$cgi{list},$cgi{open}}' > "$TEST_TEMPLATE"
+testcase 'A list of <b>words</b>' P=text list="words" text="A list of words" open="<b>"
+testcase 'A list of <span>words</span>' P=text list="words" text="A list of words" open="<span>"
+
+printf '$highlight{$cgi{text},$cgi{list},$cgi{open},$cgi{close}}' > "$TEST_TEMPLATE"
+testcase 'A list of <b>words</b>' P=text list="words" text="A list of words" open="<b>" close="</b>"
+testcase 'A *list* of *words*' P=text list="words${tab}list" text="A list of words" open="*" close="*"
+
 rm "$OMEGA_CONFIG_FILE" "$TEST_INDEXSCRIPT" "$TEST_TEMPLATE"
 rm -rf "$TEST_DB"
 if [ "$failed" = 0 ] ; then
--- a/xapian-applications/omega/query.cc
+++ b/xapian-applications/omega/query.cc
@@ -702,7 +702,7 @@ p_nottag(unsigned int c)
 // FIXME: shares algorithm with indextext.cc!
 static string
 html_highlight(const string &s, const string &list,
-		const string &bra, const string &ket)
+	       const string &bra, const string &ket)
 {
     if (!stemmer) {
 	stemmer = new Xapian::Stem(option["stemmer"]);
@@ -1563,6 +1563,7 @@ eval(const string &fmt, const vector<string> &param)
 			ket += '>';
 		    }
 		}
+
 		value = html_highlight(args[0], args[1], bra, ket);
 		break;
 	    }
--- a/xapian-applications/omega/query.cc
+++ b/xapian-applications/omega/query.cc
@@ -655,24 +655,25 @@ html_strip(const string &str)
 static unordered_map<string, int> word_to_occurrence;
 
 static void build_word_map(const string& list) {
-    string word;
+    string::size_type split = 0, split2;
     int word_index = 0;
-    for (size_t i = 0; i <= list.length(); ++i) {
-	if (list[i] == '\t' || list[i] == '\0') {
-	    if (word_to_occurrence.count(word) == 0) {
-		word_to_occurrence[word] = word_index++;
-		word.clear();
-	    }
-	} else
-	    word += list[i];
+    string word;
+    while ((split2 = list.find('\t', split)) != string::npos) {
+	word = list.substr(split, split2 - split);
+	if (word_to_occurrence.emplace(make_pair(word, word_index)).second)
+	    ++word_index;
+	split = split2 + 1;
     }
+    word = list.substr(split, list.size() - split);
+    if (word_to_occurrence.emplace(make_pair(word, word_index)).second)
+	++word_index;
 }
 
 static int word_in_list(const string& word) {
-    int pos = -1;
-    if (word_to_occurrence.count(word))
-	pos = word_to_occurrence[word];
-    return pos;
+    if (!word_to_occurrence.count(word))
+	return -1;
+    else
+	return word_to_occurrence[word];
 }
 
 // Not a character in an identifier
@@ -691,8 +692,7 @@ p_nottag(unsigned int c)
 
 // FIXME: shares algorithm with indextext.cc!
 static string
-html_highlight(const string &s, const string &list,
-	       const string &bra, const string &ket)
+html_highlight(const string &s, const string &bra, const string &ket)
 {
     if (!stemmer) {
 	stemmer = new Xapian::Stem(option["stemmer"]);
@@ -756,13 +756,11 @@ html_highlight(const string &s, const string &list,
 	}
 	j = term_end;
 	term = Xapian::Unicode::tolower(term);
-	build_word_map(list);
 	int match = word_in_list(term);
 	if (match == -1) {
 	    string stem = "Z";
 	    stem += (*stemmer)(term);
 	    match = word_in_list(stem);
-	    word_to_occurrence.clear();
 	}
 	if (match >= 0) {
 	    res += html_escape(string(l, first.raw() - l));
@@ -1553,8 +1551,9 @@ eval(const string &fmt, const vector<string> &param)
 			ket += '>';
 		    }
 		}
-
-		value = html_highlight(args[0], args[1], bra, ket);
+		build_word_map(args[1]);
+		value = html_highlight(args[0], bra, ket);
+		word_to_occurrence.clear();
 		break;
 	    }
 	    case CMD_hit:
--- a/xapian-applications/omega/query.cc
+++ b/xapian-applications/omega/query.cc
@@ -654,27 +654,36 @@ html_strip(const string &str)
 
 static unordered_map<string, int> word_to_occurrence;
 
-static void build_word_map(const string& list) {
-    string::size_type split = 0, split2;
-    int word_index = 0;
-    string word;
-    while ((split2 = list.find('\t', split)) != string::npos) {
-	word = list.substr(split, split2 - split);
+class WordList {
+    string list;
+  public:
+    WordList() : list("") { }
+
+    void build_word_map(const string& list2) {
+	// Don't build map again if passed list of terms is same as before.
+	if (list == list2) return;
+	word_to_occurrence.clear();
+	string::size_type split = 0, split2;
+	int word_index = 0;
+	string word;
+	while ((split2 = list2.find('\t', split)) != string::npos) {
+	    word = list2.substr(split, split2 - split);
+	    if (word_to_occurrence.emplace(make_pair(word, word_index)).second)
+		++word_index;
+	    split = split2 + 1;
+        }
+	word = list2.substr(split, list2.size() - split);
 	if (word_to_occurrence.emplace(make_pair(word, word_index)).second)
 	    ++word_index;
-	split = split2 + 1;
+	list = list2;
     }
-    word = list.substr(split, list.size() - split);
-    if (word_to_occurrence.emplace(make_pair(word, word_index)).second)
-	++word_index;
-}
 
-static int word_in_list(const string& word) {
-    if (!word_to_occurrence.count(word))
-	return -1;
-    else
-	return word_to_occurrence[word];
-}
+    int word_in_list(const string& word) {
+	auto it = word_to_occurrence.find(word);
+	if (it == word_to_occurrence.end()) return -1;
+	return it->second;
+    }
+};
 
 // Not a character in an identifier
 inline static bool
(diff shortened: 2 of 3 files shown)


xapian-applications/omega/query.cc
```diff
@@ -651,24 +652,28 @@ html_strip(const string &str)
...
-    while ((split2 = list.find('\t', split)) != string::npos) {
-	if (word.size() == split2 - split) {
-	    if (memcmp(word.data(), list.data() + split, word.size()) == 0)
-		return count;
-	}
-	split = split2 + 1;
-	++count;
+    for (size_t i = 0; i <= list.length(); ++i) {
+	if (list[i] == '\t' || list[i] == '\0') {
+	    if (word_to_occurrence.find(word) == word_to_occurrence.end()) {
```
jaylett: Wouldn't it be clearer to use `.count() == 0` here?

xapian-applications/omega/query.cc
```diff
@@ -651,24 +652,28 @@ html_strip(const string &str)
...
 
-// FIXME split list into hash or map and use that rather than linear lookup?
-static int word_in_list(const string& word, const string& list)
-{
-    string::size_type split = 0, split2;
+static unordered_map<string, int> word_to_occurrence;
+
+static void build_word_map(const string& list) {
+    string word;
     int count = 0;
```
jaylett: I think this variable should definitely now be called something like `word_index`. Otherwise it lends the impression that we're putting instance counts as values in the map, rather than the (word) index within the list.

xapian-applications/omega/query.cc
```diff
@@ -651,24 +652,28 @@ html_strip(const string &str)
...
-	    return count;
-    }
-    return -1;
+}
+
+static int word_in_list(const string& word) {
+    int pos = -1;
+    if (word_to_occurrence.find(word) != word_to_occurrence.end())
+	pos = word_to_occurrence[word];
+    word_to_occurrence.clear();
```
jaylett: Doesn't this guarantee that the second call to `word_in_list()` below will always return `-1`? That suggests to me that some more tests are needed.

Additionally, it might be clearer to do this by returning the word list `unordered_map` from `build_word_map()`, and explicitly freeing it in `html_highlight()` (or, perhaps better, heap allocate it within `html_highlight()` and pass it in to the other functions).

ivmarkp: Yeah, you are right. I missed that. `word_to_occurrence.clear();` should be moved inside `html_highlight()` after the second (and last call) of `word_in_list()` in line 765.

xapian-applications/omega/query.cc
ivmarkp: Okay, will rename it as `word_index`.

xapian-applications/omega/query.cc
ivmarkp: I usually prefer `.find()` over `.count()` because it expresses the intent of _finding_ something more clearly. I'll change it to `.count()` if you like. What do you suggest?

ivmarkp:
@jaylett Thanks for the review. I'll make the changes and update the PR shortly.

xapian-applications/omega/query.cc
jaylett: Your intent here isn't to find it, just to check if it exists, so `.count()` makes more sense to me. Unless @ojwb has reasons for avoiding it, I think it will make for clearer code.

ivmarkp: Okay, I'll change it to `.count()`. Thanks

jaylett:
@ivmarkp in future, you should make changes in response to PR comments as "fixup" commits, using `git commit --fixup <sha>`. This makes them easier to combine with the commits they're improving. (It doesn't make a huge amount of difference in this case where it's just one commit in the PR, but it starts to make a lot of difference in more complex cases.)

You do still need to add a test to cover the regression you introduced (but have now fixed).

ivmarkp:
> @ivmarkp <https://github.com/ivmarkp> in future, you should make changes
in response to PR comments as "fixup" commits

Noted. Will keep that in mind from next time. You also want me to squash
all commits into one once this PR is ready to be merged?

> You do still need to add a test to cover the regression you introduced
(but have now fixed).

Yeah, I'm trying to figure out how tests are laid out in `omegatest`. It's
shell script and I don't have much experience with shell script programs so
it'll take me a bit longer to add a new test but I'll update the PR as soon
as it's done. :)

jaylett:
You don't have to squash them down yourself, but it will save whoever merges this a bit of time. I'd do it once you have any tests in place :)

ivmarkp:
I just read your comment above and realized I shouldn't have squashed the commits by myself. Sorry about that.

xapian-applications/omega/query.cc
ojwb: I wasn't actually aware there was a `count()` method on `std::map`, etc!

I agree `map.count(key)` is more concise source code than `map.find(key) != map.end()`.  For the `std::map` implementation in GCC7, `count()` actually just inlines `map.find(key) == map.end() ? 0 : 1` so the two should end up producing equivalently good code.

However, while it's true that you aren't actually wanting to find it, you also really aren't wanting to count the number of occurrences - you just want to know if there are any, not how many there are.  And for `std::multimap` that distinction matters, as `count()` has to do more work that the find approach (because there `count()` calls `equal_range()` and then calculates the distance between the two iterators returned, and `std::multimap::iterator` is bi-directional, so that makes `count()` O(n) where n is the number of items with the requested key).

Exactly what the implementation of each is for `std::unordered_map` is less clear to me, and compiling a simple test function and looking at the code only really told me that the two approaches don't result in the same generated code.  But at least in the "multi" case it seems to me that `count()` will have to exhaustively check the bucket which the requested key is in, which `find()` can stop once it finds an instance of the key.

So for consistency I'd probably suggest we stick with `map.find(key) != map.end()` so that there's a single style rule for the idiom to use for this, independent of whether the container is a "multi" one or not.

xapian-applications/omega/query.cc
```diff
@@ -651,24 +652,27 @@ html_strip(const string &str)
...
-    if (word.size() == list.size() - split) {
-	if (memcmp(word.data(), list.data() + split, word.size()) == 0)
-	    return count;
+static unordered_map<string, int> word_to_occurrence;
+
+static void build_word_map(const string& list) {
+    string word;
+    int word_index = 0;
+    for (size_t i = 0; i <= list.length(); ++i) {
+	if (list[i] == '\t' || list[i] == '\0') {
```
ojwb: This means that `\0` within `list` also becomes a list item separator.  It would be better to avoid this change in behaviour even if it is for an obscure case.

What's wrong with just reusing the existing code which loops over `list` and just replacing the `memcmp()` calls with an insertion into `word_to_occurrence`?

xapian-applications/omega/query.cc
```diff
@@ -651,24 +652,27 @@ html_strip(const string &str)
...
+	    if (word_to_occurrence.count(word) == 0) {
+		word_to_occurrence[word] = word_index++;
+		word.clear();
+	    }
```
ojwb: If `word` isn't already present (which is likely to be the case the majority of the time I'd think) then this has to search to discover that `word` isn't present, then repeat most of that work to find where to insert `word`.

Better to just do that once, e.g. by trying to insert the entry and only incrementing if it was actually inserted.

```
if (word_to_occurrence.emplace(make_pair(word, word_index)).second) {
    // Insertion took place.
    ++word_index;
}
```

xapian-applications/omega/query.cc
```diff
@@ -651,24 +652,27 @@ html_strip(const string &str)
...
+	} else
+	    word += list[i];
     }
-    return -1;
+}
+
+static int word_in_list(const string& word) {
+    int pos = -1;
+    if (word_to_occurrence.count(word))
+	pos = word_to_occurrence[word];
```
ojwb: And this searches twice in the case when `word` is present...

xapian-applications/omega/query.cc
```diff
@@ -752,11 +756,13 @@ html_highlight(const string &s, const string &list,
 	}
 	j = term_end;
 	term = Xapian::Unicode::tolower(term);
-	int match = word_in_list(term, list);
+	build_word_map(list);
```
ojwb: This is the wrong place to build the mapping - you're doing it for every word in the text being highlighted.

You want to do it at most once for every call to `$highlight` (and it could be reused if `$highlight` is called again with the same list of terms to highlight, which is quite a likely case - e.g. highlighting matches in the document title and a sample of text from the document).

xapian-applications/omega/query.cc
ivmarkp: Nothing's wrong with the existing code but instead now I think that in order to avoid `\0` as an item separator, the existing code is a better option here. I'll make the changes.

xapian-applications/omega/query.cc
ivmarkp: I wasn't thinking about the extra overhead due to `[]` operator. Thanks for pointing that out; using `emplace` or 
`insert` like this avoids the issue.

xapian-applications/omega/query.cc
ivmarkp: Would it be fine to search once in each case? E.g.
```cpp
if (!word_to_occurrence.count(word))
    return -1;
else
    return word_to_occurrence[word];
```
Also, are you in favour of using `find` over `count`? I read a comment regarding that in one of the notification emails but it seems to be missing here.

ojwb: That comment is unhelpfully hidden by github because it's on a bit of code which has since been modified again - you need to click on "Hide outdated" under the second of James' comments to see that.

The code you quote doesn't search once in each case - it searches once in the case where `word` isn't present, but twice in the case where it is (because `count()` searches, then `operator[]` repeats that search).  In this case you definitely want to use `find()` - something like:

```
auto i = word_to_occurrence.find(word);
if (i == word_to_occurrence.end()) return -1;
return i->second;
```

xapian-applications/omega/query.cc
ojwb: That would achieve calling it exactly once, though not the reuse of the existing mapping when called again with the same list of terms.

C++ is an OO language, so the clean approach is to wrap the data (i.e. the mapping) as a private member of a new class, and make the functions which operate on it (such as `build_word_map()` and `word_in_list()`) member functions of that class.  The class could then also keep track of the list of terms which was used to build the current mapping, then `build_word_map()` could just return if passed the same list of terms as was used the last time.

ivmarkp: Yes, I agree. Defining a new class and using that would be a clean solution. I'm working on it now; will update the PR shortly.

xapian-applications/omega/query.cc
ivmarkp: > That comment is unhelpfully hidden by github because it's on a bit of code which has since been modified again - you need to click on "Hide outdated" under the second of James' comments to see that.

Thanks, I see it now. Addressing that comment, I'll replace `count()` with `find()` everywhere in my changes.

> The code you quote doesn't search once in each case - it searches once in the case where word isn't present, but twice in the case where it is 

Sorry, I keep forgetting the extra linear time overhead due to `operator []`. I think it's because I've used it several times under the impression of its average case time complexity which is constant. Working on this PR has changed my perception towards it completely, especially when working on performance-centric applications in real world

ivmarkp:
Ah, `xapian-check-path` is a nice addition to travis-ci. I'll fix the formatting issue it pointed out.

ivmarkp:
@ojwb Can you take another look? I made some changes following your last review. Thanks.

xapian-applications/omega/omegatest
```diff
@@ -402,6 +402,18 @@ testcase "A${tab}AZ${tab}Z" A=1 A=2 A=3 AZ=1 AZ=2 Z=xxx
 testcase "${tab}abc" =1 abc=1
 testcase "${tab}abc${tab}def" =1 abc=1 def=7
 
+# Feature tests for $highlight{}.
+printf '$highlight{$cgi{text},$cgi{list}}' > "$TEST_TEMPLATE"
+testcase 'A <b style="color:black;background-color:#ffff66">list</b> of <b style="color:black;background-color:#99ff99">words</b>' P=text list="list${tab}words" text="A list of words" open="<b>" close="</b>"
```
ojwb: Parameters `open` and `close` are passed but unused here, which is liable to confuse the reader (I'm left unsure if you didn't mean to pass them or meant to actually use them).

ojwb: `P` also seems to be unused (or is it implicitly required for the template to work?)

xapian-applications/omega/query.cc
```diff
@@ -28,6 +28,7 @@
 #include <algorithm>
 #include <iostream>
 #include <map>
+#include <unordered_map>
 #include <set>
 #include <vector>
```
ojwb: Addition to a sorted list of headers is in the wrong place...

ivmarkp: Thanks for pointing that out. Fixed in c42147e.

xapian-applications/omega/omegatest
ivmarkp: That's right. I have removed those unused parameters with 1e0beba.

xapian-applications/omega/query.cc
```diff
@@ -651,25 +652,38 @@ html_strip(const string &str)
...
+static unordered_map<string, int> word_to_occurrence;
+
+class WordList {
+    string list;
+  public:
+    WordList() : list("") { }
+
+    void build_word_map(const string& list2) {
+	// Don't build map again if passed list of terms is same as before.
+	if (list == list2) return;
```
ojwb: This optimisation still won't work as the code currently is - you create a new `WordList` object each time `html_highlight()` is called, so `list` will get reset each time and (assuming `list2` isn't empty) the mapping always gets rebuilt.

For this to work, you need to track the `unordered_map` and the list of terms it was built from together.  So either they both need to be `static` (and for better encapsulation, both inside the class so static members rather than global statics) or both need to be non-static members of the class and there's a single static instance of the class.

Stylistically, `list2` seems ugly - it would be clearer to call the parameter `list` and the member variable something like `prev_list`.

xapian-applications/omega/query.cc
```diff
@@ -651,25 +652,38 @@ html_strip(const string &str)
...
-    while ((split2 = list.find('\t', split)) != string::npos) {
-	if (word.size() == split2 - split) {
-	    if (memcmp(word.data(), list.data() + split, word.size()) == 0)
-		return count;
+static unordered_map<string, int> word_to_occurrence;
+
+class WordList {
+    string list;
+  public:
+    WordList() : list("") { }
```
ojwb: There's no reason to use `list("")`, and it's actually a bit less efficient with current compilers (or at least for 4 year old compilers - I haven't rechecked this recently, but see: https://survex.com/~olly/blog/coding/empty-strings-in-c%2B%2B.html)

In C++, variables with a class type are always default constructed (unlike variables of fundamental types, such as int and double) and a default constructed `std::string` is an empty string, so there's no need to initialise `list` explicitly at all here.  If you **really** want to then `list()` should be exactly equivalent to the implicit initialisation which happens anyway.

xapian-applications/omega/query.cc
```diff
@@ -752,11 +765,12 @@ html_highlight(const string &s, const string &list,
 	}
 	j = term_end;
 	term = Xapian::Unicode::tolower(term);
-	int match = word_in_list(term, list);
+	WordList w;
+	int match = w.word_in_list(term);
```
ojwb: This confusingly depends on you having previously created a `WordList` and called `build_word_map` on it, or else the mapping would be empty.

It would be more obvious how this worked to if you either pass in `list` as before and construct the `WordList` in this function, or pass in the `WordList` object that the caller constructs, or have a single static `WordList` object rather than constructing two for every `$highlight` expanded.

xapian-applications/omega/query.cc
```diff
@@ -651,25 +652,38 @@ html_strip(const string &str)
...
-// FIXME split list into hash or map and use that rather than linear lookup?
-static int word_in_list(const string& word, const string& list)
-{
-    string::size_type split = 0, split2;
-    int count = 0;
-    while ((split2 = list.find('\t', split)) != string::npos) {
-	if (word.size() == split2 - split) {
-	    if (memcmp(word.data(), list.data() + split, word.size()) == 0)
-		return count;
+static unordered_map<string, int> word_to_occurrence;
```
ojwb: Previous @jaylett said:
> I think this variable should definitely now be called something like `word_index`. Otherwise it lends the impression that we're putting instance counts as values in the map, rather than the (word) index within the list.

And you replied:
> Okay, will rename it as `word_index`

But you've not done that...

ivmarkp: Actually, I already did that for the `count` integer variable because I thought it was meant for that variable instead of the `word_to_occurrence`?

xapian-applications/omega/query.cc
ivmarkp: Thanks for sharing your excellent explanation. Lesson learnt: no need to explicitly initialise std::string (infact, it's better not to) :)

xapian-applications/omega/query.cc
ivmarkp: > This optimisation still won't work as the code currently is - you create a new WordList object each time html_highlight() is called, so list will get reset each time and (assuming list2 isn't empty) the mapping always gets rebuilt.

I see the mistake made here. As you suggested, `list` must be made static and also the map be encapsulated inside the class.

> Stylistically, list2 seems ugly - it would be clearer to call the parameter list and the member variable something like prev_list.

Yes, agreed. Will make this change as well.

xapian-applications/omega/query.cc
ivmarkp: > This confusingly depends on you having previously created a WordList and called build_word_map on it, or else the mapping would be empty.

Yes, since the map has global scope and declared as static, it's unlikely that it would become empty once built here unless explicitly cleared using `.clear()` or something similar.

> It would be more obvious how this worked to if you either pass in list as before and construct the WordList in this function, or pass in the WordList object that the caller constructs, or have a single static WordList object rather than constructing two for every $highlight expanded.

Sure, I think it would be better to just pass in `list` as before and construct the `WordList` here. I'll make the changes accordingly.

xapian-applications/omega/query.cc
ojwb: Oh sorry, I misread which line James was talking about.

xapian-applications/omega/query.cc
```diff
@@ -1547,7 +1563,6 @@ eval(const string &fmt, const vector<string> &param)
 			ket += '>';
 		    }
 		}
-
```
ojwb: The combined result of adding and removing code here is an arbitrary whitespace change, completely unrelated to the patch, which is a bad thing.

Otherwise, looking good now.

xapian-applications/omega/query.cc
```diff
@@ -688,7 +702,7 @@ p_nottag(unsigned int c)
 // FIXME: shares algorithm with indextext.cc!
 static string
 html_highlight(const string &s, const string &list,
-	       const string &bra, const string &ket)
+		const string &bra, const string &ket)
```
ojwb: This line is now mis-indented (it's not clear in the diff, but `html_highlight(` is 15 characters, and you've changed the line after it to be indented with 2 tab characters, which is a width of 16 characters.

ivmarkp: Yeah, I see now. Will fix this.

xapian-applications/omega/query.cc
ivmarkp: Oh, I didn't realise there was a new line here previously. Fix add that.

ojwb: As well as being good practice to review the diff before submitting a patch, for similar reasons it's a good idea to review the combined diff (i.e. diff against the branch point) when amending a patch.  

Assuming the branch was made from `master` at some point, then this will diff against the common ancestor of the current branch and `master` (NB **three** dots):

```
git diff master...
```

ivmarkp: Thanks for the information. There must be many aspects of git that are very useful that I haven't used proactively and this is definitely one of them. Will make use of it from now on.

PR #150, opened 2017-03-20
change all i++ to ++i under xapian-letor

Hell0Kitty:
I am a GSoC 2017 Candidate, interested in your Letor Project.

While I am making myself familiar with xapian, I found some improvements can be made:

* change all ++i under Letor part to i++;
I will keep updating if found others things to be improved.

Yu Shining (IRC nickname Ficus)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1 @@
+# Gittest
--- a/README.md
+++ b/README.md
@@ -1,1 +1,2 @@
 # Gittest
+# Gittest
--- a/README.md
+++ b/README.md
@@ -1,2 +1,3 @@
 # Gittest
 # Gittest
+# Ficus_Xapian_contribution
--- a/xapian-letor/tests/harness/testsuite.cc
+++ b/xapian-letor/tests/harness/testsuite.cc
@@ -788,7 +788,7 @@ test_driver::parse_command_line(int argc, char **argv)
 
     while (argv[optind]) {
 	test_names.push_back(string(argv[optind]));
-	optind++;
+        ++optind;
     }
 
 #ifdef HAVE_VALGRIND
--- a/xapian-letor/tests/harness/testutils.cc
+++ b/xapian-letor/tests/harness/testutils.cc
@@ -181,7 +181,7 @@ mset_expect_order_(const Xapian::MSet &A, bool beginning,
     }
 
     Xapian::MSetIterator j = A.begin();
-    for (size_t i = 0; i < expect.size(); i++, j++) {
+    for (size_t i = 0; i < expect.size(); ++i, ++j) {
 	TEST_AND_EXPLAIN(*j == expect[i],
 			 "Mset didn't contain expected result:\n"
 			 << "Item " << i << " was " << *j
@@ -208,7 +208,7 @@ test_mset_order_equal(const Xapian::MSet &mset1, const Xapian::MSet &mset2)
 		     << mset1.size() << " != " << mset2.size());
     Xapian::MSetIterator i = mset1.begin();
     Xapian::MSetIterator j = mset2.begin();
-    for (; i != mset1.end(); i++, j++) {
+    for (; i != mset1.end(); ++i, ++j) {
 	TEST_AND_EXPLAIN(*i == *j,
 			 "Msets have different contents -\n" <<
 			 mset1 << "\n !=\n" << mset2);


Hell0Kitty:
//
   All the changes are made under Letor/test part

jaylett:
Firstly, have you been through the developers guide linked from our GSoC student notes? It talks about good commit practices. You haven't explained why you wanted to make these changes (on counters, in `for` loops, without side effects, won't this be the same but slightly less readable?).

Secondly, you've included three commits you didn't intend to, including one which says it makes a substantive change but then doesn't.

ojwb:
@jaylett It's actually recommended by our style guide:

```
* Prefer "++i;" to "i++;", "i += 1;", or "i = i + 1".  For simple integer
  variables these should generate equivalent (if not identical) code, but if i
  is an iterator object then the pre-increment form can be more efficient in
  some cases with some compilers.  It's simpler and more consistent to always
  use the pre-increment form (unless you make use of the old value which the
  post-increment form returns).  For the same reasons, prefer "--i;" to "i--;",
  "i -= 1;", or "i = i - 1;".
```

So the logic is that `++i` can be more efficient than `i++` for iterators, and that we prefer style rules that don't require considering the type of the variable involved to know what to write.

Modern compilers can probably optimise away the extra overhead from `i++` if the `operator++(int)` method is inlined and written suitably - as HACKING says elsewhere:

```
The postfix operator++ for iterators should be implemented inline in terms
  of the prefix form as described by Joe Buck on the gcc mailing list
~ - excerpt from http://article.gmane.org/gmane.comp.gcc.devel/50201 ::
  
          class some_iterator {
          public:
              // ...
              some_iterator& operator++();
  
              some_iterator operator++(int) {
                  some_iterator tmp = *this;
                  operator++();
                  return tmp;
              }
          };
  
      The compiler is allowed to assume that the copy constructor only does
      a copy, and to optimize away unneeded copy operations.  The result
      in this case should be that, for some_iterator above, using the
      postfix operator without using the result should give code equivalent
      to using the prefix operator.
  
      Now, for [GCC 3.4], you'll find that the dead uses of tmp are only
      completely optimized away if tmp has only one data member that can fit in a
      register.  [GCC 4.0 will do] better, and you should find that this style
      comes very close to eliminating any penalty from "incorrect" use of the
      postfix form.
  
  Xapian's PostingIterator, TermIterator, PositionIterator, and ValueIterator all
  have only one data member which fits in a register.
```

Our minimum GCC requirement is 4.7 for 1.4.x, and 4.8 for git master.

Also the most common place where you need to increment iterators is in a loop over a container, and with C++11 `for (iterator i = c.begin(); i != c.end(); ++i) { do_something(*i); }` loop is better written as `for (auto&& i : c) { do_something(i); }` anyway.

So possibly things have moved on and this should be reconsidered, though there's the issue of third-party iterators which don't implement `operator++(int)` as an inline wrapper.

It's certainly something that people often "get wrong" in submitted patches, and I can understand why as `i++` is a common idiom for "increment i" at least in C code where the side-effect or returning the old value can clearly be trivially optimised away if it's unused.

But as things stand, the final commit seems useful as it fixes deviations from our style guide.  The other 3 commits should be removed (`git rebase -i` and just delete them from the list, then `git push --force` to update the branch for this PR).

ojwb:
The final commit actually introduces a new deviation from the style guide as one of the changed lines is now indented with 8 spaces instead of a tab, which is why the CI reports a failure:

https://travis-ci.org/xapian/xapian/jobs/213023127

jaylett:
@ojwb I'm not opposed to the change, but it needs some justification in the commit message IMHO. I understand why it's important for general iterators, but I don't really think of integers (which two of the changes are) as iterators, hence having a justifying commit message would I think be helpful.

Certainly:

> all i++ changed to ++i under xapian-letor

isn't a great message because it just repeats what the change is (with the marginal detail that we're changing all of them). "Prefer postfix increment for iterators" might be a better subject, for instance, followed by some explanation noting the style guide.

ojwb:
> @ojwb I'm not opposed to the change, but it needs some justification in the
> commit message IMHO.

Indeed, I wasn't intended to disagree with that.  I was just filling
in the background.

> all i++ changed to ++i under xapian-letor
>
> isn't a great message because it just repeats what the change is (with the
> marginal detail that we're changing all of them). "Prefer postfix increment
> for iterators" might be a better subject, for instance, followed by some
> explanation noting the style guide.

Yes, except that should be "Prefer prefix"!

jaylett:
D'oh! Prefer prefix, indeed :)

Hell0Kitty:
Thank you very much for your reviews!
Can I just open a new pull request with a proper commit message, and replace the white space with tab?

ojwb:
Can I just open a new pull request with a proper commit message, and
replace the white space with tab?

Much better to fix up the branch locally and force push it to update the
existing PR.

PR #151, opened 2017-03-22
Prefer Prefix!

Hell0Kitty:
I correct the previous pull request (#150), so that now tab is used instead of white spaces.


ojwb:
Thanks for the patch - rebased and applied in 8d91211bba7e.

PR #152, opened 2017-04-22
New functionality in weighting scheme API

ivmarkp:
This PR adds the functionality to create Weight object from parameter string and also adds the support for newer weighting schemes added in Xapian 1.4.1 (BM25+, PL2+, Dir+) in Omega.

A static method named `create` is defined in Xapian::Weight that takes "scheme" and a Registry object and does the following set of things:

* obtains an appropriate weighting scheme object by looking at the first  word in "scheme" string and using `Registry::get_weighting_scheme()` method,

* using the weighting scheme object, it then calls the virtual method `create_from_parameters` with a pointer to the parameter string which returns the appropriate parameterised weighting scheme object,

* finally, returns the weighting scheme object to the code in omega:
```cpp
    enq.set_weighting_scheme(*(Xapian::Weight::create(scheme)));
```

As a consequence, the existing code in omega (xapian-applications/omega/weight.cc) which does that for the older weighting schemes is replaced with just a single line of code mentioned above.


ivmarkp:
@ojwb @jaylett Travis check succeeded on linux with gcc but failed on OS X and on linux with clang. Not sure what caused the build failure but I don't have access to either of those setups so can't possibly try to resolve the issue locally. Can you please look into this? Errors:
```bash
/Users/travis/build/xapian/xapian/swig/preinst-swig: line 9: /Users/travis/build/xapian/xapian/swig/swig: No such file or directory
/Users/travis/build/xapian/xapian/swig/preinst-swig: line 9: exec: /Users/travis/build/xapian/xapian/swig/swig: cannot execute: No such file or directory
```

jaylett:
@ivmarkp the error is before that; looks like you need to include some more headers in `lmweight.cc` (and possibly other files) to support your changes:

```
weight/lmweight.cc:263:5: error: use of undeclared identifier 'errno'
    errno = 0;
    ^
weight/lmweight.cc:264:13: error: use of undeclared identifier 'strtol'; did you mean 'stol'?
    int v = strtol(*p, &end, 10);
            ^~~~~~
            stol
/usr/local/clang-3.5.0/bin/../include/c++/v1/string:4128:37: note: 'stol' declared here
_LIBCPP_FUNC_VIS long               stol  (const string& __str, size_t* __idx = 0, int __base = 10);
                                    ^
weight/lmweight.cc:264:24: error: cannot initialize a parameter of type 'size_t *' (aka 'unsigned long *') with an rvalue of type 'char **'
    int v = strtol(*p, &end, 10);
                       ^~~~
/usr/local/clang-3.5.0/bin/../include/c++/v1/string:4128:73: note: passing argument to parameter '__idx' here
_LIBCPP_FUNC_VIS long               stol  (const string& __str, size_t* __idx = 0, int __base = 10);
                                                                        ^
weight/lmweight.cc:265:22: error: use of undeclared identifier 'errno'
    if (*p == end || errno || v < 1 || v > 5)
                     ^
4 errors generated.
make[3]: *** [weight/lmweight.lo] Error 1
make[3]: Leaving directory `/home/travis/build/xapian/xapian/xapian-core'
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory `/home/travis/build/xapian/xapian/xapian-core'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/home/travis/build/xapian/xapian/xapian-core'
make: *** [all] Error 2
```

If you're on either Debian or Ubuntu, you should be able to install clang fairly easily if this becomes too tedious to wait for Travis to fail as you fix things :-)

ojwb:
If you want to reproduce on Linux, you'll also need to specify libc++ as the C++ standard library, since the clang packages default to using libstdc++ (for compatibility with code compiled using GCC), e.g. if your distro has clang 3.9 then something like:

```
./configure CXX='clang++-3.9 -stdlib=libc++'
```

But you probably just need to add the headers needed for `errno` and `strtol()`.

xapian-core/include/xapian/weight.h
```diff
@@ -29,6 +29,10 @@
 #include <xapian/types.h>
 #include <xapian/visibility.h>
 
+#include "registry.h"
+
+#include "../common/stringutils.h"
```
ojwb: You can't include headers internal to the library from a public API header (but I can't see why you need to).

xapian-core/include/xapian/weight.h
```diff
@@ -29,6 +29,10 @@
 #include <xapian/types.h>
 #include <xapian/visibility.h>
 
+#include "registry.h"
```
ojwb: Should be `<xapian/registry,h>` and be added to the group above in sorted order.

xapian-core/include/xapian/weight.h
```diff
@@ -372,6 +395,10 @@ class XAPIAN_VISIBILITY_DEFAULT Weight {
     Xapian::termcount get_wdf_upper_bound() const {
 	return wdf_upper_bound_;
     }
+
+    static void parameter_error(const char * param, const std::string & scheme);
+
+    bool double_param(const char ** p, double * ptr_val) const;
```
ojwb: It's not great to add internal stuff like this such that it has to appear in the public API header - it means we can't change the calling signature of these methods within a release series.

Neither seems to use the context of a `Weight` object - could they both be made `static` and added to `Weight::Internal` instead?

xapian-core/include/xapian/weight.h
```diff
@@ -400,6 +427,8 @@ class XAPIAN_VISIBILITY_DEFAULT BoolWeight : public Weight {
     double get_sumextra(Xapian::termcount doclen,
 			Xapian::termcount uniqterms) const;
     double get_maxextra() const;
+
+    const BoolWeight * create_from_parameters(const char * params) const;
```
ojwb: Why return a `const` pointer?  `unserialise()` doesn't.

xapian-core/weight/weight.cc
```diff
@@ -134,4 +135,81 @@ Weight::unserialise(const string &) const
...
+    else if (scheme == "inl2")
+	scheme = "Xapian::InL2Weight";
+    else if (scheme == "pl2")
+	scheme = "Xapian::PL2Weight";
+    else if (scheme == "pl2plus")
+	scheme = "Xapian::PL2PlusWeight";
+    else if (scheme == "tfidf")
+	scheme = "Xapian::TfIdfWeight";
+    else if (scheme == "trad")
+	scheme = "Xapian::TradWeight";
```
ojwb: It's not good to hard-code this list - the idea is this works for user schemes too.  Also it's `O(n)` lookup.

I'd overlooked that the current key here is the class name.  I think we need a `map` (or perhaps better `unordered_map`) to track the schemes by the short name too.

xapian-core/weight/weight.cc
```diff
@@ -134,4 +135,81 @@ Weight::unserialise(const string &) const
...
+    else
+	throw "Unknown $opt{weighting} setting: " + scheme;
```
ojwb: That exception message only makes sense in the context of omega.

xapian-applications/omega/weight.cc
```diff
@@ -23,267 +24,17 @@
...
-
-	if (scheme == "coord") {
-	    enq.set_weighting_scheme(Xapian::CoordWeight());
-	    return;
-	}
-
-	if (scheme != "bool") {
-	    throw "Unknown $opt{weighting} setting: " + scheme;
-	}
+	enq.set_weighting_scheme(*(Xapian::Weight::create(scheme)));
```
ojwb: This leaks the `Weight` object returned by `create()` (only if there are parameters, but I've addressed that in a comment elsewhere).  It needs to delete the returned object after using it (`Enquire::set_weighting_scheme()` takes a copy of the object passed in).

xapian-core/weight/weight.cc
```diff
@@ -134,4 +135,81 @@ Weight::unserialise(const string &) const
...
+
+    if (C_isspace(*p))
+	return reg.get_weighting_scheme(scheme)->create_from_parameters(p);
+    else
+	return reg.get_weighting_scheme(scheme);
```
ojwb: Looking at the code I realise there's a problem with this short-cutting in the case where there are no parameters - `create_from_parameters()` creates a new object, while this case returns a pointer to an existing object.  So the caller can't delete the returned object, but if it doesn't delete it then it'll sometimes leak.  I think we have to always call `create_from_parameters()` here.

xapian-applications/omega/weight.cc
ivmarkp: Agreed. Should've been careful here. I think after passing the object, it can simply be deleted as `
delete Xapian::Weight::create(scheme);` in the next line.

xapian-core/weight/weight.cc
ivmarkp: I see. So, here we don't want to care whether or not `scheme` has parameters or not but instead we want the `create_from_parameters()` to handle the case where it doesn't, in which case `create_from_parameters()` should return the default object e.g. Xapian::BM25Weight()?

xapian-core/weight/weight.cc
ivmarkp: Yeah, it should be removed. Anyway, with new `map` in place in `Registry::Internal` this whole `if ... else` code block will need to be removed.

xapian-core/weight/weight.cc
ivmarkp: Yeah, we need another `map` with short names as its keys. I'll get this in with a new commit.

Also, I wonder if we could settle with using `map` for now as using `unordered_map` would require making a few changes in `Registry::Internal`?

xapian-core/include/xapian/weight.h
ivmarkp: Okay, will fix this.

xapian-core/include/xapian/weight.h
ivmarkp: This was needed for `C_isspace` function in the definition of `Weight::create()` in weight.cc. I guess using simply `if (*p = ' ')` might also be fine and then this header will no longer be required.

xapian-core/include/xapian/weight.h
ivmarkp: > Neither seems to use the context of a Weight object - could they both be made static and added to Weight::Internal instead?

I tried that but build fails at weighting scheme classes seemingly due to the dependency between `create_from_parameters` in Weight subclasses and `parameter_error` in Weight::Internal which, since builds after Weight subclasses, causes these errors:

```bash
weight/bb2weight.cc:190:10: error: incomplete type ‘Xapian::Weight::Internal’ used in nested name specifier
     if (!Xapian::Weight::Internal::double_param(&p, &k))
          ^
weight/bb2weight.cc:191:2: error: incomplete type ‘Xapian::Weight::Internal’ used in nested name specifier
  Xapian::Weight::Internal::parameter_error("Parameter is invalid", "bb2");
  ^
weight/bb2weight.cc:193:2: error: incomplete type ‘Xapian::Weight::Internal’ used in nested name specifier
  Xapian::Weight::Internal::parameter_error("Extra data after parameter", "bb2");
```
What do you suggest?

xapian-core/weight/weight.cc
ojwb: Yes.

xapian-core/include/xapian/weight.h
ojwb: If the code which needs `C_isspace()` is in `weight.cc`, then add the header inclusion to `weight.cc` not here.

Including a header in a file where it's not used at all is unhelpful - as your patch currently stands, if a developer modifies `common/stringutils.h` then `make` will have to recompile every source file which directly or indirectly includes `include/xapian/weight.h`, but many of those files probably don't actually use anything from `common/stringutils.h`.

xapian-core/include/xapian/weight.h
ojwb: Those errors are because `weight/bb2weight.cc` doesn't currently include `weightinternal.h`, and so we only have the forward declaration of `Xapian::Weight::Internal` from `xapian/weight.h` and hence only know it is a class, not what any of its members or methods are.

xapian-core/weight/weight.cc
ojwb: Realistically, in practice nobody will register a lot of different `Weight` subclasses so `std::map` is fine, and probably more space efficient.

xapian-applications/omega/weight.cc
ojwb: That'll allocate a new object and then free that one, which doesn't fix the leak of the first one.

You need to only call `create()` once and store its result in a local variable.

ojwb:
a4a1be5 adds new file `xapian-bindings/python3/xapian_wrap.stamp-t` which shouldn't be in the repo (it's a temporary file for a build timestamp file).

ojwb:
There's a failure still on OS X:

```
/Users/travis/build/xapian/xapian/xapian-core/include/xapian/weight.h:431: Warning 842: Covariant return types not supported in C#. Proxy method will return Xapian::Weight *.
```

This isn't specific to OS X, it's just that the Linux CI build didn't test C# as the dependencies for it aren't installed (I've pushed a change to git master to install them, though I think you'd have to rebase onto current master for the CI to install them on your branch).

I suspect it doesn't matter that the return type is being treated as `Xapian::Weight*` for C#, in which case just telling SWIG that it can ignore this should be fine.

ivmarkp:
@ojwb Hi, thanks for pushing the change for linux CI. I'll rebase my local branch onto latest master and force push.

Also, sorry for slow response; I've been busy with ongoing exams (there's one tomorrow). Anyway, I'm working on this PR now and will be able to push some changes addressing your review comments in an hour or so.

ivmarkp:
@ojwb The issue regarding the covariant return types not being supported in C# still persists even after you pushed a change to install C# dependencies. May be the issue goes deeper than that. How should I go about fixing this?

ojwb:
The changed I pushed just fixed the CI to actually test the C# bindings on Linux - I didn't attempt to address the covariant return type issue.

As I said above:

> I suspect it doesn't matter that the return type is being treated as `Xapian::Weight*` for C#, in which case just telling SWIG that it can ignore this should be fine.

The warning says that in C#, the wrapped method will return `Xapian::Weight` rather than `Xapian::BM25Weight`, but the C# code can't usefully do much with the object returned except pass it  back to Xapian to set it as the weighting scheme to use, so that's seems OK.

So something like:

```
%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::Weight::create_from_parameters;
```

You might need to specify that for each subclass rather than just being able to do it for `Weight` - I'm not sure without trying it.

xapian-core/weight/boolweight.cc
```diff
@@ -86,4 +86,14 @@ BoolWeight::get_maxextra() const
...
 }
 
+BoolWeight *
+BoolWeight::create_from_parameters(const char * p) const
+{
+    if (*p == '\0')
+	return new Xapian::BoolWeight();
+    if (*p == ' ')
+	throw "No parameters are required for BoolWeight";
+    return new Xapian::BoolWeight();
```
ojwb: Currently `"1.7"` is a valid parameter string for `BoolWeight`, which seems wrong.  Any non-empty parameter string should probably throw an exception.

Also, the exception type should be `Xapian::InvalidArgumentError`, not `const char*`.

The same comments apply to several other weighting schemes below.

xapian-core/weight/weightinternal.h
```diff
@@ -245,6 +247,25 @@ class Weight::Internal {
...
+	return true;
+    }
+
+    inline static void parameter_error(const char * param,
+				       const std::string & scheme) {
+	string m(param);
+	m += ": '";
+	m += scheme;
+	m += "'";
+	throw m;
```
ojwb: Should throw `Xapian::InvalidArgumentError`, not `std::string`.

`param` seems a less clear name for the function parameter holding the error message than `msg` was.

xapian-core/weight/weightinternal.h
```diff
@@ -245,6 +247,25 @@ class Weight::Internal {
 
     /// Return a std::string describing this object.
     std::string get_description() const;
+
+    inline static bool double_param(const char ** p, double * ptr_val) {
```
ojwb: Methods defined in the body of the class are implicitly `inline`.

xapian-core/weight/weight.cc
```diff
@@ -134,4 +138,25 @@ Weight::unserialise(const string &) const
     throw Xapian::UnimplementedError("unserialise() not supported for this Xapian::Weight subclass");
 }
 
+const Weight *
+Weight::create(const string & s, const Registry & reg)
+{
+    const char *p = s.c_str();
+    std::string scheme = "";
```
ojwb: As I noted in the other PR, it's better not to explicitly initialise `std::string` like this.

xapian-core/weight/boolweight.cc
ivmarkp: `Weight::create` makes sure that `p` either points to a `\0` or a blank space before `create_from_parameters()` method is called with `p`. I can't see how `"1.7"` is a valid parameter string here although `" 1.7" could be a valid string in this case where `p` would be pointing to a blank space in which case an exception is thrown as specified. I might be missing your point; can you please elaborate a bit?

> Also, the exception type should be Xapian::InvalidArgumentError, not const char*.

Yeah, will fix that.

ivmarkp:
@ojwb Thanks for the help with SWIG warnings. Turns out we just need to add `%warnfilter(842);` in the SWIG interface file for the C# bindings (842 is the reference number for covariant return type warning as mentioned [here](http://www.swig.org/Doc1.3/Warnings.html). I've checked and it works fine now.

xapian-bindings/csharp/csharp.i
```diff
@@ -236,6 +236,8 @@ namespace Xapian {
 
 }
 
+%warnfilter(842);
```
ojwb: (a) Use `WARN_CSHARP_COVARIANT_RET` not `842` - the named constant is at least some-what self-documenting; the number code means nothing unless the reader digs out the corresponding SWIG documentation.
(b) This will ignore any case where there's a covariant return types - we want to only ignore those we've checked to be OK.

xapian-core/api/registry.cc
```diff
@@ -72,6 +72,9 @@ class Xapian::Registry::Internal : public Xapian::Internal::intrusive_base {
     /// Clear all registered lat-long metrics.
     void clear_lat_long_metrics();
 
+    /// Clear all registered weighting schemes with short names.
+    void clear_weighting_schemes_short();
+
```
ojwb: This should just be done by the existing `clear_weighting_schemes()` - it doesn't make sense to clear the weighting schemes from one mapping but not the other.

xapian-core/api/registry.cc
```diff
@@ -197,7 +201,7 @@ Registry::Internal::add_defaults()
     weighting_scheme_short = new Xapian::PL2Weight;
     wtschemes_short["pl2"] = weighting_scheme_short;
     weighting_scheme_short = new Xapian::PL2PlusWeight;
-    wtschemes_short["pl2plus"] = weighting_scheme;
+    wtschemes_short["pl2plus"] = weighting_scheme_short;
```
ojwb: This hard-codes the short names of the weighting schemes, which makes the whole registry approach pointless, as the user can't register their own scheme here.  Instead it should register based on calling a method on the object to get its short name.

Also, we should be able to store the same object in the short map that we already created for the standard map, rather than creating two of each.

xapian-core/weight/boolweight.cc
ojwb: > I can't see how `"1.7"` is a valid parameter string here

The `create_from_parameters()` method is in the public API too.

xapian-core/include/xapian/weight.h
```diff
@@ -315,6 +316,25 @@ class XAPIAN_VISIBILITY_DEFAULT Weight {
 	return stats_needed & UNIQUE_TERMS;
     }
 
+    /** @private @internal Return the appropriate weighting scheme object.
```
ojwb: You've marked the new `create()` method as `@private @internal` which means it won't appear in the collated API documentation.

ojwb:
If the user registers a weighting scheme with `Registry::register_weighting_scheme()` then no short name entry is added.

ivmarkp:
@ojwb Fixed with e259dcb. Please have a look.

ivmarkp:
@ojwb I think, since `register_object()` can't distinguish between `registry` map as being `wtschemes` or `wtschemes_short` map, we have to define another function just like `register_object()` to register weighting schemes with short names as well. We could may be call it `register_object_short()`. What do you think?

P.S. Sorry for the delayed response. I had a couple very busy days at college. I will now be actively working to get this PR merged as quickly as possible.

ojwb:
I think it makes more sense to just specialise the existing template for `Xapian::Weight` - that makes it easy to only create one `Xapian::Weight` object of each registered sub-type and add a pointer to that object to both maps.

xapian-core/api/registry.cc
```diff
@@ -107,6 +111,65 @@ register_object(map<string, T*> & registry, const T & obj)
...
+
+    T * clone = obj.clone();
+    if (rare(!clone)) {
+	throw Xapian::InvalidOperationError("Unable to register object - clone() method returned NULL");
+    }
+
+    r1.first->second = clone;
+
+    T * clone2 = obj.clone();
+    r2.first->second = clone2;
```
ivmarkp: @ojwb I realize we need to call `clone()` more than just once here, otherwise valgrind complains about invalid read and writes for tests `registry2` and `double_register_leak`. The problem with calling `clone()` once and storing the pointer to the returned object to both `r1.first->second` and `r2.first->second` is that with multiple calls to `register_weighting_scheme()` (as is the case in the mentioned tests) pointer gets deleted at line 144 and it then tries to delete the same pointer at line 158 and that's where errors are reported by valgrind. A simple solution is to call clone once for both `r1` and `r2` which basically makes sure the problem that I just described never occurs.

ojwb: The key issue here is really object ownership.  One map needs to own the objects and the other just keeps pointers to them, and you only delete the pointer when updating one of the maps (and also in `clear_weighting_schemes()`).

There is a wrinkle here if someone tries to register a weighting scheme with the same long name but a different shortname to one already registered (assuming the existing map is the one which owns the pointers).  It think it's helpful to reject such attempts with `InvalidOperationError` anyway, as it would lead to a confusing situation.

xapian-core/api/registry.cc
```diff
@@ -175,37 +171,36 @@ Registry::Internal::add_defaults()
...
+    weighting_scheme = new Xapian::DLHWeight;
+    wtschemes_short[weighting_scheme->short_name()] = weighting_scheme;
+    weighting_scheme = new Xapian::PL2PlusWeight;
+    wtschemes_short[weighting_scheme->short_name()] = weighting_scheme;
+    weighting_scheme = new Xapian::PL2Weight;
+    wtschemes_short[weighting_scheme->short_name()] = weighting_scheme;
+    weighting_scheme = new Xapian::DPHWeight;
+    wtschemes_short[weighting_scheme->short_name()] = weighting_scheme;
+    weighting_scheme = new Xapian::LMWeight;
+    wtschemes_short[weighting_scheme->short_name()] = weighting_scheme;
```
ivmarkp: May be I'm missing something but the object that we are referring to here is really a pointer to object and if we do not create a second one we'll have to store the same object pointer in the second map as well and this leads to a lot of errors reported by valgrind for many tests all of which originate from line 239 in this file which is trying to delete the pointer already deleted in line 233.

```cpp
Xapian::Weight * weighting_scheme;
weighting_scheme = new Xapian::BB2Weight;
wtschemes[weighting_scheme->name()] = weighting_scheme;
wtschemes_short[weighting_scheme->short_name()] = weighting_scheme;
```

xapian-bindings/csharp/csharp.i
```diff
@@ -236,7 +236,24 @@ namespace Xapian {
...
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::TradWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::TfIdfWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::InL2Weight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::IfB2Weight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::IneB2Weight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::DLHWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::PL2PlusWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::PL2Weight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::DPHWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::LMWeight::create_from_parameters;
```
ivmarkp: I agree. Will make this list sorted with a fixup commit.

xapian-core/api/registry.cc
ojwb: > May be I'm missing something but the object that we are referring to here is really a pointer to object and if we do not create a second one we'll have to store the same object pointer in the second map as well

Yes, exactly.

> and this leads to a lot of errors reported by valgrind for many tests all of which originate from line 239 in this file which is trying to delete the pointer already deleted in line 233.

Line number references aren't entirely helpful as they are somewhat specific to a particular commit.

It seems you mean in `clear_weighting_schemes()` - you shouldn't delete the objects from both maps.  As I said, make one map the owner.  Then delete only from that map.  The other map just stores a different mapping to the same pointers.  You need to make sure that entries get added/updated/removed from both maps in step, but if you do then there's no leak and no way for an object to be used after it is deleted.

It's quite common to have multiple variables pointing to the same object, but it's important in such situations to be clear who actually owns the object (and so has responsibility for deleting it), or if the object wasn't dynamically allocated, be clear that this is the case.  Failing to get this right either results in memory leaks, or in use-after-free (which often has security implications).

xapian-core/api/registry.cc
```diff
@@ -260,13 +322,16 @@ void
...
+    register_object(internal->wtschemes, internal->wtschemes_short, wt);
 }
 
 const Xapian::Weight *
 Registry::get_weighting_scheme(const string & name) const
 {
     LOGCALL(API, const Xapian::Weight *, "Xapian::Registry::get_weighting_scheme", name);
+    const char * p = name.c_str();
+    if (std::islower(*p))
+	RETURN(lookup_object(internal->wtschemes_short, name));
```
ojwb: Use `C_islower()` rather than `std::islower()` (and drop the `<cctype>` include).  Functions like `std::islower()` change behaviour depending on the currently set locale (because the definition of what a lower case letter is varies by language), and user code may well change the current locale.  To avoid behaviour depending on the current locale, xapian-core has its own `C_islower()`, etc which act like `std::islower()` in the "C" locale (i.e. as if `setlocale(LC_CTYPE, "C");` were in effect).

Reusing `get_weighting_scheme()` for both lookups like this seems OK, but means that the long name really mustn't begin with a lower case letter.  While that is the current convention, I don't think it's documented as a requirement and certainly isn't currently enforced in any way, so perhaps if the short name lookup returns `NULL` we should continue on and try the long name lookup, to avoid needlessly breaking any current user subclasses that return a lower case name.

`name[0]` is simpler and clearer than using `c_str()` and then `*`, and avoids an extra variable.  But the code should also be robust to being passed an empty string.  So something like `!name.empty() && C_islower(name[0])`

xapian-core/weight/boolweight.cc
```diff
@@ -86,4 +92,14 @@ BoolWeight::get_maxextra() const
...
 }
 
+BoolWeight *
+BoolWeight::create_from_parameters(const char * p) const
+{
+    if (*p == '\0')
+	return new Xapian::BoolWeight();
+    if (*p == ' ')
+	throw InvalidArgumentError("No parameters are required for BoolWeight");
+    return new Xapian::BoolWeight();
```
ojwb: This is needlessly complicated, and yet not really correct.  Your code does:

* Empty string -> OK
* Non-empty string with space as first character -> bad
* Non-empty string with character other than space as first character -> OK

There are really just two cases:

* Empty string -> OK
* Non-empty string -> bad

xapian-core/weight/coordweight.cc
```diff
@@ -86,4 +92,12 @@ CoordWeight::get_maxextra() const
     return 0;
 }
 
+CoordWeight *
+CoordWeight::create_from_parameters(const char * p) const
+{
+    if (*p == ' ')
+	throw InvalidArgumentError("No parameters are required for CoordWeight");
+    return new Xapian::CoordWeight();
```
ojwb: This is the same situation as `BoolWeight` (i.e. there shouldn't be any parameters), but this is handled differently - differently wrongly.  The question that's relevant is whether the string is empty or not, not whether it starts with a space.

xapian-core/weight/dlhweight.cc
```diff
@@ -203,4 +209,14 @@ DLHWeight::get_maxextra() const
...
 
+DLHWeight *
+DLHWeight::create_from_parameters(const char * p) const
+{
+    if (*p == '\0')
+	return new Xapian::DLHWeight();
+    if (*p == ' ')
+	throw InvalidArgumentError("No parameters are required for DLHWeight");
+    return new Xapian::DLHWeight();
+}
```
ojwb: Same situation as `BoolWeight`...

xapian-core/weight/dphweight.cc
```diff
@@ -160,4 +166,14 @@ DPHWeight::get_maxextra() const
...
 }
 
+DPHWeight *
+DPHWeight::create_from_parameters(const char * p) const
+{
+    if (*p == '\0')
+	return new Xapian::DPHWeight();
+    if (*p == ' ')
+	throw InvalidArgumentError("No parameters are required for DPHWeight");
+    return new Xapian::DPHWeight();
```
ojwb: And again.

xapian-core/weight/tfidfweight.cc
```diff
@@ -238,4 +244,12 @@ TfIdfWeight::get_wtn(double wt, char c) const
     return wt;
 }
 
+TfIdfWeight *
+TfIdfWeight::create_from_parameters(const char * p) const
+{
+    if (*p == '\0')
+	return new Xapian::TfIdfWeight();
+    return new Xapian::TfIdfWeight(p + 1);
```
ojwb: This blindly ignores the first character of `p`.

xapian-core/weight/weight.cc
```diff
@@ -134,4 +144,25 @@ Weight::unserialise(const string &) const
...
+    const char *p = s.c_str();
+    std::string scheme;
+
+    while (*p != ' ') {
+	if (*p == '\0') break;
+	scheme += *p;
+	p++;
+    }
+
+    return reg.get_weighting_scheme(scheme)->create_from_parameters(p);
```
ojwb: We should increment `p` here if it was a space rather than the end of the string.

xapian-applications/omega/weight.cc
```diff
@@ -23,267 +24,21 @@
...
-	if (scheme == "coord") {
-	    enq.set_weighting_scheme(Xapian::CoordWeight());
-	    return;
-	}
-
-	if (scheme != "bool") {
-	    throw "Unknown $opt{weighting} setting: " + scheme;
-	}
+	delete wt;
+	wt = NULL;
```
ojwb: Setting `wt` to `NULL` doesn't protect us much as the variable's scope ends at the `}` on the next line.  But this is liable to trigger a compiler warning such as "variable is assigned a value which is never used".

xapian-bindings/csharp/csharp.i
```diff
@@ -236,6 +236,26 @@ namespace Xapian {
...
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::InL2Weight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::LMWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::PL2PlusWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::PL2Weight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::TfIdfWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::TradWeight::create_from_parameters;
+%warnfilter(WARN_CSHARP_COVARIANT_RET) Xapian::Weight::create_from_parameters;
+
+
 %include ../generic/except.i
```
ojwb: Minor nit, but the double blank line here is inconsistent with the spacing otherwise used.

xapian-core/api/registry.cc
```diff
@@ -260,14 +322,21 @@ void
...
     LOGCALL(API, const Xapian::Weight *, "Xapian::Registry::get_weighting_scheme", name);
-    RETURN(lookup_object(internal->wtschemes, name));
+    const Xapian::Weight * wt = lookup_object(internal->wtschemes, name);
+    const Xapian::Weight * wt_short = lookup_object(internal->wtschemes_short, name);
+    if (!name.empty() && C_islower(name[0])) {
+	if (wt_short == NULL)
+	    RETURN(wt);
+	RETURN(wt_short);
+    }
+    RETURN(wt);
```
ojwb: Here you look up both `wt` and `wt_short` every time, which is wasted effort in most cases.

You only need to look up `wt_short` if the lower case initial check succeeds.

And you only need to look up `wt` after you're done with potentially checking the short name - if you found an object by the short name, you can just return it.

xapian-core/weight/weight.cc
```diff
@@ -134,4 +144,27 @@ Weight::unserialise(const string &) const
...
+
+    while (*p != ' ') {
+	if (*p == '\0') break;
+	scheme += *p;
+	p++;
+    }
+
+    if (*p == ' ')
+	return reg.get_weighting_scheme(scheme)->create_from_parameters(p++);
+    return reg.get_weighting_scheme(scheme)->create_from_parameters(p);
```
ojwb: This (a) duplicates the code needlessly and (b) doesn't actually work - `reg.get_weighting_scheme(scheme)->create_from_parameters(p++);` calls the method with the current value of `p` and **then** increments `p`.

You can just increment `p` if it points to a space, then have a single return using `p`.

xapian-core/api/registry.cc
```diff
@@ -107,6 +111,49 @@ register_object(map<string, T*> & registry, const T & obj)
...
+static inline void
+register_object(map<string, T*> & registry1, map<string, T*> & registry2,
+		const T & obj)
+{
+    string name = obj.name();
+    if (rare(name.empty())) {
+	throw Xapian::InvalidOperationError("Unable to register object - name() method returned empty string");
+    }
+
+    string short_name = obj.name();
```
ojwb: Should call `short_name()`, not `name()` again.

xapian-core/api/registry.cc
```diff
@@ -107,6 +111,49 @@ register_object(map<string, T*> & registry, const T & obj)
...
+    }
+
+    T * clone = obj.clone();
+    if (rare(!clone)) {
+	throw Xapian::InvalidOperationError("Unable to register object - clone() method returned NULL");
+    }
+
+    r1.first->second = clone;
+    r2.first->second = clone;
+}
```
ojwb: We're mostly there, but you don't seem to have addressed the wrinkle I mentioned here with sharing the object between the two maps.

To re-iterate, if I register a `Weight` object which reports the same `name()` as an already registered object but a different `short_name()`, then the object in the name map is replaced (and so the old object deleted) but the entry in the short name map for the old short name is left, and that's now a dangling pointer.  This is an obscure case, but we really shouldn't leave dangling pointers like this where they could get used as they're a potential security risk.

We need to take a bit more care to deal with this situation.

One option (the one I mentioned before) is to reject attempts to do this (and throw `Xapian::InvalidOperatorError` instead), which has the advantage of not letting people create a situation like this accidentally.  To do that we need to check that (a) entries for the two names either both do or both don't already exist (because if only one of them exists, a previous object used the same short or long name, but not both, and (b) if they both do both exist, that the pointers in them are the same (because otherwise we're trying to replace the long name entry for one weight and the short name entry for a different one).  I think the correct check is something like: `(r1.second != r2.second || (r1.second && r1.first->second != r2.first->second))`

We could reliably remove the old object from the short name map instead - in the case where we replacing the object in the name map, before we delete the old object we need to:

* check if we're replacing an existing object in the short_name map too, and if so if that object is the same one we're replacing in the name map - if so, we can just continue as you already do.
* otherwise, we need to locate the corresponding short_name entry and remove it from the short_name map (at least if there is such an entry - it could be it got overwritten by registering a second weighting scheme with the same short name) - we can do this by doing a lookup using the short name from the object we're about to delete, and then comparing pointers to make sure it is the same object (another could have been registered over the top).

So that would be something like:

```
    if (!r1.second) {
	// ...
	T * p = NULL;
	swap(p, r1.first->second);
        if (p && (r2.second || r2.first->second != p)) {
            // We're replacing an object with a different short name to us,
            // so check if the old object is still registered under its short
            // name.
            auto it = registry2.find(p->short_name());
            if (it != registry2.end() && *it == p)
                registry2.erase(it);
        }
        delete p;
    }
```

But that's more complex, and just seems to enable creating a confusing situation.  Being able to updated a registered weighting scheme is probably useful, but being able to create a situation in which two registered objects have the same short name seems mostly a trap for the unwary.

xapian-core/api/registry.cc
```diff
@@ -107,6 +111,49 @@ register_object(map<string, T*> & registry, const T & obj)
...
+{
+    string name = obj.name();
+    if (rare(name.empty())) {
+	throw Xapian::InvalidOperationError("Unable to register object - name() method returned empty string");
+    }
+
+    string short_name = obj.name();
+    if (rare(short_name.empty())) {
+	throw Xapian::InvalidOperationError("Unable to register object - short_name() method returned empty string");
+    }
```
ojwb: I think it's OK for a weighting scheme not to provide a short name, especially as allowing that means existing user weighting scheme should continue to work (they just won't provide the ability to be created from a string description).

I think just gating everything which sets or uses `r2` with `if (!short_name.empty())` achieves that.

ojwb:
The other thing this really ought to have is some feature tests in the test suite (if it had, that should have caught the `string short_name = obj.name();` bug).

xapian-core/api/registry.cc
ivmarkp: Yes, apologies for this mistake.

xapian-core/api/registry.cc
ivmarkp: So, we should go with the first option i.e. reject such attempts with `InvalidOperationError`?

Also, I can recall actually adding the check when you suggested earlier but I might have failed to commit that or may be it accidentally got lost at some point with `git stash clear` while jumping between different branches.

ivmarkp: Should we be checking for `!r1.second` instead of `r1.second` in the second condition in the check `(r1.second != r2.second || (r1.second && r1.first->second != r2.first->second))`?

ivmarkp:
It actually gets covered under the check and will throw a `Xapian::InvalidOperatorError` from an existing test named `double_register_leak` with backend "none".

xapian-core/api/registry.cc
ojwb: I think the first option probably makes most sense.

And yes, it should indeed be `!r1.second` - well spotted.

ojwb:
OK, I think we're there.  Thanks for all your work on this.

I'm on an old slow machine right now, so I'll sort out the merge tomorrow when I'm on my dev box.

ivmarkp:
Great, thanks for taking the time to review. Glad to see this getting merged.

PR #153, opened 2017-04-23
omindex: delay libmagic checks(ticket#743)

caiyulun:
Link to the Ticket: https://trac.xapian.org/ticket/743

the libmagic call to get mime type is expensive, so we should check it later, for example, we can check the size by call the stat, check the timestamps and the DB for an existing entry before call libmagic.

This PR now achieves to check the filesize before call the libmagic.

To solved next:

since check the timestamps of a file need to check through the database in some situation: when the timestamps of the file return from stat call (says T1 for  convenience) is older than the last time we run omindex, we need to  look through the database to determine if the file is indexed and get the modify time when it is indexed (says T2 for  convenience), if T2 is older than T1, we should re-index the file.

it is expensive to iterate through the DB, so we should figure out the order between libmagic check and timestamps in next step.

### DIFF
--- a/xapian-applications/omega/omindex.cc
+++ b/xapian-applications/omega/omindex.cc
@@ -93,14 +93,7 @@ index_file(const string &file, const string &url, DirectoryIterator & d,
 	urlterm = hash_long_term(urlterm, MAX_SAFE_TERM_LENGTH);
 
     string mimetype = mimetype_from_ext(mime_map, ext);
-    if (mimetype.empty()) {
-	mimetype = d.get_magic_mimetype();
-	if (mimetype.empty()) {
-	    skip(urlterm, file.substr(root.size()), "Unknown extension and unrecognised format",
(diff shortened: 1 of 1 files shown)


ojwb:
Merged to git master as d32e13545e699e6920e0e5a3ebfe8a94206de32f, thanks.

xapian-applications/omega/omindex.cc
```diff
@@ -133,6 +120,20 @@ index_file(const string &file, const string &url, DirectoryIterator & d,
 	return;
     }
 
+    // if can't get the mime type from extension,call libmagic to get it
```
ojwb: One small but quite important point - you should aim to make comments clear and unambiguous, and getting the grammar and punctuation helps that (though I realise that's often harder for non-native speakers).

So better to capitalise the sentence, give the verb a subject ("we can't" rather than "can't"), put a space after the comma, and a full stop (period) at the end.  I've pushed a commit to fix these in this case.

PR #154, opened 2017-04-24
Fix build on FreeBSD

sunpoet:
In net/tcpclient.cc, include resolver.h after safesyssocket.h to solve the following build failure.

```
--- net/tcpclient.lo ---
In file included from net/tcpclient.cc:28:
net/resolver.h:102:20: error: use of undeclared identifier 'AF_INET'
        hints.ai_family = AF_INET;
                          ^
net/resolver.h:103:22: error: use of undeclared identifier 'SOCK_STREAM'
        hints.ai_socktype = SOCK_STREAM;
                            ^
```

### DIFF
--- a/xapian-core/net/tcpclient.cc
+++ b/xapian-core/net/tcpclient.cc
@@ -25,7 +25,6 @@
 #include "tcpclient.h"
 
 #include "remoteconnection.h"
-#include "resolver.h"
 #include "str.h"
 #include <xapian/error.h>
 
@@ -36,6 +35,7 @@
 #include "safesysselect.h"
 #include "safesyssocket.h"
 #include "socket_utils.h"
+#include "resolver.h"
 
 #include <cmath>
 #include <cstring>


jaylett:
Would it be better to include `safesyssocket.h` in `net/resolver.h`, so that other clients of it would also work with any include order?

ojwb:
> Would it be better to include safesyssocket.h in net/resolver.h, so that other clients of it would also work with any include order?

Our (possibly unwritten) policy is that a header ought to include any
other headers it needs to work, so that would be the more appropriate
fix.  Will apply shortly.

But thanks very much for the report and analysis.

sunpoet:
That would be fine. I tried to add safesyssocket.h to net/resolver.h first when I got build failure. I then tried if re-order fix the problem, therefore I submitted this patch.

I'll update local patch in FreeBSD ports tree to use 40f35d3.

Thanks!

PR #155, opened 2017-05-23
Fix dangling reference to stopper in letor

AyushP123:
Since stopper was created locally in Xapian::QueryParser::initialise_queryparser, on calling the function we would get an invalid reference to the stopper. This has been fixed by allocating the stopper using new() and releasing it after it has been set in parser.

### DIFF
--- a/xapian-letor/ranker/ranker.cc
+++ b/xapian-letor/ranker/ranker.cc
@@ -225,7 +225,8 @@ parse_query_string(const string & query_line, int line_number)
 static Xapian::QueryParser
 initialise_queryparser(const Xapian::Database & db)
 {
-    Xapian::SimpleStopper mystopper(sw, sw + sizeof(sw) / sizeof(sw[0]));
+    Xapian::SimpleStopper* mystopper;
+    mystopper = new Xapian::SimpleStopper(sw, sw + sizeof(sw) / sizeof(sw[0]));
     Xapian::Stem stemmer("english");
     Xapian::QueryParser parser;
     parser.add_prefix("title", "S");
@@ -234,7 +235,7 @@ initialise_queryparser(const Xapian::Database & db)
     parser.set_default_op(Xapian::Query::OP_OR);
     parser.set_stemmer(stemmer);
     parser.set_stemming_strategy(Xapian::QueryParser::STEM_SOME);
-    parser.set_stopper(&mystopper);
+    parser.set_stopper(mystopper->release());
     return parser;
 }
 


jaylett:
@AyushP123 can you rebase this to latest master to see if that fixes the test problem? It might not, but it's a better place to start.

ojwb:
The real problem was that letor's testsuite was hard-coded to link against a build of xapian-core in a combined tree (while the letor library used the xapian-core version probed by its configure script), and assumed the library was `libxapian.la` which isn't true in development versions - e.g. it's `libxapian-1.5.la` currently.

The reason we haven't hit this previously was that CI has only run the bindings testsuite for the last few months, but vivekp noticed and I fixed that a few days ago.

It presumably works in your tree because you have a `libxapian.la` in xapian-core from building before the change to add `-1.5`.

I've pushed commits to fix letor to always use the configure-probed xapian-core and not assume it's being built as part of a combined tree, so rebasing on the latest master should now fix the CI failure.

xapian-letor/ranker/ranker.cc
```diff
@@ -234,7 +234,8 @@ initialise_queryparser(const Xapian::Database & db)
     parser.set_default_op(Xapian::Query::OP_OR);
     parser.set_stemmer(stemmer);
     parser.set_stemming_strategy(Xapian::QueryParser::STEM_SOME);
-    parser.set_stopper(&mystopper);
+    parser.set_stopper(mystopper);
+    mystopper->release();
```
ojwb: The docs say:

> You can hand ownership of a dynamically allocated `Stopper` object to Xapian by calling `release()` and then passing the object to a Xapian method.

However, you do things the other way around - i.e. pass the object to a Xapian method and **then** call `release()` on this.  

Your order doesn't work correctly - the object is now being reference counted, but nothing owns it (because it wasn't being reference counted when you passing it to the method, so that didn't add a reference).  In this case that probably means it will never be released (i.e. a memory leak), though it could also mean it got released before being done with (if `QueryParser` now passes it to something else which temporarily takes a reference and releases it again).

`release()` actually returns the object it was called on, so you can release and hand-over in one:

```
parser.set_stopper(mystopper->release());
```

xapian-letor/ranker/ranker.cc
```diff
@@ -225,7 +225,7 @@ parse_query_string(const string & query_line, int line_number)
 static Xapian::QueryParser
 initialise_queryparser(const Xapian::Database & db)
 {
-    Xapian::SimpleStopper mystopper(sw, sw + sizeof(sw) / sizeof(sw[0]));
+    Xapian::SimpleStopper* mystopper = new Xapian::SimpleStopper(sw, sw + sizeof(sw) / sizeof(sw[0]));
```
ojwb: This line is too long now.

ojwb:
Judging by the branch name, it looks like you made this change in the github online editor.

That's probably OK for making trivial typo corrections in documentation, but a really bad way to make code changes.  Even a simple code change is hard to make correctly if you don't try to compile it before submitting it (as the necessity to make a follow-up commit here demonstrates).  If there's CI that will hopefully catch such problems, but fundamentally this is not the right approach.

Also, your "fixup" commit isn't a real fixup commit - the commit message doesn't match the required pattern, so `git rebase -i --autosquash` doesn't recognise it.  This is really another consequence of using the online editor to make changes that it's not appropriate to use it for - if you were working in a clone of the repo, you'd use `git commit --fixup` and that would create a suitably formatted commit message for you.

AyushP123:
@jaylett @ojwb .. I have incorporated the changes suggested by you ... we should merge this PR

ojwb:
The code's good now.

The commit message has some issues.  In the interests of not dragging this out further, I've just fixed them, but please bear these in mind for future commit messages:

>  Fix dangling reference to stopper in letor
>
> Since stopper was created locally in Xapian::QueryParser::initialise_queryparser, on calling the function we would get an invalid reference to the stopper. This has been fixed by allocating the stopper using new() and releasing it after it has been set in parser.

First a formatting point - by convention git commit messages are line wrapped to 72 columns (using the online editor thing generally results in each paragraph being one really long line, as here).

`Xapian::QueryParser::initialise_queryparser` is method (or member) `initialise_queryparser` of class `Xapian::QueryParser` - there is no such method or member.  There's a static non-class function `initialise_queryparser` which returns `Xapian::QueryParser`.

The invalid reference isn't created **on calling** the function, but rather when the function returns.

Saying **releasing it after it has been set in the parser** is misleading - `release()` isn't called after setting now (that code was incorrect), and `release()` effectively hands over ownership whereas "releasing" suggest the object is deallocated.

`new()` would be better written as `new` (since it's not usually called function-style).  OTOH, `initialise_queryparser` _is_ a function, so probably more clearly written as `initialise_queryparser()`.

PR #156, opened 2017-06-04
Returning MSet instead of sorted Docids after re-ranking (Ticket #734)

AyushP123:
Changes made to MSet: Added two functions to the public API set_new_weights and re_rank to MSet along with a helper function set_new_weights_helper.
1. set_new_weights: assigns new weights to items
2. re_rank sorts the MSet by sorting items on the basis of their weights.

Changes made to Letor: 
1. Keeping the vector<FeatureVector> in the original order. Ranker only updates weights.
2. New weights are obtained from vector<FeatureVector> and assigned to items and it is sorted according to the wieghts subsequently.
3. Printing the docids from MSet to show the updated order after ranker re-ranks MSet

PR #157, opened 2017-06-13
Returning MSet instead of sorted Docids after re-ranking Approach-2 (Ticket #734)

AyushP123:
An Alternate approach to #156. In this approach instead of updating all the weights of an MSet we update the weight of each document one by one and update the MSet at the end. The major difference is the xapian-core api set_item_weight which can now be used to set the weight of any item and the MSet will update itself (i.e. resort and update max_attained and max_possible) automatically. This can help users to update the weights of any number of documents in any order.


AyushP123:
@jaylett , @ojwb ..... I have just created this pull request as I had the idea of an alternate approach which in my opinion is better ..... we can review this PR at a later date

ojwb:
@AyushP123 This really isn't a helpful PR to open at this point - there's too much other stuff in the history to actually see what differences you're proposing.  Also, you really ought to focus on addressing the feedback on the existing PRs rather than opening more.

Let's just close this for now, and once #156 is merged, you can rebase this on current master and open a PR which just shows the changes compared to the other approach.

PR #158, opened 2017-06-15
Add seterror command in omega

samuelharden:
Currently, it's not possible to stop the execution of a query and return an error message from the template. 

Let's say we want to check if the user passed a very high hitsperpage, which will hang the browser. We can check the CGI parameter in the template and use the seterror command to set error and prevent the query from being executed.

### DIFF
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -542,14 +542,14 @@ $set{OPT,VALUE}
 	Similarly, ``XFOO:stemmer`` specifies the stemmer to use for field
 	``XFOO``, with ``stemmer`` providing a default.
 
-$seterror{error_msg}
+$seterror{ERROR_MESSAGE}
 	set error message for the current execution, which can also be looked
 	up using ``$error``.
 
-	Using ``$seterror`` error early in template prevent running query.
-	Saves us from executing query unnecessarily.
+	Using ``$seterror`` error early in template prevents running the query.
 
-	You can use ``$seterror`` when the user enters a wrong parameter in the search.
+	For example, ``$seterror`` can be used when the user enters a wrong
+	parameter in the search.
 
 $setrelevant{docids}
 	add documents into the RSet
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -542,6 +542,15 @@ $set{OPT,VALUE}
 	Similarly, ``XFOO:stemmer`` specifies the stemmer to use for field
 	``XFOO``, with ``stemmer`` providing a default.
 
+$seterror{error_msg}
+	set error message for the current execution, which can also be looked
+	up using ``$error``.
+
+	Using ``$seterror`` error early in template prevent running query.
+	Saves us from executing query unnecessarily.
+
+	You can use ``$seterror`` when the user enters a wrong parameter in the search.
+
 $setrelevant{docids}
 	add documents into the RSet
 
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -414,16 +414,20 @@ printf '$highlight{$cgi{text},$cgi{list},$cgi{open},$cgi{close}}' > "$TEST_TEMPL
 testcase 'A list of <b>words</b>' list="words" text="A list of words" open="<b>" close="</b>"
 testcase 'A *list* of *words*' list="words${tab}list" text="A list of words" open="*" close="*"
 
-#Test setting seterror and mset size, should not run the query after setting error.
+# Test setting seterror and mset size, should not run the query after setting error.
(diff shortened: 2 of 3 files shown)


xapian-applications/omega/omegatest
```diff
@@ -414,6 +414,20 @@ printf '$highlight{$cgi{text},$cgi{list},$cgi{open},$cgi{close}}' > "$TEST_TEMPL
...
+
+#Test arguments inside seterror are evaluated
+printf '$set{error,sample error}$seterror{$opt{error}}$msize$if{$error,!$error!}' > "$TEST_TEMPLATE"
+testcase '0!sample error!' P=text
+
+#Test msize when error set after running query, should not affect running of query.
+printf '$last$if{$cgi{ERR},$seterror{$cgi{ERR}}}$msize$if{$error,!$error!}' > "$TEST_TEMPLATE"
+testcase '11' P=text
+testcase '11!boo!' P=text ERR=boo
+
```
ojwb: It would be good to test with some HTML-special characters in the error to make sure it doesn't get sent through HTML entity encoding (as the C++ `throw` approach does).  Making the string `<sample error>` in the second case would do that.

xapian-applications/omega/omegatest
```diff
@@ -414,6 +414,20 @@ printf '$highlight{$cgi{text},$cgi{list},$cgi{open},$cgi{close}}' > "$TEST_TEMPL
 testcase 'A list of <b>words</b>' list="words" text="A list of words" open="<b>" close="</b>"
 testcase 'A *list* of *words*' list="words${tab}list" text="A list of words" open="*" close="*"
 
+#Test setting seterror and mset size, should not run the query after setting error.
```
ojwb: Please put a space between the `#` and the comment text for readability (and also consistency with all the existing comments in this file and elsewhere in the codebase).  Also below.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -542,6 +542,15 @@ $set{OPT,VALUE}
...
 	``XFOO``, with ``stemmer`` providing a default.
 
+$seterror{error_msg}
+	set error message for the current execution, which can also be looked
+	up using ``$error``.
+
+	Using ``$seterror`` error early in template prevent running query.
+	Saves us from executing query unnecessarily.
+
+	You can use ``$seterror`` when the user enters a wrong parameter in the search.
```
ojwb: We should make clear that this is just an example of how you can use `$seterror`.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -542,6 +542,15 @@ $set{OPT,VALUE}
 	Similarly, ``XFOO:stemmer`` specifies the stemmer to use for field
 	``XFOO``, with ``stemmer`` providing a default.
 
+$seterror{error_msg}
+	set error message for the current execution, which can also be looked
+	up using ``$error``.
+
+	Using ``$seterror`` error early in template prevent running query.
```
ojwb: "prevents" and "the query".

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -542,6 +542,15 @@ $set{OPT,VALUE}
...
+	Saves us from executing query unnecessarily.
```
ojwb: This really just says the same thing as the previous sentence, but less clearly - I'd drop it, or rephrase if it's supposed to make a different point.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -542,6 +542,15 @@ $set{OPT,VALUE}
 	Similarly, ``XFOO:stemmer`` specifies the stemmer to use for field
 	``XFOO``, with ``stemmer`` providing a default.
 
+$seterror{error_msg}
```
ojwb: The arguments in this document are conventional upper case (i.e. `ERROR_MSG` or perhaps better to avoid the abbreviation and say `ERROR_MESSAGE` or `MESSAGE`).

xapian-applications/omega/query.cc
```diff
@@ -1116,6 +1117,7 @@ T(relevant,	   0, 1, N, Q), // is document relevant?
 T(relevants,	   0, 0, N, Q), // return list of relevant documents
 T(score,	   0, 0, N, 0), // score (0-10) of current hit
 T(set,		   2, 2, N, 0), // set option value
+T(seterror,	   1, 1, 1, 0), // set error_msg, setting it early stops query execution
```
ojwb: Make the 4th parameter `N` rather than `1`, like we do for all the existing commands which evaluate all their arguments (the processing code doesn't care if evalargs > maxargs, and by using `N` we don't have the potential trap of someone adding a new optional argument to a command and increasing maxargs, but failing to also increase evalargs).

ojwb:
Thanks, squashed and merged via git.

I've also added a note to `$error` mentioning `$seterror`.

PR #160, opened 2017-06-18
Add $hash OmegaScript command

ivmarkp:
Leveraging the existing implementation of MD5 hashing algorithm, this command will enable users to generate a unique ID for the given query text (e.g. by using $query{} command).

P.S. This feature is a subset of the clickstream logging functionality.


jaylett:
This is a small-ish change — it should have been in one commit, in my view. We can do that on merge, however; don't change the commit boundaries now.

xapian-applications/omega/query.cc
```diff
@@ -1940,6 +1955,19 @@ eval(const string &fmt, const vector<string> &param)
 		value = args[0];
 		url_prettify(value);
 		break;
+	    case CMD_qid: {
+		string qstr = args[0];
+		trim_string(qstr);
```
ojwb: What's the reason for removing leading and trailing spaces (rather than just processing the query as-is, or performing more normalisation)?

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -418,6 +418,11 @@ $prettyurl{URL}
 	URL for displaying to the user (rather than in links), but which should
 	still work if copied and pasted.
 
+$qid{TEXT}
+	unique ID for the query TEXT which would be usually taken from $query.
+	This command uses MD5 hashing algorithm to produce a 128-bit hash for
+	the given query text which is used as a unique ID.
```
ojwb: I'm not keen to tie the implementation to using MD5 in the documentation like this does.  MD5 is not a secure hash today, and organisations may (with good intent, though arguably somewhat misguidedly) impose a policy that MD5 should not be used at all.  As this stands, it would be hard for us to move `$qid` to using a more modern hash, because the documentation explicitly says that it uses MD5.  If we're going to pick a particular hash like this, MD5 is not the right choice.

Also, the name `$qid` is rather cryptic and quite specific to this particular use - where possible we prefer to add more generic functions which can be used to achieve a specific requirement but have wider applicability, and the space trimming makes `$qid` less useful as a generic way to hash something.

So I think it would make more sense to add a new omegascript command which calculates the hash of a given text string, and to keep that command generic, any normalising you want to do is left to the code which makes use of that.  So instead of `$qid{$query}` you might write:

```
$md5{$transform{$query,^ *(.*?) *$,\1}}
```

Or perhaps `$hash{md5,$transform{$query,^ *(.*?) *$,\1}}` so it's easier to add new hash algorithms and configure them from the template.

ojwb:
> it should have been in one commit, in my view

Certainly the first and third commits don't make sense separately - you'll get a link error with just the first one.

> don't change the commit boundaries now.

Indeed.

xapian-applications/omega/query.cc
ivmarkp: This is to save us from assigning different IDs to some queries that are essentially the same (queries with accidental leading or trailing spaces e.g. "string   " and "  string"). Also, as it isn't useful to log such queries with different IDs, I think it would help minimise noise in the log data.

xapian-applications/omega/docs/omegascript.rst
ivmarkp: I agree; this would be more useful as a generic command like `$hash{}` that you have proposed. I'll update the PR with the implementation of `$hash{}` command.

xapian-applications/omega/query.cc
```diff
@@ -1064,6 +1066,7 @@ T(fmt,		   0, 0, N, 0), // name of current format
 T(freq,		   1, 1, N, 0), // frequency of a term
 T(ge,		   2, 2, N, 0), // test >=
 T(gt,		   2, 2, N, 0), // test >
+T(hash,		   2, 2, N, 0), // hash a string using the specified hash function
 T(highlight,	   2, 4, N, 0), // html escape and highlight words from list
```
ivmarkp: Don't know it's showing an extra tab here? Same as in line 1066, 1064 etc. above.

ivmarkp:
`The job exceeded the maximum time limit for jobs, and has been terminated.` -- from travis failed job log. Can we restart the build so it can show us a green tick? There should be no build failure issues as far as linux is concerned.

xapian-applications/omega/query.cc
ojwb: It's because there's an extra column at the start of the patch (with `+` for added lines, `-` for removed lines and ` ` for context lines).  In the normal file, the `,` is column 7 and the tab takes us to column 8.  In the patch, the `,` is column 8 so that same tab takes us to column 16.

So don't worry about it.  It's not you.

ojwb:
> Can we restart the build so it can show us a green tick?

```
Running test: keepalive1...

The job exceeded the maximum time limit for jobs, and has been terminated.
```

Oh the irony!

Restarted.  It's in the xapian-core tests, so I don't see how this patch can be to blame.

xapian-applications/omega/query.cc
```diff
@@ -1566,6 +1566,24 @@ eval(const string &fmt, const vector<string> &param)
 		if (string_to_int(args[0]) > string_to_int(args[1]))
 		    value = "true";
 		break;
+	    case CMD_hash: {
+		string hash = args[0];
+		string data = args[1];
+		std::transform(hash.begin(), hash.end(), hash.begin(), ::tolower);
+		value = "";
```
ojwb: You can assume `value` will start empty.

xapian-applications/omega/query.cc
```diff
@@ -1566,6 +1566,24 @@ eval(const string &fmt, const vector<string> &param)
...
+		if (hash == "md5") {
+		    trim_string(data);
```
ojwb: A generic `$hash` shouldn't trim the data - we should do any normalisation in the template where we use `$hash` for this purpose.

xapian-applications/omega/query.cc
ojwb: What is using this hashed query id supposed to accomplish?

If it's to track a particular user's "search session", then trimming spaces doesn't seem necessary - users don't commonly repeat the exact same query but add or remove spaces at the start or end or it.

If we're trying to normalise query strings which are really the same, then it doesn't really seem to be doing enough.  E.g. something like `$transform{ $query ,\s+, ,g}` would normalise all runs of whitespace to a single space (and put a single space at the start and end - that's just because it's simpler to ensure a single space there than none).  Or perhaps we should go further and use the terms parsed from the query or something, so that punctuation and word order (outside of phrases) don't matter.

xapian-applications/omega/query.cc
ivmarkp: Yeah, this `trim_string()` function isn't useful anymore here so I'll go ahead and remove it.

xapian-applications/omega/query.cc
ivmarkp: Okay, will remove this statement.

xapian-applications/omega/query.cc
ivmarkp: Okay, cool.

xapian-applications/omega/query.cc
ivmarkp: > What is using this hashed query id supposed to accomplish?

Query ID basically provides a means to uniquely identify each query from clickstream log if we need to do so in the future. For now, it's being particularly helpful in preprocessing step which I'm currently onto, for obtaining the final clickstream log file.

> If it's to track a particular user's "search session", then trimming spaces doesn't seem necessary - users don't commonly repeat the exact same query but add or remove spaces at the start or end or it.

This was based on the idea that leading or trailing whitespaces in query string wouldn't make query anything more useful other than leading to some duplicate log entries so I thought it was best to just rid off them to keep log file neat.

Do you suggest using the transform command that you mentioned earlier `$transform{$query,^ *(.*?) *$,\1}` or this `$transform{ $query ,\s+, ,g}` for dealing with whitespaces?

xapian-applications/omega/omegatest
```diff
@@ -432,6 +432,15 @@ printf '$last$if{$cgi{ERR},$seterror{$cgi{ERR}}}$msize$if{$error,!$error!}' > "$
...
 testcase '11!boo!' P=text ERR=boo
 
+# Feature tests for $hash{}
+printf '$hash{md5,$cgi{text}}' > "$TEST_TEMPLATE"
+testcase '098f6bcd4621d373cade4e832627b4f6' text="test"
+testcase 'b4c216e4da73d1d01277ef46d0514821' text="simple query"
+testcase 'd41d8cd98f00b204e9800998ecf8427e' text=""
+
+printf '$hash{sha-1,$cgi{text}}' > "$TEST_TEMPLATE"
+testcase '' text ="test"
```
jaylett: Unwanted space after `text`?

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -222,6 +222,11 @@ $fmt
 $freq{term}
 	frequency of a term
 
+$hash{HASH,TEXT}
+    unique ID for TEXT string using the hashing algorithm specified by HASH.
+    Currently, this command only supports MD5 hashing algorithm which can be
+    used by specifying HASH as "md5" (without quotes).
```
jaylett: Since `HASH` must be `md5` at the moment, I suspect it'd be better to put the hash argument second, and make it optional but defaulting to md5.

It's also perhaps worth noting which characters will be emitted by `$hash{}` (ie the hex characters), since that tells people whether they'll need escaping in various contexts. Also pointing out the length in characters of an md5 hash may be useful.

Finally, I'd put md5 in a "fixed-space literal" (surrounded by double backticks) rather than enquoting it then saying "(without quotes)".

samuelharden: Currently, we only have one algorithm. It best that if it works without specifying Hash Algorithm name.

We are even doing transform operation to make it all lower case. Even though we just have only one algorithm.

samuelharden:
Can we also add a GSOC label to this?

xapian-applications/omega/docs/omegascript.rst
ivmarkp: @jaylett @samuelharden Yes, it makes sense to make `HASH` optional with md5 as default. I've made the required changes.

xapian-applications/omega/omegatest
ivmarkp: Yes, it seems like it. Will fix this.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -222,10 +222,11 @@ $fmt
 $freq{term}
 	frequency of a term
 
-$hash{HASH,TEXT}
-    unique ID for TEXT string using the hashing algorithm specified by HASH.
-    Currently, this command only supports MD5 hashing algorithm which can be
-    used by specifying HASH as "md5" (without quotes).
+$hash{TEXT[,HASH}
```
jaylett: Missing closing square bracket :)

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -222,10 +222,11 @@ $fmt
...
+    unique ID for ``TEXT`` string using the hashing algorithm specified by
+    ``HASH`` which defaults to MD5 hashing algorithm. Currently, this command
+    only supports MD5 which yields a 128-bit hash sequence of 32 hexadecimal
```
jaylett: Perhaps "…serialised as 32 hexadecimal"?

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -222,10 +222,11 @@ $fmt
 $freq{term}
 	frequency of a term
 
-$hash{HASH,TEXT}
-    unique ID for TEXT string using the hashing algorithm specified by HASH.
-    Currently, this command only supports MD5 hashing algorithm which can be
-    used by specifying HASH as "md5" (without quotes).
+$hash{TEXT[,HASH}
+    unique ID for ``TEXT`` string using the hashing algorithm specified by
+    ``HASH`` which defaults to MD5 hashing algorithm. Currently, this command
```
jaylett: I'd drop the second "hashing algorithm".

xapian-applications/omega/query.cc
```diff
@@ -1555,8 +1555,9 @@ eval(const string &fmt, const vector<string> &param)
 		    value = "true";
 		break;
 	    case CMD_hash: {
-		string hash = args[0];
-		string data = args[1];
+		string data = args[0];
+		string hash = "md5";
+		if (args.size() > 1) hash = args[1];
 		std::transform(hash.begin(), hash.end(), hash.begin(), ::tolower);
```
jaylett: No need to do this if the value was defaulted.

xapian-applications/omega/docs/omegascript.rst
ivmarkp: Oh, sorry. Will add that.

xapian-applications/omega/query.cc
ivmarkp: Ah, right. Will remove this transform operation.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -222,10 +222,10 @@ $fmt
...
 	frequency of a term
 
-$hash{TEXT[,HASH}
+$hash{TEXT[,HASH]}
     unique ID for ``TEXT`` string using the hashing algorithm specified by
-    ``HASH`` which defaults to MD5 hashing algorithm. Currently, this command
-    only supports MD5 which yields a 128-bit hash sequence of 32 hexadecimal
+    ``HASH`` which defaults to MD5. Currently, this command only supports
+    MD5 which yields a 128-bit hash sequence serialised as 32 hexadecimal
     characters.
```
ojwb: By having a default and making that `md5` I'm slightly concerned that we actually tie ourselves to supporting MD5 for longer, because changing the default causes "quiet" problems.  If we drop support for an algorithm, at least that should make templates which specify that algorithm explicitly fail with an error.  It seems likely we'll eventually end up in a situation where basically everyone has to specify the hash to use (because even now, for most uses you don't want MD5 and that's not going to become less true with time), but we can't change the default because that could break templates for someone not specifying the hash but assuming the default is MD5 (not an unreasonable assumption, as we're documenting this above).

So I'm not entirely sure if having a default for HASH is useful or not - it seems to really depend if there are cases where you want to hash something but you really don't care what hash algorithm is used, at least to the extent that you might use a different algorithm tomorrow.  What use cases did whoever suggested having a default have in mind?

Also, although the docs say HASH can be omitted, the command table in `query.cc` still marks `$hash` as taking at least 2 parameters.  If we are going to support a default, then that needs fixing and there should be a testcase to check that the defaulting actually works.

xapian-applications/omega/query.cc
```diff
@@ -1558,7 +1558,6 @@ eval(const string &fmt, const vector<string> &param)
 		string data = args[0];
 		string hash = "md5";
 		if (args.size() > 1) hash = args[1];
-		std::transform(hash.begin(), hash.end(), hash.begin(), ::tolower);
```
ojwb: While only recognising `md5` in lower case seems reasonable to me (in general, OmegaScript is case sensitive, and we expect lower case tokens for things like query parser flags, prefix settings, etc), the documentation ought to be clear on this.

Also, quietly producing empty output if the algorithm name isn't recognised doesn't seem helpful, especially as it'll mean everything will appear to hash the same.  I think that should be an error.

xapian-applications/omega/docs/omegascript.rst
ojwb: We could explicitly document that the default is volatile, and perhaps give some criterion by which it's selected (e.g. "generally believed to be the most secure of those supported" or something).  But for that we also should have some idea of when someone might actually use the default.

jaylett: If we start talking about security, we probably shouldn't use md5 at all. There's an advantage to having warning that the algorithm is going to change, as well. So perhaps dropping the default entirely would be easiest (although there's something a little weird about having an option with only one valid choice, which nonetheless has to be provided).

xapian-applications/omega/query.cc
ivmarkp: > While only recognising md5 in lower case seems reasonable to me (in general, OmegaScript is case sensitive, and we expect lower case tokens for things like query parser flags, prefix settings, etc), the documentation ought to be clear on this.

Should I include this `std::transform` operation back here with a note in the documentation that `HASH` must be lowercase?

> Also, quietly producing empty output if the algorithm name isn't recognised doesn't seem helpful, especially as it'll mean everything will appear to hash the same. I think that should be an error.

Okay, I'll add a `Xapian::InvalidOperationError` to raise an error if the algorithm name is invalid or not recognised.

xapian-applications/omega/docs/omegascript.rst
ivmarkp: Just to be sure, should we drop the default behaviour then?

ojwb: > If we start talking about security, we probably shouldn't use md5 at all.

I think the only reason we are at all is in the interests of expediency: there's already code to calculate MD5 hashes in the tree so it's much simpler to hook that up.

> there's something a little weird about having an option with only one valid choice, which nonetheless has to be provided

True, but I'd expect we'll add more options in the near future.  I'm quite keen to find a sensible way to switch to something else for duplicate removal - if a system hides duplicates based on their hash, currently someone who can add documents to that system can use hash collisions to hide results.  And once we have another hash function available we can hook that up in omega for little extra work.

> Just to be sure, should we drop the default behaviour then?

I think we should.  The main reasons I see for having a single `$hash` are (a) to control the proliferation of omegascript functions a bit (b) it means that choice of hash can be easily parameterised (if instead we had separate functions for `$md5`, `$sha256`, `$blake2b`, etc then you'd have to use `$eval` to achieve that with the same generality, which is problematic as it is essentially executing the configuration file setting as code).

samuelharden:
@ivmarkp Please set the appropirate commit messages for your commit. 

Don't keep repeating "fixup! fixup! Add the implementation of hash command". Have a message which tells something about  the changes in commit

ivmarkp:
@samuelharden actually James [advised](https://github.com/xapian/xapian/pull/147#issuecomment-287614481) earlier to make changes in PR as fixup commits. So, been using that convention since then as it makes easier to combine fixup commits with their respective original commits when finally merging.

samuelharden:
Sorry for confusion then. probably it's best for merge commits. If James suggested this, keep following it.

@jaylett  I was wondering if just having fixup! with the relevant message is better than having exactly same commit message or it's better to have the exact same commit message.

xapian-applications/omega/docs/omegascript.rst
ivmarkp: Okay, will make the required changes for that.

xapian-applications/omega/query.cc
```diff
@@ -1567,6 +1568,10 @@ eval(const string &fmt, const vector<string> &param)
 			hexhash += buf;
 		    }
 		    value = hexhash;
+		    flag = true;
+		}
+		if (flag != true) {
+		    throw Xapian::InvalidOperationError("Hashing algorithm is invalid");
```
ivmarkp: @ojwb would there be a way to test this behaviour in omegatest?

ojwb:
> was wondering if just having fixup! with the relevant message is better than having exactly same commit message or it's better to have the exact same commit message.

That's just how this git feature works - the commit message has to be `!fixup ` (or `squash! `) followed by the start of the title of commit that it fixes up.  See the docs for `git rebase --autosquash`:

```
       --autosquash, --no-autosquash
           When the commit log message begins with "squash! ..." (or "fixup!
           ..."), and there is a commit whose title begins with the same ...,
           automatically modify the todo list of rebase -i so that the commit
           marked for squashing comes right after the commit to be modified,
           and change the action of the moved commit from pick to squash (or
           fixup). Ignores subsequent "fixup! " or "squash! " after the first,
           in case you referred to an earlier fixup/squash with git commit
           --fixup/--squash.
```

Generally you just use `git commit --fixup` rather than writing such commit messages by hand.

You're right that the resulting commit messages aren't ideal for review in github, but I think the best way to address that is to strongly steer the students towards smaller chunks of work per PR.  That's so much easier to deal with overall anyway.

xapian-applications/omega/query.cc
ojwb: Don't compare a boolean value against `true` or `false` explicitly like this.  Use `if (flag)` or `if (!flag)` instead.

Though here `flag` seems unnecessary - the `throw` can just be an `else` on the `if` which checks for `md5`, which would be simpler and a bit clearer.

`Xapian::InvalidOperationError` is a xapian-core exception - here you want to throw a `std::string` with the message, and then in omegatest you can check the output to be `Exception: ` followed by that string.

It would be useful to mention the value that wasn't handled; I'd also suggest wording it like similar existing errors, such as:

```
            throw "Unknown function '" + var + "'";
```

jaylett:
Remaining issues to address I believe are:

 * [ ] documentation should make it clear that hash names are lowercased automatically, and that they are no longer defaulted
 * [ ] remove `flag` variable
 * [ ] throw a suitable string error (and test it in `omegatest`)

ojwb:
> documentation should make it clear that hash names are lowercased automatically, and that they are no longer defaulted

I believe I said this already, but I think the hash names should be required to be lower case for consistency - elsewhere omegascript handles comparable things in a case sensitive way, and only accepts lower case (for example, query parser flags, weighting scheme names, omegascript command names).  This sort of consistency makes things easier to learn and to remember - if some settings are case sensitive and some are case insensitive, you either have to also learn which are which, or just ignore the difference and pretend they're all case sensitive, in which case the fact some aren't is just a pointless distraction (e.g. if I'm an omega user and get a patch to the templates which uses `$hash{$value,MD5}` I'd have to do extra work to confirm that actually worked).

xapian-applications/omega/omegatest
```diff
@@ -432,6 +432,15 @@ printf '$last$if{$cgi{ERR},$seterror{$cgi{ERR}}}$msize$if{$error,!$error!}' > "$
 testcase '11' P=text
 testcase '11!boo!' P=text ERR=boo
 
+# Feature tests for $hash{}
+printf '$hash{$cgi{text},md5}' > "$TEST_TEMPLATE"
+testcase '098f6bcd4621d373cade4e832627b4f6' text="test"
+testcase 'b4c216e4da73d1d01277ef46d0514821' text="simple query"
+testcase 'd41d8cd98f00b204e9800998ecf8427e' text=""
```
ojwb: I think it'd be useful to keep the testcase for `" test "` or whatever it was that was here before, to confirm that whitespace normalisation is not happening (i.e. that it gives a different output to `"test"`).

xapian-applications/omega/omegatest
```diff
@@ -432,6 +432,15 @@ printf '$last$if{$cgi{ERR},$seterror{$cgi{ERR}}}$msize$if{$error,!$error!}' > "$
...
+
+printf '$hash{$cgi{text},sha-1}' > "$TEST_TEMPLATE"
+testcase '' text="test"
```
ojwb: For testing the "not a valid hash name" case, `sha-1` is a bit of a confusing choice as it looks like the name of a valid hash.  Better to pick something that's clearly not a valid hash to the reader, e.g. like `bogus` or `notavalidhash`.

ivmarkp:
Sorry for the late response. My system remained cutoff from the internet yesterday the whole time as the router malfunctioned. I'll be getting a replacement today hopefully but for now working off of mobile network.

@jaylett Thanks for compiling the remaining issues. I hope I have addressed all of them now. Also, I can't seem to access to those checkboxes so can't set them as checked.

jaylett:
@ivmarkp you can't update the checklist, but that's kind of the point — it allows us to keep track of things that are addressed to our satisfaction. (I _think_ that anyone with write access to the repo can update them, but it's possible only I can because I created it.)

This is almost there, but I think the documentation should be explicit that it uses lowercase (from the current docs I'd expect it to use uppercase), and it _must_ show that `HASH` isn't optional.

ivmarkp:
> you can't update the checklist, but that's kind of the point — it allows us to keep track of things that are addressed to our satisfaction. (I think that anyone with write access to the repo can update them, but it's possible only I can because I created it.)

Oh, I see.

> This is almost there, but I think the documentation should be explicit that it uses lowercase (from the current docs I'd expect it to use uppercase), and it must show that HASH isn't optional.

Okay, I'll add that bit and remove the square brackets around `HASH`.

ivmarkp:
@jaylett On a second thought, is it necessary that we make it explicit that `HASH` must be lowercase because we transform it to lowercase afterwards in `query.cc`

ojwb:
> we transform it to lowercase afterwards in `query.cc`

Part of my point is we should stop doing that too.

ivmarkp:
Sorry, I think I misunderstood you earlier.

xapian-applications/omega/omegatest
ojwb: I still think this would be useful...

xapian-applications/omega/query.cc
```diff
@@ -1551,6 +1554,24 @@ eval(const string &fmt, const vector<string> &param)
...
+		    string md5, hexhash;
+		    md5_string(data.c_str(), md5);
+		    for (size_t i = 0; i < md5.size(); ++i) {
+			char buf[16];
+			sprintf(buf, "%02x", static_cast<unsigned char>(md5[i]));
+			hexhash += buf;
+		    }
+		    value = hexhash;
+		}
+		else {
```
ojwb: Xapian coding convention is to "cuddle" braces, so `} else {` here.

ivmarkp: FIxed it.

xapian-applications/omega/omegatest
ivmarkp: I had already added a testcase for `" test "` in one of the recent commits. Please see line 440 below.

ojwb: Oh sorry, i guess I failed to look below the comments...

ojwb:
Thanks for the final fixes.

`git rebase -i --autosquash` fails - I think the "Remove an unwanted space" commit in the middle of a series of fixup commits is the problem, but the split into multiple commits here doesn't seem useful as is essentially one change (adding a new omegascript function) which isn't useful divisible.  So I flattened it into one commit and rebased that onto master (75745f19134d03daff184f70b7038cb3c031954a).

PR #162, opened 2017-06-24
Issues with removal of prefix when iterator incremented after deletion of word

samuelharden:
I have added two test both doing similar things. With only one difference, in one case iterator is incremented before deletion of the term in the document. In another case, it is deleted after deletion of the term from the document.

Test with incrementing after the term is failing with **Invalid read of size 8** in this case and while I was using in one of application it was giving weird results, moving iterator to someother location or starting in some case when incremented after the deletion.

Please verify if this is not expected behavior which xapian should have. I will try to solve the issue with your help and can use these test for TDD development.

### DIFF
--- a/xapian-core/tests/api_wrdb.cc
+++ b/xapian-core/tests/api_wrdb.cc
@@ -1949,3 +1949,58 @@ DEFINE_TESTCASE(modifyvalues1, writable) {
 
     return true;
 }
+
+//Test removal of specific prefix terms from the document, incrementing the iterator before deleting the term from document.
+DEFINE_TESTCASE(testprefixedtermsremoval_iteratebeforedeletion, writable) {
+    Xapian::WritableDatabase db = get_writable_database();
+    Xapian::TermGenerator indexer;
+    Xapian::Document doc;
+    indexer.set_database(db);
+    indexer.set_document(doc);
+    indexer.index_text("But other predicted events did occur, such as a sharp fall in the pound and rising inflation.", 1, "NA");
+    db.commit();
+    Xapian::TermIterator term_iterator_end = doc.termlist_end();
+    Xapian::TermIterator term_iterator = doc.termlist_begin();
+    term_iterator.skip_to("NA");
+    while (term_iterator != term_iterator_end) {
+	const string& term = *term_iterator;
+	if (startswith(term, "NA")) {
+	    ++term_iterator;
+	    doc.remove_term(term);
+	} else {
+	    break;
+	}
+    }
+    Xapian::TermIterator tlist = db.allterms_begin("NA");
+    Xapian::TermIterator tlist_end = db.allterms_end("NA");
+    TEST(tlist == tlist_end);
+    return true;
+}
+
+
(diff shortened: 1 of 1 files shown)


samuelharden:
kushal@kushal-Lenovo-Yoga-2-Pro:~/gaurav/xapian-gaurav-gsoc/xapian-core/tests$ ./runtest ./apitest testprefixedtermsremoval_iteratebeforedeletion
Running test './apitest testprefixedtermsremoval_iteratebeforedeletion' under valgrind
Running tests with backend "inmemory"...
./apitest backend inmemory: All 1 tests passed.                                
Running tests with backend "singlefile_glass"...
Running tests with backend "multi_glass"...
Running tests with backend "remoteprog_glass"...
./apitest backend remoteprog_glass: All 1 tests passed.                        
Running tests with backend "remotetcp_glass"...
./apitest backend remotetcp_glass: All 1 tests passed.                         
./apitest total: All 4 tests passed.
kushal@kushal-Lenovo-Yoga-2-Pro:~/gaurav/xapian-gaurav-gsoc/xapian-core/tests$ ./runtest ./apitest testprefixedtermsremoval_iterateafterdeletion
Running test './apitest testprefixedtermsremoval_iterateafterdeletion' under valgrind
Running tests with backend "inmemory"...
Running test: testprefixedtermsremoval_iterateafterdeletion... Invalid read of size 8
./apitest backend inmemory: 0 tests passed, 1 failed.
Running test: testprefixedtermsremoval_iterateafterdeletion... VALGRIND DETECTED A PROBLEM
./apitest backend glass: 0 tests passed, 1 failed.
Running tests with backend "singlefile_glass"...
Running tests with backend "multi_glass"...
Running tests with backend "remoteprog_glass"...
Running test: testprefixedtermsremoval_iterateafterdeletion... VALGRIND DETECTED A PROBLEM
./apitest backend remoteprog_glass: 0 tests passed, 1 failed.
Running tests with backend "remotetcp_glass"...
Running test: testprefixedtermsremoval_iterateafterdeletion... VALGRIND DETECTED A PROBLEM
./apitest backend remotetcp_glass: 0 tests passed, 1 failed.
./apitest total: 0 tests passed, 4 failed.
kushal@kushal-Lenovo-Yoga-2-Pro:~/gaurav/xapian-gaurav-gsoc/xapian-core/tests$ git branch

ojwb:
Thanks for the reproducer - the cause is that `std::map::erase()` invalidates iterators pointing to erased elements - see http://en.cppreference.com/w/cpp/container/map/erase

I'm not sure how best to solve this.

The simplest fix is probably to not use `erase()` but instead flag removed entries and then skip them when iterating.

Other that that, we could use a different data structure which doesn't invalidating iterators like this.  Not quite sure what 's suitable for that.

Or we could track all active iterators over a particular termlist and then when we erase, adjust any which point to erased entries.  This sounds overly complex, though typically it'll be rare for there to actually be more than one such iterator, so in practice it might not be too crazy.

ojwb:
Fixed in git master by 9650fc6f46fca47fbd78bc9080a88d42b558c438.  Will backport for 1.4.6.

PR #163, opened 2017-07-04
Update Letor harness to match Core harness

AyushP123:
This will allow the tests written in letor to run against all the backends. Although the changes in xapian-letor/configure.ac and xapian-letor/build/tests/Makefile.mk are not permanent, for the time being this will help.

### DIFF
--- a/xapian-letor/configure.ac
+++ b/xapian-letor/configure.ac
@@ -459,12 +459,10 @@ if true ; then
     *-yes )
       dnl For mingw and msvc we have an alternative implementation which
       dnl doesn't need fork() or socketpair().
-      dnl
       ;;
     *djgpp* | *msdos* )
       dnl DJGPP has a dummy implementation of fork which always fails.
       dnl
-      dnl For disk-based backend, use flock() for locking, which doesn't need
       dnl fork() or socketpair().
       enable_backend_remote=no
       ;;
--- a/xapian-letor/configure.ac
+++ b/xapian-letor/configure.ac
@@ -463,7 +463,9 @@ if true ; then
     *djgpp* | *msdos* )
       dnl DJGPP has a dummy implementation of fork which always fails.
       dnl
-      dnl fork() or socketpair().
+      dnl If someone actually wanted remote backend support, then DJGPP has a
+      dnl pthreads port, so using threads like we do on Windows would make more
+      dnl sense.
       enable_backend_remote=no
       ;;
     *)
--- a/xapian-letor/configure.ac
+++ b/xapian-letor/configure.ac
@@ -460,26 +460,6 @@ if true ; then
       dnl For mingw and msvc we have an alternative implementation which
       dnl doesn't need fork() or socketpair().
       dnl
-      dnl We need -lws2_32 for getaddrinfo(), etc.
-      REMOTE_LIBS=-lws2_32
-      dnl Vista is needed for the AI_ADDRCONFIG flag to getaddrinfo().
-      AC_DEFINE([WINVER], [0x600],
-		[Version of Windows to assume (0x600 => Vista).])
-      AC_DEFINE([_WIN32_WINNT], [WINVER],
-		[Version of Windows to assume.])
-      ;;
-    *djgpp* | *msdos* )
-      dnl DJGPP has a dummy implementation of fork which always fails.
-      dnl
-      dnl For disk-based backend, use flock() for locking, which doesn't need
-      dnl fork() or socketpair().
-      AC_DEFINE([FLINTLOCK_USE_FLOCK], 1, [Define to use flock() for flint-compatible locking])
-      dnl If someone actually wanted remote backend support, then DJGPP has a
-      dnl pthreads port, so using threads like we do on Windows would make more
-      dnl sense.
-      enable_backend_remote=no
-      ;;
-    *)
       dnl On Unix, we need fork and socketpair for the remotebackend.
       SAVE_LIBS=$LIBS
       AC_CHECK_FUNCS([fork], [], [
@@ -492,28 +472,11 @@ if true ; then
       ])
       AC_DEFINE([HAVE_SOCKETPAIR], [1],
 		[Define to 1 if you have the 'socketpair' function.])
-      dnl Check if extra libraries are needed for getaddrinfo or inet_ntop()
-      dnl (e.g. on Solaris).
-      dnl
-      dnl We're currently assuming that any system that is worth trying to
-      dnl support has getaddrinfo() and inet_ntop(), since these are the
-      dnl standard route for supporting IPv6, and that's pretty much essential
-      dnl for platforms to support now.
-      AC_SEARCH_LIBS([getaddrinfo], [nsl socket], [], [
-	AC_MSG_ERROR([getaddrinfo() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
-      ])
-      AC_SEARCH_LIBS([inet_ntop], [nsl socket], [], [
-	AC_MSG_ERROR([inet_ntop() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
-      ])
       REMOTE_LIBS=$LIBS
       LIBS=$SAVE_LIBS
       ;;
   esac
 
-  if test "$enable_backend_remote" = yes ; then
-    TYPE_SOCKLEN_T
-    XAPIAN_LIBS="$XAPIAN_LIBS $REMOTE_LIBS"
-  fi
 fi
 
 dnl If eatmydata is installed, we run the testsuite under it to speed it up.
--- a/xapian-letor/configure.ac
+++ b/xapian-letor/configure.ac
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
   AC_DEFINE([HAVE_VALGRIND], [1], [Define if a suitable valgrind is installed])
 fi
 
+REMOTE_LIBS=
+if true ; then
+  AC_PREPROC_IFELSE([AC_LANG_SOURCE([[
+#ifdef __WIN32__
+#error WIN32
+#endif
+  ]])], [win32=no], [win32=yes])
+
+  case $host_os-$win32 in
+    *-yes )
+      dnl For mingw and msvc we have an alternative implementation which
+      dnl doesn't need fork() or socketpair().
+      dnl
+      dnl We need -lws2_32 for getaddrinfo(), etc.
+      REMOTE_LIBS=-lws2_32
+      dnl Vista is needed for the AI_ADDRCONFIG flag to getaddrinfo().
+      AC_DEFINE([WINVER], [0x600],
+		[Version of Windows to assume (0x600 => Vista).])
+      AC_DEFINE([_WIN32_WINNT], [WINVER],
+		[Version of Windows to assume.])
+      ;;
+    *djgpp* | *msdos* )
+      dnl DJGPP has a dummy implementation of fork which always fails.
+      dnl
+      dnl For disk-based backend, use flock() for locking, which doesn't need
+      dnl fork() or socketpair().
+      AC_DEFINE([FLINTLOCK_USE_FLOCK], 1, [Define to use flock() for flint-compatible locking])
+      dnl If someone actually wanted remote backend support, then DJGPP has a
+      dnl pthreads port, so using threads like we do on Windows would make more
+      dnl sense.
+      enable_backend_remote=no
+      ;;
+    *)
+      dnl On Unix, we need fork and socketpair for the remotebackend.
+      SAVE_LIBS=$LIBS
+      AC_CHECK_FUNCS([fork], [], [
+	AC_MSG_ERROR([fork() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
+      dnl Check if -lsocket is required for socketpair (Solaris needs it).
+      dnl And on Haiku it's in -lnetwork.
+      AC_SEARCH_LIBS([socketpair], [socket network], [], [
+	AC_MSG_ERROR([socketpair() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
+      AC_DEFINE([HAVE_SOCKETPAIR], [1],
+		[Define to 1 if you have the 'socketpair' function.])
+      dnl Check if extra libraries are needed for getaddrinfo or inet_ntop()
+      dnl (e.g. on Solaris).
+      dnl
+      dnl We're currently assuming that any system that is worth trying to
+      dnl support has getaddrinfo() and inet_ntop(), since these are the
+      dnl standard route for supporting IPv6, and that's pretty much essential
+      dnl for platforms to support now.
+      AC_SEARCH_LIBS([getaddrinfo], [nsl socket], [], [
+	AC_MSG_ERROR([getaddrinfo() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
+      AC_SEARCH_LIBS([inet_ntop], [nsl socket], [], [
+	AC_MSG_ERROR([inet_ntop() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
+      REMOTE_LIBS=$LIBS
+      LIBS=$SAVE_LIBS
+      ;;
+  esac
+
+  if test "$enable_backend_remote" = yes ; then
+    TYPE_SOCKLEN_T
+    XAPIAN_LIBS="$XAPIAN_LIBS $REMOTE_LIBS"
+  fi
+fi
+
 dnl If eatmydata is installed, we run the testsuite under it to speed it up.
 dnl If EATMYDATA is set to an empty value, then skip this check and don't use
 dnl eatmydata.
--- a/xapian-letor/configure.ac
+++ b/xapian-letor/configure.ac
@@ -466,10 +466,6 @@ if true ; then
       dnl
       dnl For disk-based backend, use flock() for locking, which doesn't need
       dnl fork() or socketpair().
-      AC_DEFINE([FLINTLOCK_USE_FLOCK], 1, [Define to use flock() for flint-compatible locking])
-      dnl If someone actually wanted remote backend support, then DJGPP has a
-      dnl pthreads port, so using threads like we do on Windows would make more
-      dnl sense.
       enable_backend_remote=no
       ;;
     *)
--- a/xapian-letor/configure.ac
+++ b/xapian-letor/configure.ac
@@ -460,6 +460,19 @@ if true ; then
       dnl For mingw and msvc we have an alternative implementation which
       dnl doesn't need fork() or socketpair().
       dnl
+      ;;
+    *djgpp* | *msdos* )
+      dnl DJGPP has a dummy implementation of fork which always fails.
+      dnl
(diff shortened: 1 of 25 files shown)


xapian-letor/configure.ac
```diff
@@ -385,6 +385,12 @@ AH_BOTTOM(
 ])
 
 AC_DEFINE([XAPIAN_LIB_BUILD], [1], [Workaround for avoiding sub-header included error])
+AC_DEFINE([BUILD_BACKEND_GLASS], [1], [Workaround for avoiding setting the conditionals])
+AC_DEFINE([BUILD_BACKEND_INMEMORY], [1], [Workaround for avoiding setting the conditionals])
+AC_DEFINE([BUILD_BACKEND_REMOTE], [1], [Workaround for avoiding setting the conditionals])
```
ojwb: Why are you defining these as preprocessor symbols?

xapian-letor/configure.ac
```diff
@@ -385,6 +385,12 @@ AH_BOTTOM(
...
+AM_CONDITIONAL([BUILD_BACKEND_GLASS], true)
+AM_CONDITIONAL([BUILD_BACKEND_INMEMORY], true)
+AM_CONDITIONAL([BUILD_BACKEND_REMOTE], true)
```
ojwb: This is bogus - you're assuming these are enabled in xapian-core and the build will fail when they aren't.

Why haven't you taken the approach we discussed on IRC of wrapping the code in `#ifdef XAPIAN_HAS_GLASS_BACKEND` etc?

(Also underquoted - should be `[true]`.)

ojwb:
Please restore the branch for this PR so we can reopen it.

xapian-letor/configure.ac
AyushP123: I did try wrapping the code with `#ifdef ` but the source files for backendmanagers under the if condition in Makefile wasn't getting compiled. Did you mean that I should add all the backendmanager source files to testharness_sources in xapian-letor/tests/harness/Makefile.mk and then wrap the code with `#idef`?

xapian-letor/configure.ac
AyushP123: so that all the conditionals in the xapian-letor/harness/tests gets set to true

xapian-letor/configure.ac
AyushP123: Also @jaylett and you had a discussion on IRC .... through that discussion I  entirely certain which approach should be used.

xapian-letor/configure.ac
jaylett: These aren't preprocessor symbols; they're only used as conditionals inside the autoconf system. I think I perhaps led you in slightly the wrong direction here, however — apologies. I think you want `AM_CONDITIONAL` instead, but @ojwb is better on the autotools system than I am. (I do recommend reading the introduction to the autoconf and automake manuals so you understand how they fit together with the rest of the build system.)

jaylett: (And Olly has suggested a different approach which I don't think needs these at all, as it happens.)

xapian-letor/configure.ac
jaylett: The discussion was intended to give you some ideas to go and investigate; we are unlikely to be able to simply solve problems for you, nor should we.

You'll need to use `XAPIAN_HAS_*` to wrap code, _and_ ensure that `Makefile.mk` is suitably adjusted (which probably involves removing the conditions) so that everything is built. You've clearly considered this, so: what happened when you tried it?

AyushP123: I made the mistake of not changing the Makefile.mk in letor/harness while implementing `#ifdef` ..... I did rectify the mistake and it now works perfectly.

ojwb:
You've force-pushed a branch mid-review **again**.

Please **stop** doing this.  It makes review unnecessarily harder, as we pretty much have to start again looking at the whole diff against master rather than just checking the incremental changes in each subsequent commit.

You should never force push a branch which other people working on with you unless they've all agreed.  People sometimes state "this branch may be force pushed at any time", which doesn't seem an entirely healthy attitude, but at least then people have been warned.

Force-pushing pull request branches without consent from the people you're requesting merge your changes is a great way to annoy those people, which works against getting your changes merged.  

A big part of GSoC is teaching you how to productively contribute to FOSS projects in general.  There certainly are differences in how different project operate, but there's a lot in common, especially once you look past the details.

Developers on FOSS projects are usually busy people and often doing it in their spare time, so a key part of getting your changes accepted is making it quick and painless for them to review your patch, and making sure there's no reason for them to say "no", or to direct their efforts elsewhere.

Particularly outside of GSoC, things like not following the coding conventions used by the code-base, failing to address feedback fully, making the same mistakes over and over despite having them pointed out, force pushing branches mid-review, etc are all things that will tend to lead to your patch being left to languish.  If someone reviews 10 patches and yours takes as much time to deal with as all the others put together, they'll naturally be keener to see more patches from the people who sent the other 9 and less from you, and reviewing another patch will seem more appealing than trying to get yours into shape if you don't seem to really listen to feedback.

ojwb:
Here's the diff between `tests/harness` in the two subdirectories, from the command ```diff --exclude='*[op]' -ru xapian-core/tests/harness/ xapian-letor/tests/harness``` (the `--exclude` just ignores some generated files):

```
diff '--exclude=*[op]' -ru xapian-core/tests/harness/backendmanager_glass.cc xapian-letor/tests/harness/backendmanager_glass.cc
--- xapian-core/tests/harness/backendmanager_glass.cc	2017-06-19 16:27:55.958621273 +1200
+++ xapian-letor/tests/harness/backendmanager_glass.cc	2017-07-11 09:01:18.584215196 +1200
@@ -22,6 +22,8 @@
 
 #include "backendmanager_glass.h"
 
+#ifdef XAPIAN_HAS_GLASS_BACKEND
+
 using namespace std;
 
 std::string
@@ -63,3 +65,4 @@
     return Xapian::WritableDatabase(".glass/" + last_wdb_name,
 				    Xapian::DB_OPEN|Xapian::DB_BACKEND_GLASS);
 }
+#endif
```

It'd be more consistent to have a blank line before `#endif` here, and for a long `#if` like this we usually annotate the `#endif` with a comment, e.g. see `backendmanager_glass.h` which ends like so:

```
};

#endif // XAPIAN_INCLUDED_BACKENDMANAGER_GLASS_H
```

Those same comments apply in several places below too:

```
diff '--exclude=*[op]' -ru xapian-core/tests/harness/backendmanager_inmemory.cc xapian-letor/tests/harness/backendmanager_inmemory.cc
--- xapian-core/tests/harness/backendmanager_inmemory.cc	2017-06-19 10:49:54.320976477 +1200
+++ xapian-letor/tests/harness/backendmanager_inmemory.cc	2017-07-11 09:01:18.584215196 +1200
@@ -22,6 +22,8 @@
 
 #include "backendmanager_inmemory.h"
 
+#ifdef XAPIAN_HAS_INMEMORY_BACKEND
+
 using namespace std;
 
 std::string
@@ -42,3 +44,4 @@
 {
     return getwritedb_inmemory(vector<string>(1, file));
 }
+#endif
diff '--exclude=*[op]' -ru xapian-core/tests/harness/backendmanager_remote.cc xapian-letor/tests/harness/backendmanager_remote.cc
--- xapian-core/tests/harness/backendmanager_remote.cc	2017-07-09 10:48:26.807252734 +1200
+++ xapian-letor/tests/harness/backendmanager_remote.cc	2017-07-11 09:01:18.584215196 +1200
@@ -25,6 +25,8 @@
 #include <string>
 #include "str.h"
 
+#ifdef XAPIAN_HAS_REMOTE_BACKEND
+
 BackendManagerRemote::BackendManagerRemote(const std::string & remote_type_)
 	: remote_type(remote_type_)
 {
@@ -98,3 +100,4 @@
 
     return args;
 }
+#endif
diff '--exclude=*[op]' -ru xapian-core/tests/harness/backendmanager_remoteprog.cc xapian-letor/tests/harness/backendmanager_remoteprog.cc
--- xapian-core/tests/harness/backendmanager_remoteprog.cc	2017-06-19 16:27:55.958621273 +1200
+++ xapian-letor/tests/harness/backendmanager_remoteprog.cc	2017-07-11 09:01:18.584215196 +1200
@@ -29,6 +29,8 @@
 # include <valgrind/memcheck.h>
 #endif
 
+#ifdef XAPIAN_HAS_REMOTE_BACKEND
+
 using namespace std;
 
 std::string
@@ -102,3 +104,4 @@
 #endif
     return Xapian::Remote::open_writable(XAPIAN_PROGSRV, args);
 }
+#endif
diff '--exclude=*[op]' -ru xapian-core/tests/harness/backendmanager_remotetcp.cc xapian-letor/tests/harness/backendmanager_remotetcp.cc
--- xapian-core/tests/harness/backendmanager_remotetcp.cc	2017-06-27 09:05:02.685902489 +1200
+++ xapian-letor/tests/harness/backendmanager_remotetcp.cc	2017-07-11 09:01:18.588215194 +1200
@@ -58,6 +58,8 @@
 # include <valgrind/memcheck.h>
 #endif
 
+#ifdef XAPIAN_HAS_REMOTE_BACKEND
+
 using namespace std;
 
 // We've had problems on some hosts which run tinderbox tests with "localhost"
@@ -405,3 +407,4 @@
     }
 #endif
 }
+#endif
diff '--exclude=*[op]' -ru xapian-core/tests/harness/backendmanager_singlefile.cc xapian-letor/tests/harness/backendmanager_singlefile.cc
--- xapian-core/tests/harness/backendmanager_singlefile.cc	2017-06-19 16:27:55.958621273 +1200
+++ xapian-letor/tests/harness/backendmanager_singlefile.cc	2017-07-11 09:01:18.588215194 +1200
@@ -31,6 +31,8 @@
 #include <cstring>
 #include "safeerrno.h"
 
+#ifdef XAPIAN_HAS_GLASS_BACKEND
+
 using namespace std;
 
 BackendManagerSingleFile::BackendManagerSingleFile(const std::string & subtype_)
@@ -103,3 +105,4 @@
 {
     throw Xapian::UnimplementedError("Single-file databases don't support writing");
 }
+#endif
Only in xapian-core/tests/harness/: Makefile
diff '--exclude=*[op]' -ru xapian-core/tests/harness/Makefile.mk xapian-letor/tests/harness/Makefile.mk
--- xapian-core/tests/harness/Makefile.mk	2017-07-09 10:48:26.807252734 +1200
+++ xapian-letor/tests/harness/Makefile.mk	2017-07-11 09:01:18.584215196 +1200
@@ -41,19 +41,13 @@
 	harness/fdtracker.cc\
 	harness/utestsuite.cc
 
-if BUILD_BACKEND_GLASS
 testharness_sources +=\
 	harness/backendmanager_glass.cc\
 	harness/backendmanager_singlefile.cc
-endif
 
-if BUILD_BACKEND_INMEMORY
 testharness_sources += harness/backendmanager_inmemory.cc
-endif
 
-if BUILD_BACKEND_REMOTE
 testharness_sources +=\
 	harness/backendmanager_remote.cc\
 	harness/backendmanager_remoteprog.cc\
 	harness/backendmanager_remotetcp.cc
-endif
diff '--exclude=*[op]' -ru xapian-core/tests/harness/testrunner.cc xapian-letor/tests/harness/testrunner.cc
--- xapian-core/tests/harness/testrunner.cc	2017-07-09 10:48:26.807252734 +1200
+++ xapian-letor/tests/harness/testrunner.cc	2017-07-11 09:01:18.588215194 +1200
@@ -32,7 +32,6 @@
 #include "backendmanager_remoteprog.h"
 #include "backendmanager_remotetcp.h"
 #include "backendmanager_singlefile.h"
-
 #include "stringutils.h"
 #include <iostream>
 
```

You've removed a blank line here for no good reason (and in a change which claims to update one version to match the other).

xapian-letor/configure.ac
```diff
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
...
+#endif
+  ]])], [win32=no], [win32=yes])
+
+  case $host_os-$win32 in
+    *-yes )
+      dnl For mingw and msvc we have an alternative implementation which
+      dnl doesn't need fork() or socketpair().
+      dnl
+      dnl We need -lws2_32 for getaddrinfo(), etc.
+      REMOTE_LIBS=-lws2_32
```
ojwb: letor doesn't use `getaddrinfo()` or anything else winsock-related I can think of, so I don't think we need this here.

The only relevant part is that we special case to not check for `fork()` or `socketpair()`.

xapian-letor/configure.ac
```diff
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
...
+      dnl Vista is needed for the AI_ADDRCONFIG flag to getaddrinfo().
+      AC_DEFINE([WINVER], [0x600],
+		[Version of Windows to assume (0x600 => Vista).])
+      AC_DEFINE([_WIN32_WINNT], [WINVER],
+		[Version of Windows to assume.])
```
ojwb: Again, shouldn't be needed for letor.

xapian-letor/configure.ac
```diff
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
...
+      ;;
+    *djgpp* | *msdos* )
+      dnl DJGPP has a dummy implementation of fork which always fails.
+      dnl
+      dnl For disk-based backend, use flock() for locking, which doesn't need
+      dnl fork() or socketpair().
+      AC_DEFINE([FLINTLOCK_USE_FLOCK], 1, [Define to use flock() for flint-compatible locking])
+      dnl If someone actually wanted remote backend support, then DJGPP has a
+      dnl pthreads port, so using threads like we do on Windows would make more
+      dnl sense.
```
ojwb: This stuff about `flock()` and locking isn't relevant here.  The relevant part is that we special case because of DJGPP's dummy implementation of `fork()`.

The `AC_CHECK_FUNCS` test can't spot this because it only checks that you can compile and link a program using `fork()`, which you can under DJGPP - the issue is that `fork()` always returns an error at runtime: http://www.delorie.com/djgpp/doc/libc/libc_338.html

xapian-letor/configure.ac
```diff
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
...
+      dnl We're currently assuming that any system that is worth trying to
+      dnl support has getaddrinfo() and inet_ntop(), since these are the
+      dnl standard route for supporting IPv6, and that's pretty much essential
+      dnl for platforms to support now.
+      AC_SEARCH_LIBS([getaddrinfo], [nsl socket], [], [
+	AC_MSG_ERROR([getaddrinfo() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
+      AC_SEARCH_LIBS([inet_ntop], [nsl socket], [], [
+	AC_MSG_ERROR([inet_ntop() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
```
ojwb: Letor doesn't use `getaddrinfo()` or `inet_ntop()`.

xapian-letor/configure.ac
```diff
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
...
+      REMOTE_LIBS=$LIBS
+      LIBS=$SAVE_LIBS
+      ;;
+  esac
+
+  if test "$enable_backend_remote" = yes ; then
```
ojwb: `enable_backend_remote` is never set to `yes` anywhere - only to `no` under some conditions.  So it'll either be `no` or empty at this point and this condition will always be false.

`enable_backend_remote` is also rather confusingly named inside letor, as it isn't tracking if the remote backend is enabled.  I would just scrap it for letor - there's no need to set it to `no` in the DJGPP case, as `REMOTE_LIBS` will be left empty, and this check can just be `if test -n "$REMOTE_LIBS"`.

xapian-letor/configure.ac
```diff
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
...
+    TYPE_SOCKLEN_T
```
ojwb: Letor doesn't use `socklen_t` so there's no point probing for it.

Also you didn't copy over the file where the macro `TYPE_SOCKLEN_T` is defined, so `TYPE_SOCKLEN_T` isn't expanded when `configure` is generated from `configure.ac` and this test won't work properly anyway.  You can see this if you look at the generated `configure` scripts:

```
olly@gemse:~/git/xapian$ grep -C2 TYPE_SOCKLEN_T xapian-{core,letor}/configure
xapian-letor/configure-
xapian-letor/configure-  if test "$enable_backend_remote" = yes ; then
xapian-letor/configure:    TYPE_SOCKLEN_T
xapian-letor/configure-    XAPIAN_LIBS="$XAPIAN_LIBS $REMOTE_LIBS"
xapian-letor/configure-  fi
```

In `xapian-letor/configure` it's unexpanded, and would be an error except that's masked because the conditional check is always false.  In `xapian-core/configure` we don't find it, because it was correctly expanded.

xapian-letor/configure.ac
```diff
@@ -447,6 +447,75 @@ if test -n "$VALGRIND" ; then
...
+      AC_CHECK_FUNCS([fork], [], [
+	AC_MSG_ERROR([fork() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
+      dnl Check if -lsocket is required for socketpair (Solaris needs it).
+      dnl And on Haiku it's in -lnetwork.
+      AC_SEARCH_LIBS([socketpair], [socket network], [], [
+	AC_MSG_ERROR([socketpair() required for the remote backend - if an extra library is needed, pass LIBS=-lfoo to configure.  Or --disable-backend-remote to disable it.)])
+      ])
+      AC_DEFINE([HAVE_SOCKETPAIR], [1],
+		[Define to 1 if you have the 'socketpair' function.])
```
ojwb: The test harness uses `fork()` and `socketpair()`, so this the above is the part we want.

AyushP123:
@ojwb I sincerely apologize for the mess up regarding force pushing the entire branch. I squashed the fixups into one commit to make the commits look more cleaner. I will never rebase again till I have your and @jaylett 's permission to do so.

xapian-letor/configure.ac
AyushP123: I am not sure what needs to be eliminated here the comments below the AC_DEFINE or above it. Since the stuff about flock() and locking is mentioned above AC_DEFINE.

AyushP123: I got confused here and have made multiple commits to resolve this. Hopefully the latest commit does what you expect me to do.

xapian-letor/configure.ac
```diff
@@ -447,6 +447,47 @@ if test -n "$VALGRIND" ; then
...
+#error WIN32
+#endif
+  ]])], [win32=no], [win32=yes])
+
+  case $host_os-$win32 in
+    *-yes )
+      dnl For mingw and msvc we have an alternative implementation which
+      dnl doesn't need fork() or socketpair().
+      dnl
+      ;;
```
ojwb: You've left an empty comment line (`dnl` is essentially a comment in m4).

xapian-letor/configure.ac
```diff
@@ -447,6 +447,47 @@ if test -n "$VALGRIND" ; then
...
+    *djgpp* | *msdos* )
+      dnl DJGPP has a dummy implementation of fork which always fails.
+      dnl
+      dnl For disk-based backend, use flock() for locking, which doesn't need
+      dnl fork() or socketpair().
```
ojwb: The part about `flock()` is irrelevant for letor.

xapian-letor/configure.ac
ojwb: If you remove the test for `flock()`, you should also remove the comment which talks about where and why we want `flock()`.  You can use `git grep -w flock` to quickly see that it's not used anywhere in letor.

As I said above, the relevant part is that we special case because of DJGPP's dummy implementation of `fork()`.

ojwb:
Autosquash and rebased as 89f98545dc383255433bae6557782fdeedcf33af.

PR #165, opened 2017-07-06
Ignore syntax errors in perl, Ignore POD errors entirely, as much as …

samuelharden:
Currently, When there is even a small syntax error in POD we exit the process and the file fails. We could probably ignore the errors coming from pod. As it will give us the text to index that file than killing this file.

Will it be useful to ignore errors ?

### DIFF
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -181,7 +181,7 @@ index_add_default_filters()
     // pod2text's output character set doesn't seem to be documented, but from
     // inspecting the source it looks like it's probably iso-8859-1.
     index_command("text/x-perl",
-		  Filter("pod2text --errors=none", "text/plain", "iso-8859-1", false));
+		  Filter("perl -MPod::Text -e '$p = Pod::Text->new(); $p->no_errata_section(1); $p->parse_file($ARGV[0])'", "text/plain", "iso-8859-1", false));
     // FIXME: -e0 means "UTF-8", but that results in "fi", "ff", "ffi", etc
     // appearing as single ligatures.  For European languages, it's actually
     // better to use -e2 (ISO-8859-1) and then convert, so let's do that for
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -181,7 +181,7 @@ index_add_default_filters()
     // pod2text's output character set doesn't seem to be documented, but from
     // inspecting the source it looks like it's probably iso-8859-1.
     index_command("text/x-perl",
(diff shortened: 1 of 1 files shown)


ojwb:
Generally this seems sensible, except that `--errors=none` doesn't seem to really do what you'd expect:

```
$ pod2text --errors=none Pod-Simple/t/junk1.pod
    pie is nice

    E<puppies>

POD ERRORS
    Hey! The above document had some coding errors, which are explained
    below:

    Around line 2:
        =cut found outside a pod block. Skipping to next block.

```

So it generates a "POD ERRORS" section (and that is on `stdout`) for at least some errors, and we don't want to index that.

It looks like using `--errors=stderr` and throwing stderr away works better:

```
$ pod2text --errors=stderr Pod-Simple/t/junk1.pod 2> /dev/null
    pie is nice

    E<puppies>

```

Actually throwing away stderr isn't great as it may hide actual errors (rather than just POD formatting ones), so probably just using that without the `2>/dev/null` makes more sense - you'll see messages about formatting errors in various other formats, and they do alert the user that there's an issue with the files being indexed.

The only way I found to suppress the errors without throwing away stderr is to use the `Pod::Text` module directly:

```
$ perl -MPod::Text -e '$p = Pod::Text->new(); $p->no_errata_section(1); $p->parse_file($ARGV[0])' Pod-Simple/t/junk1.pod
    pie is nice

    E<puppies>

```

The example I used above is from the Pod::Simple source:

```
$ cat Pod-Simple/t/junk1.pod

=cut

=head9 I like pie

B<pie is nice

E<puppies>

```

ojwb:
I didn't really mean to suggest that last command was the best answer.

What's the context of this?  Are you seeing malformed POD in the wild?

samuelharden:
In one of the implementations I was using the error on stderr to see the reason of failed file, so wanted to avoid getting formatting errors on Stderr. But here in the main library. I think it makes sense to just send all the errors on stderr --errors=stderr 

Shall I change the command to send errors on stderr for the master and update the PR ?

ojwb:
I've pushed a change to add `--errors=stderr` (7a3d1492eafdb30bc11d1e6cf73ed9c8055fbf88) for now.  Just noticed I failed to credit you - sorry about that, this report was appreciated.

I've also sent a patch to podlators to make `--errors=none` work in a less surprising way (https://github.com/rra/podlators/pull/7) so if that gets accepted we can change to use `--errors=none` in the longer term.

ojwb:
The podlators patch was already merged, though we'll need to wait a few years until we can assume it's installed everywhere.

PR #164, opened 2017-07-04
Update Letor harness to match Core harness

AyushP123:
Updating Letor Harness to sync it with Core Harness would allow the tests in letor to run for all backends.

PR #166, opened 2017-07-10
Add tests to letor

AyushP123:
Writing automated test suite for letor .... Currently it tests all the aspects of ranker and a letor/api.


xapian-letor/tests/Makefile.am
```diff
@@ -48,7 +49,7 @@ BUILT_SOURCES =
 
 if MAINTAINER_MODE
 BUILT_SOURCES += api_all.h api_collated.h $(collated_apitest_sources:.cc=.h) \
- api_generated.cc
+ api_generated.cc api_letor.cc
```
ojwb: `api_letor.cc` is not a generated file so shouldn't be in this list.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
+/** @file api_generated.cc
+ * @brief test common features of API classes
+ */
+/* Warning: This file is generated by ./generate-api_generated - do not modify directly! */
```
ojwb: Lots of incorrect information just cut-and-pasted above...

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include <config.h>
+
+#include "api_letor.h"
+
+#define XAPIAN_DEPRECATED(D) D
```
ojwb: This line is to disable deprecation warnings for features we want to test in this file.  There shouldn't be any such features in letor currently.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
...
+#include <xapian-letor.h>
+#include <fstream>
+
+#include "apitest.h"
+#include "testutils.h"
+
+using namespace std;
+
+DEFINE_TESTCASE(assign_db_path1, !backend) {
+   Xapian::Ranker *ranker = new Xapian::ListNETRanker();
```
ojwb: Why create this with `new`?  It'll mostly just adds a memory leak if the test fails without completing.

(I'm sure I pointed this out previously too...)

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
...
+			     "tigers, lions, and bears would all keep each "
+			     "other busy. On the other hand, bears don't live "
+			     "in the same continent as far as I know.");
+    db.add_document(doc);
+}
+
+DEFINE_TESTCASE(ranker1, generated) {
+    Xapian::Ranker *ranker = new Xapian::ListNETRanker();
+    string a = get_database_path("apitest_ranker1",
+				       gen_uniqterms_gt_doclen_db);
```
ojwb: Indentation.

Also, `a` is a very non-descriptive variable name.

Repeated below.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
...
+    Xapian::MSet mymset = enquire.get_mset(0, 10);
+    string query ="20001 'lions'";
+    string qrel = "20001 Q0 1 1\n20001 Q0 2 2";
+    std::ofstream myfile;
+    myfile.open("query.txt");
+    myfile << query;
+    myfile.close();
+    myfile.open("qrel.txt");
+    myfile << qrel;
+    myfile.close();
```
ojwb: You're using the exact same filenames for dynamically created files in multiple testcases, which is unnecessarily hostile to running testcases in parallel.

But the contents of these files are known in advance, so why not just add them to the source tree?  That'll be faster than creating them over and over, and reduce the source code size by more than the size of a single copy of the text in a file.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
...
+    Xapian::prepare_training_file(a, "query.txt", "qrel.txt", 10);
+    ranker->set_database_path(a);
+    ranker->set_query(Xapian::Query("lions"));
+    ranker->train_model("training-data.txt", "");
```
ojwb: You seem to be relying on `training-data.txt` being implicitly generated by the code above, yet name it explicitly here.

As I've said before (but was seemingly ignored) having a default name for this file is not helpful, and I think this illustrates some reasons why.  It's unclear to the reader where this file is coming from, and the default filename causes a collision if this testcase and the next are run in parallel - both testcases will try to write out the same file and then read it back, which will cause various erratic failures depending on the exact scheduling of the two tasks.

We should scrap the default and require the user to explicitly specify the filename to use.  And testcases should specify a filename such that there's no collision (e.g. by including the testcase name in the training data filename).

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
...
+}
+
+DEFINE_TESTCASE(ranker1, generated) {
+    Xapian::Ranker *ranker = new Xapian::ListNETRanker();
+    string a = get_database_path("apitest_ranker1",
+				       gen_uniqterms_gt_doclen_db);
+    Xapian::Enquire enquire((Xapian::Database(a)));
+    enquire.set_query(Xapian::Query("lions"));
+    Xapian::MSet mymset = enquire.get_mset(0, 10);
+    string query ="20001 'lions'";
```
ojwb: whitespace (and below)

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,187 @@
...
+    delete ranker;
+    ranker = new Xapian::SVMRanker();
+    TEST_EXCEPTION(Xapian::FileNotFoundError, ranker->train_model("", ""));
+    delete ranker;
+
+    return true;
+}
+
+static void
+gen_uniqterms_gt_doclen_db(Xapian::WritableDatabase& db, const string&)
```
ojwb: This function name seems to have been copied and pasted, and makes no sense in this context.  We aren't generating a database here where `unique_terms > document length`.

xapian-letor/tests/api_letor.h
```diff
@@ -0,0 +1,6 @@
+/* Warning: This file is generated by ./collate-test - do not modify directly! */
```
ojwb: Do not commit generated files.  Add them to `.gitignore` instead.

xapian-letor/tests/apitest.cc
```diff
@@ -59,6 +59,24 @@ get_database_path(const string &dbname)
...
+
+Xapian::Database
+get_database(const std::string &dbname,
+	     void (*gen)(Xapian::WritableDatabase&,
+			 const std::string &),
+	     const std::string &arg)
+{
+    return backendmanager->get_database(dbname, gen, arg);
+}
+
```
ojwb: Shouldn't this (and below) be in PR #164?

xapian-letor/tests/api_letor.cc
AyushP123: How should I resolve this?

AyushP123: copying or assignment is not allowed

AyushP123: like this you meant Xapian::SVMRanker ranker;??

samuelharden: you can have a new object via
Xapian::Ranker ranker = XapianListNETRanker(); 
https://stackoverflow.com/questions/1764831/c-object-without-new

xapian-letor/tests/api_letor.cc
samuelharden: There needs to be a space between  ="2

It's always a good idea to run script to check these error before commiting any changes.
you can run the script by :
`git diff | ../xapian-maintainer-tools/xapian-check-patch`

xapian-letor/tests/api_letor.cc
AyushP123: No I can't assignment isn't allowed

AyushP123: Ranker is an abstract class. Cannot declare a variable for it.

xapian-letor/tests/api_letor.cc
ojwb: That's generally good advice, though this is actually a case which `xapian-check-patch` doesn't detect currently (or didn't - I just pushed a fix for it).

But you also shouldn't expect the script to absolve you of having to try to follow the coding conventions as you code.  It's likely it's never going to spot everything, at least with the way it currently works.

xapian-letor/tests/api_letor.cc
ojwb: > did you mean this Xapian::SVMRanker ranker;

Yes, like that.

> `Xapian::Ranker ranker = Xapian::ListNETRanker();`

Even if `Ranker` wasn't abstract, this would cause object slicing:

https://en.wikipedia.org/wiki/Object_slicing

AyushP123:
@ojwb @jaylett @samuelharden .... I was rebasing the current master onto this branch and I have by mistake added some commits .... my apologies

xapian-letor/tests/api_letor.cc
ojwb: You've not addressed this.

xapian-letor/tests/api_letor.cc
ojwb: ...

ojwb:
I'm not clear why you were rebasing this in the first place - were you asked to?

You also seem to have not addressed most of the points made previously.

xapian-letor/tests/api_letor.cc
```diff
@@ -33,32 +31,28 @@
...
-   ranker->set_database_path("db_path");
-   TEST_EQUAL("db_path", ranker->get_database_path());
-   delete ranker;
+DEFINE_TESTCASE(check_db_path, !backend) {
+   Xapian::ListNETRanker ranker1;
+   ranker1.set_database_path("db_path");
+   TEST_EQUAL("db_path", ranker1.get_database_path());
+   Xapian::SVMRanker ranker2;
+   ranker2.set_database_path("db_path");
+   TEST_EQUAL("db_path", ranker2.get_database_path());
```
ojwb: `ranker1` and `ranker2` is clumsy.  You could just wrap each ranker's part of the test in a block, and call them `ranker`.

xapian-letor/tests/api_letor.cc
```diff
@@ -33,32 +31,28 @@
...
+    TEST_EXCEPTION(Xapian::FileNotFoundError, ranker1.train_model("", ""));
+    Xapian::SVMRanker ranker2;
+    TEST_EXCEPTION(Xapian::FileNotFoundError, ranker2.train_model("", ""));
 
     return true;
 }
 
 static void
-gen_uniqterms_gt_doclen_db(Xapian::WritableDatabase& db, const string&)
+populate_db(Xapian::WritableDatabase& db, const string&)
```
ojwb: While this no longer inaccurately describes the generated database, it now doesn't describe it at all.  When you need more generate functions, you'll end up calling them `populate_db2()` etc which really isn't helpful.

What are the particular feature(s) of this database?

xapian-letor/tests/api_letor.cc
```diff
@@ -104,26 +98,19 @@ gen_uniqterms_gt_doclen_db(Xapian::WritableDatabase& db, const string&)
     db.add_document(doc);
 }
 
-DEFINE_TESTCASE(ranker1, generated) {
+DEFINE_TESTCASE(listnet_ranker, generated) {
     Xapian::ListNETRanker ranker;
     string a = get_database_path("apitest_ranker1",
```
ojwb: You've not address the naming of `a`, which I explicitly raised.

xapian-letor/tests/api_letor.cc
```diff
@@ -104,26 +98,19 @@ gen_uniqterms_gt_doclen_db(Xapian::WritableDatabase& db, const string&)
...
-				       gen_uniqterms_gt_doclen_db);
+				       populate_db);
```
ojwb: You've not addressed the indentation here, which I explicitly raised.

xapian-letor/tests/api_letor.cc
```diff
@@ -104,26 +98,19 @@ gen_uniqterms_gt_doclen_db(Xapian::WritableDatabase& db, const string&)
...
-    myfile.open("query.txt");
-    myfile << query;
-    myfile.close();
-    myfile.open("qrel.txt");
-    myfile << qrel;
-    myfile.close();
-    Xapian::prepare_training_file(a, "query.txt", "qrel.txt", 10);
+    Xapian::prepare_training_file(a, "testdata/query.txt",
+				  "testdata/qrel.txt", 10,
+				  "training-data.txt");
```
ojwb: You still have multiple testcases generating `training-data.txt`, which I explicitly raised.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,165 @@
...
+
+DEFINE_TESTCASE(createfeaturevector, generated)
+{
+    Xapian::FeatureList fl;
+    Xapian::Database db = get_database("apitest_ranker1",
+				       populate_db);
+    Xapian::Enquire enquire(db);
+    enquire.set_query(Xapian::Query("lions"));
+    Xapian::MSet mset;
+    TEST(mset.empty());
```
ojwb: What are you trying to test here exactly?

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,165 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include <config.h>
+
+#include "api_letor.h"
+
+#include <xapian.h>
+#include <xapian-letor.h>
+#include <fstream>
```
ojwb: No longer used.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,165 @@
...
+				       populate_db);
+    Xapian::Enquire enquire((Xapian::Database(a)));
+    enquire.set_query(Xapian::Query("lions"));
+    Xapian::MSet mymset = enquire.get_mset(0, 10);
+    Xapian::prepare_training_file(a, "testdata/query.txt",
+				  "testdata/qrel.txt", 10,
+				  "training-data.txt");
+    ranker.set_database_path(a);
+    ranker.set_query(Xapian::Query("lions"));
+    ranker.train_model("training-data.txt", "");
```
ojwb: Why are you passing `""` here?

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,165 @@
...
+    Xapian::docid doc1 = *mymset[0];
+    Xapian::docid doc2 = *mymset[1];
+    ranker.rank(mymset, "");
```
ojwb: And here?

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,165 @@
...
+DEFINE_TESTCASE(check_db_path, !backend) {
+   Xapian::ListNETRanker ranker1;
+   ranker1.set_database_path("db_path");
+   TEST_EQUAL("db_path", ranker1.get_database_path());
+   Xapian::SVMRanker ranker2;
+   ranker2.set_database_path("db_path");
+   TEST_EQUAL("db_path", ranker2.get_database_path());
+
+   return true;
+}
```
ojwb: This function is only indented by 3 spaces.

xapian-letor/tests/api_letor.cc
```diff
@@ -104,26 +98,19 @@ gen_uniqterms_gt_doclen_db(Xapian::WritableDatabase& db, const string&)
...
-    std::ofstream myfile;
-    myfile.open("query.txt");
-    myfile << query;
-    myfile.close();
-    myfile.open("qrel.txt");
-    myfile << qrel;
-    myfile.close();
-    Xapian::prepare_training_file(a, "query.txt", "qrel.txt", 10);
+    Xapian::prepare_training_file(a, "testdata/query.txt",
+				  "testdata/qrel.txt", 10,
```
ojwb: You need to support srcdir != builddir (also known as a "VPATH build", but when that's the case, the current directory will be the builddir and `testdata` will be in the srcdir, so these two files won't be found with these relative paths.

You can get the appropriate directory with `test_driver::get_srcdir()` - look at the code which loads files from `testdata` in the `xapian-core/tests` subdirectory.

xapian-letor/tests/api_letor.cc
AyushP123: I am passing "" ... so that the ranker will use default model_key. Each Ranker has a default model_key.

xapian-letor/tests/api_letor.cc
AyushP123: I am passing "" ... so that the ranker will use default model_key. Each Ranker has a default model_key.

xapian-letor/tests/api_letor.cc
AyushP123: I am trying to pass an empty MSet to the create_feature_vector to show that it doesnt fail on an empty MSet.

xapian-letor/tests/api_letor.cc
AyushP123: so should I delete the files at the end of the tests ?? .... because Ranker does need the training-data.txt to train the model

xapian-letor/tests/api_letor.cc
AyushP123: This is a temporary database which has two documents and is used by the ranker for training and ranking

xapian-letor/tests/api_letor.cc
AyushP123: Yes I will do that

xapian-letor/tests/api_letor.cc
ojwb: I can read the API docs to see that - but why did you choose to pass `""` when you wrote this testcase?

There seem to really be three cases here - there's specifying a filename and a `model_key`, there's just specifying a filename, and there's specifying a filename and an empty string for `model_key`.

Of these three, I'd imagine that the first and second are the ones API users are most likely to use, with probably second being more common.  The main case I can see where the third case is useful is when the API user's code makes `model_key` configurable to the end-user.

You're only testing one of these three cases here, and it seems to me it is the case which is probably least important.

xapian-letor/tests/api_letor.cc
ojwb: I mean specifically the code just above, where you create an empty `MSet`, then test that it's empty:

```
Xapian::MSet mset;
TEST(mset.empty());
```

That seems a test which belongs in the xapian-core testsuite rather than here, which leads me to wonder if you meant to write something different?

xapian-letor/tests/api_letor.cc
jaylett: No, you should (as has been pointed out) be using test-specific names.

xapian-letor/tests/api_letor.cc
jaylett: You should get the default by not passing this parameter to `rank()` at all.

xapian-letor/tests/Makefile.am
AyushP123: Currently I have added it to collated_apitest_sources .... if that is wrong please let me know

AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: I am indexing two documents into the db and have modified the name .... please let me know there is a better name or if the name is inappropriate

xapian-letor/tests/api_letor.cc
AyushP123: changed the name from a to db_path .... I used the name a before because I copied it from "compactnorenumber1" test in xapian-core/tests/api_compact.cc

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: I have added "query.txt" and "qrel.txt" to source tree .... :+1:

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.h
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: I have updated the name .... although I couldn't come up with something good .... if you or @jaylett have any suggestions please let me know

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: I am not sure whether I have resolved this ..... what I do generally is that, if a line crosses 80 columns, I simply put enter after the nearest comma

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: I will resolve this by eliminating TEST(mset.empty()) ... it was redundant

xapian-letor/tests/api_letor.cc
AyushP123: I have improved upon this by writing the test for the two cases out of 3 and leaving the least unimportant one which is redundant.

xapian-letor/tests/api_letor.cc
AyushP123: :+1:

xapian-letor/tests/api_letor.cc
AyushP123: :+1: .... I have eliminated the entire function

xapian-letor/tests/api_letor.cc
AyushP123: I have added `std::string data_directory = test_driver::get_srcdir() + "/testdata/";` as a prefix for those files .... let me know if you meant anything else

xapian-letor/tests/Makefile.am
ojwb: It's useful mark the comments you think you've dealt with, but commenting with the emoji sends a notification email each time, so we get a bit bombarded.

Better to use "Add your reaction" (hold the mouse pointer over a comment and some icons appear at the top right - you want the "+" and happy face one.

xapian-letor/tests/.gitignore
```diff
@@ -22,3 +22,7 @@
 /perflog.xml
 /submitperftest
 /zlib-vg.so
+/listnet_ranker_training.txt
+/svm_ranker_training.txt
+/api_letor.h
+/scorer_output.txt
```
ojwb: It's perhaps not entirely obvious, but the existing list groups similar files together, and is sorted within each group - so it'd be better to add `/api_letor.h` to the end of the existing group of `/api_*` files above.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,185 @@
...
+DEFINE_TESTCASE(listnet_ranker, generated)
+{
+    Xapian::ListNETRanker ranker;
+    TEST_EXCEPTION(Xapian::FileNotFoundError, ranker.train_model(""));
+    string db_path = get_database_path("apitest_ranker1",
+				       db_index_two_documents);
+    Xapian::Enquire enquire((Xapian::Database(db_path)));
+    enquire.set_query(Xapian::Query("lions"));
+    Xapian::MSet mymset = enquire.get_mset(0, 10);
+    std::string data_directory = test_driver::get_srcdir() + "/testdata/";
```
ojwb: You don't actually need the `std::` here (because `using namespace std;` is in effect), and it isn't used on e.g. `db_path` above.

xapian-letor/tests/api_letor.cc
```diff
@@ -0,0 +1,185 @@
...
+				"qrel.txt", "ListNet_Ranker",
+				"scorer_output.txt", 10, ""));
+    TEST_EXCEPTION(Xapian::FileNotFoundError,
+		   ranker.score("", data_directory + "qrel.txt",
+				"ListNet_Ranker", "scorer_output.txt", 10));
+    TEST_EXCEPTION(Xapian::FileNotFoundError,
+		   ranker.score(data_directory + "qrel.txt", "",
+				"ListNet_Ranker", "scorer_output.txt", 10));
+    ranker.score(data_directory + "query.txt", data_directory + "qrel.txt",
+		 "ListNet_Ranker", "scorer_output.txt", 10);
```
ojwb: I'd suggest just creating the two filenames once, e.g.:

```
string data_directory = test_driver::get_srcdir() + "/testdata/";
string query_txt = data_directory + "query.txt";
string qrel_txt = data_directory + "qrel.txt";
```

And then you can just use `query_txt` and `qrel_txt` everywhere.  That'll make the code a bit more readable, and probably be more efficient (because the compiler probably isn't smart enough to hoist a common subexpression of type `std::string`).

xapian-letor/tests/api_letor.cc
ojwb: Naming stuff well is not always easy - as the saying goes:

> There are two hard problems in computer science: cache invalidation, naming things, and off-by-one errors

I don't have a better name, so happy to go with what we have now.

xapian-letor/tests/api_letor.cc
ojwb: Wrapping after the comma's good, but in a parameter list the wrapped line should be indented to line up just after the `(` above.  This looks good in the latest version.

xapian-letor/tests/api_letor.cc
ojwb: Cool.

xapian-letor/tests/api_letor.cc
ojwb: OK, that's good.

(The naming in compactnorenumber1 also isn't great, but there are 4 databases (`a`, `b`, `c`, and `d`) so the reader can hopefully see they're basically arbitrary labels.  Copying that naming scheme when you have just one database doesn't make much sense.)

ojwb:
Thanks, looks good to me now.  When the tests are run under valgrind, there's a leak which #171 fixes, so we should really merge that first (so as not to have `make check` broken on master).

As I noted on IRC, libsvm is spewing some output to the terminal - looks like calling `void svm_set_print_string_function(void (*print_func)(const char *));` with an empty function should solve that.  That's essentially cosmetic, so let's deal with that as a separate PR.

ojwb:
Looking good now - autosquashed and merged to master.  Thanks.

PR #167, opened 2017-07-10
Stopword removal

richhiey1996:
KMeans has been added with a method named set_stopper. This gives the user full control on the kind of stopwords they would want to use. The provided stopper is given to the TermListGroup which uses it to filter out stopword terms.

### DIFF
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -57,7 +57,7 @@ TermListGroup::TermListGroup()
     LOGCALL_CTOR(API, "TermListGroup", NO_ARGS);
 }
 
-TermListGroup::TermListGroup(const MSet &docs, const Stopper *stopper = NULL)
+TermListGroup::TermListGroup(const MSet &docs, const Stopper *stopper)
 {
     LOGCALL_CTOR(API, "TermListGroup", docs | stopper);
     for (MSetIterator it = docs.begin(); it != docs.end(); ++it)
@@ -80,7 +80,7 @@ DummyFreqSource::get_doccount() const
 }
 
 void
-TermListGroup::add_document(const Document &document, const Stopper *stopper = NULL)
+TermListGroup::add_document(const Document &document, const Stopper *stopper)
 {
     LOGCALL_VOID(API, "TermListGroup::add_document", document | stopper);
 
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -52,11 +52,6 @@ Clusterer::~Clusterer()
     LOGCALL_DTOR(API, "Clusterer");
 }
 
-TermListGroup::TermListGroup()
-{
-    LOGCALL_CTOR(API, "TermListGroup", NO_ARGS);
-}
-
 TermListGroup::TermListGroup(const MSet &docs, const Stopper *stopper)
 {
     LOGCALL_CTOR(API, "TermListGroup", docs | stopper);
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -52,11 +52,16 @@ Clusterer::~Clusterer()
     LOGCALL_DTOR(API, "Clusterer");
 }
 
-TermListGroup::TermListGroup(const MSet &docs)
+TermListGroup::TermListGroup()
 {
-    LOGCALL_CTOR(API, "TermListGroup", docs);
+    LOGCALL_CTOR(API, "TermListGroup", NO_ARGS);
+}
+
+TermListGroup::TermListGroup(const MSet &docs, const Stopper *stopper = NULL)
+{
+    LOGCALL_CTOR(API, "TermListGroup", docs | stopper);
     for (MSetIterator it = docs.begin(); it != docs.end(); ++it)
-	add_document(it.get_document());
+	add_document(it.get_document(), stopper);
     num_of_documents = docs.size();
 }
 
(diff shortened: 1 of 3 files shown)


xapian-core/include/xapian/cluster.h
```diff
@@ -570,6 +592,9 @@ class XAPIAN_VISIBILITY_DEFAULT KMeans : public Clusterer {
     /// Specifies the maximum number of iterations that KMeans will have
     unsigned int max_iters;
 
+     /// Used to calculate term frequencies of the required terms
+    TermListGroup tlg;
```
ojwb: The comment is overindented.

xapian-core/cluster/cluster.cc
```diff
@@ -81,11 +102,19 @@ TermListGroup::add_document(const Document &document)
 
     TermIterator titer(document.termlist_begin());
 
+    bool allow_stopword_removal = stopper.get() ? true : false;
     for (; titer != document.termlist_end(); ++titer) {
+	const string &term = *titer;
+
+	// Remove stopwords by using the Xapian::Stopper object
+	// that is set
+	if (allow_stopword_removal && (*stopper)(term))
```
ojwb: I'd scrap the flag and just test `if (stopper.get() && (*stopper)(term))`.  It's clearer to the reader what's happening, and that the potential `NULL` dereference is guarded against, and it should be efficient - `opt_intrusive_ptr`'s `get()` method is inlined and just returns the `px` while `operator*` is inlined and returns `*px`, so this boils down to `if (px && (*px)(term))`.

xapian-core/cluster/kmeans.cc
```diff
@@ -72,7 +79,7 @@ void
 KMeans::initialise_points(const MSet &source)
 {
     LOGCALL_VOID(API, "KMeans::initialise_points", source);
-    TermListGroup tlg(source);
+    tlg.construct_tlg(source);
```
ojwb: This prolongs the lifetime of the `TermListGroup` and the (potentially large) `unordered_map` it contains.

I'd be inclined to keep `tlg` as a local here, and instead make `stopper` a member of `KMeans` which gets passed as an optional parameter to the `TermListGroup` constructor.

That also avoids having to worry about the semantics of calling `construct_tlg()` more than once on the same object.

richhiey1996: So the reason I did it like this is because of storing the stopper as a pointer rather than an object. I wasn't able to pass the stopper object around because it is an `opt_intrusive_ptr` and passing it as a pointer of Stopper class raised an error. So it made sense to make it a member of TermListGroup because it would only be used there.

> This prolongs the lifetime of the TermListGroup and the (potentially large) unordered_map it contains.
It does make sense to thus add it in KMeans itself due to the map being potentially huge. But in that case, I'm confused about whether I should be storing the Stopper as an object or as a pointer. I saw the previous implementations in the queryparser where it has been stored as a pointer and followed the same. Is there a specific reason it was stored that way? Or would it not harm to have the Stopper object itself?

xapian-core/cluster/cluster.cc
richhiey1996: Oh yes I forgot that these methods get inlined.
I'll wait for your comment on the query raised below to make these changes.

xapian-core/cluster/kmeans.cc
ojwb: You can't really pass around an object directly in C++ - the choices are passing a pointer, passing a reference (which is essentially syntactic sugar for passing a pointer, except a `NULL` reference isn't valid), or passing by value (which actually means the object gets copied - that's cheap for simple types such as `int`, `double`, but passing a complex type by value is best avoided - certainly for containers like `std::map`, etc which could be arbitrarily large.  Since a `Stopper` can be an arbitrary user subclass, it could be arbitrarily expensive to copy.

Another reason you can't pass a `Stopper` by value is object slicing - see https://en.wikipedia.org/wiki/Object_slicing for a fuller explanation, but essentially if you pass a `FooStopper` to a `func(Stopper x)` then `x` will actually be a copy of just the `Stopper` base class.

The key consideration is who owns the pointed to `Stopper` object (and in particular who is responsible for making sure it gets destroyed at a suitable point, and that it only gets destroyed once).  Due to the `opt_intrusive_ptr` mechanism, a `Stopper` object might be reference counted (in which case the lifetime is automatically handled), or it isn't reference counted (in which case the caller is responsible).

The second case allows allocating it on the stack (which isn't possible with the non-"opt" `intrusive_ptr`) - e.g.:

```
{
    SimpleStopper stopper;
    stopper.add("no");
    QueryParser qp;
    QueryParser.set_stopper(&stopper);
    /// Use qp.
}
```

Here `stopper` lives until we leave the enclosing block, as does `qp`.

The way that `opt_intrusive_ptr` works is that `Stopper` has `opt_intrusive_base` as a base class, and that holds the reference count (this in the "intrusive" bit - some other approaches hold the reference count externally and can be used to reference count almost any class, but the trade-off is that the intrusive approach is simpler).  The reference count is `0` when the pointer isn't reference counted and one more than the number of references when it is.

So you can make `Xapian::Internal::opt_intrusive_ptr<Xapian::Stopper> stopper` a member of `KMeans` and then just pass `stopper.get()` to the `TermListGroup` constructor, which can take a `Stopper*`.

In this case, you don't actually even need to assign that `Stopper*` to another `opt_intrusive_ptr` in that constructor - we know the object must be valid while we're executing the `TermListGroup` constructor as either it's not reference counted, or the caller holds a reference (tricky cases involving a second thread aren't a concern, as we explicitly don't support using the same Xapian API object concurrently from multiple threads).  But if you were going to store such an object as a member, you can just put it into another `opt_intrusive_ptr`, which would either just hold it, or (if it's reference counted) increment the reference count and then decrement it again if that `opt_intrusive_ptr` is destroyed or assigned a new pointer.

xapian-core/include/xapian/cluster.h
```diff
@@ -162,15 +163,20 @@ class XAPIAN_VISIBILITY_DEFAULT TermListGroup : public FreqSource {
...
     /** Constructor
      *
      *  @param docs	MSet object used to construct the TermListGroup
+     *  @param stopper	Xapian::Stopper object to identify stopwords
      */
-    explicit TermListGroup(const MSet &docs);
+    explicit TermListGroup(const MSet &docs, const Stopper *stopper = NULL);
+
+    /// Default constructor
+    TermListGroup();
```
ojwb: Why did you add a default constructor?  It doesn't seem useful for this class (if you really wanted an empty `TermListGroup` for some reason, `TermListGroup(MSet())` would give you one).

richhiey1996: Yes I guess it's safe to remove it. I'll add that change.

xapian-core/include/xapian/cluster.h
```diff
@@ -605,6 +611,14 @@ class XAPIAN_VISIBILITY_DEFAULT KMeans : public Clusterer {
      */
     ClusterSet cluster(const MSet &mset);
 
+    /** Set the Xapian::Stopper object to be used for identifying stopwords.
+     *  Stopwords are discarded while calculating term frequency for terms
+     *
```
ojwb: There should really be a blank (except for the `*`) line between the brief description and the rest of the doc comment.

I missed this before, so rather than delay the merge, I've merged and then fixed it as a follow-up commit: 659b03cde562b55fb080dc9bb8c6114683e3a752

ojwb:
Thanks.  Autosquashed, rebased and pushed.

PR #168, opened 2017-07-10
Stemming

richhiey1996:
Added stemming and a subclass of Xapian::Stopper named Xapian::StemStopper for removal of stopwords as well as their stemmed equivalents.

### DIFF
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -88,6 +88,11 @@ TermListGroup::add_document(const Document &document, const Stopper *stopper)
 	if (stopper && (*stopper)(term))
 	    continue;
 
+	// Remove unstemmed terms since document vector should
+	// contain only stemmed terms
+	if (term[0] != 'Z')
+	    continue;
+
 	unordered_map<string, doccount>::iterator i;
 	i = termfreq.find(term);
 	if (i == termfreq.end())
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -618,10 +618,10 @@ Cluster::Internal::recalculate()
     centroid.divide(size());
 }
 
-StemStopper::StemStopper(const Stem &stemmer_)
+StemStopper::StemStopper(const Stem &stemmer_, stem_strategy strategy)
+    : stem_action(strategy), stemmer(stemmer_)
 {
-    LOGCALL_CTOR(API, "StemStopper", stemmer_);
-    stemmer = stemmer_;
+    LOGCALL_CTOR(API, "StemStopper", stemmer_ | strategy);
 }
 
 string
@@ -641,6 +641,19 @@ void
 StemStopper::add(const string &term)
 {
     LOGCALL_VOID(API, "StemStopper::add", term);
-    stop_words.insert(term);
-    stop_words.insert('Z' + stemmer(term));
+    switch (stem_action) {
+	case STEM_NONE:
+	    stop_words.insert(term);
+	    break;
+	case STEM_ALL_Z:
+	    stop_words.insert('Z' + stemmer(term));
+	    break;
+	case STEM_ALL:
+	    stop_words.insert(stemmer(term));
+	    break;
+	case STEM_SOME:
+	    stop_words.insert(term);
+	    stop_words.insert('Z' + stemmer(term));
+	    break;
+    }
 }
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -307,15 +307,16 @@ Point::Point(const TermListGroup &tlg, const Document &document_)
 	string term = *it;
 	double termfreq = tlg.get_termfreq(term);
 
-	// If the term exists in only one document within the MSet,
-	// or if it is a filter term, then these terms are not used
-	// for document vector calculations
-	if (wdf < 1 || termfreq <= 1)
+	// If the term exists in only one document, or if it exists in
+	// every document within the MSet, or if it is a filter term, then
+	// these terms are not used for document vector calculations
+	if (wdf < 1 || termfreq <= 1 || size == termfreq)
 	    continue;
 
 	double tf = 1 + log((double)wdf);
 	double idf = log(size / termfreq);
 	double wt = tf * idf;
+
 	weights[term] = wt;
 	magnitude += wt * wt;
     }
--- a/xapian-core/cluster/cluster.cc
+++ b/xapian-core/cluster/cluster.cc
@@ -616,3 +616,30 @@ Cluster::Internal::recalculate()
     }
     centroid.divide(size());
 }
+
+StemStopper::StemStopper(const Stem &stemmer_)
+{
+    LOGCALL_CTOR(API, "StemStopper", stemmer_);
+    stemmer = stemmer_;
+}
+
+string
+StemStopper::get_description() const
+{
+    string desc("Xapian::StemStopper(");
+    unordered_set<string>::const_iterator i;
+    for (i = stop_words.begin(); i != stop_words.end(); ++i) {
+	if (i != stop_words.begin()) desc += ' ';
+	desc += *i;
+    }
+    desc += ')';
+    return desc;
+}
+
+void
+StemStopper::add(const string &term)
+{
+    LOGCALL_VOID(API, "StemStopper::add", term);
+    stop_words.insert(term);
+    stop_words.insert('Z' + stemmer(term));
+}
--- a/xapian-core/include/xapian/cluster.h
+++ b/xapian-core/include/xapian/cluster.h
@@ -33,9 +33,32 @@
 #include <xapian/visibility.h>
 
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 namespace Xapian {
+
+/// Stopper subclass which checks for both stemmed and unstemmed stopwords
+class XAPIAN_VISIBILITY_DEFAULT StemStopper : public Xapian::Stopper {
+    std::unordered_set<std::string> stop_words;
+    Xapian::Stem stemmer;
+
+  public:
+    /** Constructor
+     *
+     *  @param stemmer	The Xapian::Stem object to set.
+     */
+    explicit StemStopper(const Xapian::Stem &stemmer);
+
+    std::string get_description() const;
+
+    bool operator()(const std::string & term) const {
+	return stop_words.find(term) != stop_words.end();
+    }
+
+    /// Add a single stop word and its stemmed equivalent
+    void add(const std::string &term);
+};
 
 /** Class representing a set of documents in a cluster
  */
--- a/xapian-core/include/xapian/cluster.h
+++ b/xapian-core/include/xapian/cluster.h
@@ -40,15 +40,16 @@ namespace Xapian {
 
 /// Stopper subclass which checks for both stemmed and unstemmed stopwords
 class XAPIAN_VISIBILITY_DEFAULT StemStopper : public Xapian::Stopper {
-    std::unordered_set<std::string> stop_words;
-    Xapian::Stem stemmer;
-
   public:
+    /// Stemming strategies
+    typedef enum { STEM_NONE, STEM_SOME, STEM_ALL, STEM_ALL_Z } stem_strategy;
+
     /** Constructor
      *
      *  @param stemmer	The Xapian::Stem object to set.
+     *  @param strategy The stemming strategy to be used.
      */
-    explicit StemStopper(const Xapian::Stem &stemmer);
+    explicit StemStopper(const Xapian::Stem &stemmer, stem_strategy strategy = STEM_SOME);
 
     std::string get_description() const;
 
@@ -58,6 +59,11 @@ class XAPIAN_VISIBILITY_DEFAULT StemStopper : public Xapian::Stopper {
 
     /// Add a single stop word and its stemmed equivalent
     void add(const std::string &term);
+
+  private:
+    stem_strategy stem_action;
+    std::unordered_set<std::string> stop_words;
+    Xapian::Stem stemmer;
 };
 
 /** Class representing a set of documents in a cluster
--- a/xapian-core/tests/api_cluster.cc
+++ b/xapian-core/tests/api_cluster.cc
@@ -29,14 +29,38 @@
 #include "testsuite.h"
 #include "testutils.h"
 
+static void
+make_generated_db(Xapian::WritableDatabase &db, const std::string &)
+{
+    const std::unordered_set<std::string>
+    test_strings({"This line is about a cluster. Cluster is important and is everywhere",
+		  "We need to search for special cluster. Cluster cluster cluster",
+		  "Computer cluster is a special example of a cluster. Used to search fast",
+		  "Another example of cluster is a star cluster. Star cluster has a lot of stars"});
+
+    Xapian::TermGenerator indexer;
+    Xapian::Stem stemmer("english");
+    indexer.set_stemmer(stemmer);
+
+    for (std::unordered_set<std::string>::const_iterator it = test_strings.begin(); it != test_strings.end(); ++it)
+    {
+	Xapian::Document document;
+	const std::string &document_data = *it;
+	document.set_data(document_data);
+	indexer.set_document(document);
+	indexer.index_text(document_data);
+	db.add_document(document);
+    }
+}
 /** Test for cosine distance
  *  Cosine distance = 1 - (cosine of the angle between two vectors).
  *  Thus, if two vectors are equal, the distance between them will be zero
  *  and if two vectors are unequal, the distance will be 1 >= dist >= 0.
  */
-DEFINE_TESTCASE(cosine_distance1, backend)
+DEFINE_TESTCASE(cosine_distance1, generated)
 {
-    Xapian::Enquire enquire(get_database("apitest_cluster"));
+    Xapian::Database db = get_database("cosine_distance1", make_generated_db);
+    Xapian::Enquire enquire(db);
     enquire.set_query(Xapian::Query("cluster"));
 
     Xapian::MSet matches = enquire.get_mset(0, 4);
@@ -63,9 +87,9 @@ DEFINE_TESTCASE(cosine_distance1, backend)
 /** Round Robin Test
  *  Test that none of the returned clusters are empty
  */
-DEFINE_TESTCASE(round_robin1, backend)
+DEFINE_TESTCASE(round_robin1, generated)
 {
-    Xapian::Database db(get_database("apitest_cluster"));
+    Xapian::Database db = get_database("round_robin1", make_generated_db);
     Xapian::Enquire enq(db);
     enq.set_query(Xapian::Query("cluster"));
     Xapian::MSet matches = enq.get_mset(0, 4);
--- a/xapian-core/tests/api_cluster.cc
+++ b/xapian-core/tests/api_cluster.cc
@@ -30,28 +30,27 @@
 #include "testutils.h"
 
 static void
-make_generated_db(Xapian::WritableDatabase &db, const std::string &)
+make_stemmed_cluster_db(Xapian::WritableDatabase &db, const std::string &)
 {
-    const std::unordered_set<std::string>
-    test_strings({"This line is about a cluster. Cluster is important and is everywhere",
-		  "We need to search for special cluster. Cluster cluster cluster",
-		  "Computer cluster is a special example of a cluster. Used to search fast",
-		  "Another example of cluster is a star cluster. Star cluster has a lot of stars"});
+    static const char* const test_strings[] = {
+	"This line is about a cluster. Cluster is important and is everywhere",
+	"We need to search for special cluster. Cluster cluster cluster",
+	"Computer cluster is a special example of a cluster. Used to search fast",
+	"Another example of cluster is a star cluster. Star cluster has a lot of stars"
+    };
 
     Xapian::TermGenerator indexer;
     Xapian::Stem stemmer("english");
     indexer.set_stemmer(stemmer);
-
-    for (std::unordered_set<std::string>::const_iterator it = test_strings.begin(); it != test_strings.end(); ++it)
-    {
+    for (const std::string& document_data : test_strings) {
 	Xapian::Document document;
-	const std::string &document_data = *it;
 	document.set_data(document_data);
 	indexer.set_document(document);
 	indexer.index_text(document_data);
 	db.add_document(document);
     }
 }
+
 /** Test for cosine distance
  *  Cosine distance = 1 - (cosine of the angle between two vectors).
  *  Thus, if two vectors are equal, the distance between them will be zero
@@ -59,7 +58,7 @@ make_generated_db(Xapian::WritableDatabase &db, const std::string &)
  */
 DEFINE_TESTCASE(cosine_distance1, generated)
 {
-    Xapian::Database db = get_database("cosine_distance1", make_generated_db);
+    Xapian::Database db = get_database("stemmed_cluster", make_stemmed_cluster_db);
     Xapian::Enquire enquire(db);
     enquire.set_query(Xapian::Query("cluster"));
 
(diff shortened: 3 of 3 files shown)


xapian-core/include/xapian/cluster.h
```diff
@@ -28,14 +28,43 @@
...
+
+  public:
+    /// Default constructor
+    StemStopper();
+
+    /** Set the stemmer.
+     *
+     *  @param stem	The Xapian::Stem object to set.
+     */
+    void set_stemmer(const Xapian::Stem &stemmer);
```
ojwb: I think it would make more sense to just take the stemmer as a constructor parameter.  It's not useful to use this class without setting a stemmer, and it's not useful to change the stemmer mid-use, as you'll end up with a mixture of stems from different algorithms.

Also the `@param` doesn't match the actual parameter name.

xapian-core/include/xapian/cluster.h
```diff
@@ -28,14 +28,43 @@
...
 #include <xapian/visibility.h>
 
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 namespace Xapian {
 
+/** Implementation of the stopper class that can be used to identify
+ *  stemmed as well as unstemmed stopwords
```
ojwb: Ideally this should summarise the class in a single line.  Brevity is certainly important, so I'd avoid verbiage which doesn't really add much like "Implementation of the" and "can be used to".

Perhaps something like "Stopper subclass which checks for both stemmed and unstemmed stopwords".

An example of a context in which these brief descriptions are used is:

https://xapian.org/docs/apidoc/html/annotated.html

Some of the existing ones make better use of the limited space than others, and a few clearly include too much irrelevant information (`DLHWeight` for example).

ojwb:
Looks like there's a genuine test failure on the Linux CI builds (the OS X builds seem to be suffering from infrastructure problems currently).

ojwb:
For the use you have for this new stopper subclass, you only want the stemmed terms in the `Stopper` don't you?  So adding the unstemmed terms here is just going to be unnecessary bloat.

I wonder if this class should also take a `stem_strategy`?  The values have obvious meanings:

- `STEM_NONE`: only adds unstemmed terms (`Stemmer` ignored)
- `STEM_SOME`: what you have currently implemented
- `STEM_ALL`: doesn't add unstemmed terms, stemmed terms added without a `Z` prefix
- `STEM_ALL_Z`: doesn't add unstemmed terms, stemmed terms added with a `Z` prefix

xapian-core/include/xapian/cluster.h
richhiey1996: I tried to make it consistent with other `set_stemmer` methods. But I too believe it makes more sense as a constructor. Will make these changes.

xapian-core/include/xapian/cluster.h
richhiey1996: Yes I'll make that a little easier to read.

xapian-core/include/xapian/cluster.h
richhiey1996: Or should we allow initialization within the constructor as well as provide a set_stemmer methods?

richhiey1996:
> I wonder if this class should also take a stem_strategy? 

I was wanting to discuss this once the implementation was done. I see it as a useful feature to add to the Stopper because it can be used in multiple ways then. The reason I let the unstemmed terms in the set remain is because I thought it would be useful to check sometimes against the stemmed, and sometimes against the unstemmed term. If we add in the `stem_strategy`, we could add only those terms that are required for a particular use case. I believe it's a good idea. Can I go ahead with it?

xapian-core/include/xapian/cluster.h
ojwb: API consistency is good (and good to think about), but it's not the only consideration.

As I said, it's not useful to change the stemmer mid-use, so providing both seems to just be providing a method for which there's no sensible use.  The difference for the other classes with `set_stemmer()` is that it is useful to be able to change the stemmer - e.g. I might use the same `QueryParser` object to parse many queries, and have a control in the UI which selects the stemming language to use.

A more extreme example - arguably it would be more consistent if every Xapian API class had a `set_stemmer()` method, since several classes do.  But this greater consistency doesn't help the API user - instead of remembering which classes support `set_stemmer()`, they instead need to remember which classes support `set_stemmer()` in a useful way...

ojwb:
> Can I go ahead with it?

Sure, though I'd suggest prioritising getting #167 merged as that touches some of the same code as this.

xapian-core/include/xapian/cluster.h
```diff
@@ -38,22 +38,17 @@
...
-    StemStopper();
-
-    /** Set the stemmer.
+    /** Constructor
      *
-     *  @param stem	The Xapian::Stem object to set.
+     *  @param stemmer	The Xapian::Stem object to set.
      */
-    void set_stemmer(const Xapian::Stem &stemmer);
+    StemStopper(const Xapian::Stem &stemmer);
```
ojwb: Should be marked `explicit`.

richhiey1996:
> Sure, though I'd suggest prioritising getting #167 merged as that touches some of the same code as this.

Yes. This PR will require changing tests for cosine distance too (since removing stems by checking for the 'Z' prefix breaks it). So I'll get back to this once stop word removal is done.

xapian-core/include/xapian/cluster.h
```diff
@@ -28,14 +28,38 @@
...
+    std::string get_description() const;
+
+    bool operator()(const std::string & term) const {
+	return stop_words.find(term) != stop_words.end();
+    }
+
+    /// Add a single stop word and its stemmed equivalent
+    void add(const std::string &term);
+};
+
```
samuelharden: Just wondering why do we even stem the stop word. We can always check if it's a stopword before stemming. I am not sure where exactly we will need StemStopper?

What is the use-case for StemStopper, I don't see any changes where it's used in this pull request, so curious is what's the use-case.

ojwb: > We can always check if it's a stopword before stemming

The plan is to use a `Stopper` here to allow filtering out already-stemmed terms from documents we're about to cluster.  The stemming happened at index time, so we can't really check before stemming in this case.

> I don't see any changes where it's used in this pull request, so curious is what's the use-case.

Hmm, there really ought to be a feature test added for `StemStopper` to ensure it works as intended, and continues to do so.

richhiey1996: Yes as @ojwb mentioned, StemStopper is used to allow filtering of stopwords which are stemmed when we are creating the document term lists. For creating document term lists, we are using two types of dimensionality reduction - stopword removal and stemming. Since document termlists already contain stemmed terms, that isn't a problem but identifying stopwords which are stemmed is what is needed. Thus if a user inputs the stopwords that they want to filter with, we can stem and remove the stopwords, thus giving us a smaller and more representative termlist.

richhiey1996: > Hmm, there really ought to be a feature test added for StemStopper to ensure it works as intended, and continues to do so.

I will be fixing the broken test and adding a test for StemStopper shortly.

xapian-core/cluster/cluster.cc
```diff
@@ -304,6 +311,11 @@ Point::Point(const TermListGroup &tlg, const Document &document_)
 	double tf = 1 + log((double)wdf);
 	double idf = log(size / termfreq);
 	double wt = tf * idf;
+
+	// If weight of a term is zero, it doesn't contribute to the termlist
+	if (wt == 0)
+	    continue;
```
ojwb: `wdf >= 1`so `log(wdf) >= 0` so `tf >= 1`.

Hence to get `wt == 0` we need `idf == 0`, which only happens when `size == termfreq`.  Perhaps we should just check that above, and avoid two calls to `log()` (which is quite slow) in this case?

Is `termfreq > size` possible?  That would lead to negative `wt` which seems undesirable, so if it can happen we should probably exclude that too.

xapian-core/cluster/cluster.cc
```diff
@@ -604,3 +616,44 @@ Cluster::Internal::recalculate()
...
+    const string &stemmed_term = 'Z' + stemmer(term);
+    if (stem_action == STEM_NONE)
+	stop_words.insert(term);
+    else if (stem_action == STEM_ALL_Z)
+	stop_words.insert(stemmed_term);
+    else {
+	stop_words.insert(term);
+	stop_words.insert(stemmed_term);
+    }
+}
```
ojwb: This handles `STEM_ALL` the same as `STEM_SOME` which is incorrect.

I'd suggest handling an `enum` with a `switch` rather than an `if` cascade.  That way we'll get a warning (at least with GCC) if a new `enum` value is added and nobody adds the code here to handle it.

This also always calls the stemmer, but under `STEM_NONE` the result is never used, so that's wasted effort.

And finally, it's probably better stylistically to use curly brackets on all or none of the actions in an `if`/`else if`/`else` chain.

xapian-core/include/xapian/cluster.h
```diff
@@ -28,14 +28,45 @@
...
+#include <unordered_set>
 #include <vector>
 
 namespace Xapian {
 
+/// Stopper subclass which checks for both stemmed and unstemmed stopwords
+class XAPIAN_VISIBILITY_DEFAULT StemStopper : public Xapian::Stopper {
+  public:
+    /// Stemming strategies
+    typedef enum { STEM_NONE, STEM_SOME, STEM_ALL_Z } stem_strategy;
```
ojwb: Oh, I see - you don't have `STEM_ALL` here.  I think this enum ought to match the existing `stem_strategy` enums in `QueryParser` and `TermGenerator`.

xapian-core/tests/api_cluster.cc
```diff
@@ -80,3 +104,27 @@ DEFINE_TESTCASE(round_robin1, backend)
...
+    TEST(stopper('Z' + stemmer(term)));
+
+    term = "and";
+    stopper.set_stemming_strategy(Xapian::StemStopper::STEM_NONE);
+    stopper.add(term);
+    TEST(stopper(term));
+    TEST(!stopper('Z' + stemmer(term)));
+
+    return true;
+}
```
ojwb: There's no newline character at the end of this file (the red slashed circle and arrow thing in the git hub diff view shows this).

xapian-core/include/xapian/cluster.h
```diff
@@ -28,14 +28,45 @@
...
+    std::string get_description() const;
+
+    bool operator()(const std::string & term) const {
+	return stop_words.find(term) != stop_words.end();
+    }
+
+    /// Add a single stop word and its stemmed equivalent
+    void add(const std::string &term);
+
+    void set_stemming_strategy(stem_strategy strategy);
```
ojwb: It probably makes more sense to make this an optional parameter on the constructor (default `STEM_SOME`).  It doesn't seem useful to be able to change it (like it's not useful to change the stemmer used for a given `StemStopper` object), and more convenient to be able to set up the stopper in one go.

xapian-core/tests/api_cluster.cc
```diff
@@ -80,3 +104,27 @@ DEFINE_TESTCASE(round_robin1, backend)
...
+    TEST(!stopper(term));
+    TEST(stopper('Z' + stemmer(term)));
+
+    term = "and";
+    stopper.set_stemming_strategy(Xapian::StemStopper::STEM_NONE);
+    stopper.add(term);
+    TEST(stopper(term));
+    TEST(!stopper('Z' + stemmer(term)));
+
+    return true;
```
samuelharden: Shall we also add a test for non-stop word term as well here?

xapian-core/tests/api_cluster.cc
```diff
@@ -80,3 +104,27 @@ DEFINE_TESTCASE(round_robin1, backend)
...
+    Xapian::Stem stemmer("english");
+    Xapian::StemStopper stopper(stemmer);
+    std::string term = "the";
+    stopper.add(term);
+    TEST(stopper(term));
+    TEST(stopper('Z' + stemmer(term)));
+
+    term = "because";
+    stopper.set_stemming_strategy(Xapian::StemStopper::STEM_ALL_Z);
+    stopper.add("because");
```
ojwb: Here you repeat `"because"` but above and below you use `term`.

xapian-core/tests/api_cluster.cc
ojwb: Yes, that would be good.

xapian-core/tests/api_cluster.cc
```diff
@@ -29,14 +29,38 @@
...
 #include "testutils.h"
 
+static void
+make_generated_db(Xapian::WritableDatabase &db, const std::string &)
+{
+    const std::unordered_set<std::string>
+    test_strings({"This line is about a cluster. Cluster is important and is everywhere",
+		  "We need to search for special cluster. Cluster cluster cluster",
+		  "Computer cluster is a special example of a cluster. Used to search fast",
+		  "Another example of cluster is a star cluster. Star cluster has a lot of stars"});
```
richhiey1996: Ah yes sure. I'll make this change.

xapian-core/cluster/cluster.cc
richhiey1996: > Is termfreq > size possible? That would lead to negative wt which seems undesirable, so if it can happen we should probably exclude that too.

That doesn't seem like a possible case because the maximum termfreq of any term will be equal to the number of documents according to how it is being calculated. But I guess it's a common case for a few words to have size ==  termfreq (excluding stopwords). These could be terms that are common in context of the corpus (sports related documents could contain the word 'play' in all documents). So it looks useful to remove these terms before calculating their weight.

xapian-core/tests/api_cluster.cc
```diff
@@ -63,9 +87,9 @@ DEFINE_TESTCASE(cosine_distance1, backend)
 /** Round Robin Test
  *  Test that none of the returned clusters are empty
  */
-DEFINE_TESTCASE(round_robin1, backend)
+DEFINE_TESTCASE(round_robin1, generated)
 {
-    Xapian::Database db(get_database("apitest_cluster"));
+    Xapian::Database db = get_database("round_robin1", make_generated_db);
```
richhiey1996: Yes sure I'll add these changes in.

xapian-core/include/xapian/cluster.h
richhiey1996: Currently, the main idea being use to check whether a term is stemmed or not is checking it's first term. If `term[0] == 'Z'`, we allow the term to be added in if it's not a stopword. STEM_NONE and STEM_ALL seem inappropriate here because they were the reason the tests were failing before. If we do STEM_ALL, that would store stemmed terms without the 'Z' prefix which would lead to an empty termlist. In this case, do we need to change the way we look for stemmed terms?

xapian-core/include/xapian/cluster.h
richhiey1996: Yes I'll add that in.

xapian-core/tests/api_cluster.cc
richhiey1996: Yes that would be useful.

xapian-core/tests/api_cluster.cc
richhiey1996: Ah sorry about this. I'll change this.

xapian-core/tests/api_cluster.cc
richhiey1996: > (The first const attaches to the char *, meaning that the elements of the array are pointers to constant character data; the second const means the pointers in the array are also constant).

This is something new to me. I never knew we could use the keyword twice in once declaration. Thanks for this.

xapian-core/cluster/cluster.cc
jaylett: I doubt there are many in that situation, so having code to remove those terms seems unnecessary. If it were to make a big difference, then something like PCA (which we believe will help more generally) would catch this.

Moving the check around as Olly has suggested will make the system faster.

xapian-core/include/xapian/cluster.h
jaylett: You should support the same members as for `TermGenerator::set_stemming_strategy()` and similar, because otherwise it's confusing. I believe that `STEM_ALL` could do `stop_words.insert(stemmer(term))` to behave consistently.

xapian-core/cluster/cluster.cc
```diff
@@ -644,12 +637,17 @@ StemStopper::add(const string &term)
...
-	stop_words.insert(stemmed_term);
-    else {
-	stop_words.insert(term);
-	stop_words.insert(stemmed_term);
+    switch (stem_action) {
+	case STEM_NONE:
+	    stop_words.insert(term);
+	    break;
+	case STEM_ALL_Z:
+	case STEM_ALL:
```
jaylett: I don't think these should be handled the same way. `STEM_ALL_Z` will generate `Z<stem>` terms, but `STEM_ALL` will generate `<stem>` terms only. So `STEM_ALL` should `stop_words.insert(term)` not `stemmed_term` — I believe.

ojwb: Yes.

ojwb: And we shouldn't call the stemmer unless we're going to actually use the result, as I believe I already said.  The stemming algorithm is pluggable by the user, so could be arbitrary expensive.  Even the built-in stemming algorithms generally do a non-trivial amount of work.

ojwb:
Ping?  I think this is quite close, but there are a couple of unaddressed points since your last commit.

richhiey1996:
Yes I have added these changes in. Sorry about the delay with this PR.

xapian-core/cluster/cluster.cc
```diff
@@ -604,3 +612,42 @@ Cluster::Internal::recalculate()
...
+void
+StemStopper::add(const string &term)
+{
+    LOGCALL_VOID(API, "StemStopper::add", term);
+    switch (stem_action) {
+	case STEM_NONE:
+	    stop_words.insert(term);
+	    break;
+	case STEM_ALL_Z:
+	    stop_words.insert('Z' + stemmer(term));
```
jaylett: I believe you still need a `break` here.

richhiey1996: Oh that was a bad mistake. Sorry.

xapian-core/tests/api_cluster.cc
```diff
@@ -29,14 +29,39 @@
...
 #include "testutils.h"
 
+static void
+make_stemmed_cluster_db(Xapian::WritableDatabase &db, const std::string &)
+{
+    static const char* const test_strings[] =
+		{"This line is about a cluster. Cluster is important and is everywhere",
+		  "We need to search for special cluster. Cluster cluster cluster",
+		  "Computer cluster is a special example of a cluster. Used to search fast",
+		  "Another example of cluster is a star cluster. Star cluster has a lot of stars"};
```
ojwb: Non-standard formatting - should be:

```
    static const char* const test_strings[] = {
        "...",
        "..."
    };
```

xapian-core/tests/api_cluster.cc
```diff
@@ -29,14 +29,39 @@
...
+
+    Xapian::TermGenerator indexer;
+    Xapian::Stem stemmer("english");
+    indexer.set_stemmer(stemmer);
+    unsigned int count = sizeof(test_strings)/sizeof(test_strings[0]);
+    for (unsigned int i = 0; i < count; ++i)
+    {
```
ojwb: `{` should be on the same line as the `for`.

richhiey1996: Yes I have added these changes in now.

xapian-core/cluster/cluster.cc
jaylett: @richhiey1996 the fact that this needed changes but you haven't changed the tests suggests that the tests aren't quite right. Can you take a pass over the test for this thinking about the various different cases you need to explicitly check?

xapian-core/tests/api_cluster.cc
```diff
@@ -80,3 +105,35 @@ DEFINE_TESTCASE(round_robin1, backend)
...
+
+    Xapian::StemStopper stopper_all_z(stemmer, Xapian::StemStopper::STEM_ALL_Z);
+    Xapian::StemStopper stopper_all(stemmer, Xapian::StemStopper::STEM_ALL);
+    term = "because";
+    stopper_all.add(term);
+    stopper_all_z.add(term);
+    TEST(!stopper_all_z(term));
+    TEST(stopper_all_z('Z' + stemmer(term)));
+    TEST(!stopper_all(term));
+    TEST(stopper_all('Z' + stemmer(term)));
```
ojwb: The `'Z' + ` here looks wrong to me - `stopper_all` should only stop the stemmed form of `"because"` **without** a `Z` prefix.  So as-is, this test should be failing I think (I can't tell without pulling the code and building it because there's a merge conflict and that seems to suppress CI runs - not really sure why).

As well as checking that the stoppers don't stop the unstemmed term and do stop the correctly prefixed stemmed term, it would be good to also check that each doesn't stop the incorrectly prefixed stemmed term - i.e. `stopper_all_z` doesn't stop when there's no `Z` prefix, and that `stopper_all` doesn't stop when there is.

xapian-core/tests/api_cluster.cc
```diff
@@ -80,3 +105,35 @@ DEFINE_TESTCASE(round_robin1, backend)
...
+
+    Xapian::StemStopper stopper_none(stemmer, Xapian::StemStopper::STEM_NONE);
+    term = "and";
+    stopper_none.add(term);
+    TEST(stopper_none(term));
+    TEST(!stopper_none('Z' + stemmer(term)));
+
+    term = "cluster";
+    TEST(!stopper(term));
+    TEST(!stopper('Z' + stemmer(term)));
```
ojwb: A useful test, but it would be clearer to group this with the other `STEM_SOME` cases at the start.

xapian-core/tests/api_cluster.cc
ojwb: I just checked out your branch, and indeed this check fails for me.

You really should be running `make check` before pushing changes.

xapian-core/tests/api_cluster.cc
```diff
@@ -29,14 +29,39 @@
...
+	"This line is about a cluster. Cluster is important and is everywhere",
+	"We need to search for special cluster. Cluster cluster cluster",
+	"Computer cluster is a special example of a cluster. Used to search fast",
+	"Another example of cluster is a star cluster. Star cluster has a lot of stars"
+    };
+
+    Xapian::TermGenerator indexer;
+    Xapian::Stem stemmer("english");
+    indexer.set_stemmer(stemmer);
+    unsigned int count = sizeof(test_strings)/sizeof(test_strings[0]);
```
ojwb: Missing spaces around `/`.

It'd be neater to use C++11's range-based `for` though:

```
    for (const std::string& document_data : test_strings) {
        // ...
    }
```

xapian-core/cluster/cluster.cc
richhiey1996: Yes I had the same tests for STEM_ALL and STEM_ALL_Z. I'll add those changes in. I hadn't run a make check before pushing.

xapian-core/tests/api_cluster.cc
richhiey1996: Yes I didn't run a make check before pushing the last commit. I'll add the suggested changes in.

xapian-core/tests/api_cluster.cc
richhiey1996: Great. I'll group this with STEM_SOME.

xapian-core/tests/api_cluster.cc
richhiey1996: Oh yes this would be better. I'll add this in.

ojwb:
LGTM.

@jaylett: you requested changes, are you happy for me to merge?

jaylett:
@richhiey1996 I don't know if it's obvious to @ojwb how to resolve the conflict, so you may want to rebase and squash, then force push this branch to provide a final clean branch for merge.

richhiey1996:
@ojwb @jaylett  I've rebased to latest master and squashed all the fixup commits after resolving the conflict.

PR #169, opened 2017-07-13
Fixing letor bugs and improving ranker api

AyushP123:
Following Changes have been made as discussed on irc:
1. Renamed pure virtual function "virtual void Ranker::train_model" and all its instances to train.
2. FeatureList::create_feature_vector now returns an empty vector<FeatureVector> if the input MSet is empty.
3. Initialized the stats_needed variable in Feature constructor.

These were the few issues I found while testing. I will continue to discuss and resolve issues as they keep coming up.

### DIFF
--- a/xapian-letor/api/featurelist.cc
+++ b/xapian-letor/api/featurelist.cc
@@ -98,6 +98,8 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
 				    const Xapian::Database & letor_db) const
 {
     LOGCALL(API, std::vector<FeatureVector>, "FeatureList::create_feature_vectors", mset | letor_query | letor_db);
+    if (mset.empty())
+	return vector<FeatureVector>();
     std::vector<FeatureVector> fvec;
 
     for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
--- a/xapian-letor/feature/feature.cc
+++ b/xapian-letor/feature/feature.cc
@@ -26,7 +26,7 @@
 
 namespace Xapian {
 
-Feature::Feature()
+Feature::Feature() : stats_needed()
 {
     LOGCALL_CTOR(API, "Feature", NO_ARGS);
 }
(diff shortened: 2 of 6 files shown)


xapian-letor/api/featurelist.cc
```diff
@@ -98,7 +98,7 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
 				    const Xapian::Database & letor_db) const
 {
     LOGCALL(API, std::vector<FeatureVector>, "FeatureList::create_feature_vectors", mset | letor_query | letor_db);
-    if(mset.empty())
+    if( mset.empty())
```
ojwb: Still wrong - the space goes before the `(`.  Running `xapian-check-patch` before you push would save you a lot of time with this sort of thing.

ojwb:
> 1. Renamed pure virtual function "virtual void Ranker::train_model" and all its instances to train.
> 2. FeatureList::create_feature_vector now returns an empty vector<FeatureVector> if the input MSet is empty.
> 3. Initialized the stats_needed variable in Feature constructor.

When you find yourself making a list of different things you fixed like this, that's really an indication that each change should be a separate commit, with the brief commit message for each being essentially what you say in this list, and ideally a longer description explaining why (especially for the first one where the issue with subclasses having a different `train_model()` which hides the virtual one is the motivation, but that won't be obvious to the reader).  Then the commit log tells the story that this list currently does.

The key point to remember is that each commit should really be a single self-contained change.

This makes for easier review, because the reviewer can look at each commit in turn rather than having to keep 3 different changes in their head as they look through the changes.

And it is helpful later when trying to track down what change introduced a bug - `git bisect` automates binary chopping through the git history but can only narrow down to a single commit which caused a particular regression.  If the commits are each a small self-contained change that's much more helpful than being able to narrow it down to a commit which mixes several changes, which you then need to further pull apart by hand to work out which is to blame.

Also, when cherry-picking changes between branches, you can pick a whole commit or a list of commits, but cherry-picking half a commit requires a lot of manual fiddling.

ojwb:
Thanks, rebased and pushed as a335b2de66bfcc6f9cc5e2fcf94201079463be68 and preceding commits.

PR #170, opened 2017-07-23
SimplifiedDBN clickmodel implementation

ivmarkp:
This PR adds the implementation of Simplified Dynamic Bayesian Network click model based on the algorithm 1 in the paper: [A dynamic bayesian network click model for web search ranking](http://dl.acm.org/citation.cfm?id=1526711).


xapian-applications/omega/clickmodel/simplifiedDBN.cc
```diff
@@ -0,0 +1,208 @@
...
+	for (size_t j = 0; j <= sessions[i][CLICKS].length(); ++j) {
+	    if (sessions[i][CLICKS][j] == '\0') {
+		clicks.push_back(click);
+	    } else if (sessions[i][CLICKS][j] != ',') {
+		// Get clickstring of the form "docid:click_count".
+		clickstring += sessions[i][CLICKS][j];
+
+		// Get click count string.
+		string delimiter = ":";
+		int delimiterPos = clickstring.find(delimiter);
```
ojwb: Searching for a character will be more efficient that searching for a `std::string` - I'd probably eliminate `delimiter` entirely and just use `clickstring.find(':');` but if you want to name the thing, `char delimiter = ':';` would also work.

ivmarkp:
@ojwb Sorry, I forgot to ran `make` on this branch before pushing the code and thus the build failure. I'll try to fix the that and then will resolve the issues raised by you in the last review.

xapian-applications/omega/clickmodel/simplifiedDBN.cc
```diff
@@ -0,0 +1,208 @@
...
+    for (size_t i = 0; i < sessions.size(); ++i) {
+	for (size_t j = 0; j < sessions[i][DOCIDS].length(); ++j) {
+	    if (sessions[i][DOCIDS][j] != ',') {
+		docid += sessions[i][DOCIDS][j];
+	    } else {
+		docids.push_back(docid);
+		docid.clear();
+	    }
+	}
+    }
```
ivmarkp: No, that was a bug. Thanks for the catch. Fixed now.

xapian-applications/omega/clickmodel/simplifiedDBN.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * For more information of DBN click model, see the following paper:
+ *
+ * Olivier Chapelle and Ya Zhang. 2009. A dynamic bayesian network click
+ * model for web search ranking. In Proceedings of the 18th international
+ * conference on World wide web (WWW '09). 
+ */
+class SimplifiedDBN {
+	///
+	map<string, map<string, map<char, double>>> urlRelevances;
```
ivmarkp: > It's not clear to me what these mean, and it's overkill to use a map for two fixed values.

`'a'` and `'s'` refer to the model parameters: attractiveness probabilities and satisfaction probabilities as mentioned in the paper.

> I'd suggest defining constants instead of using single characters, for example (except give them meaningful names instead of using A and S):

Yes, thanks for the suggestion of using `enum`. I've made the required changes.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -73,9 +75,9 @@ getClicksList(vector<vector<string>> &sessions)
 		clickstring += sessions[i][CLICKS][j];
 
 		// Get click count string.
-		string delimiter = ":";
-		int delimiterPos = clickstring.find(delimiter);
-		clickcount = clickstring.substr(delimiterPos + 1, clickstring.length() - 1 - delimiterPos);
+		int delimiter_pos = clickstring.find(':');
+		clickcount = clickstring.substr(delimiter_pos + 1,
+			     clickstring.length() - 1 - delimiter_pos);
```
ojwb: If you omit the second argument of `substr()`, it defaults to `string::npos` which means "the rest of the string":

http://en.cppreference.com/w/cpp/string/basic_string/substr

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -90,7 +92,7 @@ getClicksList(vector<vector<string>> &sessions)
 }
 
 vector<vector<string>>
-SimplifiedDBN::buildSessions(string logfile)
+SimplifiedDBN::build_sessions(string logfile)
```
ojwb: Xapian convention is to pass `std::string` parameters by const reference.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -148,50 +150,48 @@ SimplifiedDBN::buildSessions(string logfile)
...
-	vector<int> clicks = getClicksList(sessions);
+	vector<int> clicks = get_click_list(sessions);
 
-	int lastClickedPos = clicks.size() - 1;
+	int last_clicked_pos = clicks.size() - 1;
 
 	for (size_t j = 0; j < clicks.size(); ++j)
 	    if (clicks[j] != 0)
-		lastClickedPos = j;
+		last_clicked_pos = j;
```
ojwb: If the final entry in `clicks` is non-zero, then `last_clicked_pos` will be `clicks.size() - 1` (this is also the case if all entries in `clicks` are `0`).

But then in the loop below, on the final iteration `x` will be `last_clicked_pos + 1` which is `clicks.size()` and then `clicks[x]` results in undefined behaviour.

I think you want `x <= last_clicked_pos` below, which would copy over all clicks up to and including the final non-zero one.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,213 @@
...
+
+	int last_clicked_pos = clicks.size() - 1;
+
+	for (size_t j = 0; j < clicks.size(); ++j)
+	    if (clicks[j] != 0)
+		last_clicked_pos = j;
+
+	vector<int> session_clicks;
+	for (int x = 0; x <= last_clicked_pos + 1; ++x)
+	    session_clicks.push_back(clicks[x]);
```
ojwb: There doesn't seem to be any need to make a partial copy of `clicks` here - the loop below could easily just loop over clicks from `0` to `last_clicked_pos` (or `last_clicked_pos + 1` if that's really correct).

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,213 @@
...
+	    } else {
+		url_rel_fractions[qid][docids[k]]['a'][0] += 1;
+	    }
+	}
+    }
+
+    for (auto i = url_rel_fractions.begin(); i != url_rel_fractions.end(); ++i) {
+	string qid = i->first;
+	for (auto j = i->second.begin(); j != i->second.end(); ++j) {
+	    auto rel_fractions = j->second;
```
ojwb: Simpler to use C++11 range `for` here - the above two lines could instead be:

```
    for (auto&& rel_fractions : i->second) {
```

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    file.close();
+    return sessions;
+}
+
+void
+SimplifiedDBN::train(vector<vector<string>> &sessions)
+{
+    map<string, map<string, map<char, array<double, 2>>>> url_rel_fractions;
```
ojwb: This `map<char, array<double, 2>>` should also get the `double[PARAM_COUNT_]` treatment.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,213 @@
...
+							     rel_fractions['s'][0]);
+	}
+    }
+}
+
+vector<int>
+SimplifiedDBN::get_predicted_relevances(vector<vector<string>> &sessions)
+{
+    vector<int> relevances;
+    double a, s;
```
ojwb: Better to declare variables at the point of initialisation when you can, then there are no undefined values which could be used by mistake.  And you trivially can here - just merge these declarations into the assignments in the loop.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,213 @@
...
+
+	session_entry.clear();
+    }
+
+    file.close();
+    return sessions;
+}
+
+void
+SimplifiedDBN::train(vector<vector<string>> &sessions)
```
ojwb: Shouldn't `sessions` be `const`?  It doesn't seem like it should be modified by this function.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: Ok, got it. Will fix this.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: There is `clickstring.length() - 1 - delimiter_pos` as the second argument of `substr()` here. Did I miss something else here?

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: You are right. It should really be `x <= last_clicked_pos` instead of `x <= last_clicked_pos + 1` here. Thanks for catching this mistake.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: Yes, that's right. We don't need `sessions_clicks` if the loop below goes from `0` to `last_clicked_pos`.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: Okay, will make the required changes here.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: Yes, indeed.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: Yes, it should be `const`.

ivmarkp:
@jaylett How do you suggest writing tests for the SDBN class?

One way is to write a standalone test program that calls the class functions and compares their outputs for a bunch of cases with whatever is expected using standard C++ functions (string etc.) OR we could copy Xapian harness test suite files over and make use of available macros such as `TEST_STRINGS_EQUAL`, `TEST_EQUAL` etc. Although, we might end up copying a lot of code over for essentially using a couple macros or so in the end.

jaylett:
One approach would be to put this class into core, and then put the tests there with it. @ojwb may have some thoughts on this. I don't believe that this class is _only_ useful for omega users, it's just that omega can produce the input files out of the box. It wouldn't be difficult, for instance, for a django+xapian user to record the same log entries, and then they'd want a tool for converting to letor input format — but may not have installed omega at all.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,207 @@
...
+}
+
+static vector<string>
+get_docid_list(const vector<vector<string>> &sessions)
+{
+    vector<string> docids;
+    string docid;
+    for (size_t i = 0; i < sessions.size(); ++i) {
+	for (size_t j = 0; j <= sessions[i][DOCIDS].length(); ++j) {
+	    if (sessions[i][DOCIDS][j] != ',' && sessions[i][1][j] != '\0') {
```
ojwb: You're hard-coding `1` instead of `DOCIDS` in the second use here.

I'd factor out the character lookup, so it's clearer to the reader that it's the same character being tested each time and then appended, i.e.:

```
char ch = sessions[i][DOCIDS][j];
if (ch != ',' && ch != '\0') {
   docid += ch;
```

ojwb:
I'm not sure this really makes much sense as an API addition to xapian-core - adding it to xapian-letor would make more sense if we want to provide this as an API.  But I wonder if it's best just being a stand-alone tool for now, given how it's intended to be used.

We could test that stand-alone tool from a script (like we do for `omegatest`) or we could write unit tests for the class interface (it doesn't need to be a public API for that to work - we unit test a lot of components of omega like this - see `xapian-applications/omega/*test.cc`.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,207 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include <config.h>
+
+#include "simplifieddbn.h"
+
+#include <algorithm>
+#include <iomanip>
+#include <iostream>
```
ojwb: It doesn't look like `<iostream>` is used in this file.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,207 @@
...
+SimplifiedDBN::get_predicted_relevances(const vector<vector<string>> &sessions)
+{
+    vector<int> relevances;
+
+    vector<string> docids = get_docid_list(sessions);
+
+    for (size_t i = 0; i < docids.size(); ++i) {
+	double attr_prob = url_relevances[sessions[i][QID]][docids[i]][PARAM_ATTR_PROB];
+	double sat_prob = url_relevances[sessions[i][QID]][docids[i]][PARAM_SAT_PROB];
+	relevances.push_back(attr_prob * sat_prob);
```
ojwb: Why does this method return `vector<int>` when the computation gives a floating point result?

xapian-applications/omega/clickmodel/docs/clickmodel.rst
```diff
@@ -0,0 +1,66 @@
...
+``SimplifiedDBN`` class provides train function that takes a search session
+as its input:
+
+	``void SimplifiedDBN::train(vector<vector<string>> &sessions)``
+
+To generate search sessions from the log data file, ``SimplifiedDBN`` class
+also provides a separate function which takes the path to the log data file
+as its input:
+
+	``vector<vector<string>> SimplifiedDBN::buildSessions(string logfile)``
```
ojwb: `build_sessions`

xapian-applications/omega/clickmodel/docs/clickmodel.rst
```diff
@@ -0,0 +1,66 @@
...
+Predicting the Document Relevance
+=================================
+
+``SimplifiedDBN`` class provides a function to get the predicted relevance of
+all documents in a search session based on the trained SDBN model:
+
+	vector<int> get_predicted_relevances(vector<vector<string>> &sessions);
+
+It takes a search session as its input and returns a list of predicted relevances
+of each document in that search session.
```
ojwb: "It takes a search session as its input and returns a list of predicted relevances of each document in that search session" doesn't really tell me anything very useful.

The reader can see from just looking at the method prototype that you pass it "sessions"; also the method name is `get_predicted_relevances()` and it returns `vector<int>`, so it's expected that the returned value is a list of predicted relevances.

To be useful, you need to tell the reader things they can't trivially work out for themselves from the names of the function and its parameters.

For example, how does one create a suitable `vector<vector<string>>` value to pass as the argument?  I'm guessing from the `build_sessions()` method, but I shouldn't need to guess.

And what are the values in the returned `vector<int>` and what would one do with them?  Despite reading through the code, I've no idea about that.  Without knowing that, I can't usefully write code to use this method.

I also think it would be better to put this documentation as extended documentation comments for the corresponding methods in the header file.  Currently the header file just has a brief description for each method, then the user needs to somehow know to come and look here to find more detailed information.

Also, splitting the information up like this makes it more likely that only one part gets updated if a method signature is changed.  Documentation comments in the headers work well because people modifying the method signature are more likely to see they need updating.  As a case in point, the documentation above still says `vector<vector<string>> SimplifiedDBN::buildSessions(string logfile)`, but the header currently has `vector<vector<string>> build_sessions(const string logfile);`.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,207 @@
...
+		clickstring.clear();
+		clicks.push_back(click);
+	    }
+	}
+    }
+    return clicks;
+}
+
+vector<vector<string>>
+SimplifiedDBN::build_sessions(const string logfile)
```
ojwb: `const` **reference**, not just `const`.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,207 @@
...
+	    if (sessions[i][CLICKS][j] == '\0') {
+		clicks.push_back(click);
+	    } else if (sessions[i][CLICKS][j] != ',') {
+		// Get clickstring of the form "docid:click_count".
+		clickstring += sessions[i][CLICKS][j];
+
+		// Get click count string.
+		int delimiter_pos = clickstring.find(':');
+		clickcount = clickstring.substr(delimiter_pos + 1,
+			     clickstring.length() - 1 - delimiter_pos);
```
ojwb: Second argument of `substr()` can be omitted when you want the rest of the string.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,207 @@
...
+
+		// Convert click count string to an integer.
+		stringstream ss(clickcount);
+		ss >> click;
```
ojwb: This splits and parses the number in `clickstring` for every character appended to it.

I think you want to merge the `,` and `\0` cases into one, and split and parse the number there instead.  Essentially the same loop structure that you have in `get_docid_list()` above.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,207 @@
...
+    for (size_t i = 0; i < sessions.size(); ++i) {
+	for (size_t j = 0; j <= sessions[i][CLICKS].length(); ++j) {
+	    if (sessions[i][CLICKS][j] == '\0') {
+		clicks.push_back(click);
+	    } else if (sessions[i][CLICKS][j] != ',') {
+		// Get clickstring of the form "docid:click_count".
+		clickstring += sessions[i][CLICKS][j];
+
+		// Get click count string.
+		int delimiter_pos = clickstring.find(':');
```
ojwb: `string::find()` doesn't return `int`.

xapian-applications/omega/clickmodel/docs/clickmodel.rst
ivmarkp: @ojwb I've addressed all pending review comments and have also added tests the way it was suggested earlier.  Also, thanks for highlighting what's missing in this documentation. Will consider all the points that you made here to make the required changes in it.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -43,60 +43,61 @@ SimplifiedDBN::name()
...
 vector<vector<string>>
-SimplifiedDBN::build_sessions(const string logfile)
+SimplifiedDBN::build_sessions(const string &logfile)
 {
     ifstream file;
     file.open(logfile, ios::in);
 
+    if (!file) {
+	cerr << "ERROR: Specified file " << logfile;
+	throw runtime_error(" does not exist.");
```
jaylett: Surely this doesn't make sense — half the error message goes to `cerr`, but half is in the exception message. Probably the whole thing should be in the exception message, so that client code can catch it and do something appropriate.

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -0,0 +1,130 @@
...
+};
+
+struct sessions_testcase {
+    const string logfile;
+    const sessions_items sessions;
+};
+
+string sample_log1 = "test1.log";
+string sample_log2 = "test2.log";
+string sample_log3 = "test3.log";
```
ivmarkp: These strings should really be full system path to these files so no point in hardcoding their values. So, currently `sdbntest` will fail due to that reason. I think we need something like `test_driver::get_srcdir()` from `xapian-core/testsuite.h` here too or is there any other way to achieve the same goal?

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -0,0 +1,130 @@
...
+    SimplifiedDBN sdbn;
+
+    // Tests for SimplifiedDBN::build_sessions method.
+    for (size_t i = 0; i < sizeof(sessions_tests) / sizeof(sessions_tests[0]); ++i) {
+	vector<vector<string>> result;
+	try {
+	    result = sdbn.build_sessions(sessions_tests[i].logfile);
+	} catch (std::exception &ex) {
+	    cout << ex.what() << endl;
+	    exit(1);
```
jaylett: Rather than exit in this loop, you should probably note that a failure has occurred and keep going, so all tests are run. (Otherwise if you break things you'll fix the first failure, but not know about the second until you run the tests again — which is a little annoying. With a small number of tests it's not bad, but if this gets more complex it could slow things down a lot.)

ivmarkp:
@jaylett Can you please elaborate a bit on how a utility command that you mentioned earlier should be like from the context of omega users and what sort of functionalities it would have?

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: This is actually to let the user know which file could not be opened (might be helpful when running multiple tests). Should I still remove `cerr` message?

jaylett:
@ivmarkp it needs to be a command that can be invoked by users (in the way that `omindex` can be invoked, for instance). Users should be able to pass it input filenames and an output filename and have it build the file needed to train letor.

xapian-applications/omega/clickmodel/simplifieddbn.cc
jaylett: You need to think of this as a library function: what would a _user_ of the library expect when something goes wrong? An exception to be raised? An error printed to `cerr`? Both? I've recommended the first of those.

ivmarkp: Oh okay, fixed this.

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -0,0 +1,130 @@
...
+    {sample_log1, {"821f03288846297c2cf43c34766a38f7",
+		   "45,36,14,54,42",
+		   "45:0,36:0,14:0,54:2,42:0"}},
+    {sample_log2,{}}
+};
+
+static bool
+almost_equal(const double &x, const double &y,
+			 const double epsilon = 0.001) {
+    return (abs(x - y) < epsilon) ? true : false;
```
ojwb: `operator<` returns `bool`, so the whole ` ? true : false` part is completely redundant.

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -0,0 +1,130 @@
...
+
+static sessions_testcase sessions_tests[] = {
+    {sample_log1, {"821f03288846297c2cf43c34766a38f7",
+		   "45,36,14,54,42",
+		   "45:0,36:0,14:0,54:2,42:0"}},
+    {sample_log2,{}}
+};
+
+static bool
+almost_equal(const double &x, const double &y,
```
ojwb: There's no need to pass fundamental types like `double` by reference.  On most processors a `double` can be passed in a register, so `const double &` could well be slower - unless the compiler optimises it away, it'll mean storing the value to memory and passing a pointer.

Also, there's no good reason to mark `epsilon` as `const` here either.

xapian-applications/omega/clickmodel/docs/clickmodel.rst
ojwb: I can see comments which haven't been addressed.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ojwb: Not addressed yet.

xapian-applications/omega/clickmodel/simplifieddbn.cc
ojwb: Not addressed yet.

xapian-applications/omega/clickmodel/docs/clickmodel.rst
ojwb: Not addressed yet.

xapian-applications/omega/clickmodel/simplifieddbn.cc
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    file.close();
+    return sessions;
+}
+
+void
+SimplifiedDBN::train(vector<vector<string>> &sessions)
+{
+    map<string, map<string, map<char, double[PARAM_COUNT_]>>> url_rel_fractions;
```
ojwb: No, I said to **get rid of** the `map<char, ...>` part.

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -0,0 +1,135 @@
...
+    SimplifiedDBN sdbn;
+
+    // Tests for SimplifiedDBN::build_sessions method.
+    for (size_t i = 0; i < sizeof(sessions_tests) / sizeof(sessions_tests[0]); ++i) {
+	vector<vector<string>> result;
+	try {
+	    result = sdbn.build_sessions(sessions_tests[i].logfile);
+	} catch (std::exception &ex) {
+	    cout << ex.what() << endl;
+	    continue;
```
ojwb: Now you're just reporting the exception, but otherwise ignoring it.  So now the exception won't cause the test to fail, which isn't what James suggested at all.

You need something like a `failure_count` variable, do `++failure_count` here, then at the end of the test you can check `failure_count` and if it's non-zero, report how many failures there were and  `exit(1)`.

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	for (auto&& x : result) {
+	    if (x[QID] != sessions_tests[i].sessions.qid) {
+		cerr << "Query ID mismatch occurred. "
+		     << "Expected: " << sessions_tests[i].sessions.qid
+		     << " Received: "<< x[QID] << endl;
+		exit(1);
```
ojwb: You're still bailing out on the first error here and below.

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -0,0 +1,135 @@
...
+	} catch (std::exception &ex) {
+	    cout << ex.what() << endl;
+	    continue;
+	}
+
+	for (auto&& x : result) {
+	    if (x[QID] != sessions_tests[i].sessions.qid) {
+		cerr << "Query ID mismatch occurred. "
+		     << "Expected: " << sessions_tests[i].sessions.qid
+		     << " Received: "<< x[QID] << endl;
```
ojwb: Missing space between `"` and `<<` above (and below).

xapian-applications/omega/clickmodel/simplifieddbn.cc
ivmarkp: Not sure if that can be done with enum that we have because here `a` and `s` chars need to be mapped to two different values which is why an array of size 2 is required here and it's not possible to declare arrays as enum elements. What do you suggest?

ivmarkp:
I think it was a mistake to use Github's interface to resolve merge conflicts in `Makefile.am`. I didn't expect it to merge 'master' altogether into this branch.

Although, it says "This merge commit was added into this branch cleanly." and there are no new changes added with the commit but it may cause problem during rolling up of fixup commits during the final merge.

Should I rollback locally to the last fixup commit and force push to erase the merge commit here?

xapian-applications/omega/generate-qrel-file.cc
```diff
@@ -0,0 +1,134 @@
...
+	auto docid = docids.begin();
+
+	while (rel != doc_relevances.end() || docid != docids.end()) {
+	    file_q << session[0] << ' ' << "Q0"
+	    << ' ' << *docid << ' ' << *rel << ' ' << endl;
+
+	    if (rel != doc_relevances.end())
+		++rel;
+	    if (docid != docids.end())
+		++docid;
```
jaylett: These conditionals should always be true, because you won't enter the while loop again otherwise. So this might actually be a `for` loop. (Although then we'd have to be confident that there's at least one docid & hence relevance within the session; so we'd need to make sure that sessions that were created by a search but where we never saw a click aren't returned by the SDBN system. There should be a test for that if so, and it should be documented behaviour.)

xapian-applications/omega/generate-qrel-file.cc
```diff
@@ -0,0 +1,134 @@
...
+    // Extract doc relevances and doc ids from each session and write
+    // to the qrel file in the required format.
+    for (auto&& session : sessions) {
+	doc_relevances = sdbn.get_predicted_relevances(session);
+	vector<string> docids = get_docid_list(session);
+
+	auto rel = doc_relevances.begin();
+	auto docid = docids.begin();
+
+	while (rel != doc_relevances.end() || docid != docids.end()) {
```
ivmarkp: The way `relevances` vector gets filled in `SimplifiedDBN::get_predicted_relevances` makes sure that the docids and relevances are in the same order. Also, there are tests in `sdbntest.cc` for the same function to actually verify that.

Although, I think to get both the docids and relevances from a single iterator, we can store them as strings of the form `"docid:relevance"` (similar to how clicks are stored in the final log file by `postprocess`) but that would incur additional overhead of parsing relevances from such strings whenever we want the relevances.

ivmarkp:
> I suspect doing some explicit modelling of sessions as objects will help a little with the SDBN implementation. (For instance you'd have `vector<Session>` rather than `vector<vector<string>>`, which is more readable and probably easier to work with when you look at iterators over docid+relevance rather than doing them separately as you currently have.)

Are you suggesting to declare a vector of strings named "Sessions" and use it in where we want to use sessions vector? E.g.

```
vector<string> Session;
vector<Session> training_sessions;

File: xapian-applications/omega/generate-qrel-file.cc (hunk 31 above)
Comment (jaylett) [inline xapian-applications/omega/generate-qrel-file.cc]: You could store them as `std::pair<string, double>`? Right now, you're exposing details of how you store and manage the data internally to the class, in the class's API, which isn't terribly good OO or API design.

**Comment by jaylett** [comment]
@ivmarkp I'm suggesting writing a _class_ called `Session`, so each session becomes an object that you can manipulate. Crucially, this means that someone looking at `vector<Session>` will know what it is. `vector<vector<string>>` gives no such indication of the semantics of your data.

File: xapian-applications/omega/docs/clickmodel.rst [hunk 32]
```diff
@@ -0,0 +1,75 @@
...
+=================================
+
+``SimplifiedDBN`` class provides ``get_predicted_relevances`` function to get
+the predicted relevance of all documents in a given search session based on
+the trained SDBN model.
+
+Use ``build_sessions`` function to generate a list of search sessions from an
+input log data file as described above and then pass any particular search
+session from the list for which you want the predicted relevances of the
+documents in that session to ``get_predicted_relevances``function which creates
```
jaylett: Missing space after the second pair of backticks. (If you look at this in github, it's pretty good at syntax highlighting rst and can help you track down this kind of thing — even if you're local editor does not. Building the documentation using `rst2html` or `rst2html.py` is also helpful in reviewing documentation.)

ivmarkp: Fixed it.

xapian-applications/omega/generate-qrel-file.cc
ojwb: Actually, they won't always be true as the `while` test uses `||`.  But that's broken as just above we unconditionally dereference both `docid` and `rel` when one or the other could be an end iterator.

Overall I agree that this all reflects a flawed design, and you should really get back a pair of the docid and its associated relevance.

xapian-applications/omega/generate-qrel-file.cc
```diff
@@ -0,0 +1,134 @@
...
+	    file_q << session[0] << ' ' << "Q0"
+	    << ' ' << *docid << ' ' << *rel << ' ' << endl;
```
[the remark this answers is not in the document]
ojwb: `" Q0 "` rather than `' ' << "Q0" << ' '` even...

Also, why is there a space before the end of the line?

xapian-applications/omega/generate-qrel-file.cc
ivmarkp: This was fixed in 5fe3674 with the suggested use of `std::pair`.

xapian-applications/omega/generate-qrel-file.cc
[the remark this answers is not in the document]
ivmarkp: Certainly, no need for that space. I'll remove that with a fixup commit and change `' ' << "Q0" << ' '` to `" Q0 "`.

ivmarkp:
@jaylett The utility command that you listed in the checklist above has been added as `generate-qrel-file` command. Can you mark that as done in the checklist? Also, if the PR mostly looks good now, I'd get started with resolving build errors on clang.

xapian-applications/omega/clickmodel/tests/sdbntest.cc
ivmarkp: @ojwb @jaylett Ping?

jaylett: What do you think? Have you tried copying over the code from the test driver? What other way could work, looking at the code?

ivmarkp: I haven't tried the test driver code yet but it looks like it might work. But I'm not sure if copying over all the code just for using a single function is a wise thing to do so just wanted to ask if there might be another way of getting the directory path.

jaylett: Just copy the single function? It's not ideal, but it's probably a reasonable trade-off.

ivmarkp: @jaylett I tried but it doesn't work. Is there anything else I could try? Also, build errors on clang and OSX have been fixed with the latest commit. So, the only thing remaining is resolving the standing issue with getting the correct path for test files so that `sdbntest` pass.

jaylett: @ivmarkp I can't help you if you just say "it doesn't work". Obviously you cannot copy just the single function on its own, which is currently entangled in the rest of the test driver. However, you should be able to copy and adapt the code, resolving its dependencies; or write a new function that achieves the same thing.

Looking at the function, it has few dependencies beyond the STL. It's using `startswith()` and `endswith()`, plus `file_exists()`, all of which are in our common utils library (which is available in the omega build system, but you may have to link to get it).

ivmarkp: Actually, I did use the function correctly after resolving all its dependencies that you have mentioned but the problem was that the `srcdir` environment variable returns just `.` in
```c++
char *p = getenv("srcdir");
```
which should really return the path `"/home/checksum/Documents/xapian"` as found to be set in `configure` there generated by `./bootstrap`.

jaylett: @ivmarkp have a look at how the `runtest` script in xapian-core does this. You may need a similar approach (it's generated from `runtest.in`) to pass through `srcdir` for the omega tests.

ojwb: > which should really return the path `"/home/checksum/Documents/xapian"`

There's no need for the path to `srcdir` to be absolute (and in some ways a relative path is better - for example, it's less likely to exceed path name limits if you are building deep in a directory tree; it also makes it easier to compare the output of two builds in different directories).

The reason the tests fail currently is simple and nothing to do with `srcdir`.  The problem is simply that `./sdbntest` is run from the `omega` directory in the build tree, but tries to open `test1.log` (etc) with no path specified.  However, `test1.log` lives in `clickmodel/testdata/`.

If you prepend `clickmodel/testdata/` to the log filenames, then the tests pass when building in the source directory.  To fix this fully such that this also works when `$srcdir != $builddir`, you ought to prepend the value of `$srcdir` followed by `"/clickmodel/testdata/"`.

ivmarkp: @ojwb Great, thanks! Tests pass now.

ivmarkp:
@ojwb Build is failing with the error:
```
./include/xapian/valuesetmatchdecider.h:30:10: fatal error: 'xapian/matchdecider.h' file not found
#include <xapian/matchdecider.h>
```
Has there been a change recently? I don't think build ever failed with this error before on this PR.

ojwb:
> ./include/xapian/valuesetmatchdecider.h:30:10: fatal error: 'xapian/matchdecider.h' file not found
#include <xapian/matchdecider.h>

Sorry, failed to `git add` a new file.  Fix pushed and CI rerunning.

ivmarkp:
Thanks for the fix. A couple unused variables in `matcher/deciderpostlist.h` caused the last build failure:
```
In file included from matcher/deciderpostlist.cc:23:
matcher/deciderpostlist.h:43:22: error: private field 'considered' is not used [-Werror,-Wunused-private-field]
    Xapian::doccount considered = 0;
                     ^
matcher/deciderpostlist.h:46:22: error: private field 'denied' is not used [-Werror,-Wunused-private-field]
    Xapian::doccount denied = 0;
```
Can you please push a fix for that as well?

ivmarkp:
> Can you please push a fix for that as well?

Sorry, I didn't notice before but you already fixed that issue.

ivmarkp:
@ojwb Build is currently failing on OSX with error:
```
Undefined symbols for architecture x86_64:
  "gnu_getopt_internal_(int, char* const*, char const*, option const*, int*, int)", referenced from:
      gnu_getopt_long(int, char* const*, char const*, option const*, int*) in generate-qrel-file.o
ld: symbol(s) not found for architecture x86_64
```
Seemingly, the compiler is unable to locate the definition `gnu_getopt_internal_` function which, I'm guessing, should be available via `gnu_getopt.h` file included in `generate-qrel-file`.

ojwb:
> Seemingly, the compiler is unable to locate the definition gnu_getopt_internal_ function which, I'm guessing, should be available via gnu_getopt.h file included in generate-qrel-file.

Including that header file means the compiler knows the function exists, but you're getting a linker error because you aren't linking in anything which actually implements the function.

This is because you've not included `common/getopt.cc` in `generate_qrel_file_SOURCES` in `Makefile.am`.

On Linux you won't see bug manifest because GNU getopt is present in glibc, and we use it from there when it is.  But on OS X (and most other platforms) we have to provide our own implementation.

ivmarkp:
> Including that header file means the compiler knows the function exists, but you're getting a linker error because you aren't linking in anything which actually implements the function. 

> This is because you've not included common/getopt.cc in generate_qrel_file_SOURCES in Makefile.am.

 I see. Thanks for the help.

ivmarkp:
@ojwb @jaylett Is this ready to be merged?

xapian-applications/omega/clickmodel/tests/sdbntest.cc
```diff
@@ -46,10 +46,15 @@ struct sessions_testcase {
...
-string sample_log3 = string(p) + "/clickmodel/testdata/test3.log";
+static string get_srcdir() {
+    char *p = getenv("srcdir");
+    if (!p) return ".";
+    return string(p);
+}
+
+string sample_log1 = get_srcdir() + "/clickmodel/testdata/test1.log";
+string sample_log2 = get_srcdir() + "/clickmodel/testdata/test2.log";
+string sample_log3 = get_srcdir() + "/clickmodel/testdata/test3.log";
```
ojwb: I wrote:

> It would also be much better to do this initialisation inside a function rather than as global static objects. It's hard to usefully handle exceptions or report errors during global static object initialisation - a good rule of thumb is to avoid it completely where you can, and keep it simple where you actually need it.

The change above isn't what I meant - I was suggesting to avoid unnecessarily doing work during global object initialisation, not just moving the `getenv()` call into its own function which still gets called during global object initialisation

What we're trying to avoid is doing any complex work at runtime before we enter `main()` - it's hard to usefully handle exceptions or report errors before `main()` - there's no enclosing scope to catch exceptions and I believe you can't even necessarily rely on things like `cout` working properly at this stage (because `cout` is also a global object, and the order of initialisation of global objects isn't specified, so your global object may get initialised before `cout` is initialised).

What I was proposing was to move this initialisation to the start of the `main()` function.

Also, your change means we're now calling `getenv("srcdir")`, etc multiple times - that'll be easier to avoid that if you do this at the start of `main()`, because you can easily have a temporary `srcdir` variable.

ivmarkp: Alright, I'll move the filepath initialisations inside the main function along with the `sessions_tests[]` struct array. Since, `sessions_tests[]` was declared globally earlier which uses the filepaths `sample_log1` and `sample_log2`, I had to make those initialisations global as well.

xapian-applications/omega/clickmodel/session.h
```diff
@@ -0,0 +1,57 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#include <string>
```
ojwb: Sorry, I missed this before but headers really should have guards to make them safe to include multiple times (the accepted wisdom is to make header inclusion idempotent rather than try to avoid this situation, which is often hard when it happens indirectly) - i.e. `#ifndef OMEGA_INCLUDED_SESSION_H` etc.

xapian-applications/omega/clickmodel/session.h
```diff
@@ -0,0 +1,57 @@
...
+#include <vector>
+
+using namespace std;
```
ojwb: It's not good to dump everything from namespace `std` into the global namespace like this in a header which IIUC is intended for users to include in their own code.  Better to explicitly qualify with `std::` in a header (`using namespace std;` is OK for stuff that's internal to a program or library - the only real issue there is that a new C++ standard revision may result in a new symbol in `std` which collides with a symbol in the program, but a new C++ standard version seems to require a few tweaks anyway.)

xapian-applications/omega/clickmodel/session.h
```diff
@@ -0,0 +1,57 @@
...
+	session.push_back(qid);
+	session.push_back(docids);
+	session.push_back(clicks);
+    }
+
+    /// Retrieve the query id of the session.
+    string get_qid() const { return session[QID]; }
+
+    /// Retrieve the docids string of the session.
+    string get_docids()	const { return session[DOCIDS]; }
```
ojwb: Random extra spaces before `const` here.

xapian-applications/omega/docs/clickmodel.rst
```diff
@@ -0,0 +1,75 @@
...
+===============================================
+
+.. contents:: Table of Contents
+
+Introduction
+============
+
+Clicks in web search reflect user interests in retrieved documents and by the
+use of various methods, clicks can be used to improve search results. Click
+models are one such methods. These are probabilistic graphical models that
```
ojwb: "one such **method**"

ivmarkp:
@ojwb Thanks for taking a look. Have addressed your comments.

ivmarkp:
@ojwb @jaylett Hi, its been awhile. I was wondering if this is ready to be merged?

jaylett:
@ivmarkp no, because the tests are red on this PR. The first thing I'd try (unless you can reproduce the test failures) is to rebase to master and force push the branch, so the commits remain the same but you get it rebuilt (and against the latest tests &c in case any have been tweaked since you branched).

ivmarkp:
@jaylett There were some failing PL2 and PL2+ tests which now pass locally after rebase. Travis build should go fine now.

jaylett:
@ivmarkp I think all comments have now been addressed, by @ojwb may wish to do a final check.

ivmarkp:
Thanks @jaylett.

@ojwb Can you please check this when you get a chance?

ojwb:
Thanks - rebased, squashed and merged.

I've pushed a couple of follow-up commits:

* dc00d8bc354fbe2b894766ab5b474e275643735d drops `clickmodel/simplifieddbn.cc` from `omega_SOURCES` - this code shouldn't be linked into the omega CGI binary
* 8dc973cf4026fef01dc1d69e188dab0db9477993 fixes some whitespace oddities

PR #171, opened 2017-07-25
Fix memory leak in SVMRanker

AyushP123:
Introduced a new function and made appropriate calls to deallocate the
resources to eliminate memory leak.

### DIFF
--- a/xapian-letor/ranker/svmranker.cc
+++ b/xapian-letor/ranker/svmranker.cc
@@ -46,14 +46,12 @@ using namespace std;
 using namespace Xapian;
 
 static void
-clear_svm_problem(svm_problem problem, int n)
+clear_svm_problem(svm_problem problem)
 {
     delete [] problem.y;
     problem.y = NULL;
-    for (int i = 0; i < n;i++)
-    {
+    for (int i = 0; i < problem.l; i++) {
 	delete [] problem.x[i];
-	problem.x[i] = NULL;
     }
     delete [] problem.x;
     problem.x = NULL;
--- a/xapian-letor/ranker/svmranker.cc
+++ b/xapian-letor/ranker/svmranker.cc
@@ -46,15 +46,15 @@ using namespace std;
 using namespace Xapian;
 
 static void
-clear_svm_problem(svm_problem problem)
+clear_svm_problem(svm_problem *problem)
 {
-    delete [] problem.y;
-    problem.y = NULL;
-    for (int i = 0; i < problem.l; i++) {
-	delete [] problem.x[i];
+    delete [] problem->y;
+    problem->y = NULL;
+    for (int i = 0; i < problem->l; i++) {
+	delete [] problem->x[i];
     }
-    delete [] problem.x;
-    problem.x = NULL;
+    delete [] problem->x;
+    problem->x = NULL;
 }
 
 SVMRanker::SVMRanker()
@@ -156,7 +156,7 @@ SVMRanker::train(const std::vector<Xapian::FeatureVector> & training_data)
 	std::remove(templ);
     }
     svm_free_and_destroy_model(&trainmodel);
-    clear_svm_problem(prob, fvv_len);
+    clear_svm_problem(&prob);
     svm_destroy_param(&param);
     if (this->model_data.empty()) {
 	throw LetorInternalError("SVM model empty. Training failed.");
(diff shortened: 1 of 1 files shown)


xapian-letor/ranker/svmranker.cc
```diff
@@ -45,6 +45,20 @@ Proceedings of the eighth ACM SIGKDD international conference on Knowledge disco
 using namespace std;
 using namespace Xapian;
 
+static void
+clear_svm_problem(svm_problem problem, int n)
```
ojwb: This passes `problem` by value, so will copy it - it's a 3 member structure, so that's not awful, but it would really make more sense to pass it by reference.

Isn't `n` redundant information?  It seems you could just use `problem.l` instead, which would make it harder to call this function incorrectly.

xapian-letor/ranker/svmranker.cc
```diff
@@ -45,6 +45,20 @@ Proceedings of the eighth ACM SIGKDD international conference on Knowledge disco
...
+{
+    delete [] problem.y;
+    problem.y = NULL;
+    for (int i = 0; i < n;i++)
+    {
```
ojwb: Should be a space after `;` and no newline before `{`.

xapian-letor/ranker/svmranker.cc
```diff
@@ -45,6 +45,20 @@ Proceedings of the eighth ACM SIGKDD international conference on Knowledge disco
...
+	delete [] problem.x[i];
+	problem.x[i] = NULL;
```
ojwb: This is redundant, as you go on to delete `x` itself just below.

xapian-letor/ranker/svmranker.cc
AyushP123: I think I have resolved this in the latest commit

xapian-letor/ranker/svmranker.cc
AyushP123: I don't know how did i pass xapian-check-patch test on travis C.I after having i++

samuelharden: probably best to run xapian-check-patch before committing changes. 
I have also been trying to form a habit of running Valgrind before pushing a new feature. That helps in not introducing a memory leak.

AyushP123: Yes .... this change was quite small .... so I decided to review it manually .... it shows the fundamental wrong in my thinking while writing code .... I should have actually by now adapted to using xapian conventions like having a { on the same line as for, if and while and using prefix increment instead of postfix .... but I always do rush in pushing commits and keep making these mistakes for which I again apologize

ojwb: > I don't know how did i pass xapian-check-patch test on travis C.I after having i++

The script isn't infallible - it's essentially a lot of regular expression checks, and certainly not a full C++ parser (to use a C++ parser, I think we'd need to work from the source file rather than just a patch, as the patch has limited context - I've been wondering if that would be a better approach, but for now we have what we have).

Anyway. the current version of this particular check was defeated by the missing space before `i++`.  I'll push a fix for that, and also add a check so it catches the missing space after `;`.

> I always do rush in pushing commits

Don't rush to commit!  That's really the root of the problem here.

Taking your time checking things over takes a few minutes for a patch of this size.  Pushing and seeing the CI patch check fail takes longer, and also is really abusing the CI resources which are provided for free.  An extra review cycle by the maintainer takes hours or days in this context, and for many projects it can be weeks or months - I'm not exaggerating here, some examples from looking at my github PR history:

* 7 months: https://github.com/formorer/debian-indexer-xapian/pull/4
* 8 months: https://github.com/therion/therion/pull/1
* 18 months: https://github.com/dagwieers/unoconv/pull/174
* 22 months: https://github.com/madler/zlib/pull/85

Luckily I didn't mess up on any of those!

Admittedly those are some of the outliers from a fairly large set, but weeks is certainly common.

Most maintainers like bugs in their code to be fixed for them and would love more developers to contribute to their projects, so there's a clear reason to review and merge patches.  But in many cases, they're struggling to find time to review all the patches promptly.  If you submit clean patches they can just merge right away you're likely to get reviewed more promptly than people submitting scrappy patches which need many iterations.  If you can show you're a safe pair of hands, you'll often get the ability to just push changes directly without a review up front.  But if you don't address all the comments before pushing a new version, or you make the same mistakes time after time, you'll tend to just drift down the review pile.

ojwb:
Auto-squashed and pushed as 659c87f60bba55dc9709e128de436e8be037f711.

PR #172, opened 2017-07-25
Eliminate default name for training file

AyushP123:


ojwb:
`git grep 'training-data.txt'` shows another method with this filename as a default parameter value, plus two documentation comments that need updating...

ojwb:
Thanks, autosquashed and merged to master.

PR #174, opened 2017-07-26
Fix TermGenerator do not stop stemmed term.

samuelharden:
From ticket 
"It seems that the TermGenerator? (with STOP_ALL and STEM_ALL strategy) do not stop stemmed term.For example, with french stemmer and "le" in the stopwords, le term "lea" will be stemmed to "le" and "le" term will be added to the document.Looking into termgenerator_internal.cc (method index_text), it seems that the stopper in never called on the stem, whatever the flags are.

We are using version 1.4.2 but the stopper is never called on the stem even on git master.
"
Fixed the issue and added a test case to fix this issue.


ojwb:
Thanks for the patch, but this isn't the right way to address this - it's really the documentation that's buggy (see the ticket for a detailed explanation).

samuelharden:
Thanks for the detailed explanation :+1:

PR #175, opened 2017-07-29
Add ListMLE ranker to letor

AyushP123:
Added listmle_ranker.cc along with a test for listmle_ranker to letor.

Have added listmle_ranker.cc from vhasu's 2014 GSoC repository ->
https://github.com/v-hasu/xapian/tree/gsoc2014/xapian-letor
 after making suitable changes to account for the elimination of
Xapian::Ranklist and changes made to Xapian::FeatureVector. Have also
added a standard test to test all the features of ListMLERanker.

### DIFF
--- a/xapian-letor/include/xapian-letor/ranker.h
+++ b/xapian-letor/include/xapian-letor/ranker.h
@@ -354,7 +354,7 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
 
 };
 
-class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
+class XAPIAN_VISIBILITY_DEFAULT ListMLERanker : public Ranker {
     /// Ranker parameters
     std::vector<double> parameters;
     /// Learning rate (Default is 0.001)
@@ -367,11 +367,11 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
      * @exception InvalidArgumentError will be thrown if training_data is
      *            empty.
      */
-    void train(const std::vector<Xapian::FeatureVector> & training_data);
+    void train(const std::vector<Xapian::FeatureVector>& training_data);
 
     /** Save ListMLE model as db metadata.
      *
-     *  ListMLE model file gets stored with each parameter value in a new line.
+     *  ListMLE model gets stored with each parameter value in a new line.
      *	e.g.
      *
      *  0.000920817564536697
@@ -381,28 +381,28 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
      *
      *  @param model_key	Metadata key using which model is to be stored.
      */
-    void save_model_to_metadata(const std::string & model_key);
+    void save_model_to_metadata(const std::string& model_key);
 
-    /** Load model from an external file.
+    /** Load model from db metadata.
      *
      *  @param model_key         Metadata key using which model is to be
      *				 loaded.
      *
-     *  @exception LetorInternalError will be thrown if no model exists
+     *  @exception InvalidArgumentError will be thrown if no model exists
      *  	   corresponding to the supplied key
      */
-    void load_model_from_metadata(const std::string & model_key);
+    void load_model_from_metadata(const std::string& model_key);
 
     /** Re-rank a std::vector<Xapian::FeatureVector> by using the
      *  model.
      *
      *  @param fvv	vector<FeatureVector> that will be re-ranked.
      *
-     *  @exception	LetorInternalError will be thrown if model file
+     *  @exception	InvalidArgumentError will be thrown if model file
      *			is not compatible.
      */
     std::vector<Xapian::FeatureVector>
-    rank_fvv(const std::vector<Xapian::FeatureVector> & fvv) const;
+    rank_fvv(const std::vector<Xapian::FeatureVector>& fvv) const;
 
   public:
     /* Construct ListMLE instance
@@ -410,8 +410,8 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
      * @param num_interations  Number of iterations (Default is 10)
      */
     explicit ListMLERanker(double learn_rate = 0.001,
-			    int num_interations = 10):
-	 learning_rate(learn_rate), iterations(num_interations) { }
+			    int num_interations = 10)
+	: learning_rate(learn_rate), iterations(num_interations) { }
 
     /// Destructor
     ~ListMLERanker();
--- a/xapian-letor/include/xapian-letor/ranker.h
+++ b/xapian-letor/include/xapian-letor/ranker.h
@@ -371,7 +371,8 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker : public Ranker {
 
     /** Save ListMLE model as db metadata.
      *
-     *  ListMLE model gets stored with each parameter value in a new line.
+     *  ListMLE model gets stored with serialised value of each parameter in
+     *  a new line.
      *	e.g.
      *
      *  0.000920817564536697
--- a/xapian-letor/include/xapian-letor/ranker.h
+++ b/xapian-letor/include/xapian-letor/ranker.h
@@ -354,6 +354,67 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
 
 };
 
+class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
+    /// Ranker parameters
+    std::vector<double> parameters;
+    /// Learning rate (Default is 0.001)
+    double learning_rate;
+    /// Number of iterations (Default is 10)
+    int iterations;
+
+    /** Method to train the model.
+     *
+     * @exception LetorInternalError will be thrown if training data is null.
+     */
+    void train(const std::vector<Xapian::FeatureVector> & training_data);
+
+    /** Method to save ListMLE model as db metadata.
+     *
+     *  ListMLE model file gets stored with each parameter value in a new line.
+     *	 e.g.
+     *
+     *  0.000920817564536697
+     *  0.000920817564536697
+     *  0
+     *  -1.66533453693773e-19
+     *
+     *  @param model_key	Metadata key using which model is to be stored.
+     */
+    void save_model_to_metadata(const std::string & model_key);
+
+    /** Method to load model from an external file.
+     *
+     *  @param model_key         Metadata key using which model is to be
+     *				 loaded.
+     *
+     *  @exception LetorInternalError will be thrown if no model exists
+     *  	   corresponding to the supplied key
+     */
+    void load_model_from_metadata(const std::string & model_key);
+
+    /** Method to re-rank a std::vector<Xapian::FeatureVector> by using the
+     *  model.
+     *
+     *  @param fvv	vector<FeatureVector> that will be re-ranked.
+     *
+     *  @exception	LetorInternalError will be thrown if model file
+     *			is not compatible.
+     */
+    std::vector<Xapian::FeatureVector>
+    rank_fvv(const std::vector<Xapian::FeatureVector> & fvv) const;
+
+  public:
+    /* Construct ListMLE instance
+     * @param learn_rate       Learning rate
+     * @param num_interations  Number of iterations
+     */
+    explicit ListMLERanker(double learn_rate = 0.001,
+			    int num_interations = 10):
+	 learning_rate(learn_rate), iterations(num_interations) { }
+
+    /// Destructor
+    ~ListMLERanker();
+};
 }
 
 #endif /* RANKER_H */
--- a/xapian-letor/include/xapian-letor/ranker.h
+++ b/xapian-letor/include/xapian-letor/ranker.h
@@ -362,16 +362,17 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
     /// Number of iterations (Default is 10)
     int iterations;
 
-    /** Method to train the model.
+    /** Train the model.
      *
-     * @exception LetorInternalError will be thrown if training data is null.
+     * @exception InvalidArgumentError will be thrown if training_data is
+     *            empty.
      */
     void train(const std::vector<Xapian::FeatureVector> & training_data);
 
-    /** Method to save ListMLE model as db metadata.
+    /** Save ListMLE model as db metadata.
      *
      *  ListMLE model file gets stored with each parameter value in a new line.
-     *	 e.g.
+     *	e.g.
      *
      *  0.000920817564536697
      *  0.000920817564536697
@@ -382,7 +383,7 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
      */
     void save_model_to_metadata(const std::string & model_key);
 
-    /** Method to load model from an external file.
+    /** Load model from an external file.
      *
      *  @param model_key         Metadata key using which model is to be
      *				 loaded.
@@ -392,7 +393,7 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
      */
     void load_model_from_metadata(const std::string & model_key);
 
-    /** Method to re-rank a std::vector<Xapian::FeatureVector> by using the
+    /** Re-rank a std::vector<Xapian::FeatureVector> by using the
      *  model.
      *
      *  @param fvv	vector<FeatureVector> that will be re-ranked.
@@ -405,8 +406,8 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
 
   public:
     /* Construct ListMLE instance
-     * @param learn_rate       Learning rate
-     * @param num_interations  Number of iterations
+     * @param learn_rate       Learning rate (Default is 0.001)
+     * @param num_interations  Number of iterations (Default is 10)
      */
     explicit ListMLERanker(double learn_rate = 0.001,
 			    int num_interations = 10):
--- a/xapian-letor/include/xapian-letor/ranker.h
+++ b/xapian-letor/include/xapian-letor/ranker.h
@@ -371,14 +371,7 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker : public Ranker {
 
     /** Save ListMLE model as db metadata.
      *
-     *  ListMLE model gets stored with serialised value of each parameter in
-     *  a new line.
-     *	e.g.
-     *
-     *  0.000920817564536697
-     *  0.000920817564536697
-     *  0
-     *  -1.66533453693773e-19
+     *  The model is serialised as a binary blob.
      *
      *  @param model_key	Metadata key using which model is to be stored.
      */
--- a/xapian-letor/ranker/listmle_ranker.cc
+++ b/xapian-letor/ranker/listmle_ranker.cc
@@ -44,19 +44,22 @@ et al. "Listwise Approach to Learning to Rank - Theory and Algorithm."
 using namespace std;
 using namespace Xapian;
 
-ListMLERanker::~ListMLERanker() {
+ListMLERanker::~ListMLERanker()
+{
     LOGCALL_DTOR(API, "ListMLERanker");
 }
 
-static bool labelComparer(const FeatureVector & firstfv,
-			  const FeatureVector& secondfv) {
+static bool
+label_comparer(const FeatureVector& firstfv,
+	       const FeatureVector& secondfv)
+{
     return firstfv.get_label() > secondfv.get_label();
 }
 
 static double
-calculateInnerProduct(vector<double> & parameters,
-		      vector<double> feature_sets) {
-
+calculate_inner_product(const vector<double>& parameters,
+			const vector<double>& feature_sets)
+{
     double inner_product = 0.0;
     for (size_t i = 0; i < parameters.size(); ++i) {
 	inner_product += parameters[i] * feature_sets[i];
@@ -65,9 +68,9 @@ calculateInnerProduct(vector<double> & parameters,
 }
 
 static vector<double>
-calculateGradient(vector<FeatureVector> & sorted_feature_vectors,
-		  vector<double> & new_parameters) {
-
+calculate_gradient(const vector<FeatureVector>& sorted_feature_vectors,
+		   const vector<double>& new_parameters)
+{
     //need to be optimized,how to get the length of the features
     vector<double> gradient(sorted_feature_vectors[0].get_fcount(), 0);
 
@@ -77,7 +80,7 @@ calculateGradient(vector<FeatureVector> & sorted_feature_vectors,
     double expsum = 0.0;
 
     for (size_t i = 0; i < list_length; ++i) {
-	double temp = exp(calculateInnerProduct(
+	double temp = exp(calculate_inner_product(
 			      new_parameters,
 			      sorted_feature_vectors[i].get_fvals()));
 	exponents.push_back(temp);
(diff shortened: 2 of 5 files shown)


xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,67 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+    /// Ranker parameters
+    std::vector<double> parameters;
+    /// Learning rate (Default is 0.001)
+    double learning_rate;
+    /// Number of iterations (Default is 10)
+    int iterations;
+
+    /** Method to train the model.
+     *
+     * @exception LetorInternalError will be thrown if training data is null.
```
ojwb: That sounds like `InvalidArgumentError` not `LetorInternalError`.

Also, the meaning of "is null" isn't clear.  It can't be "is a NULL pointer", as a reference can't be NULL.  Does it mean "if `training_data.empty()`"?  If so, it'd be better to be explicit and say that.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,67 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+     */
+    void train(const std::vector<Xapian::FeatureVector> & training_data);
+
+    /** Method to save ListMLE model as db metadata.
```
ojwb: Don't document a method as "Method to save ListMLE model ...".  We know it's a method, better to just say "Save ListMLE model".

Same goes for all the other "Method to X".

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,67 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+     *  @exception	LetorInternalError will be thrown if model file
+     *			is not compatible.
+     */
+    std::vector<Xapian::FeatureVector>
+    rank_fvv(const std::vector<Xapian::FeatureVector> & fvv) const;
+
+  public:
+    /* Construct ListMLE instance
+     * @param learn_rate       Learning rate
+     * @param num_interations  Number of iterations
```
ojwb: Default values of parameters ought to be documented.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,199 @@
...
+using namespace Xapian;
+
+ListMLERanker::~ListMLERanker() {
+    LOGCALL_DTOR(API, "ListMLERanker");
+}
+
+static bool labelComparer(const FeatureVector & firstfv,
+			  const FeatureVector& secondfv) {
+    return firstfv.get_label() > secondfv.get_label();
+}
```
ojwb: Should be a newline before the function name, and one before the `{`.

Be consistent about spaces around `&`, particularly with a parameter list.  `const FeatureVector&` is the more "C++" way, so probably best to use that.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,199 @@
...
+
+static double
+calculateInnerProduct(vector<double> & parameters,
+		      vector<double> feature_sets) {
```
ojwb: Newline before `{`.  Please check for and fix any further instances.

`parameters` isn't modified, so should be `const`.

`feature_sets` should be passed by const reference - as-is, it'll get copied.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,199 @@
...
+#include <stdlib.h>
+
+using namespace std;
+using namespace Xapian;
+
+ListMLERanker::~ListMLERanker() {
+    LOGCALL_DTOR(API, "ListMLERanker");
+}
+
+static bool labelComparer(const FeatureVector & firstfv,
```
ojwb: Xapian coding style is that functions and methods should be lower case with `_` between words, not "camel case" like here (only class names use camel case).

There seems to be a lot of instances of this problem here.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    vector<double> first_place_in_ground_truth_feature_sets =
+	    sorted_feature_vectors[0].get_fvals();
+
+    for (size_t i = 0; i < first_place_in_ground_truth_feature_sets.size() - 1;
+	 ++i) {
+	gradient[i] -= first_place_in_ground_truth_feature_sets[i];
+
+    }
```
ojwb: Unwanted random blank line at end of block.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,68 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
 
 };
 
+class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
```
jaylett: Put a space either side of the `:` in class inheritance. (If the line gets too long, then wrap before the `:`; see `xapian-core/include/xapian/cluster.h` for examples of both styles).

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,68 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+    /** Train the model.
+     *
+     * @exception InvalidArgumentError will be thrown if training_data is
+     *            empty.
+     */
+    void train(const std::vector<Xapian::FeatureVector> & training_data);
+
+    /** Save ListMLE model as db metadata.
+     *
+     *  ListMLE model file gets stored with each parameter value in a new line.
```
jaylett: It's no longer a model file, so this comment needs updating. Just "The model gets stored with…" would be fine.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,68 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+    std::vector<Xapian::FeatureVector>
+    rank_fvv(const std::vector<Xapian::FeatureVector> & fvv) const;
+
+  public:
+    /* Construct ListMLE instance
+     * @param learn_rate       Learning rate (Default is 0.001)
+     * @param num_interations  Number of iterations (Default is 10)
+     */
+    explicit ListMLERanker(double learn_rate = 0.001,
+			    int num_interations = 10):
```
jaylett: As with class inheritance, I'd split before the `:` for the constructor initializers.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,211 @@
...
+
+    size_t list_length = sorted_feature_vectors.size();
+
+    vector<double> exponents;
+    double expsum = 0.0;
+
+    for (size_t i = 0; i < list_length; ++i) {
+	double temp = exp(calculate_inner_product(
+			      new_parameters,
+			      sorted_feature_vectors[i].get_fvals()));
```
jaylett: This kind of line wrapping doesn't come up very often, so I'm not 100% sure if we even have a consistent style, but I'd put the indented lines at lot less indented, possibly only 4 characters in (ie one indent) from the "parent" line. (You can then put `new_parameters` and the `sorted_feature_vectors…` expression on the same line.)

As it stands, scanning down the file my brain comes to an abrupt stop because of the enormous indentation gap, rather than reading it as a single (albeit nested) function call.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,211 @@
...
+						new_parameters);
+    update_parameters(new_parameters, gradient, learning_rate);
+}
+
+void
+ListMLERanker::train(const vector<FeatureVector>& training_data)
+{
+    LOGCALL_VOID(API, "ListMLERanker::train", training_data);
+    if (training_data.empty()) {
+	throw Xapian::InvalidArgumentError("Training data is empty."
```
jaylett: There needs to be a space between the two sentences in the final string. However, I don't think we need to call out the training _file_ here (because in future we might train from something other than a file — this class only knows it was given an empty training data vector). So maybe just something like `"Cannot train: no training data"` would be sufficient.

xapian-letor/tests/api_letor.cc
```diff
@@ -182,6 +182,48 @@ DEFINE_TESTCASE(svm_ranker, generated)
...
+    string db_path = get_database_path("apitest_listmle_ranker",
+				       db_index_two_documents);
+    Xapian::Enquire enquire((Xapian::Database(db_path)));
+    enquire.set_query(Xapian::Query("lions"));
+    Xapian::MSet mymset = enquire.get_mset(0, 10);
+    string data_directory = test_driver::get_srcdir() + "/testdata/";
+    string query = data_directory + "query.txt";
+    string qrel = data_directory + "qrel.txt";
+    Xapian::prepare_training_file(db_path, query, qrel, 10,
+				  "listmle_ranker_training.txt");
```
jaylett: The same comment would apply about the `filename` argument to `::prepare_training_file()` as to the other path arguments: they should be relative to `test_driver::get_srcdir()`. This change should be made in this PR. (Even if we're going to rework this in a future change.)

xapian-letor/tests/api_letor.cc
```diff
@@ -182,6 +182,48 @@ DEFINE_TESTCASE(svm_ranker, generated)
...
+    TEST_EQUAL(doc2, *mymset[0]);
+    TEST_EQUAL(doc1, *mymset[1]);
+    mymset = enquire.get_mset(0, 10);
+    ranker.train_model("listmle_ranker_training.txt", "ListMLE_Ranker");
+    ranker.rank(mymset, "ListMLE_Ranker");
+    TEST_EQUAL(doc2, *mymset[0]);
+    TEST_EQUAL(doc1, *mymset[1]);
+    TEST_EXCEPTION(Xapian::LetorInternalError,
+		   ranker.score(query, qrel, "ListMLE_Ranker",
+				"scorer_output.txt", 10, ""));
```
jaylett: I'm not 100% certain what `Ranker::score()` is actually used for, so hopefully you can explain that in the next step when you add more detailed usage documentation. (This doesn't require any changes to this PR.)

xapian-letor/ranker/listmle_ranker.cc
ojwb: Formatting as James suggests is probably most consistent with existing code, but actually I'd probably write it differently to avoid the issue:

```
        const auto& fvals = sorted_feature_vectors[i].get_fvals();
        double temp = exp(calculate_inner_product(new_parameters, fvals));
```

Though I'd also pick a more descriptive variable name than `temp`, perhaps `exponent`.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -97,15 +100,16 @@ calculateGradient(vector<FeatureVector> & sorted_feature_vectors,
...
 
     return gradient;
 }
 
 static void
-updateParameters(vector<double> & new_parameters, vector<double> & gradient,
-		 double & learning_rate) {
+update_parameters(vector<double>& new_parameters,
+		  const vector<double>& gradient,
+		  const double& learning_rate)
```
ojwb: Passing a primitive type by `const` reference is rarely sensible.  A `double` can fit in a CPU register on any modern CPU, so copying it should be cheap.  Similarly below.

(Passing it by a non-`const` reference can be useful if you want to return a value via a parameter, but we don't here).

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,68 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+     *  0.000920817564536697
+     *  0.000920817564536697
+     *  0
+     *  -1.66533453693773e-19
+     *
+     *  @param model_key	Metadata key using which model is to be stored.
+     */
+    void save_model_to_metadata(const std::string & model_key);
+
+    /** Load model from an external file.
```
ojwb: As with the method above, this no longer involves a file AIUI.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,68 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+     *
+     *  @param model_key         Metadata key using which model is to be
+     *				 loaded.
+     *
+     *  @exception LetorInternalError will be thrown if no model exists
+     *  	   corresponding to the supplied key
```
ojwb: `InvalidArgumentError` seems more appropriate here (and below).

The documentation of `LetorInternalError` is rather vague as to its intended purpose ("LetorInternalError indicates a runtime problem of some sort"), but given the name I'd expect it to be for things like unexpected errors from libsvm.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,211 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
+/*
+ListMLE is adapted from the paper:
+Fen Xia, Tie-Yan Liu, Jue Wang, Wensheng Zhang, Hang Li
+et al. "Listwise Approach to Learning to Rank - Theory and Algorithm."
+*/
+#include <config.h>
```
ojwb: Xapian convention is to prefix lines in a multi-line comment like the above with `* ` (like the licensing boilerplate comment above it).

But this would be better located in the doc comment for the file (at the very start of the file), as a paragraph after the `@brief` line).

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,211 @@
...
+    Database letor_db(get_database_path());
+    string key = model_key;
+    if (key.empty()) {
+	key = "ListMLE.model.default";
+    }
+    string model_data = letor_db.get_metadata(key);
+    // Throw exception if no model data associated with key
+    if (model_data.empty()) {
+	throw LetorInternalError("No model found. Check key.");
+    }
```
ojwb: This comment seems to merely paraphrase the surrounding code, without adding any additional insights.  I would avoid such comments - it's not uncommon to find such comments which don't match what the code actually does.

"Why" or "how" are often useful things to explain in comments.

But "what" generally isn't unless it's pointing out something not clear from the code.  A "what" comment can actually be a sign that the code would benefit from being written in a clearer way.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,211 @@
...
+ListMLERanker::train(const vector<FeatureVector>& training_data)
+{
+    LOGCALL_VOID(API, "ListMLERanker::train", training_data);
+    if (training_data.empty()) {
+	throw Xapian::InvalidArgumentError("Training data is empty."
+					   "Check the training file");
+    }
+    int feature_cnt = training_data[0].get_fcount();
+
+    //initialize the parameters for neural network
```
ojwb: For readability (and to follow Xapian coding conventions) put a space after the comment characters, and make the comment a sentence (capital letter and full stop):

```
    // Initialize the parameters for neural network.
```

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,211 @@
...
+ */
+static void
+batch_learning(vector<FeatureVector> feature_vectors,
+	       vector<double> & new_parameters,
+	       const double& learning_rate)
+{
+    sort(feature_vectors.begin(), feature_vectors.end(), label_comparer);
+    vector<double> gradient = calculate_gradient(feature_vectors,
+						new_parameters);
+    update_parameters(new_parameters, gradient, learning_rate);
```
ojwb: `const auto& gradient = ...` would help the compiler avoid having to copy the vector.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,211 @@
...
+}
+
+std::vector<FeatureVector>
+ListMLERanker::rank_fvv(const std::vector<FeatureVector>& fvv) const
+{
+    LOGCALL(API, std::vector<FeatureVector>, "ListMLERanker::rank_fvv", fvv);
+    std::vector<FeatureVector> testfvv = fvv;
+    for (size_t i = 0; i < testfvv.size(); ++i) {
+	double listmle_score = 0;
+	std::vector<double> fvals = testfvv[i].get_fvals();
```
ojwb: Similarly `const auto& fvals = ...` would help avoid a vector copy.

xapian-letor/ranker/listmle_ranker.cc
AyushP123: I am not sure whether I have resolved this properly as I have introduced two blank lines .... which I am not sure should be there

xapian-letor/ranker/listmle_ranker.cc
AyushP123: I have written a new string .... please let me know if it should be better

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -381,37 +381,37 @@ class XAPIAN_VISIBILITY_DEFAULT ListMLERanker: public Ranker {
      *
```
ojwb: This isn't how the model is saved - we actually use `Xapian::serialise_double()` (which gives a much more compact result, and can round-trip values exactly, which is harder to reliably achieve by writing out values as text strings because the same real values aren't exactly representable in base 2 and base 10).

I don't think this is the right place to document exactly how the model is serialised though (and such documentation probably needs to go into more detail than just telling us it's a list of parameters).  API users don't need to worry about how the model is serialised, because they shouldn't need to try to unpick the data - they'll use the xapian-letor API to unserialise it.

Same issue with the doc comment for the same method of `ListNetRanker`.

xapian-letor/ranker/listmle_ranker.cc
ojwb: The citation should be in the same comment, or else it will just be a comment rather than part of the documentation.

For an example of a similar situation, see xapian-core/api/editdistance.cc which appears in the collated source documentation as:

https://xapian.org/docs/sourcedoc/html/editdistance_8cc.html#details

xapian-letor/ranker/listmle_ranker.cc
ojwb: I wasn't talking about the exception string.  This is about **comments**, in particular the one in the code I commented on:

```
// Throw exception if no model data associated with key
```

The code following the comment clearly throws an exception if `model_data.empty()`, and `model_data` was read using key `key` on the line before the comment.  So what does this comment tell the reader that isn't clear from reading the code?

xapian-letor/tests/api_letor.cc
```diff
@@ -194,19 +194,20 @@ DEFINE_TESTCASE(listmle_ranker, generated)
     string data_directory = test_driver::get_srcdir() + "/testdata/";
     string query = data_directory + "query.txt";
     string qrel = data_directory + "qrel.txt";
-    Xapian::prepare_training_file(db_path, query, qrel, 10,
-				  "listmle_ranker_training.txt");
+    string training = test_driver::get_srcdir()
+	    + "listmle_ranker_training.txt";
```
ojwb: Xapian convention is to break a line after an operator, not before (so `+` should be at the end of the first line, not the start of the second).

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -71,7 +72,7 @@ static vector<double>
 calculate_gradient(const vector<FeatureVector>& sorted_feature_vectors,
 		   const vector<double>& new_parameters)
 {
-    //need to be optimized,how to get the length of the features
+    // Need to be optimized,how to get the length of the features
```
ojwb: This sort of "todo note" comment is better marked with a "FIXME" tag.  These are easy to search for (e.g. `git grep FIXME`) and stand out more when reading the code (some editors even colour the "FIXME" text specially, but just as a all-caps word it tends to stand out).

`xapian-check-patch` will warn about adding a "FIXME", but that's a warning not an error and the CI won't fail because of it.  The reason for it is help ensure that "FIXME" notes added during development with the intention of addressing them before pushing the patch haven't just been overlooked (long-term FIXMEs are OK, things you meant to fix as part of development aren't).

I'm not clear what this note is actually about though - the implementation of `get_fcount()` just seems to be a call to `std::vector::size()`, which is a cheap operation.  How could this be optimised?

xapian-letor/ranker/listmle_ranker.cc
AyushP123: I have added the citation to the @brief comment

xapian-letor/ranker/listmle_ranker.cc
AyushP123: I did not pay attention to this comment .... have eliminated it

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -354,6 +354,69 @@ class XAPIAN_VISIBILITY_DEFAULT SVMRanker: public Ranker {
...
+     *
+     * @exception InvalidArgumentError will be thrown if training_data is
+     *            empty.
+     */
+    void train(const std::vector<Xapian::FeatureVector>& training_data);
+
+    /** Save ListMLE model as db metadata.
+     *
+     *  ListMLE model gets stored with serialised value of each parameter in
+     *  a new line.
```
ojwb: You've just done `s/each parameter value/serialised value of each parameter/`here, which means it's now not correct for either how the code used to be, nor how it is now - there are no newlines between serialised values (as you can clearly see if you look at the implementation).

And the example still shows how this might have been before we changed to using `serialise_double()`.

As I said, I don't think details of the serialisation belong here, but we definitely don't want **incorrect** details of the serialisation here.

I'd just say "The model is serialised as a binary blob."  That tells the API user that printing it out isn't going to be useful, or at least if they do, then seeing binary data doesn't indicate a problem.  Beyond that, they really don't need to worry about the format of the serialisation.  By analogy, we don't document the format of the serialisation returned by e.g. `Document::serialise()` in its doc comment.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -0,0 +1,207 @@
...
+
+#include "xapian-letor/ranker.h"
+
+#include "debuglog.h"
+#include "serialise-double.h"
+
+#include <xapian.h>
+#include <xapian/intrusive_ptr.h>
+#include <xapian/types.h>
+#include <xapian/visibility.h>
```
ojwb: If you're `#include <xapian.h>` it's pointless to then include sub-headers - do one or the other (outside of xapian-core, using `<xapian.h>` seems better, as that's the documented API header - the others are implementation details of xapian-core and other code ideally shouldn't assume their content or even existence).

xapian-letor/tests/api_letor.cc
```diff
@@ -182,6 +182,49 @@ DEFINE_TESTCASE(svm_ranker, generated)
...
+    string db_path = get_database_path("apitest_listmle_ranker",
+				       db_index_two_documents);
+    Xapian::Enquire enquire((Xapian::Database(db_path)));
+    enquire.set_query(Xapian::Query("lions"));
+    Xapian::MSet mymset = enquire.get_mset(0, 10);
+    string data_directory = test_driver::get_srcdir() + "/testdata/";
+    string query = data_directory + "query.txt";
+    string qrel = data_directory + "qrel.txt";
+    string training = test_driver::get_srcdir() + "/" +
+	    "listmle_ranker_training.txt";
```
ojwb: James mentioned on IRC, that you should combine the two strings - in C++ `x + "/" + "listmle_ranker_training.txt"` ends up being done as two separate string concatenations.

Which is a fair point if you're going to concatenate strings here, but actually, this is a generated file, so you should go in the build directory (i.e. the current directory) and not in `srcdir`.  When the build directory != `srcdir`, you should be able to run a build with a read-only `srcdir`.  This allows configuring the same source tree multiple ways, so you can test different build configurations in parallel.

It would be a good idea to read up on this a bit - see:

https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html

This testcase should work like the one using `"listnet_ranker_training.txt"`, which is correctly generated in the build directory.

ojwb:
Thanks, autosquashed and rebased onto master, and pushed.

PR #177, opened 2017-08-04
Test prepare_training_file separately

AyushP123:
Testing Xapian::prepare_training_file in the tests for rankers was
redundant and excessive. Have added a separate test for
Xapian::prepare_training_file and the file generated by it is stored in
testdata which is used by the tests for the rankers.


xapian-letor/tests/api_letor.cc
```diff
@@ -107,6 +107,8 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
     string qrel = data_directory + "qrel.txt";
     Xapian::prepare_training_file(db_path, query, qrel, 10,
 				  "training_output.txt");
+
+    return true;
```
jaylett: Would it be a good idea to do more than test that `::prepare_training_file()` doesn't raise an exception or abort? We should test that it produces the desires output.

AyushP123: ::prepare_training_file() just creates a feature vector corresponding to the <document, query> and appends the document relevance from the qrel file. Checking that it produces the desire output will be basically checking whether the features returned are correct and the correct relevance label is appended. I will be copy pasting the loops thats are implemented in the features to test whether the correct features are returned.

ojwb: Copy and pasting your implementation and comparing the output is obviously not a good way to write your tests - any bug in the implementation will be present in the copy and pasted version, and not be detected.  It would at least catch someone changing the implementation in a way which changes the output, but it's not a good approach.

But James does have a point - before this change, the output from `prepare_training_file()` was used as input to a test.  Now that output is totally ignored, and `prepare_training_file()` could produce literally any output and this test would pass.  That makes this a fairly poor test - we only ensure that it doesn't throw an exception, crash, or (if run under valgrind) leak memory or do something invalid that valgrind can detect.

Also writing a test not being easy isn't really a good reason not to write one.

We really want to ensure the output is **plausible** - testing the output can be read back in would at least catch bugs which result in output with an invalid format.  Comparing the output file with `testdata/training_data.txt` to ensure it's essentially the same would be better.  I don't think you can reliably just `diff` the files though, as that's probably too sensitive to rounding differences, but reading the two files together and checking corresponding real numbers only differ by a small amount should work.

AyushP123:
@jaylett , @ojwb : I am currently just checking the difference of the values between the two files and comparing the rest.

xapian-letor/tests/api_letor.cc
```diff
@@ -110,13 +112,50 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
+						     file1_data[j][i].size() -
+						     t)).c_str());
+	    t = file2_data[j][i].find_first_of(":");
+	    double number2 = stod((file1_data[j][i].substr(t + 1,
+						     file1_data[j][i].size() -
+						     t)).c_str());
+	    TEST(abs(number1 - number2) < 0.001);
+
+	}
+    }
```
ojwb: It would be better (simpler, clearer, not requiring temporary storage proportional to the size of the data being compared) to just read the two files together and compare the items as you read them.

xapian-letor/tests/api_letor.cc
```diff
@@ -110,13 +112,50 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
+	TEST_EQUAL(file2_data[j].size(), 22);
+	TEST_EQUAL(file1_data[j][0], file1_data[j][0]);
+	TEST_EQUAL(file1_data[j][1], file1_data[j][1]);
+	TEST_EQUAL(file1_data[j][21], file1_data[j][21]);
+	for (size_t i = 1; i < 21; ++i) {
+	    size_t t = file1_data[j][i].find_first_of(":");
+	    double number1 = stod((file1_data[j][i].substr(t + 1,
+						     file1_data[j][i].size() -
+						     t)).c_str());
+	    t = file2_data[j][i].find_first_of(":");
```
ojwb: The second argument of `substr()` will default to `std::string::npos` if omitted, which means "the rest of the string" (as any value longer than the remaining size of the string does0, so there's no need to calculate how long the rest of the string is and the code is much clearer if you don't.

You're also actually off-by-one in that calculation here - e.g. for `1:2`, `size() - t` is 2, when there's only one character more.

xapian-letor/tests/api_letor.cc
```diff
@@ -110,13 +112,50 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
+	file2_data.push_back(temp);
+    }
+    TEST_EQUAL(file1_data.size(), file2_data.size());
+    for (size_t j = 0; j < file1_data.size(); ++j) {
+	TEST_EQUAL(file1_data[j].size(), 22);
+	TEST_EQUAL(file2_data[j].size(), 22);
+	TEST_EQUAL(file1_data[j][0], file1_data[j][0]);
+	TEST_EQUAL(file1_data[j][1], file1_data[j][1]);
+	TEST_EQUAL(file1_data[j][21], file1_data[j][21]);
+	for (size_t i = 1; i < 21; ++i) {
```
ojwb: Why do you start at 1 here when you've already checked 1 as a special case above?

xapian-letor/tests/api_letor.cc
```diff
@@ -110,13 +112,50 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
+	    size_t t = file1_data[j][i].find_first_of(":");
```
ojwb: Better to use `':'` than `":"` here.  The compiler may be smart enough to spot this is a single character constant string and it can do a character search, but why rely on that rather than express it directly as a search for a character?

xapian-letor/tests/api_letor.cc
```diff
@@ -112,47 +112,28 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
+	istringstream iss1(file1);
+	istringstream iss2(file2);
+	string temp1;
+	string temp2;
+	int i = 0;
+	while ((iss1 >> temp1) && (iss2 >> temp2)) {
+	    if (i == 0 || i == 1 || i == 21) {
+		TEST_EQUAL(temp1, temp2);
+	    }
+	    else {
```
ojwb: Should all be on one line: `} else {`

xapian-letor/tests/api_letor.cc
```diff
@@ -98,6 +102,60 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+		size_t t1 = temp1.find_first_of(':');
+		size_t t2 = temp2.find_first_of(':');
+		TEST(abs(stod(temp1.substr(t1 + 1)) -
+			 stod(temp2.substr(t2 + 1))) <
+		     0.001);
+	    }
+	    i++;
+	}
+	if (i < 22) {
```
ojwb: Shouldn't this check `i != 22` - too many fields would also be wrong...

xapian-letor/tests/api_letor.cc
```diff
@@ -98,6 +102,60 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+	    throw Xapian::InternalError("File Generated with incorrect "
+					"inputs");
```
ojwb: Better to use `FAIL_TEST("message");` to fail a test.

Or simpler use the `TEST` family of macros to perform the check:

```
TEST_REL(i, !=, 22);
```

(This will also report what value `i` actually had if the comparison fails).

And for the checks below:

```
TEST(!(iss2 >> temp2));
```

etc.

xapian-letor/tests/api_letor.cc
```diff
@@ -22,6 +22,10 @@
 
 #include "api_letor.h"
 
+#include <fstream>
+#include <sstream>
```
ojwb: `<sstream>` doesn't seem to be used in the current version.

xapian-letor/tests/api_letor.cc
```diff
@@ -22,6 +22,10 @@
...
+#include <stdlib.h>
```
ojwb: Should use `<cstdlib>` here.

xapian-letor/tests/api_letor.cc
```diff
@@ -98,6 +102,47 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+    string training_data = data_directory + "training_data.txt";
+    Xapian::prepare_training_file(db_path, query, qrel, 10,
+				  "training_output.txt");
+    ifstream if1(training_data);
+    ifstream if2("training_output.txt");
+    string file1;
+    string file2;
+    while (getline(if1, file1)) {
+	 TEST(getline(if2, file2));
+	istringstream iss1(file1);
```
ojwb: There's an extra space before `TEST` on line 120 (compare it to the next line).

ojwb:
Thanks - rebased and squashed onto master.  Tests pass locally, so pushed.

PR #178, opened 2017-08-06
Move RoundRobin to the tests

richhiey1996:
I've removed the RoundRobin class from the public API and added it within the test file for clustering.
I created this branch off the stemming branch since some changes in this branch were dependent on some changes done in the stemming branch. So once that gets merged, this PR should be clean.


xapian-core/tests/api_cluster.cc
```diff
@@ -51,6 +51,58 @@ make_stemmed_cluster_db(Xapian::WritableDatabase &db, const std::string &)
     }
 }
 
+/** Round Robin clusterer:
+ *  This clusterer is a minimal clusterer which will cluster documents as -
+ *  ith document goes to the (i % k)th cluster where k is the number of clusters and
+ *  0 <= i < N; where N is the number of documents
+ */
+class XAPIAN_VISIBILITY_DEFAULT RoundRobin : public Xapian::Clusterer {
```
ojwb: `XAPIAN_VISIBILITY_DEFAULT` is no longer appropriate now this code has been moved - that macro marks a class as something to make visible outside the shared library, and this class is no longer in the shared library at all.

Otherwise, looks good to me.

richhiey1996: Oh yes I'll remove that right away.

richhiey1996: Should I rebase and squash?
Also, even after the stemming branch has been merged, the commits diff doesn't seem to change. It looks like commits from the stemming branch are still here. Is there any way to remove them to have a cleaner history on this PR? I was of the idea that the commit diff reduces when the same commits are added to the target branch.

ojwb: No need to rebase and squash - those commits just drop out when I do `git rebase master` in prepartion for merging, as they're already on master.

AIUI, a github PR is essentially just a git branch (plus some github-specific attached data), and contains exactly the commits on that branch.  So to append commits to the PR you can just commit to the branch and push those changes, and you can change the content or order of commits in the PR, but to do that you have to rewrite the history of the branch (which can be unhelpful for people reviewing it).

You branched the PR branch from an older version of master, and the commits that the branch and master have in common won't disappear from the branch just because the two follow the same history to start with - the branch point remains on the same commit.  To get rid of that shared history, you'd have to change that branch point (which `git rebase` can do).

If you're **only** changing the branch point, then with care I think you may actually be able to not rewrite history (because each the commit still has the same ancestors and the same content and meta data, you're only changing which one is the branch point).  I'm not sure I've actually tried this though.

Anyway, thanks for the quick fixup - squashed and merged to master and pushed.

richhiey1996: Oh I see. I guess my confusion arises mainly because bitbucket handles this in a different way. Thanks @ojwb!

PR #179, opened 2017-08-25
Add ERR Scorer to xapian-letor

AyushP123:
Have integrated the ERR Scorer from [vhasu's 2014 repository](https://github.com/v-hasu/xapian/tree/gsoc2014/xapian-letor). It meets
xapian coding standards. The changes made to xapian-letor have been
accomodated while integrating this scorer. Tests have also been added
for the scorer.

### DIFF
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -65,22 +65,41 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
     ~NDCGScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
+
 };
 
-/** ERRScore class
- *  ERR Score is adapted from the paper: goo.gl/Cs3ydQ
+/** Expected Reciprocal Rank(ERR) Scorer.
+ *  ERR Scorer is adapted from the paper:
+ *  http://olivier.chapelle.cc/pub/err.pdf
  *  Chapelle, Metzler, Zhang, Grinspan (2009)
  *  Expected Reciprocal Rank for Graded Relevance
  */
-class XAPIAN_VISIBILITY_DEFAULT ERRScore: public Scorer {
+class XAPIAN_VISIBILITY_DEFAULT ERRScore : public Scorer {
+    /* max_grade is the maximum grade that a label can be mapped to.
+     * The relevance label that user sets for the documents can be arbitary
+     * and huge. Computing the probability of relevance for the documents,
+     * which involves computing pow(2, label), can be time consuming.
+     * To reduce the time complexity the labels are mapped to a smaller space
+     * i.e. {0, 1, 2,...., max_grade}.
+     * By default max_grade is 4.
+     */
+    int max_grade = 4;
+
   public:
     /// Default constructor
     ERRScore();
+
+    /// Function to set max_grade
+    void set_max_grade(int max_grade_);
+
+    /// Function that returns the value set for max_grade.
+    int get_max_grade();
 
     /// Destructor
     ~ERRScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
+
 };
 
 }
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -68,6 +68,19 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
 
 };
 
+/// ERRScore class
+class XAPIAN_VISIBILITY_DEFAULT ERRScore: public Scorer {
+  public:
+    /// Default constructor
+    ERRScore();
+
+    /// Destructor
+    ~ERRScore();
+
+    double score(const std::vector<FeatureVector> & fvv) const;
+
+};
+
 }
 
 #endif /* SCORER_H */
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -65,7 +65,6 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
     ~NDCGScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
-
 };
 
 /** ERRScore class
@@ -82,7 +81,6 @@ class XAPIAN_VISIBILITY_DEFAULT ERRScore: public Scorer {
     ~ERRScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
-
 };
 
 }
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -75,21 +75,9 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
  *  Expected Reciprocal Rank for Graded Relevance
  */
 class XAPIAN_VISIBILITY_DEFAULT ERRScore : public Scorer {
-    /* max_grade is the maximum grade that a label can be mapped to.
-     * The relevance label that user sets for the documents can be arbitary
-     * and huge. Computing the probability of relevance for the documents,
-     * which involves computing pow(2, label), can be time consuming.
-     * To reduce the time complexity the labels are mapped to a smaller space
-     * i.e. {0, 1, 2,...., max_grade}.
-     */
-    int max_grade;
-
   public:
-    /** Construct an ERRScore.
-     *  @param max_grade_ max_grade value for the ERRScore.
-     *                    By default max_grade is set to 4.
-     */
-    explicit ERRScore(int max_grade_ = 4) : max_grade(max_grade_) { }
+    /// Default constructor
+    ERRScore();
 
     /// Destructor
     ~ERRScore();
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -81,15 +81,15 @@ class XAPIAN_VISIBILITY_DEFAULT ERRScore : public Scorer {
      * which involves computing pow(2, label), can be time consuming.
      * To reduce the time complexity the labels are mapped to a smaller space
      * i.e. {0, 1, 2,...., max_grade}.
-     * By default max_grade is 4.
      */
-    int max_grade = 4;
+    int max_grade;
 
   public:
-    /// Default constructor
-    ERRScore();
-
-    ERRScore(int max_grade_);
+    /** Construct an ERRScore.
+     *  @param max_grade_ max_grade value for the ERRScore.
+     *                    By default max_grade is set to 4.
+     */
+    explicit ERRScore(int max_grade_ = 4) : max_grade(max_grade_) { }
 
     /// Destructor
     ~ERRScore();
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -52,7 +52,6 @@ class XAPIAN_VISIBILITY_DEFAULT Scorer : public Xapian::Internal::intrusive_base
 
     /// Don't allow copying.
     Scorer(const Scorer & o);
-
 };
 
 /// NDCGScore class
@@ -65,7 +64,6 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
     ~NDCGScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
-
 };
 
 /** Expected Reciprocal Rank(ERR) Scorer.
@@ -83,9 +81,7 @@ class XAPIAN_VISIBILITY_DEFAULT ERRScore : public Scorer {
     ~ERRScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
-
 };
-
 }
 
 #endif /* SCORER_H */
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -68,7 +68,11 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
 
 };
 
-/// ERRScore class
+/** ERRScore class
+ *  ERR Score is adapted from the paper: goo.gl/Cs3ydQ
+ *  Chapelle, Metzler, Zhang, Grinspan (2009)
+ *  Expected Reciprocal Rank for Graded Relevance
+ */
 class XAPIAN_VISIBILITY_DEFAULT ERRScore: public Scorer {
   public:
     /// Default constructor
--- a/xapian-letor/include/xapian-letor/scorer.h
+++ b/xapian-letor/include/xapian-letor/scorer.h
@@ -89,11 +89,7 @@ class XAPIAN_VISIBILITY_DEFAULT ERRScore : public Scorer {
     /// Default constructor
     ERRScore();
 
-    /// Function to set max_grade
-    void set_max_grade(int max_grade_);
-
-    /// Function that returns the value set for max_grade.
-    int get_max_grade();
+    ERRScore(int max_grade_);
 
     /// Destructor
     ~ERRScore();
--- a/xapian-letor/scorer/err_score.cc
+++ b/xapian-letor/scorer/err_score.cc
@@ -45,12 +45,6 @@ ERRScore::~ERRScore()
     LOGCALL_DTOR(API, "ERRScore");
 }
 
-static bool
-comp(FeatureVector i, FeatureVector j)
-{
-    return i.get_label() < j.get_label();
-}
-
 double
 ERRScore::score(const std::vector<FeatureVector> & fvv) const
 {
@@ -59,7 +53,12 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
 	return 0;
     }
     size_t length = fvv.size();
-    FeatureVector max_label = *std::max_element(fvv.begin(), fvv.end(), comp);
+    FeatureVector max_label = *max_element(fvv.begin(), fvv.end(),
+					   [](FeatureVector x,
+					   FeatureVector y) {
+					       return x.get_label() <
+						      y.get_label();
+					   });
     double max_value = exp2(max_label.get_label());
 
     // Accumulated probability, which is updated for each document.
@@ -72,9 +71,9 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
 	 * http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.74.9057&rep=rep1&type=pdf
 	 */
 	double relevance_probability = (exp2(fvv[rank - 1].get_label()) -
-					 1) / max_value;
+		1) / max_value;
 
-       /* err_score = summation over all the documents
+	/* err_score = summation over all the documents
 	* ((satisfaction probability * p) / rank).
 	* Expected Reciprocal Rank(err_score) is calculated in accordance with
 	* algorithm 2 in the paper http://olivier.chapelle.cc/pub/err.pdf
--- a/xapian-letor/scorer/err_score.cc
+++ b/xapian-letor/scorer/err_score.cc
@@ -70,7 +70,7 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
     // Accumulated probability, which is updated for each document.
     double p = 1;
     double err_score = 0;
-    int max_value = exp2((int)max_label);
+    int max_value = exp2(max_label);
     for (int rank = 1; rank <= length; ++rank) {
 
 	/* Compute the probability of relevance for the document.
--- a/xapian-letor/scorer/err_score.cc
+++ b/xapian-letor/scorer/err_score.cc
@@ -1,5 +1,9 @@
 /** @file err_score.cc
  *  @brief Implementation of ERRScore
+ *
+ *  ERR Score is adapted from the paper: goo.gl/Cs3ydQ
+ *  Chapelle, Metzler, Zhang, Grinspan (2009)
+ *  Expected Reciprocal Rank for Graded Relevance
  */
 /* Copyright (C) 2014 Hanxiao Sun
  *
@@ -41,24 +45,6 @@ ERRScore::~ERRScore()
     LOGCALL_DTOR(API, "ERRScore");
 }
 
-static vector<double>
-get_labels(vector<Xapian::FeatureVector> fvv)
-{
-    LOGCALL_STATIC(API, vector<double>, "get_labels", fvv);
-    size_t fvvsize = fvv.size();
-    vector<double> labels;
-
-    for (size_t i = 0; i < fvvsize; ++i) {
-	labels.push_back(fvv[i].get_label());
-    }
-
-    return labels;
-}
-
-/* test the err_Score() use the data from goo.gl/LxEmPZ
- * input:[3,2,4]
- * output:0.63
- */
 double
 ERRScore::score(const std::vector<FeatureVector> & fvv) const
 {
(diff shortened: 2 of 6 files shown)


xapian-letor/include/xapian-letor/scorer.h
```diff
@@ -68,6 +68,19 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
...
+class XAPIAN_VISIBILITY_DEFAULT ERRScore: public Scorer {
+  public:
+    /// Default constructor
+    ERRScore();
+
+    /// Destructor
+    ~ERRScore();
+
+    double score(const std::vector<FeatureVector> & fvv) const;
+
```
jaylett: We didn't catch this on the previous scorer (since it's been there for ages), but we prefer not to have blank lines at the end of class declarations. It's probably easiest to tidy this up when we do a pass using `clang-format`, but if you wanted to tidy it up quickly as a separate commit in this PR I'd be fine with that. (It's not a fixup commit, though.)

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,94 @@
...
+    for (size_t i = 0; i < fvvsize; ++i) {
+	labels.push_back(fvv[i].get_label());
+    }
+
+    return labels;
+}
+
+/* test the err_Score() use the data from goo.gl/LxEmPZ
+ * input:[3,2,4]
+ * output:0.63
```
jaylett: This comment doesn't belong here — it should probably be used to construct a test case.

xapian-letor/tests/api_letor.cc
```diff
@@ -179,7 +179,8 @@ DEFINE_TESTCASE(listnet_ranker, generated)
     TEST_EXCEPTION(Xapian::FileNotFoundError,
 		   ranker.score(qrel, "", "ListNet_Ranker",
 				"scorer_output.txt", 10));
-    ranker.score(query, qrel, "ListNet_Ranker", "scorer_output.txt", 10);
+    ranker.score(query, qrel, "ListNet_Ranker", "ndcg_output.txt", 10);
+    ranker.score(query, qrel, "ListNet_Ranker", "err_output.txt", 10, "ERRScore");
```
jaylett: These tests don't tell us whether `ERRScore` is implemented properly, just whether they throw an exception or not. You should be able to test scorers directly by constructing `FeatureVector`s and driving their `score()` method, rather than (as this does) driving the ranker's `score()` method.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,94 @@
...
+ */
+double
+ERRScore::score(const std::vector<FeatureVector> & fvv) const
+{
+    LOGCALL(API, double, "ERRScore::score", fvv);
+
+    // hard code for the a five-point scale, the 16 means 2^(5-1)
```
jaylett: It takes some effort to understand what's going on here, and the comment doesn't help. Digging into the original paper, I'm unconvinced that we've implemented anything particularly similar to ERR, so at the least some work is required to make it clear that we have. It'd be easier to follow if our code followed the naming convention of the paper as far as possible.

While we're here, the paper should be referenced in the header of the file, and maybe in the docstring for the class.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,94 @@
...
+
+    // compute the satisfaction probability for lable of each doc in the ranking
+    for (int i = 0; i < length; ++i) {
+	labels[i] = (pow(2, labels[i]) - 1) / MAX_LABEL;
+    }
+
+    double err_score = labels[0];
+
+    // compute the accumulated probability for each doc which user will stop at
+    for (int i = 1; i < length; ++i) {
```
jaylett: Why do we ignore the label from the first feature vector? Is this because of poor transcription from algorithm 2 of the original paper?

xapian-letor/tests/api_letor.cc
AyushP123: I have added a test for the err_scorer

xapian-letor/scorer/err_score.cc
AyushP123: I am not eliminating this comment for now as it does describe the magic number MAX_LABEL from the paper

xapian-letor/scorer/err_score.cc
AyushP123: we don't ignore the label from the first feature vector .... each label is converted to the probability that a user will be satisfied with the document .... which is the contribution from the first ranked document in the ERR

xapian-letor/include/xapian-letor/scorer.h
AyushP123: This should be a separate commit so that I can eliminate this problem with the other scorer as well??

xapian-letor/scorer/err_score.cc
jaylett: Ah, I see. Again, I think the paper's description (and find the actual paper, because that's what I'm working from and what anyone else with academic paper access will want to work from) needs unpacking in an implementation comment so we know how to read the rest of the code in the scorer.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,92 @@
...
+
+double
+ERRScore::score(const std::vector<FeatureVector> & fvv) const
+{
+    LOGCALL(API, double, "ERRScore::score", fvv);
+
+    // hard code for the a five-point scale, the 16 means 2^(5-1)
+    int MAX_LABEL = 16;
+
+    std::vector<double> labels;
```
jaylett: You know the size of this, so you can use an array of doubles instead of a vector.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,92 @@
...
+    double max_label = fvv[0].get_label();
```
jaylett: You now have `MAX_LABEL` and `max_label`. At least one of these is now incorrectly-named.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,92 @@
...
+
+    for (size_t i = 0; i < fvv.size(); ++i) {
+	double label = fvv[i].get_label();
+	labels.push_back(fvv[i].get_label());
```
jaylett: You've just dereferenced this into `label`, so use that.

xapian-letor/scorer/err_score.cc
```diff
@@ -50,42 +50,44 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
...
     }
 
-    int length = labels.size();
-
-    // compute the satisfaction probability for lable of each doc in the ranking
+    // compute the satisfaction probability for label of each doc in the ranking.
     for (int i = 0; i < length; ++i) {
-	labels[i] = (pow(2, labels[i]) - 1) / MAX_LABEL;
+	vector_intermediate[i] = (pow(2, vector_intermediate[i]) - 1) /
+				  MAX_PROB;
```
jaylett: Since we've collapsed the label space to `{0, 1, 2, 3, 4}`, we should probably pre-calculate the possible values of this expression, so we aren't doing O(n) `pow()` calls. (In fact, we can then do no `pow()` calls at all, using an accumulator in the pre-calculation loop.)

xapian-letor/scorer/err_score.cc
```diff
@@ -50,42 +50,44 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
...
-
-	// for users
-	for (int j = i - 1; j >= 0; --j) {
-	    temp_err *= (1 - labels[j]);
-	}
-	err_score += temp_err;
+    /* compute the accumulated probability p for each doc
+     * that the user will stop at.
+     * err_score = summation for each doc
+     * ((satisfaction probability * p) / rank)
```
jaylett: I'd reference "algorithm 2 in the paper" or similar explicitly, and I'd put `p` first rather than in the middle of the calculation (matching the `ERR <- …` line in algorithm 2). Given you're calling it "rank" in the comment here, why not use that as the loop variable as well? (And `rank-1` when indexing into `vector_intermediate`; or 0-index `rank` and explain why.)

xapian-letor/scorer/err_score.cc
ojwb: Also, in C++ `2^(5-1)` is `6` not `16` (since `^` is bit-wise exclusive or).  If you mean "raise to the power of", it'd be much better to write something that actually means that in C++, for example: `pow(2, 5 - 1)`

xapian-letor/scorer/err_score.cc
```diff
@@ -50,42 +50,44 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
...
+	vector_intermediate[i] = label;
 	max_label = max(max_label, label);
     }
 
-    for (size_t i = 0; i < fvv.size(); ++i) {
-	labels[i] = round((labels[i] * 4) / max_label);
+    // normalize the labels to an integer from 0 to 4 because we are using
+    // a grade point system where each document is graded from 0 to 4.
+    for (int i = 0; i < length; ++i) {
+	vector_intermediate[i] = round((vector_intermediate[i] * 4) /
```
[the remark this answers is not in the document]
ojwb: Yes, `const int FOO = 4;` is generally better C++ style than a preprocessor define like `#define FOO 4`.

There used to be a bit of a wrinkle (at least with some compilers) with constant class members still occupying space in the class unless you make them `static` but making them `static` meant you then needed to define them out-of-line.  I think C++11 has resolved that though, and now you can safely say `static const int FOO = 4;` inside a class definition.

xapian-letor/scorer/err_score.cc
ojwb: FWIW, `pow(2, x)` will very likely get optimised to `x * x` and not call `pow()` at all (IIRC, both GCC and clang do this).

xapian-letor/include/xapian-letor/scorer.h
```diff
@@ -65,7 +65,22 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
     ~NDCGScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
+};
+
+/** ERRScore class
```
ojwb: "ERRScore class" is not a good brief description.  It's just the name of the class (which in this case doesn't really usefully explain itself, since it's not clear what "ERR" means) plus the word "class".

Something like "Expected Reciprocal Rank Scorer" would at least explain the acronym.

xapian-letor/include/xapian-letor/scorer.h
```diff
@@ -65,7 +65,22 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
...
+ *  ERR Score is adapted from the paper: goo.gl/Cs3ydQ
+ *  Chapelle, Metzler, Zhang, Grinspan (2009)
+ *  Expected Reciprocal Rank for Graded Relevance
+ */
+class XAPIAN_VISIBILITY_DEFAULT ERRScore: public Scorer {
```
ojwb: Missing space before `:`

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,102 @@
...
+ERRScore::score(const std::vector<FeatureVector> & fvv) const
+{
+    LOGCALL(API, double, "ERRScore::score", fvv);
+
+    /* 2^(gmax) where gmax is the max of all the relevance grades.
+     * since a 5 point grade sytem is used, possible relevance grades are
+     * {0, 1, 2, 3, 4}, gmax = 4. For more information refer to section 4 of
+     * the paper http://olivier.chapelle.cc/pub/err.pdf
+     */
+    int MAX_PROB = 16;
```
ojwb: `const int` would be better really.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,102 @@
...
+
+    int length = fvv.size();
+
+    /* used to store values which change from labels to
+     * satisfaction probability
+     */
+    double intermediate_values[length];
+    double max_label = fvv[0].get_label();
```
ojwb: What if `fvv.empty()`?

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,102 @@
...
+
+    /* compute the accumulated probability p for each doc
+     * that the user will stop at.
+     * err_score = summation for each doc
+     * ((satisfaction probability * p) / rank)
+     */
+    for (int i = 0; i < length; ++i) {
+	err_score = err_score + (intermediate_values[i] * p / (i + 1));
+	p = p * (1 - intermediate_values[i]);
+    }
```
ojwb: It seems that the above loop could be merged with the two loops before it (we can't merge all four loops as we need to calculate `max_label` before we can scale the values).

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,102 @@
...
+	double label = fvv[i].get_label();
+	intermediate_values[i] = label;
+	max_label = max(max_label, label);
+    }
+
+    // normalize the labels to an integer from 0 to 4 because we are using
+    // a grade point system where each document is graded from 0 to 4.
+    for (int i = 0; i < length; ++i) {
+	intermediate_values[i] = round((intermediate_values[i] * 4) /
+				       max_label);
```
ojwb: Is `round()` the correct operation here?  I'm not seeing where this is explicitly specified in either the paper or the blog post, but the blog post says "the last grade 4 (perfect)", which suggests to me that only `max_label` should map to `4` - i.e. we want `floor()` not `round()` here.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,102 @@
...
+
+    int length = fvv.size();
+
+    /* used to store values which change from labels to
+     * satisfaction probability
+     */
+    double intermediate_values[length];
+    double max_label = fvv[0].get_label();
+
+    // store the labels set the by the user in intermediate_values.
```
ojwb: "set the by the user" -> "set by the user"

xapian-letor/scorer/err_score.cc
jaylett: Oh, that's handy. Although an accumulator should still be faster overall, it won't be as clear.

xapian-letor/include/xapian-letor/scorer.h
```diff
@@ -65,7 +65,6 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
     ~NDCGScore();
 
     double score(const std::vector<FeatureVector> & fvv) const;
-
```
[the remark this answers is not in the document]
AyushP123: I made these changes in the commit by mistake .... I can remove the file that caused this change and undo the change in the commit ... if you allow me to do so

xapian-letor/scorer/err_score.cc
AyushP123: throwing an InvalidArgumentError

xapian-letor/scorer/err_score.cc
AyushP123: like the example I discussed in IRC if suppose we have a 1000 documents ranked in a continuous fashion we would want all the documents from 750 - 1000 get mapped to 4.

xapian-letor/scorer/err_score.cc
AyushP123: Resolved this by throwing an InvalidArgumentError. Can't return 0 in this case because the scorer needs a label to return a legitimate score.

xapian-letor/scorer/err_score.cc
AyushP123: replaced the const variable with an variable which can be set by the user with default value of 4.

xapian-letor/include/xapian-letor/scorer.h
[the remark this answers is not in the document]
jaylett: Probably roll them back in another fixup commit, then do them again in another commit.

xapian-letor/include/xapian-letor/scorer.h
```diff
@@ -65,22 +65,41 @@ class XAPIAN_VISIBILITY_DEFAULT NDCGScore: public Scorer {
...
+     * By default max_grade is 4.
+     */
+    int max_grade = 4;
+
   public:
     /// Default constructor
     ERRScore();
 
+    /// Function to set max_grade
+    void set_max_grade(int max_grade_);
```
jaylett: Why isn't this just a constructor parameter?

AyushP123: yes, that would be correct. I was thinking that after creating an ERRScorer instance a user would like to change max_grade multiple times.

jaylett: Why? It's a lightweight class, in almost all cases it'd be clearer just to make a new one.

AyushP123: yes, I realized that after making the change .... I have pushed a new commit which introduces a parameterized constructor

xapian-letor/include/xapian-letor/scorer.h
```diff
@@ -89,11 +89,7 @@ class XAPIAN_VISIBILITY_DEFAULT ERRScore : public Scorer {
     /// Default constructor
     ERRScore();
 
-    /// Function to set max_grade
-    void set_max_grade(int max_grade_);
-
-    /// Function that returns the value set for max_grade.
-    int get_max_grade();
+    ERRScore(int max_grade_);
```
jaylett: You can default this argument so you only need one constructor; you can use an initializer list to set the `max_grade` member here (which at present will mean you don't need the constructor in the `.cc` file); and you must mark single-argument constructors `explicit` to avoid implicit conversion problems.

xapian-letor/scorer/err_score.cc
ojwb: I didn't see that discussion, but that would mean 4 isn't "perfect".

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,96 @@
...
+    for (int i = 0; i < length; ++i) {
+	double label = fvv[i].get_label();
+	intermediate_values[i] = label;
+	max_label = max(max_label, label);
+    }
+
+    // Accumulated probability, which is updated for each document.
+    double p = 1;
+    double err_score = 0;
+    int max_value = exp2(max_label);
```
ojwb: Shouldn't `max_value` be `double`?

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,96 @@
...
+    for (int rank = 1; rank <= length; ++rank) {
+
+	/* Compute the probability of relevance for the document.
```
ojwb: Superfluous blank line before the comment.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,96 @@
...
+	 * Probability of relevance is calculated in accordance with the gain
+	 * function for the Discounted Cumulative Gain in the paper:
+	 * http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.74.9057&rep=rep1&type=pdf
+	 */
+	intermediate_values[rank - 1] = (exp2(intermediate_values[rank - 1]) -
+					 1) / max_value;
```
ojwb: There's no reason to stuff the calculated probability of relevance back into `intermediate_values` - we'll never use it from there after this loop iteration.  Just put it in a local variable, and use it from there below.

That'll make for more readable code too, as this line currently ends up wrapped awkwardly due to its length.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,96 @@
...
+}
+
+double
+ERRScore::score(const std::vector<FeatureVector> & fvv) const
+{
+    LOGCALL(API, double, "ERRScore::score", fvv);
+    if (fvv.empty()) {
+	throw Xapian::InvalidArgumentError("Empty argument supplied");
+    }
+    int length = fvv.size();
```
ojwb: `int` is really not the right type for an array size - better to use `size_t`.

xapian-letor/scorer/err_score.cc
```diff
@@ -0,0 +1,96 @@
...
+
+    /* used to store values which change from labels to
+     * satisfaction probability
+     */
+    double intermediate_values[length];
```
ojwb: This uses a non-standard extension supported by GCC and some other compilers - ISO C++ requires the size of an array to be known at compile time.  We should avoid non-standard extensions (except in code which is inherently compiler-specific anyway).

(I know James said to use an array instead, but I think he overlooked that the array size here isn't a compile-time constant).

The options I can see are:

(a) Revert to using `vector<double>`.

(b) Use `double* intermediate_values = new double[length];` - the downside is you then need to explicitly call `delete []` on the returned pointer, and that's fiddly to ensure in the face of exceptions.

(c) Use `std::unique_ptr<double[]> intermediate_values{new double[length]};` - effectively (b) but using `unique_ptr` to ensure there's no leak.

(d) Do away with `intermediate_values` entirely.  Calling `fvv[i].get_label()` returns a member variable rather than performing some complex calculation, so there's not really a good reason to carefully cache the result - we could just call this again in the final loop.

If we need a temporary container, (c) is the cleanest and least overhead, but I'd suggest (d) makes more sense in this case.

xapian-letor/scorer/err_score.cc
ojwb: `NDCGScore::score()` will return 0 if passed an empty vector.

xapian-letor/scorer/err_score.cc
AyushP123: yes

xapian-letor/scorer/err_score.cc
AyushP123: yes, definitely d makes more sense

ojwb:
Thanks, LGTM.

@jaylett: OK to merge?

xapian-letor/scorer/err_score.cc
```diff
@@ -72,9 +71,9 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
 	 * http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.74.9057&rep=rep1&type=pdf
 	 */
 	double relevance_probability = (exp2(fvv[rank - 1].get_label()) -
-					 1) / max_value;
+		1) / max_value;
```
ojwb: I think this was better before.

Though probably more readable to use an intermediate variable, and so avoid needing to wrap the expression:

```
 	auto label = fvv[rank - 1].get_label();
 	double relevance_probability = (exp2(label) - 1) / max_value;
```

xapian-letor/scorer/err_score.cc
```diff
@@ -59,7 +53,12 @@ ERRScore::score(const std::vector<FeatureVector> & fvv) const
 	return 0;
     }
     size_t length = fvv.size();
-    FeatureVector max_label = *std::max_element(fvv.begin(), fvv.end(), comp);
+    FeatureVector max_label = *max_element(fvv.begin(), fvv.end(),
+					   [](FeatureVector x,
+					   FeatureVector y) {
```
ojwb: We don't have an explicit style rules for lamba functions currently, so I think we should apply the rules for functions unless there's a good reason why they don't make sense, which would mean aligning `FeatureVector y` directly under `FeatureVector x`.  I think that makes sense here, and is more readable (at a glance, it currently looks like `FeatureVector y` is a parameter of the `max_element()` invocation).

ojwb:
There's a couple of minor nits in the changes added since I OKed this, but aside from this I'm still happy with this.

@jaylett You didn't seem to respond to my "OK to merge?" above - it looks to me like your review has been fully addressed so I'm going to assume you're happy now.  If not, please yell.

jaylett:
@ojwb apologies: I'm happy, once this passes CI :)

jaylett:
(Looks like the CI problem might be the thing you were working with on master, so if that's fixed a rebase should solve it?)

ojwb:
Yes, the CI problems are unrelated to the changes here (Apple ship an 11+ year old version of Bison in xcode, and that doesn't work for building doxygen HEAD from homebrew, and my attempts to get it to use bison from homebrew aren't working; we're using doxygen HEAD to work around a bug in the current doxygen release, though I see that the homebrew doxygen package now includes a patch for this, so I've reverted master to using that - let's see if that fixes the problem).

ojwb:
Merged to master.

PR #180, opened 2017-10-17
Add flag to start snippets at first non-whitespace character.

rsto:
We were having issues with snippet generation for email subjects. E.g. the subject

    [xapian-devel] foo

would result in the snippet

     xapian-devel] foo

(note the missing bracket at the start).

This patch adds a flag to MSet to start snippets at the first non-whitespace character. For the example this would become

    [xapian-devel] foo

The default is to keep snippet-generation as-is and there's a regression test for that.

### DIFF
--- a/xapian-core/queryparser/cjk-tokenizer.cc
+++ b/xapian-core/queryparser/cjk-tokenizer.cc
@@ -53,6 +53,8 @@ CJK::codepoint_is_cjk(unsigned p)
     // Array containing the last value in each range of codepoints which
     // are either all CJK or all non-CJK.
     static const unsigned splits[] = {
+	// 0E00..0E7F; Thai, Lanna Tai, Pali
+	0x0E00 - 1, 0x0E7F,
 	// 1100..11FF; Hangul Jamo
 	0x1100 - 1, 0x11FF,
 	// 2E80..2EFF; CJK Radicals Supplement
--- a/xapian-core/tests/api_termgen.cc
+++ b/xapian-core/tests/api_termgen.cc
@@ -184,6 +184,9 @@ static const test test_simple[] = {
     // would split this differently as '申込み ！月額 円'
     { "", "申込み！月額円", "み[2] 円[4] 月額[3] 申込[1]" },
 
+    // Thai word segmentation
+    { "", "โดยตั้งใจ", "ตั้งใจ[2] โดย[1]"},
+
     // Test set_stemming_strategy():
     { "stem=en,none,!cjkwords",
 	  "Unstemmed words!", "unstemmed[1] words[2]" },


ojwb:
CI failed because `xapian-check-patch` says:

```
xapian-core/queryparser/termgenerator_internal.cc:508: error: added/changed line uses spaces for indentation rather than tab: +	        (flags & Xapian::MSet::SNIPPET_START_AT_NONSPACE &&
```

I'd also suggest parentheses around the `&` subexpression so the reader doesn't have to think about the relative precedence of `&` vs `&&`.

```
// Regression test: default snippet generation strips of non-word chars
```

Typo - should be "strips off".

I kind of wonder if this should actually be the default setting (or just how it works).  At least for paired punctuation (such as the various sorts of brackets and quotes) it'll look better if we don't chop off the opener when the closer is shown, and in the case the testcase tests we even chop it off when the full snippet is within the allowed length.

I also wonder if we should handle the end of the snippet differently under this flag - currently "end of sentence" punctuation is included, but probably paired punctuation is also useful there.

rsto:
Thanks, I fixed indentation and the typo and ran it against xapian-check-patch.

As to default setting: I think the current default is adequate for the general case. It seems to handle missing spaces between words and punctuation well. E.g. for `foo,bar` (note the missing space after `,`) we can be sure never to get a snippet that includes the `,` like `...,<b>bar</b>`.

But for paired punctuation both of your suggestions sound good to me.

rsto:
@ojwb Is there anything left to change before this could go into the master branch? We are using the patch successfully since months in production.

rsto:
There's a couple of issues with Travis:

The GCC build fails with 

    The job exceeded the maximum time limit for jobs, and has been terminated.

which seems not to be related to the codebase (e.g. I didn't see this issue with a previous run of the same code).

The OS X build fails with

    /Applications/Xcode.app/Contents/Developer/usr/bin/make  check-am
    PYTHONPATH=..:$PYTHONPATH env DYLD_LIBRARY_PATH=/Users/travis/build/xapian/xapian/xapian-core/.libs /usr/local/bin/python2 \
            -c 'import sphinx,sys;sys.exit(sphinx.main(sys.argv))' \
            -b html -d doctrees -c docs ./docs docs/html
    usage: usage: -c [OPTIONS] SOURCEDIR OUTPUTDIR [FILENAMES...]
    -c: error: argument -c: expected one argument

which also looks like an unrelated issue.

I'm not sure really what to do, this PR didn't even touch the `.travis.yml` file :-/

ojwb:
It's failing similarly on master currently, and I think it must be due to a change in the travis environment because nothing obviously related has changed recently in our code.  Not sure what's going on - the error doesn't make much sense to me, but I don't have access to OS X to debug it.

ojwb:
This seems a useful change, but I'm still a bit unsure about this being an option (or if we have an option the default being the current behaviour), and it seemed better to decide that before merging as otherwise we risk making a release without coming back to it and then we're stuck with this as an option for a couple of release cycles.

An option means the API user has to make a choice as to whether to enable this or not (or just accept the default, which is probably is what most will do, so good defaults matter).  But this doesn't seem like something where the best behaviour varies - while the particular `xapian-devel] ...` motivation given above is specific to email, in general it's odd to start the sample just after some punctuation when the matching punctuation is in the sample.  I can't really see a situation in which it's better to behave as we currently do, and it's not like the current behaviour was carefully designed to do this - it just seemed most natural to start with a word, and I didn't really think about punctuation right before the first word.

I'll try to dig through some documents to find cases where this flag would change the behaviour, so we have more data points to work from.

rsto:
> I'll try to dig through some documents to find cases where this flag would change the behaviour, so we have more data points to work from.

Thanks for the feedback. Off the top of my head, there's one scenario where the current default produces a nicer snippet:

For `Hello,foo!` (no space between `,` and `foo`) the patch produces for snippet-length 4

    ...,<b>foo</b>!

whereas the current default produces 

    ...<b>foo</b>!

I could work around this by adding a check for punctuation in the conditional. That should safe-guard against documents with sloppy punctuation setting. I'll wait for your feedback following your own analysis of documents.

rsto:
> It's failing similarly on master currently, and I think it must be due to a change in the travis environment because nothing obviously related has changed recently in our code. Not sure what's going on - the error doesn't make much sense to me, but I don't have access to OS X to debug it.

I can only guess, but this looks like an issue with the sphinx version. The GCC build on Linux uses sphinx-1.2.2 (or so I gather from the Ubuntu package name), but the OS X builds use sphinx 1.7. It's the second `-c` argument that's failing, not to `-c` argument to the python interpreter.

ojwb:
> but the OS X builds use sphinx 1.7

FWIW, it is fine with 1.6.7 locally.

> It's the second `-c` argument that's failing, not to `-c` argument to the python interpreter.

Indeed, but what is being passed seems to fit the syntax it claims to want in the error message.

Looking through the 1.7.0 changes list turns up this little gem: https://github.com/sphinx-doc/sphinx/pull/3668

So they've unhelpfully made an incompatible change so either a version check is needed (ick) or else we pass a dummy argument in `argv[0]` that older sphinx can safely ignore.

@jaylett has OS X so is taking a look.

ojwb:
#188 fixed the sphinx issue, so I've restarted the failed OS X build, and also the linux gcc one which timed out.

ojwb:
Some initialise analysis suggests that most punctuation falls nicely into two categories:

 * Cases where it seems to be usually beneficial (and rarely worse) to include the punctuation - examples:
   * opening (or undifferentiated) paired punctuation such as various quotes and brackets (e.g. `'`, `"`, `«`, `<`, `¡`, `¿`, `(`, `[`, `{`, ```)
   * currency symbols (e.g. `$`, `€`, `£`, `¥`, `¢`)
   * signed numbers (`-` and `+`) (though `-` is often used as a hyphen)
   * file paths (`/`, `\`)
   * assorted others which are commonly used as a prefix in a way which adds some context (`@`, `%`, `&`, `§`, `#`, `~`)

 * Cases where it seems almost universally undesirable or neutral to include the punctuation, which seems to include all closing paired punctuation - examples:
   *  `,`, `?`, `!`, `;`, `:`, `)`, `]`, `}`, `»`, `>`, `=`, `*`, `|`, `^`
   * closing directional quotes
   * various dashes and hyphens
   * ellipses

Based on this, I'd suggest a whitelist approach, using a list of Unicode categories (e.g. CURRENCY_SYMBOL, OPEN_PUNCTUATION, INITIAL_QUOTE_PUNCTUATION) plus a list of individual code points.

I'm not sure what's best when there are multiple such punctuation characters - perhaps just taking one, unless it's repeated (which helps cases like `\\unc\path` and `--help`)?  Certainly we want a limit on how many we include.

ojwb:
It occurred to me that the points at which you might want to start a snippet are similar to the points at which you might want to start a line, so the Unicode line-breaking algorithm is probably interesting:

http://www.unicode.org/reports/tr14/

If we're using ICU anyway we could potentially use it for this, though it's unclear to me from its API docs if it can efficiently iterate back one line break from the end of a specified string, or if internally works out all the line breaks in the string to do that (or converts the string from UTF-8 to something else or something else making it O(length of the string) or worse).  And perhaps the line breaking rules don't exactly align with what's wanted here.

ojwb:
Actually, since we know we're happy to break after whitespace we could just back up to the preceding whitespace (or the start of the text if there's none) - that would typically mean passing a very short string to ICU, and we wouldn't need to ask it at all in the common case when there's whitespace right before the first word of the snippet.

To be clear, I wasn't expecting you to try this with ICU but if you fancy having a play I'd certainly be interested to hear how it goes.

ojwb:
I've added a limit on the amount of punctuation included and adjusted it so it only considers the contiguous group of such punctuation right before the first word character.

This handles your test case in the same way as your patch with the option on, and your example with a comma in the same way as the existing code (so `bar,foo!` with length 5 gives `...<b>foo</b>!`) but I'd like to hear if there are other cases that this doesn't handle at least as well.

This seems to work at least as well in all cases as the current code and is clearly better in quite a few, so I've pushed those changes to master in 15e5efa0d0e55c687050ee2e9500fc7690a6d37d and 
b9c8758e9b7cee9e0d530069d8a4412e0dc77796 with the plan of including them in 1.4.6.

However, as well as wanting feedback on whether this addresses your needs, there's also my question above about whitespace:

> I also noticed that your version doesn't take any account of whitespace between the punctuation and the first actual word, so for `[ blah blah foo` it'll start the snippet at the `[`. Is that intentional, or just an accident of the implementation? If intentional, what's the thinking behind doing that?

We probably should consider including closing paired punctuation (and maybe some other characters) at the end of the snippet - currently we only include end of sentence punctuation there.

And longer term we should look at using ICU's implementation of the Unicode line breaking algorithm here instead or in part (or another implementation if we don't go with ICU).

rsto:
Sorry for delayed response, I've been away from work for a couple of weeks. Your changes look good to me. Considering closed paired punctuation would make this indeed even more robust for the special case of handling email subjects of mailing lists. But in practice, I don't expect this to be much of an issue, as we typically don't see whitespace between the starting `[` and the list name.  Off the top of my head I can't imagine any other case, where we would need to special-case punctuation.

With regards to ICU: how would you like to approach integrating ICU in Xapian? Currently, the discussion has been on this pull request https://github.com/xapian/xapian/pull/114, should we open an own PR or issue to track this? One year ago, I had an experimental branch for replacing UTF-8 iterators in Xapian with ICU: https://github.com/rsto/xapian/tree/utf8itor Should we start a PR from there?

ojwb:
> Off the top of my head I can't imagine any other case, where we would need to special-case punctuation.

I came across a few cases where it's arguably better to include trailing punctuation while making up testcases for the change above:

```
    input = "/opt/foo/bin/";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "/opt/<b>foo</b>/bin/");

    input = "\"foo bar\"";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "\"<b>foo</b> bar\"");

    input = "\\\\server\\share\\foo\\";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "\\\\server\\share\\<b>foo</b>\\");

    input = "«foo»";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "«<b>foo</b>»");

    input = "#include <foo>";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "...#include &lt;<b>foo</b>&gt;");

    input = "(foo)";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "(<b>foo</b>)");

    input = "{foo}";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "{<b>foo</b>}");

    input = "[foo]";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "[<b>foo</b>]");

    input = "`foo`";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "`<b>foo</b>`");

    input = "@foo@";
    TEST_STRINGS_EQUAL(mset.snippet(input, strlen(input), stem),
                       "@<b>foo</b>@");
```

Currently all of these fail because they omit the final punctuation (but add "..." instead, so actually give a longer snippet!)

Currency symbols are conventionally written after the number in some locales too.

But that's really a separate potential improvement which can be slotted in at any point, and the code as already released doesn't handle those any better.

As for ICU, I'm thinking we probably just make it an optional "experimental" dependency for the next stable release series which people wanting CJKV segmentation can enable.  ICU is rather hefty and I have a few reservations about it, plus rewriting everything to use it would delay the release series which as usual is taking longer to get out than I'd hoped.

So keeping it in #114 seems sensible to me.

ojwb:
I've pushed a commit to master which keeps some trailing punctuation, which I'll backport to 1.4.x assuming it applies easily (I expect it should so this should be in 1.4.16).

We could perhaps use ICU (if enabled) for deciding where to start/end snippets, but as noted above #114 covers that.  I think everything else here is addressed, so closing.

@rsto If you (or anyone else) find cases that could be improved, please report them.

PR #181, opened 2017-10-20
Add warning to action 'hash' for integer argument greater than 6

samuelharden:
Currently if your puts max_length less than 6 with hash argument  it throws unknown error which need to be debugged to understand we can't put that.

It might be the less frequent occurrence. But useful to send error to user about it.

### DIFF
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -495,7 +495,7 @@ else
     : # OK
 fi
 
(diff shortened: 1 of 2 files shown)


ojwb:
Thanks.

I've merged a patch based on this PR as 4c409f1e6123f437a41137018bba76eb99ab7564.  The main adjustments were that this is now an error not a warning (since we can't handle it), and that we check for it at script parse time rather than script execution time.

PR #182, opened 2017-10-20
No argument takes next action as input When command has = sign and doesn't have next command it's takes next action as input

samuelharden:
Currently if someone writes something like this

```
md5: boolean= unhtml boolean=XAFTERTRUNCATE
```

DB will contain :

```
Term List for record #2:
XAFTERTRUNCATE:E752DEB0C8E42A7D73CCA0C7818C2F23
unhtml:E752DEB0C8E42A7D73CCA0C7818C2F23
```

which seems wrong. Complete Example:

```
~$xapian-delve -r 2 fjd
Term List for record #2:
XAFTERUNHTML:E752DEB0C8E42A7D73CCA0C7818C2F23
unhtml:E752DEB0C8E42A7D73CCA0C7818C2F23
~$ cat extractmd5.script
md5: boolean= unhtml boolean=XAFTERUNHTML
```

### DIFF
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -479,7 +479,7 @@ else
     failed=`expr $failed + 1`
 fi
 
-rm "$OMEGA_CONFIG_FILE" "$TEST_INDEXSCRIPT" "$TEST_TEMPLATE"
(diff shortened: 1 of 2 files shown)


ojwb:
Thanks for raising this, and for the patch.  As we discussed on IRC last week, deprecating allowing spaces here is simpler (and it has never been documented that they were allowed).  I've done that in fc59095ff4fbef396dae30ee7bb8be381a81a058.

PR #183, opened 2017-11-23
Convert all errors to go to stderr instead of stdout in scriptindex

samuelharden:
Currently errors and warning are printed on stdout in scriptindex. Shift them to stderr.

It would be ideal to send errors on stderr to isolate from verbose or info things we print in scriptindex.

Program running scriptindex would want to separate error from info statement to record or store errors.


ojwb:
Thanks - rebased and merged as 6ebd807d9c82b98609b7c25be7b8165724111b0d.

PR #184, opened 2018-01-01
Remove needless zeroing of calloc-allocated memory

mmcco:
There were some less obvious instances, but I erred on the side of caution so that I wouldn't break anything.

Thanks for your time,
Michael McConville
University of Utah

### DIFF
--- a/xapian-core/queryparser/lemon.c
+++ b/xapian-core/queryparser/lemon.c
@@ -593,7 +593,6 @@ acttab *acttab_alloc(void){
     fprintf(stderr,"Unable to allocate memory for a new acttab.");
     exit(1);
   }
-  memset(p, 0, sizeof(*p));
   return p;
 }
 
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for the patch.

However, the maintainer for this code is actually the sqlite project, so it would be better to send a patch to them instead:

https://www.sqlite.org/src/artifact?ci=trunk&filename=tool/lemon.c

Sadly they don't have an open bug tracker - instead you're meant to email patches to a mailing list, which means you may have to chase if nobody picks them up:

https://www.sqlite.org/src/wiki?name=Bug+Reports

I'm reluctant to increase the size of the delta we're carrying against sqlite's `lemon.c` by applying a patch which doesn't actually fix a bug, especially as GCC 5 and later will optimise away redundant calls to `memset()` after `calloc()` such as these anyway.  We generally sync changes at least once per release cycle, so if it gets applied in sqlite's repo, we'll pick up the change.

PR #185, opened 2018-02-14
Fix buggy annotations

uppinder:
Fix annotations that were buggy and caused build to fail. Modifications done according to discussion
in IRC.

### DIFF
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -646,6 +646,25 @@ $stoplist
 	returns a list of any terms in the query which were ignored as
 	stopwords.
 
+$subdb[{DOCID}]
+        return the name of the sub-database containing ``DOCID`` (or the
+        current document in the histlist if ``DOCID`` is omitted).
+
+        NB: The current implementation assumes that each omega database name
+        corresponds to a single Xapian database - if a database name refers to
+        a stub database file expanding to multiple Xapian databases then this
+        command will misbehave.
+
+$subid[{DOCID}]
+        return the docid in the sub-database corresponding to ``DOCID`` in the
+        combined database (or the current document in the histlist if ``DOCID``
+        is omitted).
+
+        NB: The current implementation assumes that each omega database name
+        corresponds to a single Xapian database - if a database name refers to
+        a stub database file expanding to multiple Xapian databases then this
+        command will misbehave.
+
 $substr{STRING,START[,LENGTH]}
         returns the substring of ``STRING`` which starts at byte position
         ``START`` (the start of the string being 0) and is ``LENGTH`` bytes
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -1,7 +1,7 @@
 #!/bin/sh
 # omegatest: Test omega CGI
 #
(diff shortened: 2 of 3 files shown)


ojwb:
CI failure on OS X is also present on master and pretty clearly not related to these changes.

xapian-core/queryparser/queryparser.lt
```diff
@@ -477,8 +477,8 @@ static unsigned int yy_find_shift_action(
       if( iLookAhead<sizeof(yyFallback)/sizeof(yyFallback[0])
              && (iFallback = yyFallback[iLookAhead])!=0 ){
         LOGLINE(QUERYPARSER,
-		"FALLBACK " << ParseTokenName(iLookAhead) << " => " <<
-		ParseTokenName(iFallback));
+    "FALLBACK " << ParseTokenName(iLookAhead) << " => " <<
+    ParseTokenName(iFallback));
```
ojwb: Here (and in several places below) your patch changes the indentation of lines you haven't otherwise modified.

If there are whitespace fixes to be made, it is better to make them as a separate commit rather than mixing them with functional changes.  In this case, these changes are not useful, so please revert them.  It makes for a smoother review process if you make such changes as "fixup" commits - see https://xapian-developer-guide.readthedocs.io/en/latest/contributing/contributing-changes.html#open-a-pull-request-on-github

xapian-core/queryparser/queryparser.lt
```diff
@@ -569,16 +569,14 @@ static void yyStackOverflow(yyParser *yypParser){
...
-			   "', pending reduce " << yyNewState - YY_MIN_REDUCE);
-    }
+  if( yyNewState<YYNSTATE ){
+    LOGLINE(QUERYPARSER, zTag << " '" <<
+     yyTokenName[yypParser->yystack.back().major] <<
+     "', go to state " << yyNewState);
+  }else{
+    LOGLINE(QUERYPARSER, zTag << " '" <<
+     yyTokenName[yypParser->yystack.back().major] <<
+     "', pending reduce " << yyNewState - YY_MIN_REDUCE);
```
ojwb: This code has changed, but also has similar indentation problems on the continuation lines now.  The first non-whitespace character of the continuation lines should line up under the `z` of `zTag` as it did before (in an editor at least; in github's diff viewer it appears to be off by one column because the first line a 4 space indent, but the continuation lines start with a tab, and so the leading `-` makes a difference).

uppinder:
The commits in this PR got discarded due to force push and also I had deleted the previous fork from which this PR was generated (extremely naive move from my side). I'm not sure how to regenerate the commits for this PR. If it's not possible, then, I request #186 to be chosen instead of #185.

xapian-core/queryparser/queryparser.lt
uppinder: The reason for the indentation issues is because earlier I had set my settings to: use tabs and tab size to 4 spaces. After asking James in IRC and checking HACKING, I had adjusted the settings accordingly (8 spaces). And as I started had afresh after deleting the fork, there are no fixup commits in #186.

xapian-core/queryparser/queryparser.lt
uppinder: Addressed in comment above.

PR #187, opened 2018-02-19
Custom size range

uppinder:
Basic implementation of allowing units in range values (eg., "size:1K..2K"). Involves a test case and implementation for the same. Basically, the range values are normalized to a base unit and handled in the same way as done by NumberRangeProcessor (eg., 1K -> 1024B). Currently, only supports byte as a unit.

### DIFF
--- a/xapian-core/api/valuerangeproc.cc
+++ b/xapian-core/api/valuerangeproc.cc
@@ -471,13 +471,12 @@ UnitRangeProcessor::operator()(const string& b, const string& e)
 	    // Not a valid byte unit
 	    goto not_our_range;
 	}
-	string b_ = string(b, 0, b.size() - 1);
 
 	errno = 0;
-	const char * startptr = b_.c_str();
+	const char * startptr = b.c_str();
 	char * endptr;
 	num_b = strtod(startptr, &endptr);
-	if (endptr != startptr + b_.size() || errno) {
+	if (endptr != startptr + b.size() - 1 || errno) {
 	    // Invalid characters in string || overflow or underflow.
 	    goto not_our_range;
 	}
@@ -493,13 +492,12 @@ UnitRangeProcessor::operator()(const string& b, const string& e)
     	    // Not a valid byte unit
 	    goto not_our_range;
 	}
-	string e_ = string(e, 0, e.size() - 1);
 
 	errno = 0;
-	const char * startptr = e_.c_str();
+	const char * startptr = e.c_str();
 	char * endptr;
 	num_e = strtod(startptr, &endptr);
-	if (endptr != startptr + e_.size() || errno) {
+	if (endptr != startptr + e.size() - 1 || errno) {
 	    // Invalid characters in string || overflow or underflow.
 	    goto not_our_range;
 	}
--- a/xapian-core/api/valuerangeproc.cc
+++ b/xapian-core/api/valuerangeproc.cc
@@ -465,46 +465,73 @@ UnitRangeProcessor::operator()(const string& b, const string& e)
     // Parse the numbers to floating point.
     double num_b, num_e;
 
-    if (!b.empty()) {
-	double factor_b = check_byte_unit(b);
-	if (factor_b == -1) {
-	    // Not a valid byte unit
-	    goto not_our_range;
-	}
+    // True if b has unit, e.g. 20K..
+    bool b_has_unit = false;
 
+    if (!b.empty()) {
 	errno = 0;
 	const char * startptr = b.c_str();
 	char * endptr;
 	num_b = strtod(startptr, &endptr);
-	if (endptr != startptr + b.size() - 1 || errno) {
-	    // Invalid characters in string || overflow or underflow.
+
+	if (errno) {
+	    // overflow or underflow
 	    goto not_our_range;
 	}
-	num_b *= factor_b;
+
+	// For lower range having a unit, e.g. 100K..
+	if (endptr == startptr + b.size() - 1) {
+	    double factor_b = check_byte_unit(b);
+	    if (factor_b == -1) {
+		// Not a valid byte unit
+		goto not_our_range;
+	    }
+	    b_has_unit = true;
+	    num_b *= factor_b;
+	}
     } else {
 	// Silence GCC warning.
 	num_b = 0.0;
     }
 
     if (!e.empty()) {
-	double factor_e = check_byte_unit(e);
-	if (factor_e == -1) {
-    	    // Not a valid byte unit
-	    goto not_our_range;
-	}
-
 	errno = 0;
 	const char * startptr = e.c_str();
 	char * endptr;
 	num_e = strtod(startptr, &endptr);
-	if (endptr != startptr + e.size() - 1 || errno) {
-	    // Invalid characters in string || overflow or underflow.
+
+	if (errno) {
+	    // overflow or underflow
 	    goto not_our_range;
 	}
-	num_e *= factor_e;
+
+	// For upper range having a unit, e.g. ..100K
+	if (endptr == startptr + e.size() - 1) {
+	    double factor_e = check_byte_unit(e);
+	    if (factor_e == -1) {
+		// Not a valid byte unit
+		goto not_our_range;
+	    }
+	    num_e *= factor_e;
+
+	    // When lower range is not empty and
+	    // only upper range unit, e.g. 20..100K
+	    if (!b.empty() && !b_has_unit) {
+		num_b *= factor_e;
+	    }
+	} else {
+	    // When lower range has no unit
+	    goto not_our_range;
+	}
     } else {
 	// Silence GCC warning.
 	num_e = 0.0;
+
+	// Fail case when lower range
+	// has no unit, e.g. 200..
+	if (!b.empty() && !b_has_unit) {
+	    goto not_our_range;
+	}
     }
 
     return RangeProcessor::operator()(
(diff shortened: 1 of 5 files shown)


ojwb:
All the CI builds fail with your changes.  Looking through:

The Linux builds are failing because you need to update the bindings for your new class:

https://travis-ci.org/xapian/xapian/jobs/343166075#L9881

The OS X build is failing for unrelated reasons (something changed in the travis OS X image it seems, but it's not clear to me what's wrong, and I don't have access to OS X to investigate).

And the coding style checking script finds a number of issues:

https://travis-ci.org/xapian/xapian/jobs/343166078#L447

The final one in `tests/api_queryparser.cc` you can probably ignore - we try to keep code line length to at most 80 columns, but a lot of the data arrays in that file already exceed that, and it wouldn't make the code any more readable to wrap them all.  However that line should have a space before the closing `}`.

uppinder:
Rectified the code style errors except that final one in tests/api_queryparser.cc, as suggested by you. 

Not sure where to update the bindings for the new class, would appreciate feedback on the same.

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,82 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+    // Parse the numbers to floating point.
+    double num_b, num_e;
+
+    if (!b.empty()) {
+	double factor_b = check_byte_unit(b);
+	if (factor_b == -1) {
+	    // Not a valid byte unit
+	    goto not_our_range;
+	}
+	string b_ = string(b, 0, b.size() - 1);
```
ojwb: There's no need to create a temporary string object here - `strtod()` will stop parsing when it reaches the units, so you can just work from `b` and adjust the end check to allow for the units.

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,82 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
     return Xapian::Query(Xapian::Query::OP_INVALID);
 }
 
+static const string byte_units[4] = {
+    "B", "K", "M", "G"
+};
```
ojwb: It's better to avoid `std::string` for constant string data like this - there's a fair bit of overhead to creating a string object which you don't have with a string constant, and in a case like this it's likely the compiler will generate code that actually initialises the strings in the array at runtime, whereas really we want such data to be read-only data created entirely at compile time.

The obvious alternative of an array of `const char*` is also problematic in a shared library because each pointer in such an array requires a load-time relocation, and the more of those there are, the slower the library loads.  Because of that we try to avoid adding more load-time relocations, and to eliminate existing ones.

In this case you should be able to declare this as an array of char arrays:

```
static const char byte_units[4][2] = {
    "B", "K", "M", "G"
};
```

That's also a nicely compact representation - at least on x86-64 Linux it's just 8 bytes (8 because each entry has a trailing nul byte).

xapian-core/api/valuerangeproc.cc
```diff
@@ -439,4 +439,82 @@ NumberRangeProcessor::operator()(const string& b, const string& e)
...
+	num_b = 0.0;
+    }
+
+    if (!e.empty()) {
+	double factor_e = check_byte_unit(e);
+	if (factor_e == -1) {
+    	    // Not a valid byte unit
+	    goto not_our_range;
+	}
+	string e_ = string(e, 0, e.size() - 1);
```
ojwb: Similarly here...

xapian-core/tests/api_queryparser.cc
```diff
@@ -1726,6 +1726,39 @@ DEFINE_TESTCASE(qp_range4, !backend) {
     return true;
 }
 
+static const test test_unitrange1_queries[] = {
+    { "financial report size:100K..1M", "((financial@1 OR report@2) FILTER VALUE_RANGE 1 \\xe0&@ \\xe04)" },
+    { NULL, NULL }
+};
+
+// Test chaining of UnitRangeProcessor class.
```
ojwb: This isn't testing chaining of this class, it's just a simple feature test for the class.

xapian-core/tests/api_queryparser.cc
```diff
@@ -1726,6 +1726,39 @@ DEFINE_TESTCASE(qp_range4, !backend) {
     return true;
 }
 
+static const test test_unitrange1_queries[] = {
+    { "financial report size:100K..1M", "((financial@1 OR report@2) FILTER VALUE_RANGE 1 \\xe0&@ \\xe04)" },
```
ojwb: This really needs more test cases.

Someone changing this code in the future could completely break support for `G` and `B` and this test would still pass.

uppinder: Added more test cases, including those for optional range units.

xapian-core/tests/api_queryparser.cc
uppinder: Updated comment.

xapian-core/api/valuerangeproc.cc
uppinder: Updated.

xapian-core/api/valuerangeproc.cc
uppinder: Updated.

xapian-core/api/valuerangeproc.cc
uppinder: Changes done accordingly.

xapian-core/api/valuerangeproc.cc
```diff
@@ -471,13 +471,12 @@ UnitRangeProcessor::operator()(const string& b, const string& e)
...
 	}
-	string b_ = string(b, 0, b.size() - 1);
 
 	errno = 0;
-	const char * startptr = b_.c_str();
+	const char * startptr = b.c_str();
 	char * endptr;
 	num_b = strtod(startptr, &endptr);
-	if (endptr != startptr + b_.size() || errno) {
+	if (endptr != startptr + b.size() - 1 || errno) {
```
uppinder: For clarity, the '-1' is because endptr points to the last character of b, which could be among the byte units "B", "K", etc. I forgot the comment here but soon after, this line of code is modified in another commit.

xapian-core/include/xapian/queryparser.h
```diff
@@ -389,6 +389,48 @@ class XAPIAN_VISIBILITY_DEFAULT NumberRangeProcessor : public RangeProcessor {
...
+     *
+     *  The string supplied in str_ is the prefix plus ":" that defines the field
+     *  upon which the range query is performed.
+     *
+     *  For example, if str_ is "size:", and the range
+     *  processor has been added to the queryparser, the queryparser will
+     *  accept "size:3K..10K" as a valid range.
+     */
+    UnitRangeProcessor(Xapian::valueno slot_,
+			 const std::string &str_ = std::string())
```
ojwb: This line's indented two columns too far (though the diff viewer makes it look like it's just one column too far).

uppinder: In NumberRangeProcessor's constructor, the c of const is aligned with X of Xapian::valueno, if I'm not wrong. If that's the case then I believe second line above is only one column too far. I've added a new commit moving back one column.

uppinder:
The OS X build is failing, which I think is unrelated to this PR.

xapian-core/include/xapian/queryparser.h
ojwb: The alignment here is now good, except it's using all spaces rather than being tab-filled, as `xapian-check-patch` points out:

```
xapian-core/include/xapian/queryparser.h:416: error: added/changed line uses spaces for indentation rather than tab: +                       const std::string &str_ = std::string())
```

The Xapian convention for C and C++ code is that the indent is done with a series of 0 or more tabs (each 8 columns wide), then a series of 0 or more spaces (each one column wide), using as many tabs as possible.

The difference isn't visible in many cases, but there are benefits to having a canonical representation when there's no visible difference - especially it means that you can't have two files which look the same but aren't, or at least if you do one of them must be "wrong".  That's very helpful when dealing with merging changes, because if everything follows the rules then you avoid merge conflicts due to whitespace changes in one version.

For similar reasons, source file lines shouldn't have trailing whitespace, and the final line of the file should end in a newline, but there shouldn't be further empty lines after it.

ojwb:
This is very close now.

Your commit messages are pretty good content-wise, but the formatting is a little odd.

You seem to capitalise most words in the first line of the message as if using title-case, but that's not the norm - generally you'd only capitalise the first word (and any which would naturally be capitalised, like proper nouns, symbols which start with a capital such as class names).  So "Remove temp variables for range values" instead of "Remove Temp Variables for Range Values", etc.

Also, you're wrapping the detailed part of the commit message (which is good) but to a much narrower width than is conventional - 72 columns is more usual.

(The above are both widely used norms for git commit messages, not specific to Xapian).

There's more on commit messages in our developer guide:

https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages

Don't worry about updating the existing commit messages in this case - the follow-on commits here are really better treated as fix-ups so in this case we'll squash the branch down to one commit to merge it.  But please take this on board for future commits.

xapian-core/include/xapian/queryparser.h
uppinder: Thanks for clarification regarding the indentation style; pushed fix-up commit.

ojwb:
Thanks for the updates, and sorry it took a while for me to get back to this.

ojwb:
I noticed I missed something - one of the new function bodies is over-indented by a level.  I've now pushed 6257cee2c03d72084158921e90351675f500e75b to fix that.

Unfortunately `xapian-check-patch` isn't yet smart enough to spot such issues.

uppinder:
Thanks for being patient and thoroughly reviewing this PR; this definitely helped me get acquainted with Xapian's codebase.

PR #189, opened 2018-03-01
Fix assertion error with hash command in scriptindex.

samuelharden:
* Currently hash command doesn't work without argument and get's assertion error.
* num_argument is taken as zero when argument is not given, which should be defaulted to max safe term length.
* hash command fails with default argument of zero and raises an exception.
* set the default to MAX_SAFE_TERM_LENGTH - 1 also when argument is not provided to hash command.

Orignal Error :
scriptindex: ../../hashterm.cc:65: std::string hash_long_term(const string&, unsigned int): Assertion `max_length >= HASH_LEN' failed.
Aborted

### DIFF
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -710,6 +710,16 @@ else
     failed=`expr $failed + 1`
 fi
 
+# Regression test to check hash works without argument, was failing earlier with ASSERTION error.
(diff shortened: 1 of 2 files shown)


ojwb:
Thanks.

There was actually code to handle this case but in the wrong place!

The commit meta-data here is screwy (bad author, bad committer, and a badly formatted commit message), so I've applied these changes by hand and removed the code that was in the wrong place.

PR #190, opened 2018-03-02
Support latest homebrew python in travis

jaylett:
homebrew has now completed its switch to have python3 as the default
python; that means that we need to change some package and command
references. Also, it means that python2 isn't linked into the path
when you install via homebrew, so we have to find it explicitly in
its install location by setting PYTHON2 appropriately.


ojwb:
Thanks.  I merged this on Friday, but github didn't notice so closing.

PR #192, opened 2018-03-13
 Handling negative indices when maxlen <= ind.size()

addy007-icy:
Setting output to an empty string to avoid negative index reference. 
Ticket link: https://trac.xapian.org/ticket/754

### DIFF
--- a/xapian-applications/omega/sample.cc
+++ b/xapian-applications/omega/sample.cc
@@ -50,12 +50,12 @@ generate_sample(const string & input, size_t maxlen,
 	if (output.size() >= maxlen) {
 	    // Need to truncate output.
 	    if (last_word_end <= maxlen / 2) {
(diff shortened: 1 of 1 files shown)


samuelharden:
Travis build also fails with:

`$ git diff `[ "$TRAVIS_PULL_REQUEST" = false ] && echo master || echo "$TRAVIS_BRANCH"`...|xapian-maintainer-tools/xapian-check-patch
xapian-applications/omega/sample.cc:53: error: added/changed line has space before tab: +	    	if (maxlen - ind.size() <= 0) {
xapian-applications/omega/sample.cc:54: error: added/changed line has space before tab: +	   		output.resize(0);
error count:	2
`

samuelharden:
If you are solving one of the bugs, it will also be great to reference that in commit message

PR #193, opened 2018-03-13
BugFix: #754 Exception in generate_sample with 0 as sample size

addy007-icy:
Handling the case when maxlen <= ind.size() by returning an empty string to avoid referencing negative indices.


ojwb:
Closing this - please keep this in PR#191.

PR #194, opened 2018-03-21
Check document length bounds against document length

uppinder:
Cross-check the document length upper and lower bounds with the
document length in xapian-check.

### DIFF
--- a/xapian-core/backends/glass/glass_dbcheck.cc
+++ b/xapian-core/backends/glass/glass_dbcheck.cc
@@ -742,7 +742,7 @@ check_glass_table(const char * tablename, const string &db_dir, int fd,
 	    if (doclen < version_file.get_doclength_lower_bound() ||
 		doclen > version_file.get_doclength_upper_bound()) {
 		if (out) {
-		    *out << "doclen not within bounds";
+		    *out << "doclen not within bounds" << endl;
 		}
 		++errors;
 		continue;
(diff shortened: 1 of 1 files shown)


xapian-core/backends/glass/glass_dbcheck.cc
```diff
@@ -738,6 +738,16 @@ check_glass_table(const char * tablename, const string &db_dir, int fd,
 		continue;
 	    }
 
+	    // Check doclen with doclen lower and upper bounds
+	    if (doclen < version_file.get_doclength_lower_bound() ||
+		doclen > version_file.get_doclength_upper_bound()) {
+		if (out) {
+		    *out << "doclen not within bounds";
```
ojwb: Thanks.

This looks good, except that this really should write `endl` too so that the message is on its own line rather than running into whatever is output next.  Existing similar messages do this.

ojwb:
Thanks, merged.

One little note - when there's already a ticket for the issue being addressed it's a good idea to reference that ticket in the commit message - e.g. with a line which says:

```
Closes https://trac.xapian.org/ticket/617
```

That allows people to later find the ticket from the commit if they want more background on a change (for example, if there turns out to be a problem with it).  Often bug trackers actually monitor commit messages and will automatically link to a commit referencing the ticket and even automatically close it if the reference has "Closes" in front.  We haven't set that up for our trac so far, but the reference is still useful for humans in the future.

I added such a note to the merge commit's message in this case, which will do.

PR #195, opened 2018-04-01
Extend test of behavior of database objects after database close

guruhegde:
Changes:
Added tests to cover termlist, alltermlist, positionlist iterator. Ran the api test on all backend types. No failure. 
I will update this PR with additional tests w.r.t ticket #337.

Questions:
* Should I squash the commits to make single one?
* It is mentioned in the comment that iterator increment may not raise exception. Is it because block of values will be fetched from database when iterator object is assigned with begin method?

### DIFF
--- a/xapian-core/tests/api_closedb.cc
+++ b/xapian-core/tests/api_closedb.cc
@@ -307,7 +307,7 @@ DEFINE_TESTCASE(closedb5, transactions) {
 	Xapian::WritableDatabase wdb = get_writable_database();
 	wdb.close();
 	TEST_EXCEPTION(Xapian::InvalidOperationError,
-		wdb.commit_transaction());
+		       wdb.commit_transaction());
 	try {
 	    wdb.begin_transaction();
 	} catch (const Xapian::DatabaseError &) {
--- a/xapian-core/tests/api_closedb.cc
+++ b/xapian-core/tests/api_closedb.cc
@@ -325,6 +325,11 @@ DEFINE_TESTCASE(closedb7, writable) {
     // is a no-op, and so doesn't have to fail.  Similarly we may be able to
     // call db.begin_transaction(), but we can't make any changes inside that
     // transaction.
+    try {
+	db.commit();
+    } catch (const Xapian::DatabaseError &) {
+    }
+
     TEST_EXCEPTION(Xapian::DatabaseError,
 		   db.add_document(Xapian::Document()));
     TEST_EXCEPTION(Xapian::DatabaseError,
--- a/xapian-core/tests/api_closedb.cc
+++ b/xapian-core/tests/api_closedb.cc
@@ -51,6 +51,8 @@ struct closedb1_iterators {
     Xapian::PostingIterator pl1;
     Xapian::PostingIterator pl2;
     Xapian::PostingIterator plend;
+    Xapian::TermIterator tl1;
+    Xapian::TermIterator tlend;
 
     void setup(Xapian::Database db_) {
 	db = db_;
@@ -60,6 +62,8 @@ struct closedb1_iterators {
 	pl2 = db.postlist_begin("paragraph");
 	++pl2;
 	plend = db.postlist_end("paragraph");
+	tl1 = db.termlist_begin(1);
+	tlend = db.termlist_end(1);
     }
 
     int perform() {
@@ -95,12 +99,17 @@ struct closedb1_iterators {
 	COUNT_CLOSEDEXC(db.reopen());
 
 	TEST_NOT_EQUAL(pl1, plend);
+	TEST_NOT_EQUAL(tl1, tlend);
 
 	COUNT_CLOSEDEXC(db.postlist_begin("paragraph"));
 
 	COUNT_CLOSEDEXC(TEST_EQUAL(*pl1, 1));
 	COUNT_CLOSEDEXC(TEST_EQUAL(pl1.get_doclength(), 28));
 	COUNT_CLOSEDEXC(TEST_EQUAL(pl1.get_unique_terms(), 21));
+
+	COUNT_CLOSEDEXC(TEST_EQUAL(*tl1, "a"));
+	COUNT_CLOSEDEXC(TEST_EQUAL(tl1.get_wdf(), 2));
+	COUNT_CLOSEDEXC(TEST_EQUAL(tl1.get_termfreq(), 3));
 
 	// Advancing the iterator may or may not raise an error, but if it
 	// doesn't it must return the correct answers.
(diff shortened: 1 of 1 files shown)


ojwb:
> Should I squash the commits to make single one?

The commit boundaries seem good to me.

Our dev guide has some [useful pointers](https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commits) on this.  In this case you can think of each commit you've made as a single complete change, though all of them together also arguably make a single complete change.  The finer granularity has some advantages (e.g. someone later tracking down where a bug was introduced by using `git bisect` will end up with a smaller change to have to manually look through) so that looks good to me.

If you were in a situation where combining the commits made sense, it's better to make the amendments as "fixup" commits (see the dev guide for more details) which the reviewer can squash down automatically before merging - if you as the submitter squash commits together and rewrite history that makes things harder for reviewers as they can't trivially see the incremental changes made in response to review comments.  Generally you shouldn't rewrite history unilaterally - only do it if you've not pushed the branch publicly or if there's clear agreement to rewrite history from everyone using the branch.

> It is mentioned in the comment that iterator increment may not raise exception. Is it because block of values will be fetched from database when iterator object is assigned with begin method?

Essentially, though it may be that data is read in batches rather than only by the begin method.  The iterator doesn't actually see that the database has been closed until it tries to read the next batch.  We could make the iterator explicitly check with its associated database object if it has been closed before every operation, but that's a lot of essentially pointless checking which adds overhead.  So instead the semantics we specify are that you'll either get the answer you would have got just before the database was closed or an exception saying the database is closed.  This is true for any object which internally links to a database object, not just iterators.  In some cases, it may be due to caching rather than batched data.

> I will update this PR with additional tests w.r.t ticket #337.

OK - I'll hold off merging for now.

guruhegde:
Thanks for the tips on commit boundaries and telling about batched read. 

In #337, commit() for writable database, How can I test the behavior?  
For remote type, TEST_EXCEPTION() can be used.  How to test no op behavior? 

Similarly for transaction API. Any advice?

guruhegde:
I tried this, not sure whether this is the way.  

`+    if (startswith(get_dbtype(), "remote")) {
+       TEST_EXCEPTION(Xapian::DatabaseError,
+                      db.commit());
+       TEST_EXCEPTION(Xapian::DatabaseError,
+                      db.begin_transaction(true));
+    } else {
+       try {
+           db.commit();
+       } catch (const Xapian::DatabaseError &) {
+           return false;
+       }
+    }
`

xapian-core/tests/api_closedb.cc
```diff
@@ -242,6 +299,20 @@ DEFINE_TESTCASE(closedb5, transactions) {
...
     }
+
+    {
+	// commit_transaction() throws InvalidOperationError when
+	// not in a transaction. begin_transaction() is no-op or
+	// throws DatabaseError
+	Xapian::WritableDatabase wdb = get_writable_database();
+	wdb.close();
+	TEST_EXCEPTION(Xapian::InvalidOperationError,
+		wdb.commit_transaction());
```
ojwb: Xapian coding style is to align just after the `(` on the line above in a case like this (as is done in every existing wrapped `TEST_EXCEPTION(...)` in this file).

xapian-core/tests/api_closedb.cc
```diff
@@ -242,6 +299,20 @@ DEFINE_TESTCASE(closedb5, transactions) {
...
+	try {
+	    wdb.begin_transaction();
```
ojwb: If `begin_transaction()` succeeds it would be good to test that `commit_transaction()` does something appropriate.

xapian-core/tests/api_closedb.cc
```diff
@@ -242,6 +299,20 @@ DEFINE_TESTCASE(closedb5, transactions) {
...
+	} catch (const Xapian::DatabaseError &) {
+	}
+    }
```
ojwb: Equivalent tests to those added above could be done with `cancel_transaction()` in place of `commit_transaction()`.

xapian-core/tests/api_closedb.cc
```diff
@@ -242,6 +299,20 @@ DEFINE_TESTCASE(closedb5, transactions) {
 	Xapian::Database db = get_writable_database_as_database();
 	TEST_EQUAL(db.get_doccount(), 0);
     }
+
```
ojwb: The additions below don't really fit the current description of this testcase: "If a transaction is active, close() shouldn't implicitly commit()."

I would suggest they'd fit better in closedb7 (description: "Test WritableDatabase methods") where there's even an existing comment that discusses how `begin_transaction()` should behave, except the condition on closedb7 doesn't include `transactions`.

I think probably move the "If a transaction is active, close() shouldn't implicitly commit()." comment to the first block, add a more generic comment for the testcase such as "Test the effects of close() on transactions" and relocate the part of the comment in closedb7 that talks about `begin_transaction()` to the comment block below.

xapian-core/tests/api_closedb.cc
guruhegde: Done

xapian-core/tests/api_closedb.cc
guruhegde: Checked begin_transaction() returns successfully, based on that commit_transaction() check is done.  Let me know if it needs any update.

xapian-core/tests/api_closedb.cc
guruhegde: Done

xapian-core/tests/api_closedb.cc
guruhegde: Addressed the changes suggested for comments. Let me know if all ok :)

xapian-core/tests/api_closedb.cc
```diff
@@ -49,7 +49,6 @@ struct closedb1_iterators {
     Xapian::Database db;
     Xapian::Document doc1;
     Xapian::PostingIterator pl1;
-    Xapian::PostingIterator pl2;
```
guruhegde: I couldn't see why pl2 is present. So removed. Let me know if it's there for some purpose, I will undo this change.

xapian-core/tests/api_closedb.cc
guruhegde: Sorry about the style error. I keep forgetting to run style check script locally.  Will take care of this in future.

xapian-core/tests/api_closedb.cc
ojwb: Good spot.  The history isn't very informative (it was added like this) but I think the idea was to test an iterator that had been advanced before the database was closed (as well as `pl1` which hasn't been).  That seems useful as it might affect the behaviour, but to be useful we need to actually check `pl2` in similar ways to `pl1` below (just with different values in the checks because it is one further along).

PR #197, opened 2018-04-10
Add C++11 move support for smart pointer classes

guruhegde:
Address issue #642 -
* Add move support for intrusive_ptr, intrusive_ptr_nonnull and opt_intrusive_ptr.
* Add unit tests

### DIFF
--- a/xapian-core/include/xapian/intrusive_ptr.h
+++ b/xapian-core/include/xapian/intrusive_ptr.h
@@ -236,7 +236,33 @@ public:
 
     ~intrusive_ptr_nonnull()
     {
-        if(--px->_refs == 0 ) delete px;
+        if(px != 0 && --px->_refs == 0 ) delete px;
+    }
+
+    intrusive_ptr_nonnull(intrusive_ptr_nonnull && rhs) : px( rhs.px )
+    {
+        rhs.px = 0;
+    }
+
+    intrusive_ptr_nonnull & operator=(intrusive_ptr_nonnull && rhs)
+    {
+        this_type( static_cast< intrusive_ptr_nonnull && >( rhs ) ).swap(*this);
+        return *this;
+    }
+
+    template<class U> friend class intrusive_ptr_nonnull;
+
+    template<class U>
+    intrusive_ptr_nonnull(intrusive_ptr_nonnull<U> && rhs) : px( rhs.px )
+    {
+        rhs.px = 0;
+    }
+
+    template<class U>
+    intrusive_ptr_nonnull & operator=(intrusive_ptr_nonnull<U> && rhs)
+    {
+        this_type( static_cast< intrusive_ptr_nonnull<U> && >( rhs ) ).swap(*this);
+        return *this;
     }
 
     intrusive_ptr_nonnull & operator=(intrusive_ptr_nonnull const & rhs)
--- a/xapian-core/include/xapian/intrusive_ptr.h
+++ b/xapian-core/include/xapian/intrusive_ptr.h
@@ -430,6 +430,35 @@ public:
 	if( counting && --px->_refs == 1 ) delete px;
     }
 
+    opt_intrusive_ptr(opt_intrusive_ptr && rhs)
+    : px( rhs.px ), counting( rhs.counting )
+    {
+        rhs.px = 0;
+	rhs.counting = 0;
+    }
+
+    opt_intrusive_ptr & operator=(opt_intrusive_ptr && rhs)
+    {
+        this_type( static_cast< opt_intrusive_ptr && >( rhs ) ).swap(*this);
+        return *this;
+    }
+
+    template<class U> friend class opt_intrusive_ptr;
+
+    template<class U>
+    opt_intrusive_ptr(opt_intrusive_ptr<U> && rhs)
+    : px( rhs.px ), counting( rhs.counting )
+    {
+        rhs.px = 0;
+	rhs.counting = 0;
+    }
+
+    template<class U>
+    opt_intrusive_ptr & operator=(opt_intrusive_ptr<U> && rhs)
+    {
+        this_type( static_cast< opt_intrusive_ptr<U> && >( rhs ) ).swap(*this);
+        return *this;
+    }
     opt_intrusive_ptr & operator=(opt_intrusive_ptr const & rhs)
     {
 	this_type(rhs).swap(*this);
--- a/xapian-core/include/xapian/intrusive_ptr.h
+++ b/xapian-core/include/xapian/intrusive_ptr.h
@@ -113,6 +113,32 @@ public:
         if( px != 0 && --px->_refs == 0 ) delete px;
     }
 
+    intrusive_ptr(intrusive_ptr && rhs) : px( rhs.px )
+    {
+        rhs.px = 0;
+    }
+
+    intrusive_ptr & operator=(intrusive_ptr && rhs)
+    {
+        this_type( static_cast< intrusive_ptr && >( rhs ) ).swap(*this);
+        return *this;
+    }
+
+    template<class U> friend class intrusive_ptr;
+
+    template<class U>
+    intrusive_ptr(intrusive_ptr<U> && rhs) : px( rhs.px )
+    {
+        rhs.px = 0;
+    }
+
+    template<class U>
+    intrusive_ptr & operator=(intrusive_ptr<U> && rhs)
+    {
+        this_type( static_cast< intrusive_ptr<U> && >( rhs ) ).swap(*this);
+        return *this;
+    }
+
     intrusive_ptr & operator=(intrusive_ptr const & rhs)
     {
         this_type(rhs).swap(*this);
--- a/xapian-core/include/xapian/intrusive_ptr.h
+++ b/xapian-core/include/xapian/intrusive_ptr.h
@@ -236,12 +236,12 @@ public:
 
     ~intrusive_ptr_nonnull()
     {
-        if(px != 0 && --px->_refs == 0 ) delete px;
+        if(--px->_refs == 0 ) delete px;
     }
 
     intrusive_ptr_nonnull(intrusive_ptr_nonnull && rhs) : px( rhs.px )
     {
-        rhs.px = 0;
+        ++px->_refs;
     }
 
     intrusive_ptr_nonnull & operator=(intrusive_ptr_nonnull && rhs)
@@ -255,7 +255,7 @@ public:
     template<class U>
     intrusive_ptr_nonnull(intrusive_ptr_nonnull<U> && rhs) : px( rhs.px )
     {
-        rhs.px = 0;
+        ++px->_refs;
     }
 
     template<class U>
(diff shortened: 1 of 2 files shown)


xapian-core/include/xapian/intrusive_ptr.h
```diff
@@ -236,7 +236,33 @@ template<class T> class intrusive_ptr_nonnull
 
     ~intrusive_ptr_nonnull()
     {
-        if(--px->_refs == 0 ) delete px;
+        if(px != 0 && --px->_refs == 0 ) delete px;
+    }
+
+    intrusive_ptr_nonnull(intrusive_ptr_nonnull && rhs) : px( rhs.px )
+    {
+        rhs.px = 0;
```
guruhegde: @ojwb Here for source object invariant is lost, but the object is going to be destructed. That's ok right?

ojwb:
The macOS travis build failed for external reasons, so I've restarted it.

xapian-core/include/xapian/intrusive_ptr.h
ojwb: That's an interesting question.

In typical use the object will soon be destroyed, but that's not a requirement when using `std::move()`.

http://en.cppreference.com/w/cpp/language/move_constructor says:

> Move constructors typically "steal" the resources held by the argument (e.g. pointers to dynamically-allocated objects, file descriptors, TCP sockets, I/O streams, running threads, etc.) rather than make copies of them, and leave the argument in some valid but otherwise indeterminate state.

So that suggests it's not conventional to leave the object in an invalid state like this.

I think it's probably safer to just increment the reference count here (i.e. make it do the same as the copy constructor).  That's not a big overhead, and the compiler might even be able to optimise away the increment and decrement if the object is immediately destroyed.

Your move assignment (which essentially does `swap()`) is fine, as that preserves the invariant.

xapian-core/include/xapian/intrusive_ptr.h
```diff
@@ -378,6 +430,35 @@ template<class T> class opt_intrusive_ptr
...
+	rhs.counting = 0;
+    }
+
+    template<class U>
+    opt_intrusive_ptr & operator=(opt_intrusive_ptr<U> && rhs)
+    {
+        this_type( static_cast< opt_intrusive_ptr<U> && >( rhs ) ).swap(*this);
+        return *this;
+    }
     opt_intrusive_ptr & operator=(opt_intrusive_ptr const & rhs)
```
ojwb: It'd be better to have a blank line between methods here (as elsewhere in this file).

xapian-core/tests/unittest.cc
```diff
@@ -680,6 +682,86 @@ static bool test_uuid1()
     return true;
 }
 
+// Classes used by movesupport1 test
+class A : public Xapian::Internal::intrusive_base {
+    int x = 0;
+  public:
+    A(int x_) : x(x_) {}
```
ojwb: Unless you want the compiler to implicitly convert `int` to `A`, it's a good idea to mark constructors which take a single parameter (either explicitly, or via default parameter values) as `explicit`.  Similarly for `B` below.

It's not an issue as the code is now, but if we're consistent about it we're more likely to remember to do it when it matters, and also won't get bitten by it suddenly starting to matter as the code evolves.

xapian-core/tests/unittest.cc
```diff
@@ -680,6 +682,86 @@ static bool test_uuid1()
...
+	// Same test for opt_intrusive_ptr class
+	B * b1 = new B{5};
+	b1->release();
+	Xapian::Internal::opt_intrusive_ptr<B> p1(b1);
+	Xapian::Internal::opt_intrusive_ptr<B> p3;
+
+	// Test move constructor
+	Xapian::Internal::opt_intrusive_ptr<B> p2(std::move(p1));
+	TEST_EQUAL(p2->get_x(), 5);
+	TEST_EQUAL(p2->_refs, 2);
```
ojwb: The `_refs` member is really an implementation detail, so it's violating the abstraction to look at it here.  It's only `public` at all for convenience of implementation.

If it is useful to inspect, I'd suggest that instead you do `auto before_refs = p1->_refs` before attempting to invoke the move constructor, and then `TEST_EQUAL(p2->_refs, before_refs);` here to test that the reference count hasn't changed.  That at least avoids encoding knowledge here about what the value means exactly.

But I'd think you could have `B`'s constructor take a `bool&` parameter which its destructor sets to false, and use that to test that the object hasn't been deleted at the end of this block, but has been just after this block.  That should ensure that we catch any bugs in updating the reference count in the move constructor or move assignment.

ojwb:
> I want to add move support for other classes as well. Should I work in this PR or make another one once ready?

I think this PR is close to mergable, so I'd suggest making another branch off this one to put work on other classes into.  Once this is merged, rebase that branch onto the latest master and then open a PR.

> Also I am not sure that's the place to add tests(or even tests are needed).

It's good to have some tests, and I think you're adding appropriate tests and that `unittest` is a sensible place for them.

xapian-core/include/xapian/intrusive_ptr.h
guruhegde: Done

xapian-core/tests/unittest.cc
guruhegde: I agree. Updated. Since B takes two arguments now, explicit keyword not added.

xapian-core/tests/unittest.cc
guruhegde: > But I'd think you could have B's constructor take a bool& parameter which its destructor sets to false,
I followed this. It's better. Removed check for _refs value.

xapian-core/include/xapian/intrusive_ptr.h
guruhegde: You are right, thanks! code is updated.

xapian-core/tests/unittest.cc
```diff
@@ -732,33 +741,34 @@ static bool test_movesupport1()
 	// Test move constructor
 	Xapian::Internal::intrusive_ptr_nonnull<A> p2(std::move(p1));
 	TEST_EQUAL(p2->get_x(), 5);
-	TEST_EQUAL(p1.get(), 0);
 
 	// Test move assignment
 	p3 = std::move(p2);
 	TEST_EQUAL(p3->get_x(), 5);
-	TEST_EQUAL(p2.get(), 0);
```
ojwb: Why did you remove these?

Checking `get()` is OK - it's just `_refs` that's not an implementation detail (and you still check `get()` elsewhere.

ojwb:
The appveyor failures are unrelated - I've recently updated `bootstrap` to download tarballs for the autotools from github to try to eliminate ocassional CI failures due to the mirrors we were using being unavailable, but there's an issue with the SSL certificates from github not being handled by the rather old `wget` in the appveyor mingw installed.  Please just ignore those for now.

I think this is ready, aside from the removal of two tests of `get()` in the latest changes, which I don't understand.

xapian-core/tests/unittest.cc
guruhegde: Sorry, I forgot to mention this change. Thanks for pointing this.

Now test equality doesn't hold. 

It's clear in move constructor case - `TEST_EQUAL(p1.get(), 0)`, because `NULL` assignment not done.

####first version
I should've asked this before. I couldn't understand how equality worked first time in move assignment case - `TEST_EQUAL(p2.get(), 0)`. If move assignment really swapped, then `p2.get()` contain what p3 had before assignment. That didn't happen. How `p2.get()` gives 0?

I tried tracing in gdb. It calls assignment operator. It seems there is some optimization. Step through gdb was very confusing, I couldn't get much. Any advice how to debug?

####latest version
With the latest commit, for move assignment `TEST_EQUAL(p2.get(), 0)` doesn't hold, instead  `TEST_EQUAL(p2.get(), p3.get())` holds. It seems swap is not happening really.  Just one way assignment done.

xapian-core/tests/unittest.cc
```diff
@@ -686,19 +686,29 @@ static bool test_uuid1()
...
 	return x;
     }
 };
 
 class B : public Xapian::Internal::opt_intrusive_base {
     int x = 0;
+    bool & alive;
   public:
-    B(int x_) : x(x_) {}
+    B(int x_, bool & alive_) : x(x_), alive(alive_) {
```
guruhegde: I realised, initialising variable `alive` in constructor may not be the right way. It causes no problem here. If  `class B` is made to modify `alive` only in destructor, it could be clear. I assign `true` to `alive` after object  `b1` is created.  

It is trivial, but seems better.  Should I make this change?

ojwb: We do it roughly like you do elsewhere, e.g. `DestroyedFlag` in `api_none.cc` and that seems OK to me.

Why are you thinking it's better to not handle this in the class?

Also, it probably would be a good idea to delete the copy ctor and assignment operator, to make sure that we don't get a copy created implicitly (which would share the reference and the flag would get cleared when the first copy was destroyed).

xapian-core/tests/unittest.cc
ojwb: Oh I see - you had to change this because the move operations for the nonnull case now actually copy rather than move.  That's OK, but I think deserves a comment to explain why this differs in this way from the `intrusive_ptr` test above.

Not sure about the other question - I'll take a look.

xapian-core/tests/unittest.cc
guruhegde: If new object got created with same destroy flag reference, tests will not be valid, like if new object sets the flag, if old object cleared it.  You can ignore this, it will not happen:)

I will delete copy constructor and assignment like you said.

xapian-core/tests/unittest.cc
ojwb: I think that's good - it's efficient and leaves `rhs` in a valid state, so it seems a valid approach.  And with the normal uses of `std:;move()` we should destroy `rhs` soon afterwards.

guruhegde:
I felt I need more experience with the library internals to handle the move support for other classes. I have some free time, I can look at some other issue or take up another task. 
If you recommend any that would be helpful.

ojwb:
I don't think that really needs much additional knowledge - the common cases (e.g. `Database`, `Enquire`, `MSet`, etc just have a single member which is `Xapian::intrusive_ptr<Internal> internal;` (or non-null variant) it's a matter of adding the move ctor and operator to just do `internal = std::move(o.internal)` or similar.

The implementations probably can't be in the header in most cases, since the `Internal` class is usually only defined inside the library (there's the odd exception, like `Query`).

But feel free to have a look at the other `GoodFirstBug` tickets to see if there's anything of interest.

guruhegde:
Thanks, I had followed exactly what you suggested. As I was trying to understand internals for classes(Especially in matcher folder), I thought I delay this a bit.

I had done changes to Database & WritableDatabase class. Tcl binding was not compiling successfully.  I will mention details in new PR soon.

PR #198, opened 2018-04-15
Diversification Module: GLS-MPT

uppinder:
Contains implementation of GLS with MPT as the evaluation function for diversification.

Pending tasks:
- [x] Pre-compute pairwise cosine similarities.
- [x] Add tests: Working and performance test for diversification.

### DIFF
--- a/xapian-core/diversify/Makefile.mk
+++ b/xapian-core/diversify/Makefile.mk
@@ -1,0 +1,3 @@
+lib_src +=\
+	diversify/diversify.cc
+
--- a/xapian-core/diversify/Makefile.mk
+++ b/xapian-core/diversify/Makefile.mk
@@ -1,3 +1,2 @@
 lib_src +=\
 	diversify/diversify.cc
-
--- a/xapian-core/diversify/diversify.cc
+++ b/xapian-core/diversify/diversify.cc
@@ -50,7 +50,7 @@ Diversify::get_description() const
 }
 
 void
-Diversify::initialise_points(const MSet &source)
+Diversify::initialise_points(const MSet& source)
 {
     LOGCALL_VOID(API, "Diversify::initialise_points", source);
     unsigned int count = 0;
@@ -67,7 +67,7 @@ Diversify::initialise_points(const MSet &source)
 }
 
 pair<Xapian::docid, Xapian::docid>
-Diversify::get_key(const Xapian::docid &docid_a, const Xapian::docid &docid_b)
+Diversify::get_key(const Xapian::docid& docid_a, const Xapian::docid& docid_b)
 {
     pair<Xapian::docid, Xapian::docid> key;
     if (docid_a <= docid_b) {
@@ -103,7 +103,7 @@ Diversify::compute_similarities()
 }
 
 vector<Xapian::docid>
-Diversify::compute_diff_dmset(const vector<Xapian::docid> &dmset)
+Diversify::compute_diff_dmset(const vector<Xapian::docid>& dmset)
 {
     LOGCALL(API, vector<Xapian::docid>, "Diversify::compute_diff_dmset", dmset);
     vector<Xapian::docid> diff_dmset;
@@ -126,7 +126,7 @@ Diversify::compute_diff_dmset(const vector<Xapian::docid> &dmset)
 }
 
 double
-Diversify::evaluate_dmset(const vector<Xapian::docid> &dmset)
+Diversify::evaluate_dmset(const vector<Xapian::docid>& dmset)
 {
     LOGCALL(API, double, "Diversify::evaluate_dmset", dmset);
     double score_1 = 0, score_2 = 0;
@@ -153,7 +153,7 @@ Diversify::evaluate_dmset(const vector<Xapian::docid> &dmset)
 }
 
 DocumentSet
-Diversify::get_dmset(const MSet &mset)
+Diversify::get_dmset(const MSet& mset)
 {
     LOGCALL(API, MSet, "Diversify::get_dmset", mset);
     // Return original mset if no need to diversify
--- a/xapian-core/diversify/diversify.cc
+++ b/xapian-core/diversify/diversify.cc
@@ -22,14 +22,14 @@
 #include <config.h>
 
 #include "xapian/diversify.h"
-
 #include "xapian/error.h"
+
 #include "debuglog.h"
 
-#include <vector>
-#include <limits>
 #include <algorithm>
 #include <cmath>
+#include <limits>
+#include <vector>
 
 using namespace Xapian;
 using namespace std;
--- a/xapian-core/diversify/diversify.cc
+++ b/xapian-core/diversify/diversify.cc
@@ -144,6 +144,10 @@ DocumentSet
 Diversify::get_dmset(const MSet &mset)
 {
     LOGCALL(API, MSet, "Diversify::get_dmset", mset);
+    // Nothing to do
+    if (mset.empty())
+	throw InvalidArgumentError("Given mset should not be empty");
+
     // Return original mset if no need to diversify
     if (k == 0) {
 	DocumentSet dmset;
--- a/xapian-core/diversify/diversify.cc
+++ b/xapian-core/diversify/diversify.cc
@@ -144,12 +144,8 @@ DocumentSet
 Diversify::get_dmset(const MSet &mset)
 {
     LOGCALL(API, MSet, "Diversify::get_dmset", mset);
-    // Nothing to do
-    if (mset.empty())
-	throw InvalidArgumentError("Given mset should not be empty");
-
     // Return original mset if no need to diversify
-    if (k == 0) {
+    if (k == 0 || mset.size() <= 2) {
 	DocumentSet dmset;
 	for (MSetIterator it = mset.begin(); it != mset.end(); ++it)
 	    dmset.add_document(it.get_document());
--- a/xapian-core/diversify/diversify.cc
+++ b/xapian-core/diversify/diversify.cc
@@ -183,12 +183,12 @@ Diversify::get_dmset(const MSet &mset)
 
     // Merge main_dmset and diff_dmset into final dmset
     DocumentSet dmset;
-    for (Point doc : main_dmset) {
+    for (auto doc : main_dmset) {
 	dmset.add_document(doc.get_document());
     }
 
     vector<Point> diff_dmset = compute_diff_dmset(main_dmset);
-    for (auto doc : main_dmset)
+    for (auto doc : diff_dmset)
 	dmset.add_document(doc.get_document());
 
     return dmset;
--- a/xapian-core/diversify/diversify.cc
+++ b/xapian-core/diversify/diversify.cc
@@ -1,0 +1,176 @@
+/** @file diversify.cc
+ *  @brief Diversification API
+ */
+/* Copyright (C) 2018 Uppinder Chugh
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
+#include <config.h>
+
+#include "xapian/diversify.h"
+
+#include "xapian/error.h"
+#include "debuglog.h"
+
+#include <vector>
+#include <limits>
+#include <algorithm>
+#include <cmath>
+
+using namespace Xapian;
+using namespace std;
+
+Diversify::Diversify(unsigned int k_,
+		     double lambda_,
+		     double b_,
+		     double sigma_sqr_)
+    : k(k_), lambda(lambda_), b(b_), sigma_sqr(sigma_sqr_)
+{
+    LOGCALL_CTOR(API, "Diversify", k_ | lambda_ | b_ | sigma_sqr_);
+}
+
+string
+Diversify::get_description() const
+{
+    return "Diversify()";
+}
+
+void
+Diversify::initialise_points(const MSet &source)
+{
+    LOGCALL_VOID(API, "Diversify::initialise_points", source);
+    TermListGroup tlg(source);
+    for (MSetIterator it = source.begin(); it != source.end(); ++it) {
+	points.push_back(Point(tlg, it.get_document()));
+	weights[it.get_document().get_docid()] = it.get_weight();
+    }
+}
+
+vector<Point>
+Diversify::compute_diff_dmset(const std::vector<Point> &dmset)
+{
+    LOGCALL(API, vector<Point>, "Diversify::compute_diff_dmset", dmset);
+    vector<Point> diff_dmset;
+    for (auto point : points) {
+	bool found_point = false;
+	Xapian::docid point_docid = point.get_document().get_docid();
+	for (auto doc : dmset) {
+	    if (point_docid == doc.get_document().get_docid()) {
+		found_point = true;
+		break;
+	    }
+	}
+
+	if (!found_point) {
+	    diff_dmset.push_back(point);
+	}
+    }
+
+    return diff_dmset;
+}
+
+double
+Diversify::evaluate_dmset(const vector<Point> &dmset)
+{
+    LOGCALL(API, double, "Diversify::evaluate_dmset", dmset);
+    double score_1 = 0, score_2 = 0;
+
+    for (auto doc : dmset)
+	score_1 += weights[doc.get_document().get_docid()];
+
+    vector<Point> diff_dmset = compute_diff_dmset(dmset);
+
+    Xapian::CosineDistance d;
+    for (auto point : diff_dmset) {
+	double min_dist = numeric_limits<double>::max();
+	unsigned int pos = 1;
+	for (auto doc : dmset) {
+	    double sim = d.similarity(point, doc);
+	    double weight = 2 * b * sigma_sqr * (1 / log(1 + pos)) * (1 - sim);
+	    min_dist = min(min_dist, weight);
+	    ++pos;
+	}
+	score_2 += min_dist;
+    }
+
+    return -lambda * score_1 + (1 - lambda) * score_2;
+}
+
+DocumentSet
+Diversify::get_dmset(const MSet &mset)
+{
+    LOGCALL(API, MSet, "Diversify::get_dmset", mset);
+    // Return original mset if no need to diversify
+    if (k == 0) {
+	DocumentSet dmset;
+	for (MSetIterator it = mset.begin(); it != mset.end(); ++it)
+	    dmset.add_document(it.get_document());
+	return dmset;
+    }
+
+    if (k > mset.size())
+	k = mset.size();
+
+    initialise_points(mset);
+    vector<Point> main_dmset(points.begin(), points.begin() + k);
+    vector<Point> curr_dmset = main_dmset;
+
+    while (true) {
+	bool found_better_dmset = false;
+	for (unsigned int i = 0; i < main_dmset.size(); ++i) {
+	    auto curr_doc = main_dmset[i];
+	    unsigned int best_score = evaluate_dmset(curr_dmset);
+
+	    vector<Point> diff_dmset = compute_diff_dmset(curr_dmset);
+
+	    bool found_better_doc = false;
+	    for (unsigned int j = 0; j < diff_dmset.size(); ++j) {
+		vector<Point> temp_dmset = curr_dmset;
+		temp_dmset[i] = diff_dmset[j];
+		unsigned int score = evaluate_dmset(temp_dmset);
+		if (score < best_score) {
+		    curr_doc = temp_dmset[i];
+		    best_score = score;
+		    found_better_doc = true;
+		}
+	    }
+	    if (found_better_doc) {
+		curr_dmset[i] = curr_doc;
+		found_better_dmset = true;
+	    }
+	}
+
+	// Terminate algorithm when there's no change in current
+	// document matchset
+	if (!found_better_dmset)
+	    break;
+
+	main_dmset = curr_dmset;
+    }
+
+    // Merge main_dmset and diff_dmset into final dmset
+    DocumentSet dmset;
+    for (Point doc : main_dmset) {
+	dmset.add_document(doc.get_document());
+    }
+
+    vector<Point> diff_dmset = compute_diff_dmset(main_dmset);
+    for (auto doc : main_dmset)
+	dmset.add_document(doc.get_document());
+
+    return dmset;
+}
--- a/xapian-core/diversify/diversify.cc
+++ b/xapian-core/diversify/diversify.cc
@@ -60,6 +60,22 @@ Diversify::initialise_points(const MSet &source)
     }
 }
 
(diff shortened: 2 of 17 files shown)


xapian-core/diversify/diversify.cc
```diff
@@ -0,0 +1,195 @@
...
+}
+
+void
+Diversify::initialise_points(const MSet &source)
+{
+    LOGCALL_VOID(API, "Diversify::initialise_points", source);
+    TermListGroup tlg(source);
+    for (MSetIterator it = source.begin(); it != source.end(); ++it) {
+	points.push_back(Point(tlg, it.get_document()));
+	weights[it.get_document().get_docid()] = it.get_weight();
```
ojwb: `*it` gives you the docid directly (and without creating an unnecessary temporary `Document` object).

xapian-core/diversify/diversify.cc
```diff
@@ -0,0 +1,195 @@
...
+    LOGCALL_VOID(API, "Diversify::compute_similarities", NO_ARGS);
+    Xapian::CosineDistance d;
+    for (unsigned int i = 0; i < points.size() - 1; ++i) {
+	for (unsigned int j = i + 1; j < points.size(); ++j) {
+	    Xapian::docid docid_i = points[i].get_document().get_docid();
+	    Xapian::docid docid_j = points[j].get_document().get_docid();
+	    auto key_ij = make_pair(docid_i, docid_j);
+	    auto key_ji = make_pair(docid_j, docid_i);
+	    double sim = d.similarity(points[i], points[j]);
+	    pairwise_sim[key_ij] = pairwise_sim[key_ji] = sim;
```
ojwb: If you only stored the similarly for `(x, y)` where `x` <= `y` then you would roughly halve the amount of storage needed for `pairwise_sim` and greatly reduce the cache pressure when you doing all the lookups in it.  The only cost would be that you need to compare the two docids when doing a lookup, and do the lookup such that the smaller one is first in the pair.  I'd expect that's a significant win overall.

xapian-core/diversify/diversify.cc
```diff
@@ -0,0 +1,195 @@
...
+
+#include "xapian/diversify.h"
+
+#include "xapian/error.h"
+#include "debuglog.h"
+
+#include <vector>
+#include <limits>
+#include <algorithm>
+#include <cmath>
```
ojwb: It's a Xapian (and wider) coding convention to sort entries within a block of header includes - it makes it easier to see at a glance if a header is being included, and that in turn helps to avoid pointlessly including the same header more than once.

xapian-core/include/xapian/diversify.h
```diff
@@ -0,0 +1,124 @@
...
+
+#if !defined XAPIAN_IN_XAPIAN_H && !defined XAPIAN_LIB_BUILD
+#error "Never use <xapian/diversify.h> directly; include <xapian.h> instead."
+#endif
+
+#include "xapian/cluster.h"
+#include <xapian/attributes.h>
+#include <xapian/mset.h>
+#include <xapian/types.h>
+#include <xapian/visibility.h>
```
ojwb: Should be `<>` for all of these in an installed header, and they should be in sorted order.

xapian-core/include/xapian/diversify.h
```diff
@@ -0,0 +1,124 @@
...
+
+#include <map>
+#include <vector>
+#include <unordered_map>
```
ojwb: This block should also be in sorted order.

xapian-core/include/xapian/diversify.h
```diff
@@ -0,0 +1,124 @@
...
+    /** Constructor specifying the number of diversified search results
+     *
+     *  @param  k_	Number of required diversified results
+     */
+    explicit Diversify(unsigned int k_,
+		       double lambda_ = 0.5,
+		       double b_ = 5.0,
+		       double sigma_sqr_ = 1e-3);
+
+    /** Implements diversifcation
```
ojwb: Typo.

xapian-core/diversify/Makefile.mk
```diff
@@ -0,0 +1,3 @@
+lib_src +=\
+	diversify/diversify.cc
+
```
ojwb: It's preferable to avoid extra blank lines at the end of files (unless the file format requires them, which isn't the case here).

xapian-core/diversify/diversify.cc
```diff
@@ -0,0 +1,195 @@
...
+	main_dmset = curr_dmset;
+    }
+
+    // Merge main_dmset and diff_dmset into final dmset
+    DocumentSet dmset;
+    for (Point doc : main_dmset) {
+	dmset.add_document(doc.get_document());
+    }
+
+    vector<Point> diff_dmset = compute_diff_dmset(main_dmset);
```
ojwb: `diff_dmset`  isn't used here - I suspect `main_dmset` in the line below should be `diff_dmset` (or else it seems we are just adding the same documents twice).

xapian-core/diversify/diversify.cc
uppinder: Yes, it definitely is. Thanks for pointing this out, pushed required changes in new commit.

xapian-core/diversify/diversify.cc
uppinder: I didn't know this, thanks for pointing this out. Fixed in new commit.

xapian-core/include/xapian/diversify.h
uppinder: Changes pushed in new commit.

xapian-core/include/xapian/diversify.h
uppinder: Fixed.

xapian-core/include/xapian/diversify.h
uppinder: Fixed.

xapian-core/diversify/Makefile.mk
uppinder: Fixed in a new commit.

xapian-core/diversify/diversify.cc
uppinder: Will create modifications for this soon, thanks for pointing it out.

xapian-core/diversify/diversify.cc
uppinder: You're correct, I must've overlooked this. Fixed in a new commit.

xapian-core/diversify/diversify.cc
```diff
@@ -144,6 +144,10 @@ DocumentSet
 Diversify::get_dmset(const MSet &mset)
 {
     LOGCALL(API, MSet, "Diversify::get_dmset", mset);
+    // Nothing to do
+    if (mset.empty())
+	throw InvalidArgumentError("Given mset should not be empty");
+
```
ojwb: I'd say if there's nothing to do then we should do nothing.  An empty `MSet` seems a totally reasonable answer to "give me a diversified version of this empty `MSet`".  It's already exactly as diverse as it can be.

If we throw an exception for an empty `MSet`, that forces the caller to check write code like:

```
Xapian::MSet diversified_mset;
if (mset.empty()) {
    diversified_mset = mset;
} else {
    diversified_mset = diversifier.get_dmset(mset);
}
```

In fact, I think the check in `get_dmset()` should be `if (mset.size() <= 2) return mset;` since diversification of a single entry `MSet` can only be a no-op, and swapping the entries in a two entry `MSet` won't make it any more diverse so the existing ordering from weighting or whatever should be respected.  Only once we get to three entries could reordering them conceivably improve diversity.

uppinder: Thanks for the tip. Agreed, diversification is fruitfully applicable to only three or more documents. I have added the necessary changes in a fixup commit.

xapian-core/diversify/diversify.cc
```diff
@@ -53,19 +53,22 @@ void
...
+	if (count < k) {
+	    main_dmset.push_back(*it);
+	    ++count;
+	}
     }
 }
 
 pair<Xapian::docid, Xapian::docid>
-Diversify::get_key(const Xapian::Point &doc_a, const Xapian::Point &doc_b)
+Diversify::get_key(const Xapian::docid &docid_a, const Xapian::docid &docid_b)
```
ojwb: Passing values of types where copying is expensive by `const` reference is good, but `Xapian::docid` is just an unsigned integer type for which copying is very cheap - likely cheaper than passing a pointer and dereferencing it how passing by `const` reference is probably implemented.

xapian-core/diversify/diversify.cc
```diff
@@ -81,54 +84,63 @@ Diversify::compute_similarities()
 {
     LOGCALL_VOID(API, "Diversify::compute_similarities", NO_ARGS);
     Xapian::CosineDistance d;
-    for (unsigned int i = 0; i < points.size() - 1; ++i) {
-	for (unsigned int j = i + 1; j < points.size(); ++j) {
-	    double sim = d.similarity(points[i], points[j]);
-	    auto key = get_key(points[i], points[j]);
+    for (auto p_a : points) {
+	Xapian::docid pointid_a = p_a.first;
+	Xapian::Point point_a = p_a.second;
```
ojwb: Here (and also just below) you're copying the `Xapian::Point` object, and that's unlikely to be a cheap copy - `Xapian::Point` is a subclass of `Xapian::PointType` which has a member of type`std::unordered_map<std::string, double>` (plus another member and `Xapian::Point` adds another but those are cheaper to copy).

You can avoid the copy by making these `const` references - i.e. `const Xapian::Point& point_a = pa_second;`

xapian-core/include/xapian/diversify.h
```diff
@@ -46,17 +46,20 @@ class XAPIAN_VISIBILITY_DEFAULT Diversify {
...
     unsigned int k;
 
-    /// Store given mset as a vector of points
-    std::vector<Point> points;
+    /// MPT parameters
+    double lambda, b, sigma_sqr;
 
-    /// Store relevance score for each document of given mset
-    std::unordered_map<Xapian::docid, double> weights;
+    /// Store each document from given mset as a point and its relevance score
```
ojwb: The comment here suggests the scores are stored in this member, i.e. `points` but the scores are actually in the member below - `scores`.

xapian-core/tests/perftest/perftest_diversify.cc
```diff
@@ -0,0 +1,65 @@
...
+{
+    Xapian::Database db;
+    db = backendmanager->get_database("apitest_diversify");
+
+    logger.testcase_begin("perfdiversify1");
+    Xapian::Enquire enq(db);
+
+    Xapian::Query query("java");
+
+    logger.searching_start("Diversifcation");
```
ojwb: Typo in message.

xapian-core/tests/perftest/perftest.h
```diff
@@ -95,6 +98,23 @@ class PerfTestLogger {
      */
     void searching_end();
 
+    /** Log the start of a diversifcation run.
```
ojwb: Typo: "diversifcation"

xapian-core/tests/perftest/perftest.h
```diff
@@ -95,6 +98,23 @@ class PerfTestLogger {
...
+     */
+    void diversifying_start(const std::string & description);
+
+    /** Log the start of diversification.
+     */
+    void diversify_start();
+
+    /** Log the completion of diversification.
+     */
+    void diversify_end(const unsigned int k,
```
ojwb: I'm not sure there's a good reason to make `k` const here - it's passed by value.

xapian-core/diversify/diversify.cc
```diff
@@ -0,0 +1,218 @@
...
+
+    vector<Xapian::docid> diff_dmset = compute_diff_dmset(dmset);
+
+    for (auto point_id : diff_dmset) {
+	double min_dist = numeric_limits<double>::max();
+	unsigned int pos = 1;
+	for (auto doc_id : dmset) {
+	    auto key = get_key(point_id, doc_id);
+	    double sim = pairwise_sim[key];
+	    double weight = 2 * b * sigma_sqr * (1 / log(1 + pos)) * (1 - sim);
```
ojwb: The compiler is fairly limited in what optimisations it can do for floating point calculations (unless you use options like GCC's `funsafe-math-optimizations` which don't conform to the language standards and so aren't on by default in sensible compilers).  So if you write:

```
2 * b * sigma_sqr * (1 / log(1 + pos)) * (1 - sim)
```

the compiler may be forced to generate an extra multiplication and use of a constant `1` compared to:

```
2 * b * sigma_sqr / log(1 + pos) * (1 - sim)
```

The second version also seems clearer to read if anything.

xapian-core/include/xapian/diversify.h
```diff
@@ -0,0 +1,133 @@
...
+
+    /** Initial diversified document set
+     *
+     *  Return top-k documents of mset as vector of Points, which
+     *  represent the initial diversified document set.
+     *
+     *  @param source	MSet object containing the documents of which
+     *			top-k are to be diversified
+     */
+    void initialise_points(const Xapian::MSet &source);
```
ojwb: The current code base uses rather a mixture of placements of the `&` for a reference type (and similarly for `*` for pointer types) - you'll probably see all of `const Xapian::MSet &source`, `const Xapian::MSet & source` and `const Xapian::MSet& source`.

However we're aiming to standardise on the latter as it seems to be the preferred modern C++ style - if you could try to follow that in new code that'd be great.

xapian-core/tests/perftest/perftest_diversify.cc
uppinder: Fixed in a fixup commit.

xapian-core/tests/perftest/perftest.h
uppinder: Fixed in a fixup commit.

xapian-core/include/xapian/diversify.h
uppinder: I must've forgot to correct the location of the comment after some code refactoring. Fixed in a fixup commit.

xapian-core/tests/perftest/perftest.h
uppinder: I agree that ```k``` is passed by value so a ```const``` here is redundant. But after having looked at [this](https://stackoverflow.com/questions/8714250/isnt-const-redundant-when-passing-by-value), I thought it would be a better practice to disallow modifications of ```k``` locally inside ```diversify_end``` as it's going to be only logged.

xapian-core/include/xapian/diversify.h
uppinder: I had wondered why there was such a mixture in the code base. Anyways, pushed a fixup commit for the modifications.

xapian-core/diversify/diversify.cc
uppinder: Thanks for pointing this out. Pushed a fixup commit for the same.

xapian-core/diversify/diversify.cc
uppinder: I thought GCC would optimise the expression by default and thus chose to use a more verbose representation, thanks for the sharing this information. I personally feel that the latter expression requires more mental processing while keeping the left to right precedence of (*) in mind. Although, this might be because I'm not used to that.

xapian-core/tests/perftest/perftest.h
ojwb: There's no good reason to mark it `const` here, even if you mark it `const` in the definition (for a pass-by-value parameter the `const`-ness doesn't need to match between declaration and definition).  It's not part of the API that it's unchanged inside the method - that makes no difference to the calling code.

I probably wouldn't bother on the definition myself.  I see the argument that it can help the reader comprehend a large function, but this one is only 4 statements/10 lines.

xapian-core/diversify/diversify.cc
ojwb: Both multiplication and division have the same precedence in C/C++ (I think that's true almost universally): http://en.cppreference.com/w/cpp/language/operator_precedence

You could write it as `2 * b * sigma_sqr * (1 - sim) / log(1 + pos)` if the order of evaluation doesn't matter (it clearly doesn't matter in real arithmetic, but with floating point it could matter if the intermediate values overflow or underflow the exponent range of `double` - that's unlikely to be a concern here though).

BTW, you can see the code generation difference here: https://godbolt.org/g/99cd3X

In fact, even with `-ffast-math` (which turns on `-funsafe-math-optimizations` and more) the code generated is better for: `2 * b * sigma_sqr / log(1 + pos) * (1 - sim)`

xapian-core/include/xapian/diversify.h
ojwb: The oldest code was written by Richard and myself back in 1999.  Then we were both fairly new to C++ and from a C background where it's common to write `char *p`.  One argument for that over the `char* p` form is that you can write:

```
char *p, q;
```

which is the same as:

```
char *p;
char q;
```

whereas:

```
char* p, q;
```

reads as if q is a `char*`.

Also:

```
char *p, *q;
```

reads OK, but where do you put the `*` for q with the modern C++ style?  About the best I can see is:

```
char* p, * q;
```

But such mixed declarations are more naturally split onto separate lines in C++ which encourages initialisation at the point of declaration.

At some point we switched to a space either side - not sure I really remember the logic as to why.

It would be good to do a mass clean-up at some point, but such things can be disruptive - it makes backporting patches harder, unmerged patches may stop applying cleanly, and it makes tools like `git blame` less useful (though with git you can use `-w` to ignore whitespace so it can see through this sort of change - we used to use subversion for version control and I don't think that allowed that).  Therefore we tend to hold off such things for a suitable time (such as shortly before cutting the next stable release branch, as then the older release branch will be moving to a very conservative backporting policy).

ojwb:
The appveyor CI failures are because I updated master for Unicode 11 but the appveyor builds use a pre-generated copy of unicode-data.cc (to avoid needing a working Tcl install) and I hadn't updated that.  Now done so those failures should go away.

xapian-core/diversify/diversify.cc
ojwb: This one is still pending.

xapian-core/include/xapian/diversify.h
```diff
@@ -0,0 +1,134 @@
...
+    compute_diff_dmset(const std::vector<Xapian::docid>& dmset);
+
+    /** Evaluate a diversified mset
+     *
+     *  Evaluate a diversified mset using MPT algorithm
+     *
+     *  @param dmset	Set of points representing candidate diversifed
+     *			set of documents
+     */
+    double evaluate_dmset(const std::vector<Xapian::docid>& dmset);
```
ojwb: I'm happy to leave this to a follow-up PR, but I think that `Xapian::Diversify` should use the same PIMPL idiom which most of the Xapian API classes do - the currently exceptions are classes with a virtual method which the user can subclass (functor classes such as `Xapian::RangeProcessor`) and some classes which are just inherently cheap to copy and for which the implementation is fairly simple and so unlikely to be something we'd want to change during a release series (e.g. `Xapian::MSetIterator`).

It's surprising to have a new API class which is expensive to copy when the existing ones aren't (and the documentation clearly says they're cheap to copy).  Also, exposing the implementation ties our hands when it comes to changing the implementation during each stable release series, and also means more exported symbols from the shared library (which increases the load time cost), and potentially more STL headers implicitly included by `#include <xapian.h>`).

xapian-core/include/xapian/diversify.h
```diff
@@ -0,0 +1,134 @@
...
+
+  public:
+    /** Constructor specifying the number of diversified search results
+     *
+     *  @param  k_	Number of required diversified results
+     */
+    explicit Diversify(unsigned int k_,
+		       double lambda_ = 0.5,
+		       double b_ = 5.0,
+		       double sigma_sqr_ = 1e-3);
```
ojwb: It'd be good to document the other parameters too.

uppinder: Thanks for pointing this out. Modifications pushed in a fixup commit.

xapian-core/diversify/diversify.cc
uppinder: Oops, my bad. Modification pushed in a fixup commit. Thanks for pointing it out.

xapian-core/tests/perftest/perftest.h
uppinder: Thanks for clarifying this up. Modification pushed in a fixup commit.

xapian-core/diversify/diversify.cc
uppinder: Ah, I see. In hindsight, there was never any emphasis on such micro-level optimisations in my university courses. Thanks for pointing this out, definitely something I didn't know earlier.

xapian-core/include/xapian/diversify.h
uppinder: This did cross my mind but being my first such PR, I didn't want to handle the added complexity which in hindsight I'm not sure was the right decision. But I don't think implementing the PIMPL idiom in this case would be much of a hassle now.

xapian-core/tests/testdata/apitest_diversify.txt
```diff
@@ -0,0 +1,21 @@
+The island Java lies between Sumatra to the west and Bali to the East.
```
ojwb: This new file needs to be listed `EXTRA_DIST` in `tests/Makefile.am` or else it won't be included in the tarball build by `make dist` and then someone building a release from a distribution tarball won't be able to run these new tests.

An easy way to test for this (and various other issues) locally is to run `make distcheck` - that makes a distribution tarball, then unpacks it, tries to build it as an out-of-tree build, tries to install it into a temporary location, checks that the testsuite can be run against the installed version (this is a no-op for xapian-core, but omega and the bindings support it), tries to uninstall it, and then checks that the uninstall removed all the files which were installed, and then checks that the tree unpacked from the tarball can also `make dist` successfully.

We don't currently have a build for this on the CI infrastructure that runs for PRs, though perhaps we should.

We do already have coverage for this once code is merged though - there is a cron job which runs `make distcheck` on master and RELEASE/1.4 branches a few times an hour (if there are any changes) and if it works makes a new snapshot available at https://oligarchy.co.uk/xapian/master/ or https://oligarchy.co.uk/xapian/RELEASE/1.4/ depending on which branch it is from.

xapian-core/tests/perftest/perftest.cc
```diff
@@ -383,6 +384,47 @@ PerfTestLogger::searching_end()
...
+{
+    Assert(diversifying_started);
+    double elapsed(RealTime::now() - diversifying_timer);
+    write("    <diversify>"
+	       "<time>" + str(elapsed) + "</time>"
+	       "<dset>"
+	       "<k>" + str(k) + "</k>"
+	       "<size>" + str(dset.size()) + "</size>"
+	       "</dset>"
+	       "</diversify>\n");
```
ojwb: The continuation lines here are overindented - they should line up vertically like:

```
    write("    <diversify>"
          "<time>" + str(elapsed) + "</time>"
          "<dset>"
```

ojwb:
Sorry, I just did a final review and noticed a couple of small issues I'd missed before.

xapian-core/tests/testdata/apitest_diversify.txt
uppinder: I've noticed ```apitest_cluster.txt``` isn't included in it either, so I believe that should be added too. I've pushed a commit to include ```apitest_diversify.txt``` for now.

xapian-core/tests/perftest/perftest.cc
uppinder: Modification pushed in a fixup commit.

xapian-core/tests/testdata/apitest_diversify.txt
ojwb: > I've noticed `apitest_cluster.txt` isn't included in it either, so I think that should be added too.

That one doesn't actually appear to be used anywhere (which would explain why we aren't seeing failures due to it not being included in tarballs).  It looks like its content was inlined into `api_cluster.cc` (with slight tweaks to two of them).

Good spot - I'll push a change to just drop it once I've merged this.

PR #199, opened 2018-04-16
Move support for API classes

guruhegde:
This PR address [ticket #642](https://trac.xapian.org/ticket/642)

I used the `default` keyword for implementation of move ctor and assignment. Currently copy ctor and assignment op have user defined, as it is necessary for pre c++11. I also thought this doesn't require us to include `<utility>` to use `std::move` fn(not sure whether this point is really useful). Let me know whether this is fine.

Please review when you have time. Thanks!

Can you comment on the commit boundaries I formed? Just curious to know whether I can improve on that.

### DIFF
--- a/xapian-bindings/xapian-headers.i
+++ b/xapian-bindings/xapian-headers.i
@@ -54,6 +54,7 @@
     %ignore NS::CLASS::CLASS(Internal*);
     %ignore NS::CLASS::CLASS(Internal&);
     %ignore NS::CLASS::operator=;
+    %ignore NS::CLASS::CLASS(CLASS &&);
 %enddef
 #else
 %define STANDARD_IGNORES(NS, CLASS)
@@ -62,6 +63,7 @@
     %ignore NS::CLASS::CLASS(Internal&);
     %ignore NS::CLASS::operator=;
     %ignore NS::CLASS::CLASS(const CLASS &);
+    %ignore NS::CLASS::CLASS(CLASS &&);
 %enddef
 #endif
 
--- a/xapian-core/api/positioniterator.cc
+++ b/xapian-core/api/positioniterator.cc
@@ -26,6 +26,8 @@
 #include "omassert.h"
 #include "backends/positionlist.h"
 
+#include <utility>
+
 using namespace std;
 
 namespace Xapian {
@@ -76,10 +78,19 @@ PositionIterator::operator=(const PositionIterator & o)
     RETURN(*this);
 }
 
-PositionIterator::PositionIterator(PositionIterator &&) = default;
+PositionIterator::PositionIterator(PositionIterator && o)
+    : internal(o.internal)
+{
+    o.internal = nullptr;
+}
 
 PositionIterator &
-PositionIterator::operator=(PositionIterator &&) = default;
+PositionIterator::operator=(PositionIterator && o)
+{
+    if (this != &o)
+	swap(internal, o.internal);
+    RETURN(*this);
+}
 
 Xapian::termpos
 PositionIterator::operator*() const
--- a/xapian-core/api/positioniterator.cc
+++ b/xapian-core/api/positioniterator.cc
@@ -26,8 +26,6 @@
 #include "omassert.h"
 #include "backends/positionlist.h"
 
-#include <utility>
-
 using namespace std;
 
 namespace Xapian {
@@ -75,15 +73,6 @@ PositionIterator::operator=(const PositionIterator & o)
     if (internal)
 	decref();
     internal = o.internal;
-    RETURN(*this);
-}
-
-
-PositionIterator &
-PositionIterator::operator=(PositionIterator && o)
-{
-    if (this != &o)
-	swap(internal, o.internal);
     RETURN(*this);
 }
 
--- a/xapian-core/api/positioniterator.cc
+++ b/xapian-core/api/positioniterator.cc
@@ -78,11 +78,6 @@ PositionIterator::operator=(const PositionIterator & o)
     RETURN(*this);
 }
 
-PositionIterator::PositionIterator(PositionIterator && o)
-    : internal(o.internal)
-{
-    o.internal = nullptr;
-}
 
 PositionIterator &
 PositionIterator::operator=(PositionIterator && o)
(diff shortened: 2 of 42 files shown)


xapian-bindings/xapian-headers.i
```diff
@@ -54,6 +54,7 @@
     %ignore NS::CLASS::CLASS(Internal*);
     %ignore NS::CLASS::CLASS(Internal&);
     %ignore NS::CLASS::operator=;
+    %ignore NS::CLASS::CLASS(CLASS &&);
```
guruhegde: Without this change, I get the following error.
```
 Warning 509: Overloaded method Xapian::Document::Document(Xapian::Document const &) effectively ignored,
Warning 509: as it is shadowed by Xapian::Document::Document(Xapian::Document &&).
```
I don't know much about SWIG. From what's already there, I added this fix. Please tell whether this is the way.

xapian-core/api/positioniterator.cc
```diff
@@ -76,10 +78,19 @@ PositionIterator::operator=(const PositionIterator & o)
     RETURN(*this);
 }
 
-PositionIterator::PositionIterator(PositionIterator &&) = default;
+PositionIterator::PositionIterator(PositionIterator && o)
+    : internal(o.internal)
+{
```
guruhegde: Should I add LOG same as copy ctor & assignment op(except object pointer)?

xapian-bindings/xapian-headers.i
ojwb: That seems appropriate to me - these objects are reference counted handles, so it's not useful to explicitly wrap ways to create a copy in the target language.

For some reason Tcl needs the copy ctor wrapped.  I'm not sure we bothered to work out why, but it shouldn't need the move ctor since it doesn't currently have it.

xapian-core/api/positioniterator.cc
```diff
@@ -76,6 +78,20 @@ PositionIterator::operator=(const PositionIterator & o)
...
+    o.internal = nullptr;
+}
+
+PositionIterator &
+PositionIterator::operator=(PositionIterator && o)
+{
+    if (this != &o)
+	swap(internal, o.internal);
+    RETURN(*this);
```
ojwb: This should just be `return *this;` - only use `RETURN` if there's debug logging of the function/method call (i.e.  `LOGCALL()` or similar).

(These methods aren't really interesting to log, so I think it's fine not to have logging for them.)

xapian-core/api/positioniterator.cc
```diff
@@ -76,6 +78,20 @@ PositionIterator::operator=(const PositionIterator & o)
     RETURN(*this);
 }
 
+PositionIterator::PositionIterator(PositionIterator && o)
+    : internal(o.internal)
+{
+    o.internal = nullptr;
+}
```
ojwb: This should be OK to have inline in the public API header (`include/xapian/positioniterator.h`).

That's beneficial for a trivial implementation like this one as it means that the compiler can inline it and optimise the result.  It also means we don't create another external symbol in the shared library (the fewer external symbols there are, the quicker it loads).  One extra isn't a big deal by itself, but where they can be easily avoided we try to.

ojwb:
> I used the default keyword for implementation of move ctor and assignment. Currently copy ctor and assignment op have user defined, as it is necessary for pre c++11. I also thought this doesn't require us to include <utility> to use std::move fn(not sure whether this point is really useful). Let me know whether this is fine.

I think `default` is better - it's shorter, clearer, and there's no scope for a slightly incorrect attempt at a default implementation.  It'd be good to change the `default`-equivalent implementations now that we require C++11 in both the library and public headers.

Avoiding including `<utility>` from the public API headers would be a bonus, but I think it's not something we can easily avoid entirely.

It'd also be good to switch to using `delete` where we currently just make them private as that helps the compiler give better error messages.

> Can you comment on the commit boundaries I formed? Just curious to know whether I can improve on that.

I just looked through the combined diff so far.  I really should get on with something else right now, but I'll have a look when I get a chance.

xapian-core/include/xapian/database.h
```diff
@@ -29,6 +29,7 @@
 
 #include <iosfwd>
 #include <string>
+#include <utility>
```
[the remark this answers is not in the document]
guruhegde: We can use static_cast to avoid it. should I go for that?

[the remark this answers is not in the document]
guruhegde: But I guess it is not a straightforward cast. I have to understand it better first.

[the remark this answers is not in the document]
guruhegde: Custom impl of move is not trivial it seems.  Please ignore my comment.

xapian-core/api/positioniterator.cc
guruhegde: Fixed

xapian-core/api/positioniterator.cc
guruhegde: Thanks for the tip. Done.

xapian-core/api/positioniterator.cc
ojwb: I don't think logging is all that useful for these.  The compiler is allowed (and sometimes required) to eliminate calls to them, so logging wouldn't correspond to the source exactly anyway.

The only reason I can think that logging might be interesting is to try to find where copies aren't eliminated to try to address that.  But by inlining we open things up for the compiler to optimise in more ways, and without the programmer having to study logs.  The savings aren't likely to be very large anyway, as copying just copies a pointer and adjusts reference counts.

Overall having these inline in the API headers seems a better trade-off than putting them into the library so they can have debug logging.

xapian-core/include/xapian/database.h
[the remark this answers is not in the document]
ojwb: The cast would be `static_cast<PositionIterator&&>(obj)` (when `obj`'s class is `PositionIterator`).

I think that's worse than just including `<utility>` though - the code is much clearer using `std::move()`.

guruhegde:
I think most of the classes in public API covered. Let me know if any other work is pending on this.

ojwb:
Sorry getting back to this now.

The newer changes look good to me, will merge this now.

> Can you comment on the commit boundaries I formed? Just curious to know whether I can improve on that.

They look great to me.

ojwb:
`git rebase -i --autosquash`-ed and merged to master in 32a7fd61a43f4e817578a462e55c53406f68e4ac.

Thanks for all your work on this.

PR #200, opened 2018-04-20
Test behavior on remote backend

guruhegde:
As discussed in #758, this is still work in progress, not ready yet. I would like to get some feedback and guidance.


guruhegde:
Problem I see:  on_SIGCHLD handler is called while tcp server is still running.(I verified this with checking the running process in terminal). But SIGCHLD is received only after child exits. Here forked process is replaced with server process using execl call. Does that send SIGCHLD signal? 

Next in on_SIGCHLD function, waitpid returns child process id that means child doesn't exist.
Can you help me understand this?

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -264,6 +264,11 @@ BackendManager::clean_up()
 {
 }
 
+void
+BackendManager::killall_server()
+{
+}
```
ojwb: If there's no implementation for the current backend, doing nothing isn't a good option - better to throw an exception so we quickly see if testcases requiring this feature are run with backends which we don't support it for.

xapian-core/tests/harness/backendmanager_remotetcp.cc
```diff
@@ -404,3 +404,22 @@ BackendManagerRemoteTcp::clean_up()
     }
 #endif
 }
+
+void
+BackendManagerRemoteTcp::killall_server()
+{
+    for (unsigned i = 0; i < sizeof(pid_to_fd) / sizeof(pid_fd); ++i) {
```
ojwb: This needs wrapping with `#ifdef HAVE_FORK` (currently the CI appveyor builds fail because of this).

It'd be good to implement this for the `__WIN32__` case too - I think that would require noting down `procinfo.hProcess` above so we can use that to kill the child process here with `TerminateProcess()` or something.  Perhaps that's better left until we have per-server killing, but while it's not implemented we should throw an exception for that case rather than quietly doing nothing.

ojwb:
> There are actually two process(out of which one is in zombie state, so that caused SIGCHLD signal)

The zombie state means the process no longer exists, but there's still an entry in the OS process table for it so that its exit status can be read by its parent process:

https://en.wikipedia.org/wiki/Zombie_process

When there's only a zombie process entry for a child process, the OS will deliver `SIGCHLD` to the parent, and in our case that runs `on_SIGCHLD()` which calls `waitpid(-1, &status, WNOHANG)` repeatedly to get reap any zombies and remove corresponding entries from `pid_to_fd[]`.

I think the underlying issue here is that `xapian-tcpsrv --one-shot` still calls `fork()` rather than just servicing the connection in the parent process.  When servicing multiple requests we want a child process for each so they can be handled in parallel, but that's not necessary for `--one-shot` mode, and it's unhelpful here as it means the child of the test program isn't the process actually servicing the request.

guruhegde:
I made changes to execute server kill. Kill interface still doesn't take any backend.  I guess we can modify them later. Let me know if I can do any change on this.

What test cases should I cover? Most of them like we did for `close()`?

guruhegde:
Sorry, I was updating commit message of last few, by mistake rebased everything.

xapian-core/tests/harness/backendmanager_remotetcp.cc
```diff
@@ -79,6 +79,9 @@ struct pid_fd {
 
 static pid_fd pid_to_fd[16];
```
guruhegde: any other way to avoid global variable?

xapian-core/net/tcpserver.h
```diff
@@ -105,6 +105,11 @@ class XAPIAN_VISIBILITY_DEFAULT TcpServer SOCKET_INITIALIZER_MIXIN {
     /** Accept a single connection, service requests on it, then stop.  */
     void run_once();
 
+    /** Function overload to call fork-based implementation.
+     *  Accept a single connection, service requests on it in a child process,
+     *  then stop.  */
+    void run_once(bool);
```
guruhegde: I used this trick to make a minimum change to make it work(so I don't break anything). Please suggest if you have a clean fix in mind.

ojwb: I think it's better to refactor it so the forking is handled outside of `run_once()` (which then means that `run_once()` would then works the same way cross-platform).  I've pushed that change as 4ae447e7382ff4983fe15d02663b28edc01d3479 so you can just drop this commit.

xapian-core/tests/harness/backendmanager_remotetcp.cc
ojwb: It is at least `static`, so it's only global for this one source file.

It needs to be accessible from the `SIGCHLD` handler and from the code which launches new `xapian-tcpsrv` processes, and the only argument that a signal handler gets passed is the signal number, so we can't arrange to pass in a pointer to a structure.

We could have a function which returned a pointer to a structure which was static within that function, but that doesn't seem to really make things cleaner, it just adds a wrapper layer which makes things less clear if anything.

I'd say that when you're dealing with signal handlers and the like, the code is inevitably low level system programming, and a global variable isn't such a bad thing when what's being represented is (as here) information which is naturally a singleton (it wouldn't make sense to have two different `pid_to_fd` structures).

ojwb:
I think this work is both interesting and useful (I wouldn't be at all surprised if it uncovers at least one bug in handling a remote server dying as that's not a case which gets much exercise), but for now I think let's focus on your GSoC project.

Once I've cleared the pile of patches to review I'll look over this again.  And if you find you can't progress on the GSoC project because you're waiting for us, feel free to work on this some more (but do first make sure we're aware you're blocked!)

guruhegde:
> Probably worth just trying that to see by search-and-replace. 

I had tried this and there was some issue. But with GSoC project, I didn't pursue this. Will work on this in couple of weeks.

guruhegde:
> BTW I fixed this in 4ae447e back in April.

I am aware of this, have pulled that commit. The problem was something related new exception. The expected exception was not happening. I really don't remember the details. Will update soon.

guruhegde:
Testcase shown until point where it fails:
```
// Test WritableDatabase methods.
 DEFINE_TESTCASE(closedb7, writable) {
     Xapian::WritableDatabase db(get_writable_database());
     db.add_document(Xapian::Document());
     /* db.close(); */
     kill_server();
 
     // Since we can't make any changes which need to be committed,
     // db.commit() is a no-op, and so doesn't have to fail.
     try {
         db.commit();
     } catch (const Xapian::NetworkError &) {
     }
 
     TEST_EXCEPTION(Xapian::DatabaseError,
                    db.add_document(Xapian::Document()));
     
     return true;
 }
```
Here, db.commit() is throwing NetworkError as expected and caught in the catch block. 
`db.add_document()` call is throwing this error:
```
unning test: closedb7... POSSIBLE UNRELEASED MEMORY - RETRYING TEST NetworkError: write failed (context: remote:tcp(127.0.0.1:1239)) (Broken pipe)
Expecting exception DatabaseError
```
This error is different from usual `NetworkError: Connection closed unexpectedly`.  So failure is at different point in code I guess.

When `DatabaseError` is replaced with `NetworkError`, TEST_EXCEPTION is not successful. It gives 
this error:
`Running test: closedb7... DatabaseError: Database has been closed
Expecting exception NetworkError`
I didn't understand this behavior. Any idea why this is happening and how to go about this?

guruhegde:
Another testcase I would like to discuss:

```
DEFINE_TESTCASE(closedb10, writable && metadata) {
     Xapian::WritableDatabase db(get_writable_database());
     db.set_metadata("foo", "FOO");
     db.set_metadata("bar", "BAR");
     /* db.close(); */
     kill_server();
 
     TEST_EXCEPTION(Xapian::NetworkError,
                    db.set_metadata("test", "TEST"));
     TEST_EXCEPTION(Xapian::NetworkError,
                    db.get_metadata("foo"));
     TEST_EXCEPTION(Xapian::NetworkError,
                    db.get_metadata("foo"));
     TEST_EXCEPTION(Xapian::NetworkError,
                    db.metadata_keys_begin());
 
     return true;
 }
```
This gives following error:
```
Running test: closedb10... POSSIBLE UNRELEASED MEMORY - RETRYING TEST FAILED
Expecting exception NetworkError
api_closedb.cc:501: Expected "Xapian::NetworkError"

./apitest backend remotetcp_glass: 0 tests passed, 1 failed.
```
I replaced only first `TEST_EXCEPTION` with `try` and `catch` block and test case pass. `NetworkError`  exception thrown by  `db.set_metadata()`  failed with `TEST_EXCEPTION`.

Passing version:
```
DEFINE_TESTCASE(closedb10, writable && metadata) {
     Xapian::WritableDatabase db(get_writable_database());
     db.set_metadata("foo", "FOO");
     db.set_metadata("bar", "BAR");
     /* db.close(); */
     kill_server();
 
     try {
         db.set_metadata("test", "TEST");
     } catch (Xapian::NetworkError &) {
     }
 
     /* TEST_EXCEPTION(Xapian::NetworkError, */
                    /* db.set_metadata("test", "TEST")); */
     TEST_EXCEPTION(Xapian::NetworkError,
                    db.get_metadata("foo"));
     TEST_EXCEPTION(Xapian::NetworkError,
                    db.get_metadata("foo"));
     TEST_EXCEPTION(Xapian::NetworkError,
                    db.metadata_keys_begin());
 
     return true;
```
Any explanation?

guruhegde:
Testcase `closedb1` also gives  error: `NetworkError: write failed (context: remote:tcp(127.0.0.1:1239)) (Broken pipe)`.  But `closedb1` is relatively long testcase and requires changing at multiple places, I didn't test further. Once we know the behavior of above two failed testcases, I will test `closedb1`.

Apart from that all other testcases where `remote_tcp` backend is used, replacing action and expected exception works. 
For the rest of the backend types, we get `InvalidOperationError: Server kill operation not supported for non-remotedatabase type`  as expected. So we need to skip testing for non-remote backend type and `remoteprog_glass` type.

ojwb:
For closedb7, it looks to me like it fails with `DatabaseError` when run initially, but when rerun by the testharness it fails with `NetworkError`.  On the first run you show:

```
Running test: closedb7... POSSIBLE UNRELEASED MEMORY - RETRYING TEST NetworkError: write failed (context: remote:tcp(127.0.0.1:1239)) (Broken pipe)
Expecting exception DatabaseError
```

So it passes, but the test harness sees (via the valgrind integration) that some memory has been allocated but not released for which there's still a pointer to the memory.  This can simply be due to some lazy allocation either in Xapian or a library it uses (e.g. the first call to the libc DNS resolver might cause some data structures to be allocated), but it could also be a genuine memory leak, so what the test harness does in this case is to rerun the same testcase and see if more memory is allocated (that would very strongly suggest a leak) or not (which probably means that it was just some lazy allocation).

So when you change the expected exception class, you're changing which of the two runs it fails on.

I'm not sure why it behaves differently on each run - I'd say that's a bug, though not sure if it's in the library or the test harness.

It looks like closedb10 is similar, but throws `NetworkError` on the first try and doesn't throw on the second, which seems likely to be connected.  I'd guess closedb1 could well be related too.

ojwb:
> Should I file a bug with this behavior?

Please do, unless you can work out what is causing it and come up with a fix.

PR #201, opened 2018-05-18
Generate symbol pair tuple from presenation mathml expression

ghost:
Hi,

As part of the ticket, #761 implemented basic parsing functionality. This PR is still work in progress. 

Please have a look when you have time and give your feedback.

TODO:
Parsing:
- [x] Implement layout schemes such as superscript, subscript, root etc.
- [ ] Add matrix and vector support.
- [x]  Sub-expression.
- [x] Handle `<mrow>` element.
- [x] Handle namespace prefix if present.

In the current implementation, I used `std::vector` to store symbols. Please suggest if there is a better way.


guruhegde:
Windows build failure
```
Running test: closedb5... DatabaseCreateError: Cannot create directory '.glass/dbw' (No child processes)
```
I checked history of builds, somehow one test failed on directory related operation for remotetcp backend. Any idea?

ojwb:
We do see that "No child processes" failure erratically.  It doesn't make a whole lot of sense as creating a directory shouldn't need a child process.  Unfortunately it's hard for me to debug (I don't use those platforms myself) but nobody else seems very interested in keeping them working lately.

I had another look at the code and I wonder if that's a stale errno value from a previous command - I've pushed 2fbb24a4fcc239058ce3e712ebf094087a2e78d5 which may help.  It probably won't fix the error, but it may mean we get a more helpful errno value.

xapian-core/include/xapian/mathtermgenerator.h
```diff
@@ -0,0 +1,89 @@
...
+    MathTermGenerator();
+
+    /// Destructor.
+    ~MathTermGenerator();
+
+    /** Index some math expression in a std::string.
+     *
+     * @param expr	The expression to index.
+     */
+    void index_math(const char * expr);
```
ojwb: Documentation comment says `std::string`, but the method actually takes a `const char *`.

Worse, if I try to pass a `std::string` as the documentation suggests I can, my code will fail to compile.

xapian-core/queryparser/mathtermgenerator.cc
```diff
@@ -0,0 +1,72 @@
...
+MathTermGenerator::symbol_count()
+{
+    return internal->symbol_count();
+}
+
+string
+MathTermGenerator::get_description() const
+{
+    string s("Xapian::MathTermGenerator(doc=");
+    s += ")";
```
ojwb: There's no value inserted after `doc=` here.

xapian-core/queryparser/mathtermgenerator_internal.h
```diff
@@ -0,0 +1,74 @@
...
+struct Symbol {
+    std::string label;
+    edge e;
+    // Symbols like fraction, matrix are represented over multiple lines.
+    // For example, consider a / b. a goes to top row, b goes to bottom
+    // row and fraction symbol is added in middle row.
+    // TODO Think of better way
+    std::vector<Symbol> trow;	// Top row.
+    std::vector<Symbol> brow;	// Bottom row.
+    explicit Symbol(std::string l, edge e_) : label(l), e(e_) { }
```
ojwb: `explicit` is only useful when the constructor has a single-argument form - this one doesn't.

jaylett: I believe that is no longer true in C++11. (See [this article](https://abseil.io/tips/142) for instance.) I think this means we should be making many more of our constructors explicit, so you can't do things like:

```c++
```

ojwb: Thanks, I wasn't aware of that.

The case for `explicit` for single argument constructors is rather stronger since pretty much any value passed or assigned could potentially be implicitly converted without it being obvious at the point it happens.  For the multi-argument case it needs to be a braced list it seems.

The "Constructors That Change Types, But Not Values [are OK as implicit]" idea which the article you link to suggests seems a sensible rule (or at least a rule of thumb).

guruhegde: Thanks olly, james for pointing this out. I wasn't aware of that, forgot to remove `explicit` after changing constructor.

So we keep this as it is.

xapian-core/include/xapian/mathtermgenerator.h
guruhegde: I updated the comment.

I will add an interface which takes `std::string` as argument once parsing work is done.

guruhegde:
I forgot to fix some style errors. I will update it next time.
I just realised  commit `bf3b84d` should have been delayed for later stage. From now every time, style check fails. Should I push a fixup commit?

guruhegde:
@olly, I rebased with the master branch. Hence I had to force commit. Should I keep my branch up to date? (I read in some project developer guide that rebase w.r.t master before push) Commits on master branch were not conflicting and not relevant at least in this case, so rebase was not needed I guess. What do you recommend for the future?

guruhegde:
I couldn't understand these two style errors how to fix. any idea?
```
xapian-core/queryparser/mathtermgenerator_internal.cc:108: error: Multiple spaces: +    while (*ch != '\0' && *ch != ' ' &&  *ch != '>') {
xapian-core/queryparser/mathtermgenerator_internal.cc:181: error: Expression split before operator '||', should be after: +			|| tag.compare("semantics") == 0) {
```

jaylett:
The first means there's a repeated space in the middle of the expression (after the second `&&`); the second means you should put the line break after `||` not before it, ie `||` should be the last thing on the previous line, not the first on that line.

ojwb:
> I had another look at the code and I wonder if that's a stale errno value from a previous command - I've pushed 2fbb24a which may help. It probably won't fix the error, but it may mean we get a more helpful errno value.

Sadly that didn't seem to fix the error nor give us a more helpful errno value.

xapian-core/include/xapian/mathtermgenerator.h
ojwb: My suggestion would be to just provide an interface taking `const std::string&`, which will automatically accept a `const char*` and create a temporary `std::string` from it.

That's what we do in almost all other cases in the API (there's the odd exception such as `Utf8Iterator` in cases where it seems likely people might have large amounts of text in buffer and creating a temporary `std::string` would be problematic).

ojwb:
> I forgot to fix some style errors. I will update it next time.
> I just realised commit bf3b84d should have been delayed for later stage. From now every time, style check fails. Should I push a fixup commit?

I'd say leave it - it's not a big deal (though you'll get nagging mails from travis for each push I guess).

ojwb:
I've just pushed a tweak to `xapian-check-patch` to allow long `#error` and `#warning` lines so future runs shouldn't moan about this (travis actually tests the result of merging the branch and git master).

ojwb:
The CI builds are failing on some of your new testcases, for example:

```
Running test: mathquery1... FAILED
api_query.cc:782: ((q.get_length()) == (3))
Expected 'q.get_length()' and '3' to be equal: were 0 and 3
```

and:

```
Running test: mathqueryparse1... FAILED
api_queryparser.cc:3024: ((q.get_length()) == (3))
Expected 'q.get_length()' and '3' to be equal: were 0 and 3
```

guruhegde:
This PR has become monolithic. I am thinking to build on top of this. How would you like to proceed? 
There are few things I can work with the existing review comments:-
  * switching to `std::string` in parser.
  * wrapping parser functions in class, this is not suggested, but  will be refactor without changing the logic much.
  * Testing the feature, I can check how the performance evaluation done, get the dataset etc.

xapian-core/queryparser/mathtermgenerator_internal.cc
```diff
@@ -350,16 +352,16 @@ MathTermGenerator::Internal::index_math(const char * ch, const string & prefix)
...
 
-void
+vector<string>
 MathTermGenerator::Internal::generate_symbol_pair_list()
 {
     // If main row has size n, then number of tuples along main row is
     // T = (n * (n+1))/ 2. if main row has trow and brow of each size 2, then
     // total pairs is T + 4 * T. This computation is not accurate, an attempt to
     // avoid reallocation.
     unsigned long approx_pairs_count = (mrow.size() * (mrow.size() + 1)) * 3;
```
ojwb: `unsigned long` is the wrong type here really - the `size()` methods will return `std::vector<Foo>::size_type` but that should be the same as `size_t`.  So `size_t` or just `auto` is a better choice.

ojwb:
All good questions - I've been wondering how best to handle merging this project, and been meaning to raise that.

Perhaps you, me and @samuelharden should arrange an IRC meeting to discuss?

xapian-core/include/xapian/mathtermgenerator.h
```diff
@@ -0,0 +1,101 @@
...
+
+    /// Get the current document.
+    const Xapian::Document & get_document() const;
+
+    /** Index some math expression in std::string.
+     *
+     * @param expr	The expression to index.
+     * @param prefix	The term prefix to use (default is no prefix).
+     */
+    void index_math(const std::string & expr,
```
samuelharden: Probably comment can be more descriptive here:
Since this is a new function: expression to index seems a bit less on content.
you could add something like what type of expressions are supported and probably if it's one a link to the specification of that instruction.

As this is new for users, for them to understand what exactly will this function will do.

Just a suggestion, probably @ojwb or @jaylett  can tell if it's suitable to add these things in comemnts.

xapian-core/include/xapian/mathtermgenerator.h
```diff
@@ -0,0 +1,101 @@
...
+		    const std::string & prefix = std::string());
+
+    /** Generate and return symbol pair list from math expression in string.
+     *
+     * @param expr	The expression to generate symbol pair.
+     */
+    std::vector<std::string> get_symbol_pair_list(const std::string & expr);
```
samuelharden: Probably use would benefit from more explanation here. about what is a symbol pair list and how does it look.

xapian-core/include/xapian/mathtermgenerator.h
```diff
@@ -0,0 +1,101 @@
...
+
+    // For testing purpose.
+    std::vector<std::string> get_labels_list();
```
samuelharden: If it's just for testing purpose why is it checked in.

And even if it's for testing and you want to checkin. Better to tell what it returns and what to expect from this function.

xapian-core/queryparser/mathtermgenerator_internal.cc
```diff
@@ -0,0 +1,435 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include <config.h>
+
+#include "mathtermgenerator_internal.h"
+
+/* #define DEBUGGING */
```
samuelharden: Probably this shouldn't be checkedin

xapian-core/queryparser/mathtermgenerator_internal.cc
```diff
@@ -0,0 +1,435 @@
...
+		    if (move_to_next_open_tag(tag) &&
+			    tag.compare("mrow") == 0) {
+			if (move_to_next_open_tag(tag))
+			    mrow.back().brow.emplace_back(get_label(tag),
+				    BELOW);
+			while (next_tag(tag).compare("mrow") != 0) {
+			    if (move_to_next_open_tag(tag))
+				// TODO This only works if all the elemments are
+				// token elements.
+				mrow.back().brow.emplace_back(get_label(tag),
```
samuelharden: This seems a bit oddly indented.

xapian-core/include/xapian/mathtermgenerator.h
guruhegde: Sure. More explanation will be better. I will update.

xapian-core/queryparser/mathtermgenerator_internal.cc
guruhegde: This macro is useful for now. I don't know if there is a simple way to avoid part of the file from Git. Please suggest if you have any idea.

xapian-core/queryparser/mathtermgenerator_internal.cc
```diff
@@ -0,0 +1,435 @@
...
+		    }
+		    // Parse denominator.
+		    if (move_to_next_open_tag(tag) &&
+			    tag.compare("mrow") == 0) {
+			if (move_to_next_open_tag(tag))
+			    mrow.back().brow.emplace_back(get_label(tag),
+				    BELOW);
+			while (next_tag(tag).compare("mrow") != 0) {
+			    if (move_to_next_open_tag(tag))
+				// TODO This only works if all the elemments are
```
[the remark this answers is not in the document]
guruhegde: Thanks for the idea.  Recursion seems to be the better way to handle sub-expression.

xapian-core/queryparser/mathtermgenerator_internal.cc
```diff
@@ -0,0 +1,435 @@
...
+			}
+		    } else {
+			mrow.back().brow.emplace_back(get_label(tag),
+				BELOW);
+		    }
+		} else if (tag.compare("mroot") == 0) {
+		    // Add root symbol.
+		    mrow.emplace_back("R", NEXT);
+		    // Parse base.
+		    if (move_to_next_open_tag(tag))
```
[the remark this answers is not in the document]
guruhegde: We should check. I didn't handle the similar case in few other places as well.  

* `<mroot>` can have strictly two child elements . It is not validated here. 
* Sometimes `<mrow>` is used to group set of child elements as one element. Such cases are also not handled.

xapian-core/include/xapian/queryparser.h
```diff
@@ -1066,6 +1066,20 @@ class XAPIAN_VISIBILITY_DEFAULT QueryParser {
...
+     * @param flags		Zero or more QueryParser::feature_flag
+     * 		specifying what features the QueryParser should support.
+     * 		Combine multiple values with bitwise-or (|)
+     * 		(default FLAG_DEFAULT).
+     *	@param default_prefix	The default term prefix to use (default none.
+     *		For example, you can pass "A" when parsing an "Author" field.
+     */
+    Query parse_math_query(const std::string & query_string,
+			   unsigned flags = FLAG_DEFAULT,
+			   const std::string &default_prefix = std::string());
```
samuelharden: Any thoughts on what and when flags and default_prefix would be useful in math query and indexing?

If we can't think any good usecase of flags and default_prefix, does it make sense to have them in function for sake of consistency with parse_query function?

xapian-core/queryparser/queryparser.cc
```diff
@@ -172,6 +173,20 @@ QueryParser::parse_query(const string &query_string, unsigned flags,
...
 
+Query
+QueryParser::parse_math_query(const std::string & query_string, unsigned flags,
+			      const string &)
+{
+    if (query_string.empty()) return Query();
+
+    MathTermGenerator termgen;
+    auto query_terms = termgen.get_symbol_pair_list(query_string);
+    if (flags && FLAG_BOOLEAN)
```
samuelharden: Do we have a boolean term in math ml?

FLAG_BOOLEAN is for distinguishing boolean and probabilistic term in normal term. 
Boolean terms always have wdf(word document frequency) of 1 and other normal terms can have higher wdf which is used in ranking.

Is there a plan to use that concept there and how we would describe boolean in math ml or math aware queries?

xapian-core/queryparser/mathtermgenerator_internal.h
```diff
@@ -0,0 +1,98 @@
...
+    char edge;
+    // Symbols like fraction, matrix are represented over multiple lines.
+    // For example, consider a / b. a goes to top row, b goes to bottom
+    // row and fraction symbol is added in middle row.
+    std::vector<Symbol> trow;	// Top row.
+    std::vector<Symbol> brow;	// Bottom row.
+    Symbol() = default;
+    explicit Symbol(std::string l, char e) : label(l), edge(e) {
+	trow.reserve(EST_SUB_SYMBOLS_COUNT);
+	brow.reserve(EST_SUB_SYMBOLS_COUNT);
```
samuelharden: Do we really need to reserve the space in trow and brow, and does it help?

guruhegde: I used the `reserve()` method to avoid the reallocation with a decent guess on approx. size of the container.  I didn't check whether it is really helpful. At a later stage, I will check this with examples.

xapian-core/queryparser/queryparser.cc
ojwb: Also, `flags && FLAG_BOOLEAN` is buggy as it's doing a boolean AND (`flags & FLAG_BOOLEAN` is a bit-wise AND).  As written, this just tests if flags is non-zero.

ojwb: > Boolean terms always have wdf(word document frequency) of 1 and other normal terms can have higher wdf which is used in ranking.

Minor correction here - boolean terms generally have wdf of 0 (there's no hard requirement as such but that's what `Xapian::Document::add_boolean_term()` does, and it makes sense that adding more filters shouldn't affect the document length, which Xapian defines as the sum of the wdf for all the terms indexing the document).

guruhegde: Sorry, It is a mistake. It should have been bitwise `&` like olly pointed out and `FLAG_DEFAULT` instead of `FLAG_BOOLEAN`.  I pushed a fix by removing flag usage, as the default flag is sufficient for now.

As per usage of `FLAG_BOOLEAN`, I think `dice coefficient` weight metric performs set intersection, hence metric treats each math term as a `BOOLEAN` for the score computation right?

For the following case flag can be used I think:
Given query can be searched with or without unification of symbols. For this case, can we use flag to set the option?

xapian-core/include/xapian/queryparser.h
guruhegde: Right, `flags` parameter is not needed. I removed the usage. Please check another comment for a possible use case for flags.

xapian-core/include/xapian/mathtermgenerator.h
guruhegde: I will update the comment with proper explanation.

guruhegde: Done

xapian-core/include/xapian/mathtermgenerator.h
guruhegde: Done. Let me know if it needs further explanation

xapian-core/include/xapian/mathtermgenerator.h
guruhegde: Done. Let me know if it needs further explanation

jaylett:
@guruhegde you should read up on git merge conflicts, but the stuff from the <<< line to === is your changes, and from there to >>> is the other changes that conflict with yours, that have happened since you branched from master.

In this case it looks like there's a single line edit on master (the "probabilistic" line changes to the "free-text" line), so you want to keep that change and also include your new function prototype (`parse_math_query()`).

ojwb:
A little background - I did a pass over the code and docs and changed uses of "probabilistic" which really just meant "ranked" or "weighted".

Originally the weighting schemes we supported were all probabilistic, but we now support weighting schemes which aren't (and uses can implement their own weighting scheme too).

guruhegde:
@samuelharden olly suggested to keep the `#error ...` line as it is even though it doesn't conform to the style. Aprart from that no other issues.

xapian-core/tests/api_query.cc
```diff
@@ -691,140 +693,32 @@ DEFINE_TESTCASE(notandor1, backend) {
...
-};
-
 DEFINE_TESTCASE(mathquery1, writable) {
     Xapian::WritableDatabase db = get_writable_database();
     Xapian::MathTermGenerator termgen;
 
-    auto t = index_doc;
-    for (unsigned i = 0; i < sizeof(index_doc) / sizeof(char *) - 1; ++i) {
+    std::vector<std::string> docs;
+    ifstream infile("testdata/apitest_allformulae.txt");
```
ojwb: This only works when building in the source tree - you need to use: `test_driver::get_srcdir() + "/testdata/apitest_allformulae.txt"`

xapian-core/tests/api_query.cc
```diff
@@ -691,140 +693,32 @@ DEFINE_TESTCASE(notandor1, backend) {
...
+	    continue;
+	}
+	formula.append(line);
+    }
+
+    if (!formula.empty())
+	docs.push_back(formula);
+
+
+    for (unsigned i = 0; i < docs.size(); ++i) {
```
ojwb: Why not just process each formula as you read it?  Avoids the overhead of storing them all in the vector.

xapian-core/tests/api_query.cc
guruhegde: Thanks. Fixed

xapian-core/tests/api_query.cc
guruhegde: Right, that's simple. Thanks for pointing this out. Fixed.

xapian-core/queryparser/mathtermgenerator_internal.cc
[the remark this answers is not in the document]
guruhegde: I implemented a way to handle sub-expression. Tried to make overall parsing recursive. I feel the solution I provided is not good and needs improvement. If you have any suggestion or input about the approach, that would be helpful.

ojwb:
This is an interesting feature.  It probably has a fairly narrow appeal (but then so do some existing interesting features), however if it's useful to you it's probably very useful.

Is there likely to be further development in the near future?  If not, perhaps we should just merge it to a branch in the git.xapian.org repo and close this MR.

Or if someone's up for mentoring, we could perhaps make it a GSoC project for this year (if we get our act together quickly).  @guruhegde  @gauravaror Any takers?

gauravaror:
I probably won't have time until July to be involved as I am just finishing up the final semester of masters. Having an ongoing semester during GSOC doesn't help.

I will still try to be involved but not quite sure If I can mentor full project myself. If @guruhegde  is willing, I can probably help him a bit.
@ojwb

ojwb:
I wrote:
> Is there likely to be further development in the near future? If not, perhaps we should just merge it to a branch in the git.xapian.org repo and close this MR.

There's been no further work on this, so I've recreated the branch from this PR locally (which was a bit fiddly because Guruprasad's github account no longer exists and github only seems to provide it as a patch series and you have to poke around to find the branch point), reverted the merge of the `DiceCoeffWeight` branch (which was merged long ago), merged current master to the branch, fixed the testcases for changes on master since, checked it builds and the new tests pass, and have pushed to our repo as branch mathml-expression-indexing.

We have https://trac.xapian.org/ticket/761 open to track this so I'm going to close this PR now.  Anyone wanting to progress this should start from the branch in our repo.

PR #202, opened 2018-06-03
Added tests to xapian-letor

addy007-icy:
Test commit to identify failure, as well as 
"Conditional jump or move depends on uninitialised value(s)"


xapian-letor/tests/err_output21.txt
```diff
@@ -0,0 +1,2 @@
+Ranking score for qid:1 = -nan
+Average ranking score = -nan
```
addy007-icy: Produced by a test case which one qrel was missing

xapian-letor/tests/api_letor.cc
```diff
@@ -185,6 +454,139 @@ DEFINE_TESTCASE(listnet_ranker, generated)
...
+//     TEST_EXCEPTION(Xapian::LetorInternalError,
+//			ranker.score(query, qrel, "ListNet_Ranker",
+//			"scorer_output2.txt", 10, ""));
+//     TEST_EXCEPTION(Xapian::FileNotFoundError,
+//				ranker.score("", qrel, "ListNet_Ranker",
+//				"scorer_output2.txt", 10));
+//     TEST_EXCEPTION(Xapian::FileNotFoundError,
+//			ranker.score(qrel, "", "ListNet_Ranker",
+//						"scorer_output.txt2", 10));
+//     ranker.score(query, qrel, "ListNet_Ranker", "ndcg_output2.txt", 10);
```
addy007-icy: https://pastebin.com/EnQTu8Ve
Would be the failure when any of the missing-qrel tests is uncommented.

xapian-letor/tests/err_output21.txt
AyushP123: Which is that test case there is nowhere in the whole file that err_output21 is being created??.

xapian-letor/tests/api_letor.cc
```diff
@@ -143,6 +317,92 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
+			 stod(temp2.substr(t2 + 1))) <
+		     0.001);
+	    }
+	    i++;
+	}
+	TEST_REL(i, ==, 22);
+	TEST(!(iss2 >> temp2));
+    }
+    TEST(!getline(if2, file2));
+
```
AyushP123: You should follow one guideline. In xapian-core there is no blank linke before the return statement so you should avoid adding blank lines before the return statement. Do not follow xapian-letor for coding standards as it is not formatted properly. You should ask @ojwb about more on this. **Similarly look for return statements in all the tests that you have added and make sure that you have removed the blank lines before return.**

xapian-letor/tests/api_letor.cc
```diff
@@ -227,6 +579,93 @@ DEFINE_TESTCASE(svm_ranker, generated)
...
+    TEST_EXCEPTION(Xapian::FileNotFoundError,
+		   ranker.score(qrel, "", "SVM_Ranker",
+				"scorer_output1.txt", 10));
+    ranker.score(query, qrel, "SVM_Ranker", "ndcg_output1.txt", 10);
+    ranker.score(query, qrel, "SVM_Ranker", "err_output1.txt", 10, "ERRScore");
+
+    return true;
+}
+
+
```
AyushP123: Blank Line.

xapian-letor/tests/api_letor.cc
```diff
@@ -264,11 +703,102 @@ DEFINE_TESTCASE(listmle_ranker, generated)
...
     ranker.score(query, qrel, "ListMLE_Ranker", "ndcg_output.txt", 10);
-    ranker.score(query, qrel, "ListMLE_Ranker", "err_output.txt", 10, "ERRScore");
+    ranker.score(query, qrel, "ListMLE_Ranker", "err_output.txt", 10,
+								"ERRScore");
 
     return true;
 }
 
+
+
```
AyushP123: Blank Line.

xapian-letor/tests/api_letor.cc
```diff
@@ -227,6 +579,93 @@ DEFINE_TESTCASE(svm_ranker, generated)
...
+				"scorer_output3.txt", 10));
+    ranker.score(query, qrel, "SVM_Ranker", "ndcg_output3.txt", 10);
+    ranker.score(query, qrel, "SVM_Ranker", "err_output3.txt", 10, "ERRScore");
+
+    return true;
+}
+
+// listmle ranker check /////
+
+
```
AyushP123: Blank Line.

xapian-letor/tests/api_letor.cc
```diff
@@ -264,11 +703,102 @@ DEFINE_TESTCASE(listmle_ranker, generated)
...
+		   ranker.score(qrel, "", "ListMLE_Ranker",
+				"scorer_output1.txt", 10));
+    ranker.score(query, qrel, "ListMLE_Ranker", "ndcg_output1.txt", 10);
+    ranker.score(query, qrel, "ListMLE_Ranker", "err_output1.txt", 10,
+								"ERRScore");
+
+    return true;
+}
+
+
```
AyushP123: Blank Line.

xapian-letor/tests/api_letor.cc
```diff
@@ -264,11 +703,102 @@ DEFINE_TESTCASE(listmle_ranker, generated)
...
+		   ranker.score(qrel, "", "ListMLE_Ranker",
+				"scorer_output3.txt", 10));
+    ranker.score(query, qrel, "ListMLE_Ranker", "ndcg_output3.txt", 10);
+    ranker.score(query, qrel, "ListMLE_Ranker", "err_output3.txt", 10,
+								"ERRScore");
+
+    return true;
+}
+
+
```
AyushP123: Blank Line.

xapian-letor/tests/api_letor.cc
```diff
@@ -180,11 +440,147 @@ DEFINE_TESTCASE(listnet_ranker, generated)
...
+    TEST_EXCEPTION(Xapian::FileNotFoundError,
+		   ranker.score(qrel, "", "ListNet_Ranker",
+				"scorer_output1.txt", 10));
+    ranker.score(query, qrel, "ListNet_Ranker", "ndcg_output1.txt", 10);
+    ranker.score(query, qrel, "ListNet_Ranker", "err_output1.txt", 10,
+								"ERRScore");
+
+    return true;
+}
+
```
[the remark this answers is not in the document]
ojwb: > One way to go about it would be to add a boolean value "set_label" in FeatureVector which would be set by default as False and will only be set True when the label has been assigned. This bool can be accessed by the Scorers which can then extract the labels which have been set and give the corresponding score.

Rather than have a flag member variable to say if the `label` member variable is valid or not, I'd suggest setting `label` to some invalid value (if `0.0` isn't a valid label, that makes a good default for a double as it's typically cheap to initialise to - all zero bits in the commonly used IEEE floating point representation - and cheap to check against; a negative value is another option; or `HUGE_VAL` if both there are both positive and negative valid values).

If you have a thing and a flag to say if the thing is valid, then you have two variables whose states are entwined but for which some combinations of states are "impossible", and you then need to worry about what happens if there's a bug and we end up in one of those "impossible" states.  With C++17 there's [`std::optional`](http://en.cppreference.com/w/cpp/utility/optional) which provides a class to encapsulate that nicely, but we can't assume C++17 yet.  We could potentially provide an equivalent class ourselves but we don't have one currently.

Adding a flag (or using `std::optional`) also increases the size of the class a little.

When all values of the thing are valid, we will need some sort of flag to add an invalid state, but it's generally nicer to avoid a flag if we can.

xapian-letor/tests/api_letor.cc
```diff
@@ -81,6 +81,79 @@ db_index_two_documents(Xapian::WritableDatabase& db, const string&)
...
+    db.add_document(doc);
+    doc.clear_terms();
+    termgenerator.index_text("Score score score score score score", 1, "S");
+    termgenerator.index_text("it might have an absurdly high rank in the qrel "
+			     " file or might have no rank at all in another.  "
+			     "look out for this as a testcase, as might be edgy"
+			     " Good luck and may this be with you. ", 1, "XD");
+    termgenerator.index_text(" Another irrelavent paragraph to make sure the tf"
+			     " values are down, but this increases idf values "
+			     " but let's see how this works out");
```
ojwb: Where the long strings are split above it seems random whether there's a space at the end of the first part, at the start of the second part, or both.  Functionally it doesn't matter in this case as `index_text()` doesn't care (if these were shown to the user they'd see double spaces in the "both" case, so it would matter then), but it looks messy in the source code and distracts the reader a bit.

There are also several misspellings in the sample data here.  Unless they're actually necessary for the testcase it would be better to avoid them - it makes checking for typos in messages and code comments harder.

I noticed at least "irrelavent", "scorring", "advise" (a valid word, but the wrong one - should be "advice" here), "The will are", "dem", and "sens", but there may well be others.

xapian-letor/tests/api_letor.cc
```diff
@@ -81,6 +81,79 @@ db_index_two_documents(Xapian::WritableDatabase& db, const string&)
...
+    doc.clear_terms();
+    termgenerator.index_text("Document has nothing to do with score", 1, "S");
+    termgenerator.index_text(" This is just to check if score is given a higher"
+			     " score if it is in the subject or not. Nothing "
+			     " special, just juding scores by the look of it.  "
+			     " some more scores but a bad qrel should be enough"
+			     "to make sure it is ranked down" , 1, "XD");
+    termgenerator.index_text(" Score might be something else too, but this para"
+			     " refers to score only at an abstract. Scores are "
+			     " in general scorring. Score it!");
```
ojwb: More random spaces around splits and more typos.

xapian-letor/tests/api_letor.cc
```diff
@@ -94,26 +167,169 @@ DEFINE_TESTCASE(createfeaturevector, generated)
     TEST(fv.empty());
```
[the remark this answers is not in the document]
ojwb: > Just before you run FeatureList::create_feature_vectors, you can check for an empty MSet. This will indicate that an empty feature vector is created for an empty MSet

Not sure I follow - are you saying we should check `Xapian::MSet mset;` actually gives an empty `MSet`?

Such a check primary belongs in the xapian-core testsuite (I actually don't see any test of exactly that there currently).  There's an argument for testing things you rely on from the layers you're building on, and `TEST(mset.empty());` is a succinct way to document exactly what the testcase is relying on.  In more complex cases it would certainly be reasonable, and also if we've seen the feature in question work incorrectly!  "A default constructed `Xapian::MSet` is empty" seems both obvious and very likely to work correctly, but I wouldn't object to such a check even here.

> Also I think that adding this test condition only here is enough. You need not test the creation of an empty FeatureVector by running FeatureList::create_feature_vectors on empty MSet in the remaining tests as it would be redundant. Although I think @ojwb might better advise in this regard.

I think I agree - unless I'm missing some detail, this particular part of this testcase is testing the exact same thing as the equivalent part of the "tigers" testcase AyushP123 noted below.  Improving test coverage is the main goal, not making the tests take longer to run.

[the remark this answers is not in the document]
ojwb: BTW, you can use C++11 `auto` here to avoid having to write out the long type name and avoid having to wrap the line:

```
    auto fv = fl.create_feature_vectors(mset, Xapian::Query("lions"), db);
```

xapian-letor/tests/api_letor.cc
```diff
@@ -94,26 +167,169 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+    Xapian::FeatureList fl;
+    Xapian::Database db = get_database("apitest_ranker3",
+				       db_index_one_document);
+    Xapian::Enquire enquire(db);
+    enquire.set_query(Xapian::Query("llamas"));
+    Xapian::MSet mset;
+    vector<Xapian::FeatureVector> fv =
+	    fl.create_feature_vectors(mset, Xapian::Query("llamas"), db);
+    TEST(fv.empty());
+
```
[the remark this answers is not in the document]
ojwb: Using blanks lines judiciously within a function (e.g. to split groups of statements doing distinct tasks) can aid the reader trying to follow the code.  If you look at existing tests (and certainly the more recent ones which adhere better to current coding conventions) you'll see blank lines used in this way - for example, see `xapian-core/tests/api_spelling.cc`.

It's perhaps a little unfair to call this example "random" - one could argue that this is splitting "test behaviour with an empty `MSet`" from "test behaviour with a non-empty `MSet`".  (As you note, that first part is redundant with earlier testcases, but that aside).

This particular case isn't important, but please do use blank lines where it helps readability.

xapian-letor/tests/api_letor.cc
```diff
@@ -94,26 +167,127 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+	int i = 0;
+	while ((iss1 >> temp1) && (iss2 >> temp2)) {
+	    if (i == 0 || i == 1 || i == 21) {
+		TEST_EQUAL(temp1, temp2);
+	    } else {
+		size_t t1 = temp1.find_first_of(':');
+		size_t t2 = temp2.find_first_of(':');
+		TEST(abs(stod(temp1.substr(t1 + 1)) -
+			 stod(temp2.substr(t2 + 1))) <
+		     0.001);
```
ojwb: There's a `TEST_EQUAL_DOUBLE()` macro which would be better to use if it work here (it has a tighter tolerance thatn `0.001`).

xapian-letor/tests/api_letor.cc
```diff
@@ -94,26 +167,127 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+    string file2;
+    while (getline(if1, file1)) {
+	TEST(getline(if2, file2));
+	istringstream iss1(file1);
+	istringstream iss2(file2);
+	string temp1;
+	string temp2;
+	int i = 0;
+	while ((iss1 >> temp1) && (iss2 >> temp2)) {
+	    if (i == 0 || i == 1 || i == 21) {
```
ojwb: It would be good to have a comment explaining why 0, 1 and 21 are special.  I'm somewhat familiar with the code and I'm struggling to think - someone new to the code would likely have no idea.

xapian-letor/tests/api_letor.cc
```diff
@@ -94,26 +167,127 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+    Xapian::prepare_training_file(db_path, query, qrel, 10,
+				  "training_output_data_one_doc.txt");
+    ifstream if1(training_data);
+    ifstream if2("training_output_data_one_doc.txt");
+    string file1;
+    string file2;
+    while (getline(if1, file1)) {
+	TEST(getline(if2, file2));
+	istringstream iss1(file1);
+	istringstream iss2(file2);
```
ojwb: The names `file1` and `file2` are rather misleading - I'd suggest calling these variables `line1` and `line2` since they hold the contents of a line from the corresponding stream.

xapian-letor/tests/api_letor.cc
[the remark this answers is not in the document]
addy007-icy: The singular "tigers" case was a case in which the code is tested for a single document, with a data-rich environment, looking over a case for a single file in the db might be a common mistake.
So I just added one for anyone overlooking that.

xapian-letor/tests/err_output21.txt
addy007-icy: Was commented out, force push to be deleted.

xapian-letor/tests/api_letor.cc
```diff
@@ -180,11 +440,147 @@ DEFINE_TESTCASE(listnet_ranker, generated)
 		   ranker.score(qrel, "", "ListNet_Ranker",
 				"scorer_output.txt", 10));
     ranker.score(query, qrel, "ListNet_Ranker", "ndcg_output.txt", 10);
-    ranker.score(query, qrel, "ListNet_Ranker", "err_output.txt", 10, "ERRScore");
+    ranker.score(query, qrel, "ListNet_Ranker", "err_output.txt", 10,
+								"ERRScore");
```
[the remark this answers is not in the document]
addy007-icy: Was breaking the line limit of 80, what would be the better way to represent? 
 
```
ranker.score(query, qrel, "ListNet_Ranker", 
            "err_output.txt", 10, "ERRScore"); 
```
@ojwb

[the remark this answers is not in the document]
jaylett: @addy007-icy the point Ayush was making was that going through and fixing formatting issues as part of other commits/PRs isn't the best way to do things.

To answer your question, that's not quite how I'd do it — align the opening quote on the second line with the 'q' of 'query' in the first. (And I probably would break before `"ERRScore"` and not before `"err_output.txt"` because I think it'll be clearer what's going on compared to the line above.)

xapian-letor/tests/api_letor.cc
addy007-icy: 0th, 1st and 21st string taken as input, are absolute strings, "1", "qid:20001" and "#docid=1" respectively . 
So we can just check them as normal strings, whereas the other double values would have to tested as TEST_DOUBLE()

xapian-letor/tests/api_letor.cc
```diff
@@ -62,7 +87,7 @@ db_index_two_documents(Xapian::WritableDatabase& db, const string&)
     termgenerator.index_text("This is the paragraph of interest. Tigers are "
 			     "massive beasts - I wouldn't want to meet a "
 			     "hungry one anywhere. Lions are scary even when "
-			     "lyin' down. Bears are scary even when bare. "
+			     "lyin' down. Bears are scary even when bare."
```
ojwb: I suggested it would be better to fix the cases where a string literal is split and there was space both before and after the split.

But this case here (and the one below) aren't instances of that.

Once the compiler joins the adjacent string literals we now get "[...] Bears are scary even when bare.Together I suspect they'd be less scary, [...]" which is missing a space before "Together".

This change is not useful.

xapian-letor/tests/api_letor.cc
```diff
@@ -234,11 +227,10 @@ DEFINE_TESTCASE(createfeaturevectorthree, generated)
     return true;
 }
 
-
 DEFINE_TESTCASE(preparetrainingfileonedb, generated)
 {
     string db_path = get_database_path("apitest_listnet_ranker1",
-				       db_index_one_document);
+				      db_index_one_document);
```
ojwb: This change looks wrong to me - the github diff viewer confuses things a bit as the second line with a leading tab appears to have a one column shift to the left, but when viewed in an editor with the correct settings, the first `d` on the line above should align below the first `"` on the line above that.

xapian-letor/tests/api_letor.cc
```diff
@@ -291,37 +321,34 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
+    string temp2;
+    int i = 0;
+    while ((iss1 >> temp1) && (iss2 >> temp2)) {
+	if (i == 0 || i == 1 || i == 21) {
+	TEST_EQUAL(temp1, temp2);
+	} else {
+	size_t t1 = temp1.find_first_of(':');
+	size_t t2 = temp2.find_first_of(':');
+	TEST_EQUAL_DOUBLE(stod(temp1.substr(t1 + 1)),
+			  stod(temp2.substr(t2 + 1)));
```
ojwb: This block is now underindented (it looked OK before to me).  More cases of this below.

xapian-letor/tests/api_letor.cc
```diff
@@ -291,37 +321,34 @@ DEFINE_TESTCASE(preparetrainingfile, generated)
...
-
     return true;
 }
 
-
 DEFINE_TESTCASE(preparetrainingfilethree_missingQrel, generated)
 {
-    string db_path = get_database_path("apitest_listnet_ranker3	",
-				       db_index_three_documents);
+    string db_path = get_database_path("apitest_listnet_ranker3 ",
```
ojwb: This string is what the database gets called - it really shouldn't include **any** whitespace, otherwise we end up with database filenames with spaces in, and they're just unnecessarily harder to work with if you need to poke test databases by hand during development.

xapian-letor/tests/api_letor.cc
```diff
@@ -507,42 +524,36 @@ DEFINE_TESTCASE(listnet_ranker_three_correct, generated)
...
     mymset = enquire.get_mset(0, 10);
     ranker.train_model(training_data, "ListNet_Ranker");
     ranker.rank(mymset, "ListNet_Ranker");
     TEST_EQUAL(doc1, *mymset[1]);
     TEST_EQUAL(doc2, *mymset[0]);
-    // TEST_EQUAL(doc3, *mymset[2]);
     TEST_EXCEPTION(Xapian::LetorInternalError,
 		   ranker.score(query, qrel, "ListNet_Ranker",
-				"scorer_output3.txt", 10, ""));
+				"scorer_output.txt", 10, ""));
```
ojwb: It seems you've changed all the testcases to use the same filename - `scorer_output.txt`.  Before they used separate filenames, which meant they could be run concurrently.  Now if they're run concurrently they'll overwrite each others output.  We really don't want to create obstacles to parallelising the testsuite.

ojwb: Oh, or are these written into the path set by `ranker.set_database_path(db_path)`?

xapian-letor/tests/api_letor.cc
```diff
@@ -243,12 +239,12 @@ DEFINE_TESTCASE(preparetrainingfileonedb, generated)
     string file2;
     while (getline(if1, file1)) {
     TEST(getline(if2, file2));
-    istringstream iss1(file1);
-    istringstream iss2(file2);
+    istringstream line1(file1);
+    istringstream line2(file2);
```
ojwb: This isn't the change I proposed - it's `file1` that's poorly named here, not `iss1`.  `iss1` is an `istringstream` so the name made sense; `file1` is a string holding a line from a file, so its name is misleading.

Reading through the changes you're making, I'm becoming increasingly concerned that you're not really taking the time to understand the feedback you're being given, but are just trying to hammer out some changes to push and hope that deals with it.  The problem is that inevitably ends up in follow-up feedback and an overly long and drawn-out review process with a lot of frustration on both sides - we feel we're wasting our time trying to help as you don't seem to actually listen, while you feel that you're spending forever trying to get the patch finished.

Let's try to break that cycle.

xapian-letor/tests/api_letor.cc
addy007-icy: That means that the code should look something like this? 
```
while ((line1 >> temp1) && (line2 >> temp2)) {
	if (i == 0 || i == 1 || i == 21) {
	TEST_EQUAL(temp1, temp2);
	} else {
	    size_t t1 = temp1.find_first_of(':');
	    size_t t2 = temp2.find_first_of(':');
	    TEST_EQUAL_DOUBLE(stod(temp1.substr(t1 + 1)),
			      stod(temp2.substr(t2 + 1)));
	}
	i++;
    }
```
instead of 
```
while ((line1 >> temp1) && (line2 >> temp2)) {
	if (i == 0 || i == 1 || i == 21) {
	    TEST_EQUAL(temp1, temp2);
	} else {
        size_t t1 = temp1.find_first_of(':');
	size_t t2 = temp2.find_first_of(':');
	TEST_EQUAL_DOUBLE(stod(temp1.substr(t1 + 1)),
			  stod(temp2.substr(t2 + 1)));
	}
	i++;
    }
```

xapian-letor/tests/api_letor.cc
```diff
@@ -94,26 +167,169 @@ DEFINE_TESTCASE(createfeaturevector, generated)
...
+DEFINE_TESTCASE(preparetrainingfileonedb, generated)
+{
+    string db_path = get_database_path("apitest_listnet_ranker1",
+				       db_index_one_document);
+    string data_directory = test_driver::get_srcdir() + "/testdata/";
+    string query = data_directory + "queryone.txt";
+    string qrel = data_directory + "qrelone.txt";
+    string training_data = data_directory + "training_data_one_doc.txt";
+    Xapian::prepare_training_file(db_path, query, qrel, 10,
+				  "training_output_data_one_doc.txt");
```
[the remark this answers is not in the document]
ojwb: That sounds like a lack of error checking somewhere.  Probably failing to check that `ifstream` actually opened a file - you need to call `is_open()` to see if opening the file (via the constructor or an explicit call to the `open()` method) actually worked.

That should be fixed, and a regression testcase added (one which fails before the fix and passes after).

xapian-letor/tests/api_letor.cc
addy007-icy: Sorry. I interpreted that as "A file has being read on opening the ifstream" and "A line has been read from the file" . I think I got distracted into thinking that. I'm really sorry. I will pay utmost attention onwards. 
So sorry for this.

xapian-letor/tests/api_letor.cc
ojwb: Neither of those is correct, but the second is incorrect in more ways - the first is just missing an indent of 4 columns on the first line.

Set up your editor so that a tab character in the file displays as moving to the next multiple of 8 column (so from the start of a line, a tab character moves from column 0 to 8, two tab characters to 16, etc).  It seems a tab display width of 4 is popular on Microsoft platforms, but 8 is the Unix convention and what we use.

Set the "shift indent" (or whatever your editor calls it) to 4 columns - each indent level (with a few exceptions) should be 4 columns.  So one indent from the start of a line is 4 spaces, two indent levels is a tab character, 3 is a tab character followed by 4 spaces, 4 is two tab characters, etc.

If your editor is too limited to handle the distinction (apparently some are), I'd strongly suggest setting it up for tab display width 8 and then managing the indenting by hand.  If you set tab display width to anything other than 8 you'll inevitably have a very confusing time trying to read existing code.

If you're using vim or neovim, then this should be very close to correct for Xapian code (the lines starting `"` are just comments in vim config file syntax):

```
set ts=8
set sw=4
set sts=4
" Fix indent after: case 0: {
" Indent public/protected/private by half sw
" Function return type gets no indent
" Line up in column after unclosed (
set cinoptions=l1,g0.5s,h0.5s,t0,(0
```

xapian-letor/tests/api_letor.cc
ojwb: OK, so can you explain that in a comment for the benefit of people trying to understand this in the future without them having to go digging through comments on github PRs (assuming github even still exists then).

One of those people could well be you - I think most experienced developers could tell you a tale of failing to follow some code which is rather lacking in comments, then digging into the version control history (e.g with `git blame`) to find out who wrote it, only to discover they themselves wrote it years earlier.

xapian-letor/tests/api_letor.cc
addy007-icy: And what you said makes much more sense than what I interpreted. I'll fix this right away.

xapian-letor/tests/api_letor.cc
addy007-icy: @ojwb , would this be better? To include the while above? 
```
    while (getline(if1, line1)) {
	TEST(getline(if2, line2));
	istringstream iss1(line1);
	istringstream iss2(line2);
	string temp1;
	string temp2;
	int i = 0;
	while ((iss1 >> temp1) && (iss2 >> temp2)) {
	    if (i == 0 || i == 1 || i == 21) {
		TEST_EQUAL(temp1, temp2);
	    } else {
		size_t t1 = temp1.find_first_of(':');
		size_t t2 = temp2.find_first_of(':');
		TEST_EQUAL_DOUBLE(stod(temp1.substr(t1 + 1)),
				  stod(temp2.substr(t2 + 1)));
	    }
	    i++;
	}
	TEST_REL(i, ==, 22);
	TEST(!(iss2 >> temp2));
    }
    TEST(!getline(if2, line2));
```

xapian-letor/tests/api_letor.cc
addy007-icy: They're just to test exception, so they dont matter right?

xapian-letor/tests/api_letor.cc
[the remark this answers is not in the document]
addy007-icy: Would keep it in a different commit, along with all the checks in cases when one of the qrels are missing. 

so a boolean TEST(if.is_open()) would be needed right after every open right?

xapian-letor/tests/api_letor.cc
```diff
@@ -425,7 +389,7 @@ DEFINE_TESTCASE(listnet_ranker, generated)
     Xapian::ListNETRanker ranker;
     TEST_EXCEPTION(Xapian::FileNotFoundError, ranker.train_model(""));
     string db_path = get_database_path("apitest_listnet_ranker",
-				      db_index_two_documents);
+				       db_index_two_documents);
```
addy007-icy: The 'd' comes under " now, github has it not aligned properly. This issue was there at a lot of places

xapian-letor/tests/api_letor.cc
ojwb: Yes.

xapian-letor/tests/api_letor.cc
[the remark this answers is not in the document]
ojwb: I was assuming the failure was in `Xapian::prepare_training_file()`, but I think I misunderstood where the problem is.

If the problem was a failure to check that we successfully open an `ifstream` inside the library, then `TEST(...)` wouldn't work for checking for the failure as that's a testsuite specific thing - the library code should throw an exception to signal such a problem.  Then a regression testcase would use `TEST_EXCEPTION()` to check that passing a non-existant training file to `Xapian::prepare_training_file()` causes the appropriate exception to be thrown.  And in that case it would make sense to add that regression testcase in a separate commit - probably even better to make it a separate PR so that we can try to wrap this one up and get it merged.

If it's just that we're not checking that `if1` and `if2` below are actually opened successfully, then `TEST(if1.is_open())` etc below does the trick (and that can go in this PR - as a separate commit makes sense there too).

xapian-letor/tests/api_letor.cc
```diff
@@ -81,6 +105,54 @@ db_index_two_documents(Xapian::WritableDatabase& db, const string&)
...
+    Xapian::TermGenerator termgenerator;
+    termgenerator.set_document(doc);
+    termgenerator.set_stemmer(Xapian::Stem("en"));
+    termgenerator.index_text("The will", 1, "S");
+    termgenerator.index_text("The will are considered stop words in xapian and "
+			     "would be thrown off, so the query I want to say "
+			     "is score, yes, score. The Score of a game is "
+			     "the determining factor of a game, the score is "
+			     "what matters at the end of the day. so my advise "
+			     "to everyone is to Score it!.", 1, "XD ");
```
ojwb: Here `"XD "` has an unwanted space in it, which will mean that the indexed terms will contain a space!

xapian-letor/tests/api_letor.cc
```diff
@@ -179,12 +505,14 @@ DEFINE_TESTCASE(listnet_ranker, generated)
...
-    ranker.score(query, qrel, "ListNet_Ranker", "err_output.txt", 10, "ERRScore");
-
+    ranker.score(query, qrel, "ListNet_Ranker", "ndcg_output_ListNet_3.txt=",
+		 10);
+    ranker.score(query, qrel, "ListNet_Ranker", "ndcg_output_ListNet_3.txt", 10,
+		 "ERRScore");
     return true;
 }
 
+/// Svn ranker check
```
ojwb: Typo - should be SV**M**

xapian-letor/tests/api_letor.cc
```diff
@@ -81,6 +105,54 @@ db_index_two_documents(Xapian::WritableDatabase& db, const string&)
     db.add_document(doc);
 }
 
+// To check for three documents. out of which one is irrelevent
```
ojwb: Typo: should be "irrelevant"

xapian-letor/tests/api_letor.cc
```diff
@@ -81,6 +105,54 @@ db_index_two_documents(Xapian::WritableDatabase& db, const string&)
...
+    termgenerator.increase_termpos();
+    termgenerator.index_text("Score score score is important.");
+    db.add_document(doc);
+    doc.clear_terms();
+    termgenerator.index_text("Score score score score score score", 1, "S");
+    termgenerator.index_text("it might have an absurdly high rank in the qrel "
+			     "file or might have no rank at all in another. "
+			     "Look out for this as a testcase, might be edgy "
+			     "good luck and may this be with you.", 1, "XD");
+    termgenerator.index_text("Another irrelevent paragraph to make sure the tf "
```
ojwb: Typo: should be "irrelevant"

PR #203, opened 2018-06-16
Use a brief description for xapian-core.pc.

orbea:
According to the pkg-config(1) man page the 'Description' field should be brief and pkgconf will print warnings with --validate. A multi line 'Description' is not valid and neither pkg-config or pkgconf will print more than the first line with --list-all.

This description is better suited for a man page or README.

See:
```
Description:
  This should be a brief description of the package
```
https://linux.die.net/man/1/pkg-config

pkgconf will print warnings.
```
$ pkgconf --validate xapian-core
/usr/lib64/pkgconfig/xapian-core.pc:8: warning: whitespace encountered while parsing key section
/usr/lib64/pkgconfig/xapian-core.pc:9: warning: whitespace encountered while parsing key section
/usr/lib64/pkgconfig/xapian-core.pc:10: warning: whitespace encountered while parsing key section
```

### DIFF
--- a/xapian-core/pkgconfig/xapian-core.pc.in
+++ b/xapian-core/pkgconfig/xapian-core.pc.in
@@ -4,10 +4,7 @@ libdir=@libdir@
 includedir=@incdir@
 
 Name: The Xapian Probabilistic Information Retrieval Library
-Description: Xapian is an Open Source Probabilistic Information Retrieval framework. It
-offers a highly adaptable toolkit that allows developers to easily add advanced
-indexing and search facilities to applications. This package provides the
-libraries for applications using Xapian functionality.
+Description: Xapian is an Open Source Probabilistic Information Retrieval framework.
 URL: https://xapian.org/
 Version: @VERSION@
 Cflags: -I${includedir} @abi_affecting_cxxflags@


ojwb:
This change seems to make sense, though AIUI pkgconf is a reimplementation of pkg-config and pkg-config itself seems to validate this file OK:

```
$ pkg-config --validate xapian-core
$ echo $?
0
```

If you believe that such files are invalid, you should really file a bug against pkg-config for validating them successfully.

orbea:
pkgconf is more strict and the freedesktop.org pkg-config seems to miss this one even if their documentation says it should be brief and most other examples on my system are not multiple lines. For what its worth a recent change in pkgconf caused `pkgconf --exists` to segfault on `xapian-core.pc` and other pc files because of these invalid multi line descriptions, but this was worked around in pkgconf very quickly. I figured fixing it in xapian-core made sense too.

Also `--list-all` in both pkg-config and pkgconf will print all of the pc files and their descriptions with one line for each. Obviously this truncates the description.
```
$ pkgconf --list-all | grep xapian
xapian-core                    The Xapian Probabilistic Information Retrieval Library - Xapian is an Open Source Probabilistic Information Retrieval framework. It
```

orbea:
Also if it wasn't clear, the pkg-config issue tracker is a ghost town. I much prefer to defer to the maintained and well supported pkgconf which considers these invalid.

See:
https://git.dereferenced.org/pkgconf/pkgconf/commit/5601568267939c4890bab9cbd64c30256455f29a
https://git.dereferenced.org/pkgconf/pkgconf/commit/8f6e071d5836af6c767af6a9da713d7e09f9a466
https://git.dereferenced.org/pkgconf/pkgconf/commit/6f05fec4b590cbd77800f48f8d7ae54e66c81d12

And:
https://bugs.freedesktop.org/buglist.cgi?component=src&product=pkg-config

PR #205, opened 2018-06-19
Diversification: Optimise GLS-MPT

uppinder:
This PR implements C2-GLS which is an optimised version of GLS-MPT (current diversification implementation).

Tasks:
- [x] Implement C2-GLS
- [x] Modify api_diversify.cc (functional test)
- [x] Modify perftest_diversify.cc (performance test)


xapian-core/include/xapian/diversify.h
```diff
@@ -90,6 +92,7 @@ class XAPIAN_VISIBILITY_DEFAULT Diversify {
      *	  			[1e-6,1]
      */
     explicit Diversify(unsigned int k_,
+		       double r_,
```
ojwb: The lack of a default value for `r_` here is causing the tests to fail (in the api_generated tests, because those expect that `Xapian::Diversify foo(10);` works).

If `r_` is a number of document, an integer type would seem more appropriate (e.g. `Xapian::doccount` - perhaps `k_` should be that too actually).

uppinder: Oops, forgot to modify the api test for ```Xapian::Diversify```. And also thanks for pointing out ```Xapian::doccount```, I didn't realise that was there. Pushed two fixup commits to reflect these changes.

xapian-core/diversify/diversify.cc
```diff
@@ -146,20 +133,19 @@ Diversify::Internal::compute_diff_dmset(const vector<Xapian::docid>& dmset)
 }
 
 double
-Diversify::Internal::evaluate_dmset(const vector<Xapian::docid>& dmset)
+Diversify::Internal::evaluate_dmset(const vector<Xapian::docid>& dmset,
+			  const Xapian::ClusterSet& cset)
```
ojwb: The continuation line looks underindented to me.

xapian-core/diversify/diversify.cc
```diff
@@ -186,23 +172,45 @@ Diversify::Internal::get_dmset(const MSet& mset)
...
+		}
 
-	    bool found_better_doc = false;
-	    for (unsigned int j = 0; j < diff_dmset.size(); ++j) {
-		vector<Xapian::docid> temp_dmset = curr_dmset;
-		temp_dmset[i] = diff_dmset[j];
-		double score = evaluate_dmset(temp_dmset);
+		auto temp_dmset = curr_dmset;
+		temp_dmset[i] = topc[j];
+		double score = evaluate_dmset(temp_dmset, cset);
```
ojwb: So if I follow the code correctly here we make a copy of the whole of `curr_dmset` and change one entry in the copy, test that, and then throw away the copy.

It would probably be more efficient to actually modify that single element in `curr_dmset`, test the modified `curr_dmest` and then undo the change (we'd just need to store the old value of `curr_dmset[i]` in a temporary variable to be able to undo the change).

xapian-core/diversify/diversify.cc
uppinder: Correction pushed in a fixup commit.

xapian-core/diversify/diversify.cc
uppinder: This definitely is more efficient, thanks for pointing this out. I've pushed a fixup commit to reflect this change.

ojwb:
Thanks - merged.

PR #206, opened 2018-07-07
qrel file input format to take file name/path.

addy007-icy:
As many datasets provide the file name as the unique identifier to a file,
keeping the qrel format to take the file name/path as input would be better
not interfering with any doc-id stored internally in a xapian-db


xapian-letor/tests/api_letor.cc
```diff
@@ -43,6 +43,7 @@ db_index_one_document(Xapian::WritableDatabase& db, const string&)
     termgenerator.set_document(doc);
     termgenerator.set_stemmer(Xapian::Stem("en"));
     termgenerator.index_text("Tigers are solitary animals", 1, "S");
+    termgenerator.index_text("1", 1, "U/");
```
ojwb: `TermGenerator` is for generating terms by parsing text and splitting it into words.

For boolean terms, use `Document::add_boolean_term()`.

xapian-letor/ranker/ranker.cc
```diff
@@ -114,9 +114,9 @@ static int
...
 {
     int label = -1;
-    string id = std::to_string(doc.get_docid());
-    map<string, map<string, int> >::const_iterator outerit;
-    map<string, int>::const_iterator innerit;
+    int id = doc.get_docid();
+    map<string, map<int, int> >::const_iterator outerit;
+    map<int, int>::const_iterator innerit;
 
     outerit = qrel.find(qid);
```
ojwb: It's better to declare a variable at the point where you can initialise if that's easy to do (which it is here for `outerit` and `innerit`).

Also with C++11 better in a case like this to just use `auto` rather than manually writing out the long templated typenames.

xapian-letor/ranker/ranker.cc
```diff
@@ -164,7 +164,12 @@ load_relevance(const std::string & qrel_file)
 	    if (*end) {
 		throw LetorParseError("Could not parse relevance label in Qrel file at line:" + str(qrel_count));
 	    }
-	    qrel1[token[0]].insert(make_pair(token[2], label));
+	    // Map the file name in the qrel file, to the doc-id
+	    Xapian::PostingIterator db_iter = (db.postlist_begin("U/" + token[2]));
```
ojwb: There's no need to put parentheses around the value here, and it makes it less readable if anything (because you end up with stacked `))` at the end which means the reader has to pause to match them.

xapian-letor/ranker/ranker.cc
```diff
@@ -164,7 +164,12 @@ load_relevance(const std::string & qrel_file)
...
+	    if (db_iter != db.postlist_end("U/" + token[2])) {
+		int doc_id = (*db_iter);
+		qrel1[token[0]].insert(make_pair(doc_id, label));
+	    }
```
ojwb: If we don't find the document is quietly ignoring it the appropriate response?

addy007-icy: Should we log that?

xapian-letor/tests/api_letor.cc
addy007-icy: How do I generate a URL extension (The U/"..") using `TermGenerator` other than this?

xapian-letor/ranker/ranker.cc
addy007-icy: I feel throwing an error of "File 'X' in qrel not found" would be a decent response, throw an error and ignore the current file name, and proceed to the next. As the current xapian-prepare-training file "expects" the format to be correct, if not, it shows an LetorParseError logged as:
`"Could not parse relevance label in Qrel file at line:"` after which the code exits.
But that is only when the format of the Qrel file is wrong, at cases where the input file is "not in the wrong format" it ends up writing nothing to the training-file and an empty training file is created. I think a sort-of log line or something should be helpful. Without logging anything, users would have to start digging in, where the error might have come from.

ojwb: > throw an error and ignore the current file name, and proceed to the next

Once you've thrown an exception, you can't just proceed to the next entry because throwing an exception jumps to wherever the most recent enclosing `try`/`catch` which matches is, and then execution continues from there (or the program terminates if there isn't a `try`/`catch` which matches).

Your reasoning does suggest an exception makes most sense though.

Thinking this through, the approach we have here of parsing a file in a particular format to get the relevance judgements seems a little awkward - there's not particular reason why you have to store the judgements in a file even (could be in a database, or calculated on demand - currently such setups would be forced to generate a temporary file of the required format).

As we've found here, it means that error reporting is less flexible.

Also, the code to map the letor "docid" (which is the filename in this case, but that's probably not universally true) to a `Xapian::docid` is rather application-specific.  Most Xapian databases don't have the `U`-prefix terms which this code assumes.

Ultimately I wonder if a better API here would be one which allows the caller to create a "relevance judgements" object and then add judgements to it.  The code to parse and load a "qrel format" file would then go into a tool which iterates over the qrel file and for each line of data calls this new API to add a judgement.  The application-specific code to determine the `Xapian::docid` goes in the application, and the application can log and skip lines that don't have matching documents.

So I think probably throw an exception here for now, and at some point this API can get replaced with a cleaner one.

PR #207, opened 2018-07-10
Update duplicate test in closedb10 testcase

guruhegde:
I noticed a minor error in `closedb10` testcase. I guess argument for second `get_metadata()` should be `bar`. Right?


ojwb:
Good spot.  Rebased onto master and pushed - thanks.

PR #208, opened 2018-07-14
Implementation of LCD clustering

uppinder:
Contains implementation of the LC clustering algorithm for diversification which is adapted from [here](https://www.sciencedirect.com/science/article/pii/S1570866712001074) (algorithm 2, page 81).

### DIFF
--- a/xapian-core/cluster/lc.cc
+++ b/xapian-core/cluster/lc.cc
@@ -1,0 +1,163 @@
+/** @file lc.cc
+ *  @brief LC clustering API
+ */
+/* Copyright (C) 2018 Uppinder Chugh
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
+#include <config.h>
+
+#include "xapian/cluster.h"
+#include "xapian/error.h"
+
+#include "debuglog.h"
+
+#include <algorithm>
+#include <vector>
+
+using namespace Xapian;
+using namespace std;
+
+typedef set<pair<Point, double> > PSet;
+
+LC::LC(unsigned int k_)
+    : k(k_)
+{
+    LOGCALL_CTOR(API, "LC", k_);
+    if (k_ == 0)
+	throw InvalidArgumentError("Number of required clusters should be greater than zero");
+}
+
+string
+LC::get_description() const
+{
+    return "LC()";
+}
+
+struct pcompare {
+    bool operator() (const pair<Point, double> &a,
+		     const pair<Point, double> &b) const {
+	return a.second > b.second;
+    }
+};
+
+struct dcompare {
+    bool operator() (const pair<PSet::iterator, double> &a,
+		     const pair<PSet::iterator, double> &b) const {
+	return a.second < b.second;
+    }
+};
+
+ClusterSet
+LC::cluster(const MSet &mset)
+{
+    LOGCALL(API, ClusterSet, "LC::cluster", mset);
+
+    doccount size = mset.size();
+    if (k >= size)
+	k = size;
+
+    // Store each document and it's rel score from given mset
+    set<pair<Point, double>, pcompare> points;
+
+    // Initiliase points
+    TermListGroup tlg(mset);
+    for (MSetIterator it = mset.begin(); it != mset.end(); ++it)
+	points.emplace(Point(tlg, it.get_document()), it.get_weight());
+
+    // Container for holding the clusters
+    ClusterSet cset;
+
+    // Instantiate this for computing distance between points
+    CosineDistance distance;
+
+    // First cluster center
+    PSet::iterator cluster_center = points.begin();
+
+    PSet::iterator it;
+    for (unsigned int cnum = 1; cnum <= k; ++cnum) {
+	// Container for new cluster
+	Cluster cluster;
+
+	// The original algorithm accepts a parameter 'k' which is the number
+	// of documents in each cluster, which can be hard to tune and
+	// especially when using this in conjunction with the diversification
+	// module. So, for now LC clustering accepts 'k' as the number of
+	// clusters and divides the documents equally in the first k-1 clusters
+	// and the remaining in the last cluster. This needs to be tested on a
+	// dataset to see how well this works.
+	unsigned int num_points = size / k;
+	if (cnum == k)
+	    num_points = size - (k - 1) * (size / k);
+
+	/* Select (num_points - 1) nearest points to cluster_center from
+	*  from 'points' and form a new cluster
+	*/
+
+	// Store distances of each point from current cluster center
+	// Iterator of each point is stored for fast deletion from 'points'
+	vector<pair<PSet::iterator, double> > dist_vector;
+	for (it = points.begin(); it != points.end(); ++it) {
+	    if (it == cluster_center)
+		continue;
+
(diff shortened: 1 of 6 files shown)


xapian-core/cluster/lc.cc
```diff
@@ -39,7 +39,8 @@ LC::LC(unsigned int k_)
 {
     LOGCALL_CTOR(API, "LC", k_);
     if (k_ == 0)
-	throw InvalidArgumentError("Number of required clusters should be greater than zero");
+	throw InvalidArgumentError("Number of required clusters should be \
+				    greater than zero");
```
jaylett: You don't want to do this, because it inserts a lot of spaces into the string. Instead, you can use multiple string constants; they'll be concatenated together at compile time. So something like:

```c++
    throw InvalidArgumentError("First string constant starts here "
                               "second string constant just continues");
```

Note the trailing space in the first string constant. See for instance `xapian-core/tests/harness/backendmanager.cc:230` or so.

xapian-core/cluster/lc.cc
```diff
@@ -91,7 +92,7 @@ LC::cluster(const MSet &mset)
     PSet::iterator it;
     for (unsigned int cnum = 1; cnum <= k; ++cnum) {
 	// Container for new cluster
-	Cluster cluster;
+	Cluster C;
```
jaylett: Local variable names are conventionally lower-case. Also, you should be able to come up with a more useful name than just `C`, even though the obvious one isn't possible.

uppinder: Thanks for pointing this out, I've modified the variable name used in a fixup commit.

xapian-core/cluster/lc.cc
uppinder: Could you please elaborate a bit more on "inserts a lot of spaces into the string"?

jaylett: Think about what your string constant actually contains. Before the word "greater" there will be a lot of spaces.

uppinder: Oh, I get what you mean. The backslash escapes the newline and the spaces before 'greater' get included in the string, which is not the intended effect. Thanks for the tip.

xapian-core/cluster/lcd_clusterer.cc
```diff
@@ -0,0 +1,164 @@
...
+ClusterSet
+LCDClusterer::cluster(const MSet &mset)
+{
+    LOGCALL(API, ClusterSet, "LCDClusterer::cluster", mset);
+
+    doccount size = mset.size();
+    if (k >= size)
+	k = size;
+
+    // Store each document and it's rel score from given mset
```
ojwb: Not a contraction of "it is", so should be "its".

xapian-core/cluster/lcd_clusterer.cc
```diff
@@ -0,0 +1,164 @@
...
+};
+
+ClusterSet
+LCDClusterer::cluster(const MSet &mset)
+{
+    LOGCALL(API, ClusterSet, "LCDClusterer::cluster", mset);
+
+    doccount size = mset.size();
+    if (k >= size)
+	k = size;
```
ojwb: This changes the object's `k` member, which means if you reuse an `LCDClusterer` its history can affect results - that seems surprising.

E.g. create an `LCDClusterer` with `k = 7`, cluster a 4 element MSet and you now have an `LCDClusterer` with `k = 4`.

uppinder: Valid point, this shouldn't be the intended behaviour, as you rightly point out, it's opaque to the user. I've pushed a fixup commit for the modification.

xapian-core/cluster/lcd_clusterer.cc
uppinder: Modification pushed in a fixup commit.

xapian-core/cluster/lcd_clusterer.cc
```diff
@@ -0,0 +1,164 @@
...
+	// The original algorithm accepts a parameter 'k' which is the number
+	// of documents in each cluster, which can be hard to tune and
+	// especially when using this in conjunction with the diversification
+	// module. So, for now LCD clustering accepts 'k' as the number of
+	// clusters and divides the documents equally in the first k-1 clusters
+	// and the remaining in the last cluster. This needs to be tested on a
+	// dataset to see how well this works.
+	unsigned int num_points = size / k_;
+	if (cnum == k_)
+	    num_points = size - (k_ - 1) * (size / k_);
```
ojwb: Integer division truncates towards zero, so if `k_` is 5 and `num_points` is 24, then the first 4 clusters will be of size 4 and the last of size 8.

Having one cluster twice the size of the others seems unlikely to be helpful, and I wonder if we want to aim for a more even split such that the cluster sizes differ by at most one?

In this case 4 clusters of size 5 and one of size 4, but more generally `n` clusters of size `x` and `k_ - n` of size `x - 1`.

xapian-core/cluster/lcd_clusterer.cc
```diff
@@ -0,0 +1,164 @@
...
+	for (auto it = points.begin(); it != points.end(); ++it) {
+	    if (it == cluster_center)
+		continue;
+
+	    double dist = distance.similarity(cluster_center->first, it->first);
+	    if (dist > max_dist) {
+		max_dist = dist;
+		new_cluster_center = it;
+	    }
+	}
```
ojwb: Can't we just use `dist_vector.back().first` here to find the new centre?

uppinder: You're right, this step is unnecessary. I've pushed a commit with the suggested change, thanks.

xapian-core/cluster/lcd_clusterer.cc
```diff
@@ -90,24 +90,24 @@ LCDClusterer::cluster(const MSet &mset)
...
+    // The original algorithm accepts a parameter k which is the number
+    // of documents in each cluster, which can be hard to tune and
+    // especially when using this in conjunction with the diversification
+    // module. So, for now LCD clusterer accepts k as the number of
+    // clusters and divides the documents into k clusters such that among k
+    // clusters, n clusters have x - 1 points and (k - n) have x points. This
+    // needs to be tested on a dataset to see how well this works.
+    // n * (x - 1) + (k_ - n) * x = size, where 0 <= n < k
+    unsigned n = k_ - size % k_,
+	     x = (size + n) / k_;
```
ojwb: I'd suggest making this two statements as that seems slightly more readable, and more consistent with the general coding style in the Xapian codebase.

Also, when you have a calculation such as this it's good practice to add an assertion that the results work when plugged into the original formula, something like:

```
AssertEq(n * (x - 1) + (k_ - n) * x, size);
```

This protects you from errors in your derivation, typos in the code, clumsy maintenance work in the future, etc.  The assertions compile away to nothing by default, but can be enabled by configuring with `--enable-assertions`.  When enabled, the assertion will throw a runtime exception if the condition isn't true.

Incidentally, `(size + n) / k_` gives the same answer as `(size / k_) + 1` - the `+ 1` comes from `k_/k_` and the rest is `(size - size % k_) / k_` which is just the exact version of `size / k_`.  The benefit of writing `size % k_` and `size / k_` is that the compiler should be able to spot opportunities for optimisation (e.g. see https://godbolt.org/g/AtWCw2 for the code generated by GCC 8.1 for both versions).  This is very much a micro-optimisation and since this isn't a hot loop shaving 3 instructions is basically irrelevant here - just noting it as a tip for cases where such things do matter.

xapian-core/include/xapian/cluster.h
```diff
@@ -33,6 +34,7 @@
 #include <xapian/types.h>
 #include <xapian/visibility.h>
 
+#include <set>
```
ojwb: We don't seem to use `std::set<>` in this header - better to include this from the files which actually use it instead (especially true given this is a public API header).

ojwb:
Aside from the points in my last two comments, I'm happy with this now.

@amandaJayanetti @jaylett OK with you to merge?

xapian-core/include/xapian/cluster.h
uppinder: Thanks for pointing this out. Change pushed in a fixup commit.

xapian-core/cluster/lcd_clusterer.cc
uppinder: Neat catch. I've modified the code to reflect these changes in a fixup commit.

amandaJayanetti:
Looks good to me!

ojwb:
I checked that James' feedback above has been dealt with, and the CI failure is unrelated to these changes, so I'll get this merged.

PR #209, opened 2018-07-28
Add unique terms bound stats to HoneyVersion class

guruhegde:
This address ticket #763

TODO:

- [x] Track stats in honey_compact.cc
- [x] Add test


xapian-core/backends/honey/honey_version.h
```diff
@@ -276,7 +295,9 @@ class HoneyVersion {
 		     Xapian::termcount o_doclen_ubound,
 		     Xapian::termcount o_wdf_ubound,
 		     Xapian::totallength o_total_doclen,
-		     Xapian::termcount o_spelling_wordfreq_ubound);
+		     Xapian::termcount o_spelling_wordfreq_ubound,
+		     Xapian::termcount o_uniq_terms_lbound = 0,
```
guruhegde: @olly, my comment in trac was about this change. If default values are removed, we get a compile error.

guruhegde:
There are no setters for bound stats in `HoneyVersion` class.  Shall I add new interfaces to set newly added stats?

xapian-core/tests/api_backend.cc
```diff
@@ -180,6 +180,13 @@ DEFINE_TESTCASE(dbstats1, backend) {
     // 1.3.2.
     TEST_EQUAL(db.get_wdf_upper_bound(""), 0);
 
+    if (get_dbtype() == "honey") {
+	const Xapian::termcount min_unique_len = 2;
```
guruhegde: How can I verify these values? I got these by setting wrong values first and then adjusting it, clearly, this is not the right away.

xapian-core/backends/databaseinternal.cc
```diff
@@ -104,6 +104,20 @@ Database::Internal::get_wdf_upper_bound(const string & term) const
     return cf;
 }
 
+Xapian::termcount
+Database::Internal::get_uniq_terms_lower_bound() const
+{
+    throw Xapian::UnimplementedError("This backend doesn't support"
+			       "get_uniq_terms_lower_bound");
+}
```
ojwb: However, `1` is a valid lower bound.

xapian-core/backends/databaseinternal.cc
```diff
@@ -104,6 +104,20 @@ Database::Internal::get_wdf_upper_bound(const string & term) const
...
+
+Xapian::termcount
+Database::Internal::get_uniq_terms_upper_bound() const
+{
+    throw Xapian::UnimplementedError("This backend doesn't support"
+			       "get_uniq_terms_upper_bound");
+}
```
ojwb: And the document length upper bound is a valid upper bound for the number of unique terms.

xapian-core/backends/honey/honey_version.h
```diff
@@ -129,6 +129,16 @@ class HoneyVersion {
     /// Oldest changeset removed when max_changesets is set
     mutable honey_revision_number_t oldest_changeset;
 
+    /** A lower bound on the number of unique terms in a document in this
+     *  databse
```
ojwb: Typo: "databse" (also below)

xapian-core/backends/honey/honey_version.h
ojwb: So the problem is when this gets called from `honey_compact.cc` when converting a database from glass?

So in that case, we're setting this to zero, but then working out the correct bounds as we process the termlist.  So basically this is OK, but think it's probably better to explicitly pass `0` from the caller instead of having a default which might get inadvertently used in other situations where it isn't appropriate.

guruhegde: Yes. That was the problem. I will do that

xapian-core/tests/api_backend.cc
ojwb: That is indeed far from ideal as it wouldn't catch a bug in your implementation (though it does at least mean we have a test that fails if these values change, which is better than no test at all!)

You can find the cached database this test uses in `tests/.glass/db__etext` - and we can write a small program which just iterates all the documents in it and reports the actual highest and lowest number of unique terms, which provides a more independent confirmation:

```
~/git/xapian/xapian-core/tests$ cat uniqterms.py 
#!/usr/bin/python
from xapian import *
db = Database(".glass/db__etext")
m = 9999999
M = 0
for i in xrange(1, db.get_doccount()):
    u = db.get_unique_terms(i)
    if u > 0:
        if u < m: m = u
        if u > M: M = u
print(m, M)
~/git/xapian/xapian-core/tests$ python uniqterms.py 
(2, 272)
```

guruhegde: Thanks for the script! Values match.

xapian-core/backends/databaseinternal.cc
guruhegde: I will add this. I thought, no other backend support this interface, thought setting error makes it clear if called by chance.

guruhegde: Done

xapian-core/backends/databaseinternal.cc
guruhegde: Done

xapian-core/backends/honey/honey_version.h
guruhegde: Fixed

xapian-core/backends/databaseinternal.cc
ojwb: We want to be able to call these methods from `Xapian::Weight` subclasses without having to worry about whether the backend actually tracks these values or not.

Currently `DiceCoeffWeight` uses `1` as the lower bound, so returning 1 in the default implementation means that class can just ask the database for the lower bound, and we'll use the same value as currently for backends which don't track these bounds, and the tracked bound for backends which do track it.

guruhegde:
@olly, shall I update `DiceCoefficentWeight` based on this branch?

xapian-core/backends/honey/honey_compact.cc
```diff
@@ -2033,12 +2033,16 @@ HoneyDatabase::compact(Xapian::Compactor* compactor,
 	    auto db = static_cast<const GlassDatabase*>(sources[i]);
 	    auto& v_in = db->version_file;
 	    auto& v_out = version_file_out;
+	    // Glass backend doesn't track unique term bounds, hence setting
+	    // them to 0.
```
ojwb: I think this risks misleading the reader.  Someone new to this code and trying to follow it (or indeed us in several years once we've forgotten the details) would probably think this means that a converted database has zero for the bounds.

It's true we do pass zero for them here, but then we calculate them as we convert the termlist table and fill in the correct values before we write out `version_file_out`.  The part after the "but" is probably the most interesting thing to tell the reader, as it isn't obvious without careful study of the code.

ojwb:
> shall I update `DiceCoefficentWeight` based on this branch?

Yes, that would be good.

`TfIdfWeight` also uses the number of unique terms (with the `L` wdf normalisation) but I'm not sure these bounds are useful there.  The formula is `(1 + log(wdf)) / (1 + log(len / uniq))` which ignoring wdf is maximised when when `len / uniq` is minimised.  By definition that must be >= 1 and currently we calculate the bound at 1.  We can do better if `len_min > uniq_max` but that seems unlikely to be true for any non-trivial document collection.

xapian-core/backends/honey/honey_compact.cc
guruhegde: Done

ojwb:
I was holding off merging this because it changes the format of the database incompatibly, but I realised that with a small tweak we can make it so older Xapian versions can read honey databases with the new stats (and they'll just ignore them), while new versions can read existing honey databases (and use the same weak bounds which we do for glass).

I've merged master to the branch and then made these changes, and have pushed them to the branch for this MR so the CI can check them.

ojwb:
The appveyor build is currently failing on master too, so this is good to merge.

PR #210, opened 2018-08-08
Use automake's parallel test harness

ghost:
This PR implements an initial prototype to speed up the execution of Xapian test-suite.

Please look at the ticket for some more details: https://trac.xapian.org/ticket/767

Total elapsed time for few cases (in minutes rounded off):
*  `make check` : 15
* `make check -j2` : 11
* `make check -j4`: 9

### DIFF
--- a/xapian-core/tests/run_glass
+++ b/xapian-core/tests/run_glass
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
 ./runtest ./apitest -b glass
 ./runtest ./apitest -b singlefile_glass


guruhegde:
@ojwb Travis job for gcc build fails. It was failing before commits done in this PR. You may be aware of the reason too.

ojwb:
The build failure was due to a backported patch with some valid code which GCC 4.7 doesn't like (this is the oldest GCC version 1.4.x aims to support - we configured travis to use this version to catch such problems).  I've already pushed a fix.

xapian-core/tests/run_glass
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+./runtest ./apitest -b glass
+./runtest ./apitest -b singlefile_glass
+./runtest ./apitest -b multi_glass
+./runtest ./apitest -b remoteprog_glass
+./runtest ./apitest -b remotetcp_glass
```
ojwb: By default, `sh` just continues when a command fails, and the non-zero return code is overwritten by the return code of the next command.  Here that means that only the success/failure of `./runtest ./apitest -b remotetcp_glass` will get reflected in the success/failure of the testsuite as a whole.

You can use `set -e` (best put right after the `#!` line) to tell the shell to exit if a command fails (there are some corner cases - for example, the condition of  an `if` or `while` failing sensibly won't cause the script to exit, but that's also true for commands inside a shell function called from the condition of an `if` or `while`, which tends to catch people out).

Demonstration (`$?` is 0 for success, non-zero if the command failed):

```
$ cat test.sh
#!/bin/sh
false
echo OK
$ ./test.sh; echo $?
OK
0
$ cat teste.sh 
#!/bin/sh
set -e
false
echo OK
$ ./teste.sh; echo $?
1
```

xapian-core/tests/api_anydb.cc
```diff
@@ -575,6 +575,11 @@ DEFINE_TESTCASE(topercent1, backend) {
     return true;
 }
 
+#if 0
+/** This testcase uses glass backend, even when backend intened to test is
```
ojwb: Typo: "intened"

xapian-core/tests/api_anydb.cc
```diff
@@ -575,6 +575,11 @@ DEFINE_TESTCASE(topercent1, backend) {
...
+ *  inmemory or other. This testcase may fail when run in parallel, hence
+ *  disabling for now.
+ */
 // tests the percentage values returned
 DEFINE_TESTCASE(topercent2, backend) {
     BackendManagerLocal local_manager(test_driver::get_srcdir() + "/testdata/");
```
ojwb: `git grep BackendManagerLocal` shows two other testcases which ought to have similar treatment.

ojwb:
Seems a nice speed-up for a fairly simple prototype, though unsurprisingly the improvements tail off because of the uneven slices of work.

xapian-core/tests/api_anydb.cc
guruhegde: Done. I had this in mind, but somehow other two test cases never failed, so decided to just keep them.

xapian-core/tests/run_glass
guruhegde: Thanks. Good to know this. I pushed the change.

xapian-core/tests/Makefile.am
```diff
@@ -82,8 +82,8 @@ check-honey: apitest$(EXEEXT)
 endif
 
 ## Test programs to be run
-TESTS = apitest$(EXEEXT) internaltest$(EXEEXT) stemtest$(EXEEXT) \
- unittest$(EXEEXT)
+TESTS = run_honey run_none run_inmemory run_glass \
+	internaltest$(EXTEXT) stemtest$(EXTEXT) unittest$(EXTEXT)
```
ojwb: If we do end up merging this, I notice this changes `$(EXEEXT)` to `$(EXTEXT)`, which is wrong - it's **EXEEXT** meaning the **exe**cutable file **ext**ension - on Unix it's empty (so your typo is harmless there) but on platforms where executables have a `.exe` extension this will break things.

ojwb:
I believe this was intended mainly as a prototype to check what sort of speed up we could easily get.

The numbers shows `-j4` reduces the elapsed time by 40%.  I'd hope we could get closer to 75% reduction with a parallel test harness inside apitest, etc, but 40% isn't to be sniffed at, so perhaps we should consider tidying this up and merging it as an intermediate step if we don't have a parallel test harness project in GSoC this year (or somebody working on it outside of GSoC).  Tanmay's work last year eliminated ` BackendManagerLocal` entirely, and I think most of the other blockers too.

ojwb:
Nobody's worked on a parallel test harness significantly since.

However, the test harness creates honey databases from their glass equivalent, so they can't run be trivially run in parallel and honey needs to be merged into the large glass list, which will reduce the gain from this further - basically we save the time to run `none` and `inmemory` tests (less any parallelism overheads, which are probably not significant).

I'll try updating this and see what the savings look like now, but I'm wondering if it's better to put some extra effort in and implement parallel test running in our test harness itself.  For example, that could keep track of which (database,backend) pairs are currently in use and schedule testcases to avoid conflicts.  I kind of think there's a simple first version of this which would deliver substantial gains.

ojwb:
I tried this updated and with honey added to the glass list, and the speedup was disappointing (the figures are on a different machine, but I think it was ~20%).

I also came up with a fairly simple way to schedule: https://trac.xapian.org/ticket/767#comment:9

This was an interesting experiment, but I don't think the gain from it is sufficient to outweight the reduced usability from not being able to see the logs until after the test.  Currently during development I can see when a testcase failed early in the run and interrupt it - with this I'd only see at the end, which would mean a longer wait in this situation.  Therefore closing in favour of making our harness run cases in parallel (which we have trac ticket 767 for).

PR #212, opened 2018-10-28
Support latest Travis macOS build

jaylett:
Homebrew has for a while shipped python as the python3 package (with
python2 for py2), but Travis was a little behind for a while. It
seems to have caught up, which means that upgrading the python
package before installing python2 is no longer needed.


ojwb:
Great, thanks.

PR #213, opened 2018-12-23
 Explicit DatabaseNotFoundError exception(ticket #773)

VaibhavKansagara:
This PR adds a new class _DatabaseNotFoundError_. Refer to ticket #773 for more details.

### DIFF
--- a/xapian-bindings/perl/Xapian/DatabaseNotFoundError.pm
+++ b/xapian-bindings/perl/Xapian/DatabaseNotFoundError.pm
@@ -1,0 +1,11 @@
+package Xapian::DatabaseNotFoundError;
+
+=head1 NAME
+
+Xapian::DatabaseNotFoundError -  DatabaseNotFoundError indicates that a database is not existing.
+
+=head1 DESCRIPTION
+
+
+=cut
+1;
--- a/xapian-core/backends/dbfactory.cc
+++ b/xapian-core/backends/dbfactory.cc
@@ -360,7 +360,7 @@ open_stub(WritableDatabase &db, const string &file, int flags)
     }
 
     if (db.internal->size() == 0) {
-	throw DatabaseNotFoundError(file + ": No databases listed");
+	throw DatabaseOpeningError(file + ": No databases listed");
     }
 }
 
--- a/xapian-core/backends/dbfactory.cc
+++ b/xapian-core/backends/dbfactory.cc
@@ -133,7 +133,7 @@ open_stub(Database &db, const string &file)
     if (!stub) {
 	string msg = "Couldn't open stub database file: ";
 	msg += file;
-	throw Xapian::DatabaseOpeningError(msg, errno);
+	throw Xapian::DatabaseNotFoundError(msg, errno);
     }
     string line;
     unsigned int line_no = 0;
@@ -262,7 +262,7 @@ open_stub(WritableDatabase &db, const string &file, int flags)
     if (!stub) {
 	string msg = "Couldn't open stub database file: ";
 	msg += file;
-	throw Xapian::DatabaseOpeningError(msg, errno);
+	throw Xapian::DatabaseNotFoundError(msg, errno);
     }
     string line;
     unsigned int line_no = 0;
@@ -360,7 +360,7 @@ open_stub(WritableDatabase &db, const string &file, int flags)
     }
 
     if (db.internal->size() == 0) {
-	throw DatabaseOpeningError(file + ": No databases listed");
+	throw DatabaseNotFoundError(file + ": No databases listed");
     }
 }
 
--- a/xapian-core/backends/glass/glass_version.cc
+++ b/xapian-core/backends/glass/glass_version.cc
@@ -113,7 +113,12 @@ GlassVersion::read()
 	if (rare(fd_in < 0)) {
 	    string msg = filename;
 	    msg += ": Failed to open glass revision file for reading";
+        if(errno == ENOENT || errno == ENOTDIR){
 	    throw Xapian::DatabaseNotFoundError(msg, errno);
+        }
+        else{
+        throw Xapian::DatabaseOpeningError(msg, errno);
+        }
 	}
 	close_fd = fd_in;
     }
--- a/xapian-core/backends/glass/glass_version.cc
+++ b/xapian-core/backends/glass/glass_version.cc
@@ -103,7 +103,7 @@ GlassVersion::read()
 	if (rare(lseek(fd, offset, SEEK_SET) < 0)) {
 	    string msg = "Failed to rewind file descriptor ";
 	    msg += str(fd);
-	    throw Xapian::DatabaseNotFoundError(msg, errno);
+	    throw Xapian::DatabaseOpeningError(msg, errno);
 	}
 	fd_in = fd;
     } else {
--- a/xapian-core/backends/glass/glass_version.cc
+++ b/xapian-core/backends/glass/glass_version.cc
@@ -103,7 +103,7 @@ GlassVersion::read()
 	if (rare(lseek(fd, offset, SEEK_SET) < 0)) {
 	    string msg = "Failed to rewind file descriptor ";
 	    msg += str(fd);
-	    throw Xapian::DatabaseOpeningError(msg, errno);
+	    throw Xapian::DatabaseNotFoundError(msg, errno);
 	}
 	fd_in = fd;
     } else {
@@ -113,7 +113,7 @@ GlassVersion::read()
 	if (rare(fd_in < 0)) {
 	    string msg = filename;
 	    msg += ": Failed to open glass revision file for reading";
-	    throw Xapian::DatabaseOpeningError(msg, errno);
+	    throw Xapian::DatabaseNotFoundError(msg, errno);
 	}
 	close_fd = fd_in;
     }
--- a/xapian-core/backends/glass/glass_version.cc
+++ b/xapian-core/backends/glass/glass_version.cc
@@ -113,12 +113,10 @@ GlassVersion::read()
 	if (rare(fd_in < 0)) {
 	    string msg = filename;
 	    msg += ": Failed to open glass revision file for reading";
-        if(errno == ENOENT || errno == ENOTDIR){
-	    throw Xapian::DatabaseNotFoundError(msg, errno);
-        }
-        else{
-        throw Xapian::DatabaseOpeningError(msg, errno);
-        }
+	    if (errno == ENOENT || errno == ENOTDIR) {
+		throw Xapian::DatabaseNotFoundError(msg, errno);
+	    }
+	    throw Xapian::DatabaseOpeningError(msg, errno);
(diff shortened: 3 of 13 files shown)


VaibhavKansagara:
Here i have added a file which list all the occurrences of **DatabaseOpeningError**
[occurrences.pdf](https://github.com/xapian/xapian/files/2706029/occurrences.pdf)

xapian-core/backends/dbfactory.cc
```diff
@@ -360,7 +360,7 @@ open_stub(WritableDatabase &db, const string &file, int flags)
     }
 
     if (db.internal->size() == 0) {
-	throw DatabaseOpeningError(file + ": No databases listed");
+	throw DatabaseNotFoundError(file + ": No databases listed");
```
ojwb: I'm not sure which exception is more appropriate here - this is for opening a "stub database file", which is a text file which should list one or more databases to open.  In this case the stub file was found but no databases were listed in it.

@Kronuz Does the behaviour in this situation matter for your use case?

xapian-bindings/perl/Xapian/DatabaseNotFoundError.pm
```diff
@@ -0,0 +1,11 @@
...
+
+=head1 NAME
+
+Xapian::DatabaseNotFoundError -  DatabaseNotFoundError indicates that a database is not existing.
+
+=head1 DESCRIPTION
+
+
+=cut
+1;
```
ojwb: There's no newline character at the end of this file, as the automated patch checker points out:

https://travis-ci.org/xapian/xapian/jobs/471596630

That's what the two red symbols at the end of the last line in github's diff viewer is trying to tell you.

ojwb:
The travis CI builds are failing with test failures which suggest your test suite changes aren't complete.

Run `make check` in the xapian-core directory locally for easier debugging.

(Ignore the appveyor build failures - that's currently broken on master unfortunately).

xapian-core/backends/dbfactory.cc
jaylett: I think it should be `DatabaseOpeningError`:

1. Pretty sure you can't use it to create a new database as @Kronuz would need for `DatabaseNotFoundError`;
2. It feels to me more like an invalid database than a missing one.

ojwb:
On Fri, 28 Dec 2018, 05:27 Germán Méndez Bravo <notifications@github.com
wrote:

> I’m not sure either... probably, if we can create a database there by
> opening using the CREATE flag (without having to do anything else) it
> should be DatabaseNotFoundError, otherwise it should be DatabaseOpeningError
>
That seems a sensible rule of thumb, except that I don't think we want to
differentiate the parent directory existing or not here, which would make a
difference for creating a database there (we'll create the directory for
the database itself if it doesn't exist but not parent directories).

>

Kronuz:
Yes, a missing parent directory should also result in DatabaseNotFoundError, imho.

VaibhavKansagara:
@olly,can you clarify it a little bit more.I also have a doubt about non-existent stub database files in which case _DatabaseNotFoundError_ should be thrown in my opinion. Please correct me if i am wrong.

VaibhavKansagara:
@olly,can you clarify a little bit more. I also have a doubt about non-existent stub database files in which case _DatabaseNotFoundError_ should be thrown in my opinion. Please correct me if i am wrong.

ojwb:
> can you clarify it a little bit more about the part of not differentiating the existence of parent directory.

Kronuz suggested:
> if we can create a database there by opening using the CREATE flag (without having to do anything else) it should be DatabaseNotFoundError, otherwise it should be DatabaseOpeningError 

But consider the situation of trying to open `/home/olly/foo/bar.db` - by a literal interpretation of Kronuz's rule that should throw `DatabaseNotFoundError` if `/home/olly/foo` exists and is a directory, or `DatabaseOpeningError` if `/home/olly/foo` doesn't already exist (or is not a directory) because creating a database won't automatically create any parent directories which don't exist.  I think both these cases should throw `DatabaseNotFoundError` (and Kronuz agrees).

VaibhavKansagara:
@ojwb Thank you for clarification. Are there any other changes that you would suggest?

xapian-core/backends/dbfactory.cc
ojwb: > Are there any other changes that you would suggest?

As discussed above, I think this one should stay as `DatabaseOpeningError`.

(Perhaps this case shouldn't actually be an error, but should instead just open a database with zero shards - i.e. an empty database; currently to specify an empty database in a stub file you have to put `inmemory`.  But if we change that it should really be done separately to the changes for adding `DatabaseNotFoundError`).

xapian-core/backends/glass/glass_version.cc
```diff
@@ -103,7 +103,7 @@ GlassVersion::read()
 	if (rare(lseek(fd, offset, SEEK_SET) < 0)) {
 	    string msg = "Failed to rewind file descriptor ";
 	    msg += str(fd);
-	    throw Xapian::DatabaseOpeningError(msg, errno);
+	    throw Xapian::DatabaseNotFoundError(msg, errno);
```
ojwb: I think this case isn't really "not found" - it's reading the version file for a single file database, and the fd was either passed in or opened by `backends/dbfactory.cc`so we know the file exists.  It might not be a valid Xapian database, but that I think either `DatabaseOpeningError` or `DatabaseCorruptError` would be more appropriate for that case.

It's actually hard to trigger this case in normal use since the constructor already checked that fd is seekable.  Probably you'd have to close the file descriptor behind Xapian's back, and then this would fail with `EBADF` ("Bad file descriptor"), but other things might fail first as the same file descriptor is used to read the tables.

xapian-core/backends/glass/glass_version.cc
```diff
@@ -113,7 +113,7 @@ GlassVersion::read()
 	if (rare(fd_in < 0)) {
 	    string msg = filename;
 	    msg += ": Failed to open glass revision file for reading";
-	    throw Xapian::DatabaseOpeningError(msg, errno);
+	    throw Xapian::DatabaseNotFoundError(msg, errno);
```
ojwb: This case on the other hand does seem like "not found" - e.g. you can get it if you try to open a directory which isn't a Xapian DB and specify `Xapian::DB_BACKEND_GLASS` in the flags.

ojwb:
It would be nice to ensure there's test coverage corresponding to each case where `DatabaseNotFoundError` can be thrown.

Essentially, for this change to have full test coverage, changing any `DatabaseNotFoundError` back to `DatabaseOpeningError` ought to cause a testcase to fail which I don't think is currently the case.

ojwb:
@VaibhavKansagara Can you confirm if you're OK with the patch licensing requirements?

https://trac.xapian.org/browser/git/xapian-core/HACKING#L1384

xapian-core/backends/dbfactory.cc
VaibhavKansagara: > I think both these cases should throw `DatabaseNotFoundError` (and Kronuz agrees).
>
I thought you were implying `DatabaseNotFoundError` but as you mentioned that keeping `DatabaseNotFoundError` would require furthur changes. Thank you for suggesting the changes. I will push the changes.

xapian-core/backends/glass/glass_version.cc
VaibhavKansagara: Thanks for detailed explanation. I will push the changes.

xapian-core/backends/glass/glass_version.cc
VaibhavKansagara: > This case on the other hand does seem like "not found" - e.g. you can get it if you try to open a directory which isn't a Xapian DB and specify `Xapian::DB_BACKEND_GLASS` in the flags.
>
@ojwb the value of fd_in could be negative for several reasons other than "not found", so shouldn't it be `DatabaseOpeningError`?

VaibhavKansagara:
> Essentially, for this change to have full test coverage, changing any `DatabaseNotFoundError` back to `DatabaseOpeningError` ought to cause a testcase to fail which I don't think is currently the case.
>
i will fix this shortly.
> Can you confirm if you're OK with the patch licensing requirements?
> 

xapian-core/backends/glass/glass_version.cc
ojwb: Probably "not found" is the most common reason, but there are other possibilities as you say - not having the necessary permissions is also a common one.

Checking `errno` and picking which of the two to throw would be best I guess - `ENOENT` is the obvious value to check, but `ENOTDIR` should probably also give `DatabaseNotFoundError` (you'd get that if a parent directory in the path is actual a file, e.g. trying to open `/home/olly/foo/bar.db` when `/home/olly/foo` is a file, which is very similar to the case of trying to open `/home/olly/foo/bar.db` when `/home/olly/foo` doesn't exist).

VaibhavKansagara:
@ojwb I believe I have sorted out all the bugs.Please merge this when you have time.

xapian-core/backends/dbfactory.cc
VaibhavKansagara: sorry i didn't notice.thanks fixed.

xapian-core/backends/glass/glass_version.cc
```diff
@@ -113,6 +113,9 @@ GlassVersion::read()
 	if (rare(fd_in < 0)) {
 	    string msg = filename;
 	    msg += ": Failed to open glass revision file for reading";
+	    if (errno == ENOENT || errno == ENOTDIR) {
+	    throw Xapian::DatabaseNotFoundError(msg, errno);
```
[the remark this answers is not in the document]
VaibhavKansagara: thanks fixed.

VaibhavKansagara:
@ojwb appveyor build is failing on a specific platform. I observed the same behaviour in #214. Not sure how to fix it.

ojwb:
> appveyor build is failing on a specific platform. I observed the same behaviour in #214. Not sure how to fix it.

There is currently an intermittent failure on mingw which appears to be due to a database we try to delete and recreate still being open.  It's very annoying, but I haven't managed to determine what's going on yet.  The POSIX behaviour is that you can delete an open file so maybe this actually happens on all platforms, but if that's the case I don't really understand why this doesn't ever seem to affect the mingw64 or MSVC builds.  And if it's specific to 32-bit mingw that's odd too, since mingw uses the same runtime C DLL as MSVC.

ojwb:
> There is currently an intermittent failure on mingw which appears to be due to a database we try to delete and recreate still being open. It's very annoying, but I haven't managed to determine what's going on yet.

I've marked this build as `allow_failures` in f6b8683c3823e49ed4b7532c32258dd5aa6c0a45 so it'll still be run, but failure won't cause the CI check to be marked as failing.

ojwb:
Squashed, merged and pushed.

VaibhavKansagara:
@ojwb thanks

PR #214, opened 2019-01-01
Explicit DatabaseClosedError exception(ticket #772)

VaibhavKansagara:
This PR adds a new class DatabaseClosedError. Refer to ticket #772 for more details.
Please review this PR when you have time.

### DIFF
--- a/xapian-core/common/io_utils.cc
+++ b/xapian-core/common/io_utils.cc
@@ -121,9 +121,9 @@ io_open_block_wr(const char * fname, bool anew)
 size_t
 io_read(int fd, char * p, size_t n, size_t min)
 {
-	if (fd == -2) {
-	    throw Xapian::DatabaseClosedError("Database is closed", errno);
-	}
+    if (fd == -2) {
+	    throw Xapian::DatabaseClosedError("Database has been closed", errno);
+    }
     size_t total = 0;
     while (n) {
 	ssize_t c = read(fd, p, n);
@@ -146,9 +146,9 @@ io_read(int fd, char * p, size_t n, size_t min)
 void
 io_write(int fd, const char * p, size_t n)
 {
-	if (fd == -2) {
-	    throw Xapian::DatabaseClosedError("Database is closed", errno);
-	}
+    if (fd == -2) {
+	    throw Xapian::DatabaseClosedError("Database has been closed", errno);
+    }
     while (n) {
 	ssize_t c = write(fd, p, n);
 	if (c < 0) {
@@ -163,9 +163,9 @@ io_write(int fd, const char * p, size_t n)
 size_t
 io_pread(int fd, char * p, size_t n, off_t o, size_t min)
 {
-	if (fd == -2) {
-	    throw Xapian::DatabaseClosedError("Database is closed", errno);
-	}
+    if (fd == -2) {
+	    throw Xapian::DatabaseClosedError("Database has been closed", errno);
+    }
     size_t total = 0;
 #ifdef HAVE_PREAD
     while (true) {
@@ -223,9 +223,9 @@ io_pread(int fd, char * p, size_t n, off_t o, size_t min)
 void
 io_pwrite(int fd, const char * p, size_t n, off_t o)
 {
-	if (fd == -2) {
-	    throw Xapian::DatabaseClosedError("Database is closed", errno);
-	}
+    if (fd == -2) {
+	    throw Xapian::DatabaseClosedError("Database has been closed", errno);
+    }
 #ifdef HAVE_PWRITE
     while (n) {
 	ssize_t c = pwrite(fd, p, n, o);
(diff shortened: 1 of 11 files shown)


VaibhavKansagara:
Travis CI builds is failing with an unusual error.

jaylett:
This is what's going on — it'd be helpful if you could suggest a way in which we either improve the error message, or make it clearer what the tests are expecting (say in our [developer documentation](https://xapian-developer-guide.readthedocs.io/en/latest/)):

```
$ git diff `[ "$TRAVIS_PULL_REQUEST" = false ] && echo master || echo "$TRAVIS_BRANCH"`...|xapian-maintainer-tools/xapian-check-patch
xapian-core/tests/api_closedb.cc:34: error: Line extends beyond column 80 (to column 84): +    try { CODE; } catch (const Xapian::DatabaseClosedError &) { ++closedexc_count; }
xapian-core/tests/api_closedb.cc:241: error: Line extends beyond column 80 (to column 82): +    TEST_EXCEPTION(Xapian::DatabaseClosedError, dbw1.postlist_begin("paragraph"));
```

This is two stylistic failures. What we're doing (via the `xapian-maintainer-tools/xapian-check-patch` script, which you can run yourself before making a pull request, or in this case where it's thrown errors and you want to check everything's fixed but don't want to wait for Travis to build) is to check a number of things we expect to be true, like spacing around operators and braces — or in this case, that lines aren't longer than 80 characters.

So there are two (slightly) long lines, both in `xapian-core/tests/api_closedb.cc`. If you break them over multiple lines, the script (and Travis) will be happy. (The first is a little harder because it's a macro, but the `IF_NOT_CLOSEDEXC` macro after it shows how to do that.)

Not all lines in all files obey all the styles — which is why we don't currently just check the entire repository. So when you change lines, sometimes these errors will appear even if you didn't really cause them, and even when surrounding lines have the same issues.

VaibhavKansagara:
> This is what's going on — it'd be helpful if you could suggest a way in which we either improve the error message, or make it clearer what the tests are expecting (say in our [developer documentation](https://xapian-developer-guide.readthedocs.io/en/latest/)):
>
@jaylett we could add a small fixup message.Should i change the error message?

jaylett:
@VaibhavKansagara what was unclear in the message or presentation? Then you can open a (separate) pull request to improve that. (Or file a bug explaining what you needed to know, if you can't think of a better way of describing it.)

xapian-core/common/io_utils.cc
```diff
@@ -174,7 +174,7 @@ io_pread(int fd, char * p, size_t n, off_t o, size_t min)
 	    // We get EINTR if the syscall was interrupted by a signal.
 	    // In this case we should retry the read.
 	    if (errno == EINTR) continue;
-	    throw Xapian::DatabaseError("Error reading database", errno);
+	    throw Xapian::DatabaseClosedError("Error reading database", errno);
```
ojwb: This is reporting an error trying to read a block from the database, not reporting an attempt to do something to a closed database.

If the database is closed then the file descriptor is set to `-2` to indicate this and that should be checked before we get here.

xapian-core/tests/api_closedb.cc
```diff
@@ -31,14 +31,18 @@
 using namespace std;
 
 #define COUNT_CLOSEDEXC(CODE) \
-    try { CODE; } catch (const Xapian::DatabaseError &) { ++closedexc_count; }
+    try { \
+		CODE; \
+	} catch (const Xapian::DatabaseClosedError &) { \
+		++closedexc_count; \
```
ojwb: This line and `CODE; \` above are overindented - each indent level is generally 4 columns in Xapian code, but here you've used 8.

VaibhavKansagara: Thanks for pointing this out.I will push the changes.

xapian-core/common/io_utils.cc
VaibhavKansagara: thanks fixed.

ojwb: Sorry, I wasn't clear about this - by "before we get here" I really meant "at a higher level" rather than at the start of this function.

The changes I made on master in 2a980274452b642739f799f0c066d9cdfb3048da to fix the dangling closed fd issue should mean that `fd == -2` never happens here, and indeed I can revert all the changes to `io_utils.cc` made on your branch and the tests all still pass.

And should that assumption be violated, the syscalls here will fail with `errno` set to `EBADFD` we should see a testcase failure rather than it getting quietly handled.

Don't worry about changing that - I can easily merge with those changes reverted.

ojwb:
Squashed (with the io_utils.cc changes reverted as mentioned above), merged and pushed.  Thanks!

VaibhavKansagara:
@ojwb thanks

PR #215, opened 2019-01-11
 Test MSet::snippet() with small and zero length

VaibhavKansagara:
This PR is a test for MSet::snippet() when length of the text is zero, or smaller than the length of any word in the text. For more details refer ticket #759.

### DIFF
--- a/xapian-core/tests/api_snippets.cc
+++ b/xapian-core/tests/api_snippets.cc
@@ -450,7 +450,6 @@ DEFINE_TESTCASE(snippet_small_zerolength, backend) {
 	{ "Rubbish and junk", 0, "" },
 	{ "Project R.U.B.B.I.S.H. greenlit", 0, "" },
 	{ "What a load of rubbish", 0, "" },
-	{ "What a load of rubbish", 0, "" },
 	{ "rubbish mention rubbish mention", 0, "" },
     };
 
--- a/xapian-core/tests/api_snippets.cc
+++ b/xapian-core/tests/api_snippets.cc
@@ -430,3 +430,33 @@ DEFINE_TESTCASE(snippet_start_nonspace, backend) {
 
     return true;
 }
+
+/// Test snippets with small and zero length.
+DEFINE_TESTCASE(snippet_small_zerolength, backend) {
+    Xapian::Enquire enquire(get_database("apitest_simpledata"));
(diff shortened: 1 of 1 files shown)


jaylett:
Where you're adding more commits that update errors in an earlier one, you should try to use `git commit --fixup <sha-of-earlier-commit>`. That way they're easy to collapse to a single commit when merging. (Since you'll need at least one more commit to fix the style failure, you can try it out here!)

VaibhavKansagara:
thanks @jaylett  unfortunately i commited before i read your comment but i will keep that in mind when making changes in the future.

xapian-core/tests/api_snippets.cc
```diff
@@ -430,3 +430,33 @@ DEFINE_TESTCASE(snippet_start_nonspace, backend) {
...
+	{ "mention junk rubbish", 3, "" },
+	{ "Project R.U.B.B.I.S.H. greenlit", 5, "" },
+	{ "What load rubbish", 3, "" },
+	{ "Mention rubbish", 4, "" },
+
+	// Test with zero length.
+	{ "Rubbish and junk", 0, "" },
+	{ "Project R.U.B.B.I.S.H. greenlit", 0, "" },
+	{ "What a load of rubbish", 0, "" },
+	{ "What a load of rubbish", 0, "" },
```
ojwb: Looks like there's a repeated entry here...

ojwb:
> VaibhavKansagara force-pushed the VaibhavKansagara:Test_Msnippet branch from 683e8c3 to f69a205 6 days ago

Generally you shouldn't force-push branches which you're collaborating with others on unless everyone involved agrees to it.  Rewriting history makes it confusing to follow the incremental changes, and in the context of review the simplest thing for the reviewer to do is to just restart the review from scratch, which wastes their time, and likely means you'll have to wait longer for a review.

Instead use the approach James suggested above (i.e. `git commit --fixup` with the SHA hash of the commit you want to amend, then push as normal).  Once reviewed, the person merging the changes can then automatically squash the fixup changes onto their respective commits before merging.

xapian-core/tests/api_snippets.cc
VaibhavKansagara: thanks. fixed

ojwb:
Squashed, merged and pushed.  Thanks!

PR #217, opened 2019-01-21
Add Krovetzstemmer for english language.

VaibhavKansagara:
This PR implements KrovetzStemmer for english language. The implementation of this stemmer has been taken from https://github.com/rmit-ir/KrovetzStemmer. **I have added a small testcase demonstrating the working of this stemmer.** I have also tested this stemmer offline. Below is the _diff_ file of large vocabulary of words. If you feel like going with this implementation i will add more rigorous test like those for other stemmers. The above link is not working in that case just copy and paste it in the url.
[diff.txt](https://github.com/xapian/xapian/files/2779345/diff.txt)


xapian-core/include/xapian/stem.h
```diff
@@ -30,7 +30,9 @@
 #include <xapian/intrusive_ptr.h>
 #include <xapian/visibility.h>
 
-#include <string>
+#include <iostream>
```
ojwb: Nothing in the library (and especially nothing in the public API headers) should need to be including `<iostream>`.

xapian-core/include/xapian/stem.h
```diff
@@ -171,6 +173,122 @@ class XAPIAN_VISIBILITY_DEFAULT Stem {
     }
 };
 
+class XAPIAN_VISIBILITY_DEFAULT KrovetzStemmer : public StemImplementation {
```
ojwb: I don't think this should be a public API class - instead put it inside the library and make the existing `Stem` constructor which takes a string handle "krovetz" by setting `internal` to a new `KrovetzStemmer` object.

That provides a consistent API and hides the implementation details away (which means issues which need changes to the implementation to solve can be solved without affecting the library's ABI).

xapian-core/stemmer/KrovetzStemmer.cc
```diff
@@ -0,0 +1,6846 @@
+/*==========================================================================
+ * Copyright (c) 2005 University of Massachusetts.  All Rights Reserved.
+ *
+ * Use of the Lemur Toolkit for Language Modeling and Information Retrieval
+ * is subject to the terms of the software license set forth in the LICENSE
+ * file included with this software, and also available at
+ * http://www.lemurproject.org/license.html
```
ojwb: Unfortunately that licence has a clause restricting naming of derived products which makes it incompatible with the GPL:

>   4. Products derived from this software may not be called "Lemur" or "Indri" nor may "Lemur" or "Indri" appear in their names without prior written permission of The Lemur Project. To obtain permission, contact license@lemurproject.org.

(Based on https://www.gnu.org/licenses/license-list.html#Phorum having the equivalent clause and that making that licence incompatible with the GPL.)

It's very annoying that something so minor makes it incompatible.  There are similar issues with the PHP licence and the GPL, so you can't legally distribute builds of Xapian's PHP bindings and everyone has to build it for themselves instead.

I think that sadly means we can't merge this, though it's still useful for getting your more familiar with the code.

xapian-core/stemmer/KrovetzStemmer.cc
```diff
@@ -0,0 +1,6846 @@
...
+
+    static const char *exceptions[] = {
+    "aide",
+    "bathe", "caste", "cute", "dame", "dime", "doge", "done", "dune",
+    "envelope", "gage", "grille", "grippe", "lobe", "mane", "mare",
+    "nape", "node", "pane", "pate", "plane", "pope", "programme",
+    "quite", "ripe", "rote", "rune", "sage", "severe", "shoppe",
+    "sine", "slime", "snipe", "steppe", "suite", "swinge", "tare",
+    "tine", "tope", "tripe", "twine",
+    NULL
```
ojwb: I realise this isn't code you wrote, but it illustrates a useful point to be aware of when working on a C/C++ library.

Arrays of `const char*` pointers like these are problematic for code that's in a shared library.

In simple terms, the strings themselves can be put into the read-only data section, but the array needs to contain addresses for those strings.  Those addresses will vary depend on the address that the shared library gets loaded at which varies[*].  So that that when a program using this shared library is run, the dynamic loader needs to fix up all the pointers in all of these arrays.  A single relocation is cheap, but en masse they can cause significant slowdowns.

The simplest fix is to make these arrays of fixed-width arrays - e.g this could instead be `static const char exceptions[][10] = ...` (where 10 is one more than the longest string).  The code using the array shouldn't need changing (unless it's doing icky things it really shouldn't be doing).

That means 10 bytes is used for every string (even though e.g. `"done"` which only needs 5, including the terminating zero byte), but this avoids a pointer per entry which is 8 bytes on a 64-bit platform (or 4 on a 32-bit platform), so using arrays would probably actually reduce the memory used a bit in this case.

See 2.4.3 in https://www.akkadia.org/drepper/dsohowto.pdf for more details (the whole document is a good read if you're interested in low-level details of how stuff works).

[*] If it didn't, every shared library would need to be allocated a span of addresses which didn't overlap with any other library - that's just not possible on a 32-bit platform as the total size of all libraries will exceed the address range; a 64-bit platform may have enough address range, but it's hard to come up with a system to uniquely allocate address ranges to libraries built by different people on different computers which may end up getting used together.

VaibhavKansagara: thanks for the detailed explanation and useful information.

PR #219, opened 2019-02-20
Use IEEE-754 interchange format in serialise-double

tanmaysachan:
This pull request modifies the serialise_double and unserialise_double functions to use IEEE format.

https://trac.xapian.org/ticket/673

### DIFF
--- a/xapian-core/common/serialise-double.cc
+++ b/xapian-core/common/serialise-double.cc
@@ -46,7 +46,7 @@ string serialise_double(double v)
 {
 # ifdef WORDS_BIGENDIAN
     uint64_t temp;
-    static_assert(sizeof(temp) == sizeof(v));
+    static_assert(sizeof(temp) == sizeof(v), "Check if size of double and 64 bit int is same");
     memcpy(&temp, &v, sizeof(double));
     temp = do_bswap(temp);
     return string(reinterpret_cast<const char *>(&temp), sizeof(double));
@@ -64,7 +64,7 @@ double unserialise_double(const char ** p, const char * end)
     double result;
 # ifdef WORDS_BIGENDIAN
     uint64_t temp;
-    static_assert(sizeof(temp) == sizeof(double));
+    static_assert(sizeof(temp) == sizeof(double), "Check if size of double and 64 bit int is same");
     memcpy(&temp, *p, sizeof(double));
     temp = do_bswap(temp);
     memcpy(&result, &temp, sizeof(double));
@@ -88,11 +88,10 @@ string serialise_double(double v){
      * and reduce exp by 1, since this is the way doubles
      * are stored in IEEE-754.
      *
-     * Conversion of the fractional part of the new mantissa
-     * to bits is done by repeatedly multiplying it by 2,
-     * checking if it is greater than 1, and turning the
-     * corresponding bit on, if it is.
-     *
+     * conversion of mantissa to bits is done by multiplying
+     * mantissa with 2^26, extracting the integer part and
+     * or-ing it with result, then repeating this step once
+     * more.
      */
     uint64_t result = 0;
 
@@ -104,31 +103,38 @@ string serialise_double(double v){
 
     int exp;
     v = frexp(v, &exp);
-    v *= 2.0;
 
     if (exp == 0 && v == 0.0) {
-	result = 0;
+	result = 0.0;
 	return string(reinterpret_cast<const char *>(&result),
 		      sizeof(uint64_t));
     }
+
+    v *= 2.0;
+    v -= 1.0;
 
     exp += 1022;
 
     result |= (uint64_t)exp << 52;
 
-    // mantissa bit pointer
-    uint64_t mbp = (uint64_t)1 << 51;
+# if FLT_RADIX == 2
+    double scaled_v = scalbn(v, 26);
+# else
+    double scaled_v = ldexp(v, 26);
+# endif
 
-    v -= 1.0;
+    uint64_t scaled_v_int = static_cast<uint64_t>(scaled_v);
+    result |= scaled_v_int << 26;
+    scaled_v -= static_cast<double>(scaled_v_int);
 
-    for (int i = 51; i >= 0; --i) {
-	v *= 2.0;
-	if (v >= 1.0) {
-	    result |= mbp;
-	    v -= 1.0;
-	}
-	mbp >>= 1;
-    }
+# if FLT_RADIX == 2
+    scaled_v = scalbn(scaled_v, 26);
+# else
+    scaled_v = ldexp(scaled_v, 26);
+# endif
+
+    scaled_v_int = static_cast<uint64_t>(scaled_v);
+    result |= scaled_v_int;
 
 # ifdef WORDS_BIGENDIAN
     result = do_bswap(result);
@@ -184,9 +190,9 @@ double unserialise_double(const char ** p, const char * end) {
 # else
     double result = ldexp(mantissa, exp);
 # endif
+
     if (negative) result = -result;
     return result;
 }
 
 #endif
-
--- a/xapian-core/common/serialise-double.cc
+++ b/xapian-core/common/serialise-double.cc
@@ -78,11 +78,11 @@ double unserialise_double(const char ** p, const char * end)
 #else
 
 string serialise_double(double v){
-    /* 
+    /*
      * First bit(msb) -> sign (1 means negative)
      * next 11 bits -> exponent
      * last 52 bits -> mantissa
-     * 
+     *
      * frexp gives fraction within the range [0.5, 1)
      * We multiply it by 2 to change the range to [1.0, 2.0)
      * and reduce exp by 1, since this is the way doubles
@@ -92,7 +92,7 @@ string serialise_double(double v){
      * to bits is done by repeatedly multiplying it by 2,
      * checking if it is greater than 1, and turning the
      * corresponding bit on, if it is.
-     * 
+     *
      */
     uint64_t result = 0;
 
@@ -108,7 +108,8 @@ string serialise_double(double v){
 
     if (exp == 0 && v == 0.0) {
 	result = 0;
-	return string(reinterpret_cast<const char *>(&result), sizeof(uint64_t));
+	return string(reinterpret_cast<const char *>(&result), 
+		      sizeof(uint64_t));
     }
 
     exp += 1022;
@@ -152,7 +153,7 @@ double unserialise_double(const char ** p, const char * end) {
     exp |= (second & (15 << 4)) >> 4;
     exp -= 1023;
 
-    double mantissa = 1.0; //implicit 1 in IEEE
+    double mantissa = 1.0; // implicit 1 in IEEE
     double pows2 = 0.5; // powers of 2
 
     int second_byte_counter = 1 << 3;
@@ -168,7 +169,7 @@ double unserialise_double(const char ** p, const char * end) {
 	--byte_counter;
 	int cur_byte_counter = 1 << 7;
 	for (int i = 0; i < 8; ++i) {
-	    if(cur_byte & cur_byte_counter) mantissa += pows2;
+	    if (cur_byte & cur_byte_counter) mantissa += pows2;
 	    pows2 /= 2.0;
 	    cur_byte_counter >>= 1;
 	}
--- a/xapian-core/common/serialise-double.cc
+++ b/xapian-core/common/serialise-double.cc
@@ -153,27 +153,17 @@ double unserialise_double(const char ** p, const char * end) {
     exp |= (second & (15 << 4)) >> 4;
     exp -= 1023;
 
-    double mantissa = 1.0; // implicit 1 in IEEE
-    double pows2 = 0.5; // powers of 2
+    uint64_t mantissa_bp; // variable to store bit pattern of mantissa;
+    memcpy(&mantissa_bp, *p, sizeof(double));
+    mantissa_bp &= (uint64_t(1) << 52) - 1;
 
-    int second_byte_counter = 1 << 3;
-    for (int i = 0; i < 4; ++i) {
-	if (second & second_byte_counter) mantissa += pows2;
-	pows2 /= 2.0;
-	second_byte_counter >>= 1;
-    }
+# if FLT_RADIX == 2
+    double mantissa = scalbn(mantissa_bp, -52);
+# else
+    double mantissa = ldexp(mantissa_bp, -52);
+# endif
 
-    int byte_counter = 5; // starting with the third byte onwards
-    while (byte_counter >= 0) {
-	unsigned char cur_byte = *(*p + byte_counter);
-	--byte_counter;
-	int cur_byte_counter = 1 << 7;
-	for (int i = 0; i < 8; ++i) {
-	    if (cur_byte & cur_byte_counter) mantissa += pows2;
-	    pows2 /= 2.0;
-	    cur_byte_counter >>= 1;
-	}
-    }
+    mantissa += 1.0;
 
     *p += 8;
 
@@ -184,6 +174,7 @@ double unserialise_double(const char ** p, const char * end) {
 # else
     double result = ldexp(mantissa, exp);
 # endif
+
     if (negative) result = -result;
     return result;
 }
--- a/xapian-core/common/serialise-double.cc
+++ b/xapian-core/common/serialise-double.cc
@@ -181,7 +181,7 @@ double unserialise_double(const char ** p, const char * end) {
 # if FLT_RADIX == 2
     double result = scalbn(mantissa, exp);
 # else
-    double result = ldexp(mantissa, exp >> 2);
+    double result = ldexp(mantissa, exp);
 # endif
     if (negative) result = -result;
     return result;
--- a/xapian-core/common/serialise-double.cc
+++ b/xapian-core/common/serialise-double.cc
@@ -157,22 +157,16 @@ double unserialise_double(const char ** p, const char * end) {
     memcpy(&mantissa_bp, *p, sizeof(double));
     mantissa_bp &= (uint64_t(1) << 52) - 1;
 
-# if FLT_RADIX == 2
-    double mantissa = scalbn(mantissa_bp, -52);
-# else
-    double mantissa = ldexp(mantissa_bp, -52);
-# endif
-
-    mantissa += 1.0;
-
     *p += 8;
 
-    if (exp + 1023 == 0 && mantissa == 1.0) return 0.0;
+    if (exp + 1023 == 0 && mantissa_bp == 0) return 0.0;
 
 # if FLT_RADIX == 2
-    double result = scalbn(mantissa, exp);
+    double result = scalbn(mantissa_bp, -52);
+    result = scalbn(result + 1.0, exp);
 # else
-    double result = ldexp(mantissa, exp);
+    double result = ldexp(mantissa_bp, -52);
+    result = ldexp(result + 1.0, exp);
 # endif
 
     if (negative) result = -result;
--- a/xapian-core/common/serialise-double.cc
+++ b/xapian-core/common/serialise-double.cc
@@ -30,6 +30,8 @@
 
 #include "serialise-double.h"
 
+#include "wordaccess.h"
+
 #include <cfloat>
 #include <cmath>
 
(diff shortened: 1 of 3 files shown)


xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,51 @@
...
-	    result += negative ? char(0x8e) : char(0x0e);
-	    result += char(exp + 128);
-	} else {
-	    if (exp < -32768 || exp > 32767) {
+	int exp;
+	frexp(v, &exp);
+	--exp;
+	exp >>= 3;
+	if (exp < -32768 || exp > 32767)
+	{
```
ojwb: Why have you kept this code?

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,51 @@
...
-    if (n > 1) {
-	Assert(n <= 8);
-	result[0] = static_cast<unsigned char>(result[0] | ((n - 1) << 4));
-    }
-
-    return result;
+	int check_if_big_endian = 1;
+	if (*(char *) &check_if_big_endian != 1)
+	{
+		// platform is big-endian
```
ojwb: We really don't want to have to check this at compile-time every time this function is called.

And there's no need to - the endianness is already available as a preprocessor check:

```
#ifdef WORDS_BIGENDIAN
```

xapian-core/common/serialise-double.cc
tanmaysachan: oh sorry, I was supposed to remove it. Will correct it.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,51 @@
...
+		uint64_t * temp_cast_pointer =
+			reinterpret_cast<uint64_t *>(&v);
+		*temp_cast_pointer = do_bswap(*temp_cast_pointer);
+		v = *reinterpret_cast<double *>(temp_cast_pointer);
```
ojwb: You quote often see code performing this sort of type punning, but it violates the C/C++ strict aliasing rules, which means undefined behaviour - e.g. see:

https://blog.qt.io/blog/2011/06/10/type-punning-and-strict-aliasing/

In many cases it "works", but it's really just a timebomb in the code waiting for a smarter compiler which optimises harder and takes advantage of the situation.

You're allowed to access the bytes which make up a type via a `char*` pointer, so the safe way to do what you want is to use `memcpy()`:

```
uint64_t temp;
static_assert(sizeof(temp) == sizeof(v), "Types must be the same size");
memcpy(&temp, &v, sizeof(temp));
temp = do_bswap(temp);
return string(reinterpret_cast<const char*>(&temp), sizeof(temp));
```

Modern compilers are smart enough to spot what's really going on here, so you should get machine code which efficiently transfers the bits making up a double from a floating point register to an integer register - e.g. for GCC on x86-64: https://godbolt.org/z/oxxCVC

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,51 @@
...
-	unsigned char b = static_cast<unsigned char>(exp + 7);
-	if (negative) b |= static_cast<unsigned char>(0x80);
-	result += char(b);
-    } else {
-	if (exp >= -128 && exp < 127) {
-	    result += negative ? char(0x8e) : char(0x0e);
-	    result += char(exp + 128);
-	} else {
-	    if (exp < -32768 || exp > 32767) {
+	int exp;
```
ojwb: You indentation doesn't match Xapian conventions (the `xapian-check-patch` script doesn't currently detect this case unfortunately) - probably you have your editor set wrongly.  In most cases, each indentation level should be 4 columns, you should have a tab character set to move to the next multiple of 8 columns, and indents should be "tab-filled" (so a single indent is 4 spaces; double indent a tab, and in general the indent is zero or more tabs then 0 or 4 spaces).

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,51 @@
...
+		uint64_t * temp_cast_pointer =
+			reinterpret_cast<uint64_t *>(&v);
+		*temp_cast_pointer = do_bswap(*temp_cast_pointer);
+		v = *reinterpret_cast<double *>(temp_cast_pointer);
+	}
+	size_t double_size = sizeof(double);
+	char result_temp[double_size];
+	memcpy(result_temp, &v, double_size);
+	string result(result_temp, double_size);
+	return result;
```
ojwb: There's no need to `memcpy()` in the little-endian case - just return a string from a cast pointer to the double.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,51 @@
...
-	if (exp == 15) {
-	    if (*p == end) {
-		throw Xapian::SerialisationError("Bad encoded double: short large exponent");
-	    }
-	    exp = bigexp | (static_cast<unsigned char>(*(*p)++) << 8);
-	    exp -= 32768;
-	} else {
-	    exp = bigexp - 128;
+	if (end - *p < 8)
+	{
```
ojwb: The `{` should be on the same line as the `if`.

xapian-core/common/wordaccess.h
```diff
@@ -27,6 +27,7 @@
 
 #include <cstdint>
 #include <type_traits>
+#include <cstring>
```
ojwb: This file does seem to use `memcpy()` without including the relevant header, but better not to make that unrelated fix in the same commit as changing the double serialisation.

tanmaysachan: Separated into 2 commits.

xapian-core/common/serialise-double.cc
tanmaysachan: I'm still not quite sure about the indentation, but I think the latest commit does follow what you say.

xapian-core/common/serialise-double.cc
tanmaysachan: Updated it accordingly, should be okay now. Thanks for the help!

tanmaysachan:
@ojwb I don't get why appveyor is failing

ojwb:
> tanmaysachan force-pushed the tanmaysachan:ticket673 branch 2 times, most recently from 8a6d0c8 to 611759a 6 days ago

> tanmaysachan force-pushed the tanmaysachan:ticket673 branch from 2b52ff9 to 439c8a0 6 days ago

In general it's really unhelpful to keep force-pushing a branch you've submitted for review (at least this is a small and fairly constrained change but it's still annoying).  The reviewer can't just look at the newer commits to see what you've changed in response to feedback - the simplest thing for them to do is just review the whole set of changes from scratch, which is more effort and likely to put them off.  Or they could try to work out which commit they last reviewed up to before history was rewritten, then generate a diff from there to the current branch head, review that, and then find the corresponding code in the github UI to add the review comments.  That's still extra work you've created for them which they wouldn't have had to do otherwise.

A good rule of thumb is don't force push a branch unless you're sure everyone collaborating with you on it is expecting it.  For example, sometimes people push work-in-progress branches mostly so there's a remote backup, but also so anyone interested can take a look, and they warn the branch "may be force-pushed without warning".  Or a reviewer might ask you to.

But in general it's better to make fixup commits [as described in our developer guide](https://xapian-developer-guide.readthedocs.io/en/latest/contributing/contributing-changes.htmlopen-a-pull-request-on-github).  These can then usually be squashed onto the commits they fixup automatically before merging.

xapian-core/common/serialise-double.cc
ojwb: Please don't mark review comments as resolved - that makes it much harder for the reviewer to check you've actually suitably addressed them.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,36 @@
...
+    uint64_t temp;
+    static_assert(sizeof(temp) == sizeof(v));
+    memcpy(&temp, &v, double_size);
+    temp = do_bswap(temp);
+    return string(reinterpret_cast<const char *>(&temp), double_size);
 #endif
-    // v is now in the range [1.0, 256.0)
-    exp >>= 3;
-    return exp;
+    return string(reinterpret_cast<const char *>(&v), double_size);
```
ojwb: This really want to be in a `#else`.

Otherwise in the `WORDS_BIGENDIAN` case this line is dead code.  The compiler can spot that and not generate machine code for it, but it may also generate a warning about it.  Currently GCC doesn't, but e.g. clang does with `-Wunreachable-code-return`: "warning: 'return' will never be executed"

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,36 @@
...
-    v = scalbn(v, (exp & 7) + 1);
-#else
-    v = ldexp(v, (exp & 7) + 1);
+string serialise_double(double v)
+{
+    size_t double_size = sizeof(double);
+#ifdef WORDS_BIGENDIAN
+    uint64_t temp;
+    static_assert(sizeof(temp) == sizeof(v));
+    memcpy(&temp, &v, double_size);
```
ojwb: I think it's better to eliminate `double_size` - as a reviewer (and indeed as someone trying to follow the code) what I want to be able to see is that the right amount of data is being copied.  And while `double_size` suggests it holds the size of the double, I've seen enough misleadingly named variables in my time that I would want to check the definition - it's all-to-easy for someone to update the code but miss a place that should be changed.

In this version the size is clear from the line itself:

```
    memcpy(&temp, &v, sizeof(double));
```

I's also easier for the compiler to see the size - consider if the target type is too small:

```
#include <string.h>
#include <stdint.h>

uint32_t f(const char*p) {
    uint32_t r;
    memcpy(&r, p, sizeof(double));
    return r;
}

uint32_t g(const char*p) {
    const int double_size = sizeof(double);
    uint32_t r;
    memcpy(&r, p, double_size);
    return r;
}
```

Compiling that we get a warning when `sizeof(double)` is inlined but not if it comes through a variable (not even if I make it `const`):

```
$ gcc -Wall -W -c badcopy.c
badcopy.c: In function ‘f’:
badcopy.c:6:5: warning: ‘memcpy’ forming offset [5, 8] is out of the bounds [0, 4] of object ‘r’ with type ‘uint32_t’ {aka ‘unsigned int’} [-Warray-bounds]
     memcpy(&r, p, sizeof(double));
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
badcopy.c:5:14: note: ‘r’ declared here
     uint32_t r;
              ^
```

That's a potential stack smashing vulnerability.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,36 @@
...
-	v *= 256.0;
-    } while (v != 0.0 && --maxbytes);
-
-    n = result.size() - n;
-    if (n > 1) {
-	Assert(n <= 8);
-	result[0] = static_cast<unsigned char>(result[0] | ((n - 1) << 4));
+    if (end - *p < 8) {
+		throw Xapian::SerialisationError(
+				"Bad encoded double: insufficient data");
```
ojwb: This block is over-indented.

xapian-core/common/serialise-double.cc
ojwb: There are still some problems, but I'll flag them individually.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,36 @@
...
     }
-
+    size_t double_size = sizeof(double);
+    double result;
+    memcpy(&result, *p, double_size);
```
ojwb: The above line is OK (`double_size` concerns aside) for the little-endian case, but in the big-endian case you copy the data from p into a `double` here, then below copy it from there into a `uint64_t`, then byte swap that and, then copy back to the `double`.

That first copy is an extra copy you don't need.  Maybe the compiler will be able to optimise it away, but if you eliminate it then you don't need to worry about whether it will.

And that first copy to a `double` seems icky as the value being copied is not really a bit pattern for a `double` as such - we know reversing the bytes gives a `double` value.  In practice that probably won't cause problems, but it's conceivable it could be some sort of trapping value and trigger a floating point exception if the CPU loads it.  Again if we avoid this copy, that's something we can avoid having to worry about.

ojwb:
@tanmaysachan One bit of admin - can you confirm if you're OK with the patch licensing requirements?

https://trac.xapian.org/browser/git/xapian-core/HACKING#L1392

xapian-core/common/serialise-double.cc
tanmaysachan: Ah yes, hadn't thought about the second possibility. Will fix it right away.

tanmaysachan:
> @tanmaysachan One bit of admin - can you confirm if you're OK with the patch licensing requirements?
> 
> https://trac.xapian.org/browser/git/xapian-core/HACKING#L1392

@ojwb Yes, I'm okay with the patch licensing requirements,

xapian-core/common/serialise-double.cc
tanmaysachan: Fixed it. Thanks for the insight!

xapian-core/common/serialise-double.cc
tanmaysachan: Fixed in the latest commit.

xapian-core/common/serialise-double.cc
tanmaysachan: Fixed in the latest commit.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,153 @@
...
-    v = ldexp(v, (exp & 7) + 1);
-#endif
-    // v is now in the range [1.0, 256.0)
-    exp >>= 3;
-    return exp;
+string serialise_double(double v)
+{
+# ifdef WORDS_BIGENDIAN
+    uint64_t temp;
+    static_assert(sizeof(temp) == sizeof(v));
```
ojwb: Omitting the second argument to `static_assert()` requires C++17, and currently we target C++11.  So can you add a description of what's being asserted here?

xapian-core/configure.ac
```diff
@@ -801,6 +801,7 @@ using namespace std;
     [ac_cv_func_snprintf=unknown;break]
   )
 done
+
```
ojwb: Please avoid random whitespace changes in patches - they make review a little harder, and add noise for people trying to trace back through the history of the code.

xapian-core/configure.ac
```diff
@@ -1216,6 +1217,31 @@ if test yes = "$enable_log"; then
...
+  AC_LANG_PROGRAM(
+    [[
+#include <cstdint>
+#include <cstring>
+using namespace std;
+    ]],
+    [[
+      double a = 2353.3523423;
+      uint64_t bit_pattern =
+      0b0100000010100010011000101011010001100110001101011011111011111111;
```
ojwb: Binary constants require C++14, we target C++11.  Use a hex constant instead.

xapian-core/configure.ac
```diff
@@ -1216,6 +1217,31 @@ if test yes = "$enable_log"; then
...
+      if (memcmp(&bit_pattern, &a, sizeof(double)) == 0) return 0;
+      return 1;
```
ojwb: Won't this test always fail on a big endian platform?

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,153 @@
...
+    double result = scalbn(mantissa, exp);
+# else
+    double result = ldexp(mantissa, exp);
+# endif
+    if (negative) result = -result;
+    return result;
 }
+
+#endif
+
```
ojwb: Avoid blank lines at the end of files.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,153 @@
...
+    v -= 1.0;
+
+    for (int i = 51; i >= 0; --i) {
+	v *= 2.0;
+	if (v >= 1.0) {
+	    result |= mbp;
+	    v -= 1.0;
+	}
+	mbp >>= 1;
```
ojwb: Working one bit at a time seems rather inefficient.

Can't we at least work byte-by-byte?  Multiply by 256.0, get the byte as the integer truncation of the double, then subtract the byte value from the double?  That's what the existing `serialise_double()` code does.  Here the first byte is only a partial byte (3 bits) but that's not hard to handle.

We could probably work in 4-byte chunks (multiply by 4294967296.0, get the uint32_t as the integer truncation of the double, subtract that from the double.  I think the existing code doesn't do that because it produces a variable length encoding and wants to stop when the remaining double is 0.0, but here we produce a fixed size result.

It might even be possible to just multiply by 2251799813685248.0 and cast to a uint64_t to get all the bits in one go.  We need to make sure the intermediate result is always representable as a double (remembering that double likely doesn't have IEEE properties here).

xapian-core/configure.ac
ojwb: Ah no, it works so long as the integers and `double` have the same endianness.

In theory the two could differ I think, but I don't know of any platform where they actually do so slow handling for that case isn't a big deal.

xapian-core/configure.ac
tanmaysachan: I pushed hurriedly, I apologize. Fixed.

xapian-core/common/serialise-double.cc
tanmaysachan: I couldn't really figure out a reason to prefer 32 bits over 26, So I modified it to use 26 bit chunks. Please correct me if I'm wrong.  
Multiplying by 2^51 does not seem like a safe option, so I left it.

tanmaysachan: Changed the function to serialise in one go.

xapian-core/common/serialise-double.cc
```diff
@@ -90,13 +90,23 @@ string serialise_double(double v){
...
-     * or-ing it with result, then repeating this step once
-     * more.
+     * Conversion of mantissa to bits is done by
+     * multiplying the mantissa with 2^52, converting
+     * it to a 64 bit integer representation of the original
+     * double.
      */
+
+    static_assert((uint64_t)1 << 52 <= numeric_limits<double>::max(),
+		  "Check if 2^52 can be represented by a double");
```
ojwb: When you compare an integer to a `double`, first the integer gets converted to a `double`, then the two `double` values are compared (see "Conversions" section of https://en.cppreference.com/w/cpp/language/operator_arithmetic).  But we're trying to test if this integer is representable as a `double`, so we're relying on what happens if that conversion overflows.  If we'll definitely get an "overflow" or "infinity" value which doesn't compare <= the max value, all is well, but I'm struggling to work out if we can assume that's the case.

We could perhaps test based on the max exponent supported, but that's potentially complicated by what the radix is.  We currently hard require `FLT_RADIX` to be 2 in `xapian-core/api/sortable-serialise.cc`, but that requirement could be relaxed (2 seems to be ubiquitous for `double` on modern platforms, but 16 and 10 have been used on older machines).

So perhaps the best answer is to relax this check to `<` - in practice equality won't be achieved when we want the test to succeed, and by testing `<` we don't have to worry about that the conversion to `double` might give us `max()` as the overflow value if there's no infinity.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,159 @@
...
+	result = 0.0;
+	return string(reinterpret_cast<const char *>(&result),
+		      sizeof(uint64_t));
     }
 
-    int maxbytes = min(MAX_MANTISSA_BYTES, 8);
+    bool negative = (v < 0.0);
+    if (negative) {
+	v = -v;
+	result |= (uint64_t)1 << 63;
```
ojwb: The coding guidelines in `HACKING` say to avoid C-style casts:

```
* Use C++ style casts (static_cast<>, reinterpret_cast<>, and const_cast<>)  
  or constructor-syntax (e.g. ``double(value)``) in preference to C style
  casts.  The syntax of the C++ casts is ugly, but they do make the
  intent much clearer which is definitely a good thing, and they avoid issues
  such as casting away const when you only meant to cast the type of a pointer.
```

So here `uint64_t(1)` (or `static_cast<uint64_t>(1)` is preferrable (and similarly in a couple of other places).

We'd like to be able to enable the GCC warning `-Wold-style-cast` which would issue a warning for such cases:

> -Wold-style-cast (C++ and Objective-C++ only)
>
>    Warn if an old-style (C-style) cast to a non-void type is used within a C++ program. The new-style casts (dynamic_cast, static_cast, reinterpret_cast, and const_cast) are less vulnerable to unintended effects and much easier to search for.

https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html

However, currently we've had to disable this warning due to it triggering for macros in system header files on some platforms, but at some point hopefully we can turn it on again.

xapian-core/common/serialise-double.cc
```diff
@@ -104,31 +103,38 @@ string serialise_double(double v){
 
     int exp;
     v = frexp(v, &exp);
-    v *= 2.0;
 
     if (exp == 0 && v == 0.0) {
-	result = 0;
+	result = 0.0;
```
ojwb: Why this change?  `result` is a `uint64_t` so `0` is a clearer zero value for it than `0.0`.

tanmaysachan: Sorry for the unwanted change. In the latest commit, I removed this condition altogether because I was accidentally performing this check twice. I now kept it to check at the start if V equals 0.0, and return 0 if it does.

xapian-core/common/serialise-double.cc
tanmaysachan: Fixed accordingly.

xapian-core/common/serialise-double.cc
tanmaysachan: Fixed accordingly.

xapian-core/common/serialise-double.cc
ojwb: Oh yes, this can only happen when v is 0.  Well spotted.

xapian-core/common/serialise-double.cc
```diff
@@ -38,175 +40,152 @@
...
+    while (byte_counter >= 0) {
+	unsigned char cur_byte = *(*p + byte_counter);
+	--byte_counter;
+	int cur_byte_counter = 1 << 7;
+	for (int i = 0; i < 8; ++i) {
+	    if (cur_byte & cur_byte_counter) mantissa += pows2;
+	    pows2 /= 2.0;
+	    cur_byte_counter >>= 1;
 	}
-#endif
```
ojwb: Again this is processing one bit at a time, which seems complex and inefficient.

Can we unserialise by doing the reverse the steps we use to serialise?

tanmaysachan: Fixed it in the latest commit.

ojwb:
Thanks, merged.

Your "FIXUP" commits aren't valid fixup commits that can be automatically squashed so I've had to leave them as-is.  See the documentation for `git rebase --autosquash` and `git commit --fixup` for how to make proper fixup commits.

PR #220, opened 2019-03-08
A partial patch for better tokenisation of mixed Chinese numbers

outdream:
It's a partial patch try to implement better tokenisation of mixed Chinese numbers, 
as the whole mixed Chinese numbers would be treated as a single token.
Partly fix the problem of 'mixed CJK numbers' mentioned in https://trac.xapian.org/ticket/699
And still have some commented code for unsure test cases.


xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -95,6 +95,47 @@ check_infix(unsigned ch)
     return 0;
 }
 
+/// check if ch is a Chinese character about digits
+static inline bool
+is_Chinese_digit(unsigned ch)
```
ojwb: Xapian convention is to not use upper case words in function names (so this should be `is_chinese_digit`).

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -95,6 +95,47 @@ check_infix(unsigned ch)
...
+{
+    // Below are Chinese characters reprensent digits
```
ojwb: Typo: reprensent

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -229,9 +270,23 @@ parse_terms(Utf8Iterator itor, unsigned cjk_flags, bool with_positions,
 	    do {
 		Unicode::append_utf8(term, ch);
 		prevch = ch;
-		if (++itor == Utf8Iterator() ||
-		    (cjk_flags && CJK::codepoint_is_cjk(*itor)))
+		if (++itor == Utf8Iterator())
 		    goto endofterm;
+		else if (cjk_flags) {
+		    // Only deal mixed Chinese numbers which start
+		    // with an Arabic digit and next is a Chinese digit.
```
ojwb: I think saying "Arabic digit" is confusing if you mean "0" to "9" (as I think you do here) rather than digits as used in Arabic, i.e.:

```
0660;ARABIC-INDIC DIGIT ZERO;Nd;0;AN;;0;0;0;N;;;;;
0661;ARABIC-INDIC DIGIT ONE;Nd;0;AN;;1;1;1;N;;;;;
0662;ARABIC-INDIC DIGIT TWO;Nd;0;AN;;2;2;2;N;;;;;
0663;ARABIC-INDIC DIGIT THREE;Nd;0;AN;;3;3;3;N;;;;;
0664;ARABIC-INDIC DIGIT FOUR;Nd;0;AN;;4;4;4;N;;;;;
0665;ARABIC-INDIC DIGIT FIVE;Nd;0;AN;;5;5;5;N;;;;;
0666;ARABIC-INDIC DIGIT SIX;Nd;0;AN;;6;6;6;N;;;;;
0667;ARABIC-INDIC DIGIT SEVEN;Nd;0;AN;;7;7;7;N;;;;;
0668;ARABIC-INDIC DIGIT EIGHT;Nd;0;AN;;8;8;8;N;;;;;
0669;ARABIC-INDIC DIGIT NINE;Nd;0;AN;;9;9;9;N;;;;;
06F0;EXTENDED ARABIC-INDIC DIGIT ZERO;Nd;0;EN;;0;0;0;N;EASTERN ARABIC-INDIC DIGIT ZERO;;;;
06F1;EXTENDED ARABIC-INDIC DIGIT ONE;Nd;0;EN;;1;1;1;N;EASTERN ARABIC-INDIC DIGIT ONE;;;;
06F2;EXTENDED ARABIC-INDIC DIGIT TWO;Nd;0;EN;;2;2;2;N;EASTERN ARABIC-INDIC DIGIT TWO;;;;
06F3;EXTENDED ARABIC-INDIC DIGIT THREE;Nd;0;EN;;3;3;3;N;EASTERN ARABIC-INDIC DIGIT THREE;;;;
06F4;EXTENDED ARABIC-INDIC DIGIT FOUR;Nd;0;EN;;4;4;4;N;EASTERN ARABIC-INDIC DIGIT FOUR;;;;
06F5;EXTENDED ARABIC-INDIC DIGIT FIVE;Nd;0;EN;;5;5;5;N;EASTERN ARABIC-INDIC DIGIT FIVE;;;;
06F6;EXTENDED ARABIC-INDIC DIGIT SIX;Nd;0;EN;;6;6;6;N;EASTERN ARABIC-INDIC DIGIT SIX;;;;
06F7;EXTENDED ARABIC-INDIC DIGIT SEVEN;Nd;0;EN;;7;7;7;N;EASTERN ARABIC-INDIC DIGIT SEVEN;;;;
06F8;EXTENDED ARABIC-INDIC DIGIT EIGHT;Nd;0;EN;;8;8;8;N;EASTERN ARABIC-INDIC DIGIT EIGHT;;;;
06F9;EXTENDED ARABIC-INDIC DIGIT NINE;Nd;0;EN;;9;9;9;N;EASTERN ARABIC-INDIC DIGIT NINE;;;;
```

I'm not sure there's an ideal term for "0" to "9" in this context.  [Wikipedia says](https://en.wikipedia.org/wiki/Arabic_numerals) _"Alternative names are Western Arabic numerals, Western numerals, Hindu–Arabic numerals, and Unicode calls them digits."_, but all of those seem ambiguous except "Western numerals", and that's a term I don't think I've heard used before.  "ASCII digits" is another option is we're only talking about ASCII `0` to `9` and not other forms (e.g. superscripts like `²`).

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -229,9 +270,23 @@ parse_terms(Utf8Iterator itor, unsigned cjk_flags, bool with_positions,
 	    do {
 		Unicode::append_utf8(term, ch);
 		prevch = ch;
-		if (++itor == Utf8Iterator() ||
-		    (cjk_flags && CJK::codepoint_is_cjk(*itor)))
+		if (++itor == Utf8Iterator())
 		    goto endofterm;
+		else if (cjk_flags) {
+		    // Only deal mixed Chinese numbers which start
```
ojwb: Stylistically, if you have braces `{`...`}` around one branch of an `if`, it's preferable to add them to all the branches.

However, here the `if` part transfers the flow of control away, so there's no need for the `else` anyway (and the Xapian style guide suggests omitting it in such cases).

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -229,9 +270,23 @@ parse_terms(Utf8Iterator itor, unsigned cjk_flags, bool with_positions,
...
+		    // with an Arabic digit and next is a Chinese digit.
+		    if (is_digit(prevch) && is_Chinese_digit(*itor)) {
```
ojwb: `is_digit()` checks for anything which is in Unicode category `DECIMAL_DIGIT_NUMBER`, which includes Chinese digits, etc - you can see the full list with:

```
grep ';Nd;' xapian-core/unicode/UnicodeData.txt
```

If you only want the ASCII digits, then the appropriate test is `C_isdigit()` (which works like the standard `isdigit()` function, except it is hardwired to the POSIX C locale, whereas `isdigit()` may match non-ASCII digit characters if the program calls `setlocale()`).  However, that takes an `unsigned char` so you need to check the value is in range for that, like how `U_isupper()` is defined:

```
static inline bool  
U_isupper(unsigned ch)
{
    return (ch < 128 && C_isupper(static_cast<unsigned char>(ch)));
}
```

If you are meaning to accept any digit character, then the comment above needs adjusting to reflect that.

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -229,9 +270,23 @@ parse_terms(Utf8Iterator itor, unsigned cjk_flags, bool with_positions,
...
+			do {
+			    ch = *itor;
+			    Unicode::append_utf8(term, ch);
+			} while (++itor != Utf8Iterator() &&
+			(is_digit(*itor) || is_Chinese_digit(*itor)));
```
ojwb: This line is under-indented - the first `(` should line up under the first `+` of the line above.

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -229,9 +270,23 @@ parse_terms(Utf8Iterator itor, unsigned cjk_flags, bool with_positions,
...
+			goto endofterm;
+		    }
+		    else if (CJK::codepoint_is_cjk(*itor)) {
```
ojwb: Xapian style is to put the `else` on the same line as the `}`, except that again the `else` isn't necessary here.

xapian-core/tests/api_termgen.cc
```diff
@@ -140,6 +140,11 @@ static const test test_simple[] = {
     { "", "극지라", "극[1] 극지:1 라[3] 지[2] 지라:1" },
     { "", "ウルス アップ", "ア[4] アッ:1 ウ[1] ウル:1 ス[3] ッ[5] ップ:1 プ[6] ル[2] ルス:1" },
 
+    // Test parsing mixed CJK numbers (only with Chinese now)
+    { "", "2千3百", "2千3百[1]"},
+    { "", "有2千三百", "2千三百[2] 有[1]"},
+    { "", "我有2千3百零一块钱", "2千3百零一[3] 块[4] 块钱:1 我[1] 我有:1 有[2] 钱[5]"},
+
```
ojwb: These are good tests for the cases you want the patch to handle.

But we should also ensure that it doesn't handle cases which it shouldn't.  For example, if you intend this patch to only affect the ASCII digits, then there should be tests that to check what happens to cases like the above but with `'2'` and/or `'3'` replaced by Chinese digits, Arabic-Indic digits, etc.

ojwb:
Aside from the issues noted above (which are mostly stylistic), this seems very promising.

The `QueryParser` code needs updating similarly (as you're probably aware, since you marked this as a "partial patch").

I wonder if we should also handle "pure" Chinese numbers as a single token in the same code?  It seems like splitting them into n-grams (as we currently do) will result in a lot of false matches without adding useful extra matches.  E,g, a Chinese document talking about 1919 and 2011 will match a query for 2019.

ojwb:
Also, a bit of admin - are you happy to license your contributions as described here?

https://trac.xapian.org/browser/git/xapian-core/HACKING#L1403

xapian-core/queryparser/termgenerator_internal.cc
outdream: Oh, my fault, "Arabic digit" here means "0" to "9".  
I think "ASCII digits" may be a good option, 
I will replace "Arabic digit" with it in the comment.

xapian-core/queryparser/termgenerator_internal.cc
outdream: Sorry for I misunderstood the function `is_digit()`,
what I mean is to accept only ASCII digits.
I will fix it following your suggestion.

outdream:
Thanks for your code review. 

I will fix the troubles you have mentioned first, 
then try to add support for pure Chinese digits. 

I'm glad to license my contributions.

xapian-core/queryparser/termgenerator_internal.cc
ojwb: To be clear, they are often referred to as "Arabic" as they derive from Arabic - it's just the term is ambiguous in this context.

xapian-core/queryparser/termgenerator_internal.cc
```diff
@@ -95,6 +95,53 @@ check_infix(unsigned ch)
     return 0;
 }
 
+static inline bool
+is_ascii_digit(unsigned ch)
+{
+    return (ch < 128 && C_isdigit(static_cast<unsigned char>(ch)));
```
ojwb: While the outer `(` ... `)` are OK here, they aren't required by the language and result in a stacked `)))` at the end, which makes mentally matching parentheses a bit harder.

As a rule of thumb, parentheses are probably better only used when they are required by the language, when you want to control the relative order of evaluation (e.g. `a * b + c` vs `a * (b + c)` or even `a + b + c` vs `a + (b + c)`), or where they are likely to improve clarity for someone reading the code - e.g. it's probably clearer to write `a || (b && c)` (and compilers tend to warn for `a || b && c` because a lot of people have to think about the precedence order).

outdream: Thanks, I'll fix it.

ojwb:
Thanks for the updates, and sorry for not getting back to this for a while.

This looks complete to me for handling these mixed numbers (which is what the ticket was about), but I'm not sure if we want to merge as-is, or try to come up with a broader solution.

I've been trying to get a better understanding of numbers in Chinese by reading https://en.wikipedia.org/wiki/Chinese_numerals#Whole_numbers and reviewing the IRC discussion we had.

The examples you gave of the digit and power-of-ten characters in contexts where they have other meanings (such as 九九归一) seem to be patterns which aren't valid as numbers, except for single-character numbers (which we can probably leave to ICU's word boundary finding code to discriminate).

The valid pattern for a number looks possible to express as a fairly simple state machine - in regexp terms it's along the lines of this (but extended for more powers of 10) with the additional requirement that the matched string to be at least two characters long:

```
(<digit>?<1000>)?(<digit>?<100>)?(<digit>?<10>)<digit>?
```

And we should be able to convert that to a common form for indexing (so e.g. 四十五 is indexed as the term 45, and a search for 45 or 四十五 will match documents containing either, and 4十5 could be handled in the same way for little extra work).

But I may well be missing something that means this just isn't going to work that easily.

Thoughts?

outdream:
Never mind. And have to say, the wiki surprises me a lot, it seems pretty complete and wonderful!
 The GFW really keeps me unfamiliar with such great resources :(

However, I think maybe the pattern is not enough, as some important cases are mismatched:
1. Interior zeroes
for example, 205 | [2] [100] [0] [5] | 二百零五, 
the interior [0] has no number size likes [10] or [100].

2. Year numbers
as the example you gave before, 
`"1919 and 2011 will match a query for 2019" `
the year numbers in Chinese are continuous <digit>, as 1919 -> [1][9][1][9].
I'm not sure `(<digit>+)<year>` could handle the most cases.

These are also mentioned in the wiki, with more details. And I'm afraid the pattern would go more complex if we want to support fractional values or other languages like Japanese and Korean.

For the common form of numbers, if it is decided to implement, I advise that we could index a number with both its original form and common form (and with OR relationship in query?) if they are different. By this, language of the number also contribute to the final weight.

outdream:
Besides, could you tell me would Xapian support customize tokenizer? I think it is a way to solve many cases like this. (Or is there just such functions? Sorry in advance if I missed it carelessly, if that's the case, please ignore my verbose text below...)

Below are my consideration to this question. I'm not very confident with these thoughts, please your comments and forgiveness if I'm wrong:

 I think the tokenisation of CJK languages is pretty complex, and I have found many libs concentrate on such question of Chinese (I know little about the situation of Japanese and Korean...). ICU gives a universal solution to the problem. But in a specific language, there may be other special cases likes the numbers. 

I think fixing these cases patch by patch is laborious and unsuitable, as I think Xapian concerns more on the search technology rather than language process. Allow users to use localized tokenizer lib can't just solve all the problem, but could improve it on more cases.

ojwb:
A custom tokeniser is already possible - both `TermGenerator` and `QueryParser` could be implemented using only public API functions (I'm not sure if the current implementations actually are but there's nothing special they use which isn't available via public APIs).  So someone could write a `CJKQueryParser` which parses a query string in some way and builds a `Query` object tree from it, and a `CJKTermGenerator` to index text in a compatible way.

So far for the API we've aimed to instead have a single class with flags to control behaviour, which seems more flexible as we can automatically switch mode to handle a document which is partly in Chinese and partly in English - with separate classes the user would need to scan the text first and split it up, then pass the different parts to different classes.

I'm not sure a custom tokeniser really makes the problem easier to solve though - someone still needs to write the tokenisation code whether it's in `QueryParser` and `TermGenerator` or in a custom tokeniiser.

outdream:
Thanks, I got it. 

For the custom tokenizer, sorry for my ambiguous words. I mean to extract a framework and public interfaces, to allow users easily use other libs like ICU, but more localized to specific language, to support the language better. But consider the adaption code also would be a hard work, maybe that's not a good idea... It doesn't have strict association with the question, my fault.

And for the API, is what you want a common number tokenizer which would handle Chinese numbers when the CJK flag is on? As current code mismatch the design, I don't know if I should close this branch or left for further discussion? And maybe ticket #699 on trac also should be modified.

ojwb:
Sorry for the slow response - I've been a bit backlogged with patches lately.

> For the custom tokenizer, sorry for my ambiguous words. I mean to extract a framework and public interfaces, to allow users easily use other libs like ICU, but more localized to specific language, to support the language better. But consider the adaption code also would be a hard work, maybe that's not a good idea... It doesn't have strict association with the question, my fault.

Oh I see.  No, that's a good question.

I think the hard part is probably doing it in a way with is sufficiently flexible but doesn't impose any significant overhead.  If turning text into terms during indexing is slow that makes indexing slower, potentially which limits the amount of data that can be searched.  It's probably less of a concern during query parsing, just because queries are typically much shorter.

> And for the API, is what you want a common number tokenizer which would handle Chinese numbers when the CJK flag is on? As current code mismatch the design, I don't know if I should close this branch or left for further discussion? And maybe ticket #699 on trac also should be modified.

The ideal of turning different representations of a number into a common form for indexing still seems appealing, though it seems quite complex for Chinese.  But your idea of indexing the original form too would mean that it would be OK if the common form handling didn't handle all cases (so long as it didn't mishandle them in problematic ways).  And over time people can add versions for other scripts with their own way of writing numbers (like Arabic).  Even Roman Numerals could be handled, e.g. MCMLXXIV -> 1974.

I'm leaning towards thinking we should merge what we have here for now, as it seems an improvement over the current behaviour even if we can dream of grander solutions.

And yes, #699 should be updated once we decide about this patch.

outdream:
Sorry, I'm busy with my courses these days, and haven't got it in time...
Thanks for your approval, I'm glad that my efforts would be helpful.

ojwb:
Renaming master to main should have retargetted open PRs.  That worked for other PRs but for github decided to auto-close this one instead.  I think probably because outdream has deleted their repository (unhelpful corner-case handling from github).

Working towards normalising numbers still seems useful, so I've rescued the changes to branch `mixed-chinese-numbers` in our repo and we can track this in the existing trac ticket (http://trac.xapian.org/ticket/699).

(Incidentally, the Arabic stemmer actually replaces Arabic digits with their ASCII equivalents so we're already normalising it that case.)

PR #221, opened 2019-03-15
Remove non-option arguments from the help message.

VaibhavKansagara:


ojwb:
Thanks.  Looks good - will merge when I'm next in front of a computer.

PR #222, opened 2019-03-18
Add $random and $srandom commands to OmegaScript

tanmaysachan:
Addresses https://trac.xapian.org/ticket/717.

### DIFF
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -447,14 +447,9 @@ $queryterms
 $range{START,END}
 	return list of values between ``START`` and ``END``.
 
-$random
 
 $random{HIGH}
-
-$random{LOW,HIGH}
-	return a random value in the range [``LOW``, ``HIGH``).
-	If the value of ``LOW`` is ommitted, then the range [0, ``HIGH``) is used.
-	If both the values are ommitted, then the range [0, RAND_MAX] is used.
+	return a random value in the range [0, ``HIGH``].
 
 $record[{ID}]
 	raw record contents of document ``ID``.
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -447,7 +447,6 @@ $queryterms
 $range{START,END}
 	return list of values between ``START`` and ``END``.
 
-
 $random{HIGH}
 	return a random value in the range [0, ``HIGH``].
 
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -447,6 +447,12 @@ $queryterms
 $range{START,END}
 	return list of values between ``START`` and ``END``.
 
+$random{HIGH}
+
+$random{LOW,HIGH}
+	return a random value in the range [``LOW``, ``HIGH``).
+	If the value of ``LOW`` is ommitted, the range [0, ``HIGH``) is used instead.
+
 $record[{ID}]
 	raw record contents of document ``ID``.
 
@@ -681,6 +687,9 @@ $split{SPLIT,STRING}
 	For example::
 
 	 "$split{one two three}" = "one	two	three"
+
+$srandom{SEED}
+	``SEED`` specifies a seed for random number generation.
 
 $stoplist
 	returns a list of any terms in the query which were ignored as
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -447,11 +447,14 @@ $queryterms
 $range{START,END}
 	return list of values between ``START`` and ``END``.
 
+$random
+
 $random{HIGH}
 
 $random{LOW,HIGH}
 	return a random value in the range [``LOW``, ``HIGH``).
-	If the value of ``LOW`` is ommitted, the range [0, ``HIGH``) is used instead.
+	If the value of ``LOW`` is ommitted, then the range [0, ``HIGH``) is used.
+	If both the values are ommitted, then the range [0, RAND_MAX] is used.
 
 $record[{ID}]
 	raw record contents of document ``ID``.
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -662,10 +662,6 @@ testcase '89' x=89 y=90
 printf '$random{$cgi{x}}' > "$TEST_TEMPLATE"
 testcase '0' x=1
 
-# Feature tests for srandom{}
-printf '$srandom{$cgi{x}}$random{$cgi{y}}' > "$TEST_TEMPLATE"
-testcase '193' x=123 y=200
-
 # Regression test - original implementation truncated data at first zero byte.
 printf '$hash{$cgi{text}$chr{0},md5}' > "$TEST_TEMPLATE"
 testcase 'e2a3e68d23ce348b8f68b3079de3d4c9' text="test"
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -656,11 +656,26 @@ testcase 'd41d8cd98f00b204e9800998ecf8427e' text=""
 testcase '3302c94d3500c3f695b7e09b7acd420d' text=" test "
 
 # Feature tests for random{}
-printf '$random{$cgi{x},$cgi{y}}' > "$TEST_TEMPLATE"
-testcase '89' x=89 y=90
-
-printf '$random{$cgi{x}}' > "$TEST_TEMPLATE"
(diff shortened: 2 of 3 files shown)


xapian-applications/omega/docs/omegascript.rst
```diff
@@ -447,6 +447,15 @@ $queryterms
...
 	return list of values between ``START`` and ``END``.
 
+$random
+
+$random{HIGH}
+
+$random{LOW,HIGH}
+	return a random value in the range [``LOW``, ``HIGH``).
+	If the value of ``LOW`` is ommitted, then the range [0, ``HIGH``) is used.
+	If both the values are ommitted, then the range [0, RAND_MAX] is used.
```
ojwb: "ommitted" should be "omitted" (twice).

The ticket actually proposed returning a real value, though I think that's less useful than an integer as OmegaScript doesn't really provide a way to work with real numbers.  I think the motivation for this was to enable things like A/B testing where a different version is shown to a proportion of users, and for that integer random numbers should be sufficient.

I'm not sure the no-parameter version is very useful - it returns a random number up to some platform-dependent value which you can't access from OmegaScript.  I struggle to see how I would use that.

I'd probably just provide the single-parameter version for now - `$random{2,10}` is the same as `$add{$random{8},2}` and the inclusive/exclusive bounds seem potentially confusing.

xapian-applications/omega/omegatest
```diff
@@ -655,6 +655,13 @@ testcase 'b4c216e4da73d1d01277ef46d0514821' text="simple query"
 testcase 'd41d8cd98f00b204e9800998ecf8427e' text=""
 testcase '3302c94d3500c3f695b7e09b7acd420d' text=" test "
 
+# Feature tests for random{}
+printf '$random{$cgi{x},$cgi{y}}' > "$TEST_TEMPLATE"
+testcase '89' x=89 y=90
+
+printf '$random{$cgi{x}}' > "$TEST_TEMPLATE"
+testcase '0' x=1
+
```
ojwb: This should really test that the random number returned is within the expected range for some non-degenerate cases - for example, generate 100 numbers with `$random{6}` and check each is both >= 0 and < 6.

(And it doesn't test `$random` without arguments at all, but I think that should be removed as I say above).

xapian-applications/omega/query.cc
```diff
@@ -2051,6 +2055,22 @@ eval(const string &fmt, const vector<string> &param)
 		}
 		break;
 	    }
+	    case CMD_random: {
+		int rand_number;
+		if (args.empty())
+		    rand_number = rand();
+		else if (args.size() == 1)
+		    rand_number = rand() % string_to_int(args[0]);
```
ojwb: `rand()` is generally not a good quality source of pseudo-random numbers - in C++ code, C++11 offers much better options:

https://en.cppreference.com/w/cpp/numeric/random

These also provide a way to get a completely uniform distribution across the desired range - assuming `rand()` produces a uniform distribution, using `%` introduces a slight skew if the desired range size doesn't exactly divide into `RAND_MAX` (because one extra value maps to the first values in the range).

xapian-applications/omega/docs/omegascript.rst
tanmaysachan: Changed $random in the latest commit to support exactly one argument.

xapian-applications/omega/omegatest
tanmaysachan: I wrote a new function for checking rng testcases inside the script.

xapian-applications/omega/query.cc
tanmaysachan: Added mt19937 RNG, and used uniform_int_distribution to avoid that skewing.

xapian-applications/omega/omegatest
```diff
@@ -655,6 +655,28 @@ testcase 'b4c216e4da73d1d01277ef46d0514821' text="simple query"
...
+	if [ "$rng_test_failed" -eq 0 ]; then
+            failed=`expr $failed + 1`
+            rng_test_failed=1
+        fi
+    fi
+}
+for i in `seq 100`; do
+    printf '$random{$cgi{x}}' > "$TEST_TEMPLATE"
+    rng_test '6' x=6
+done
```
ojwb: While this looks functionally correct to me, it runs omega 100 times and adding a bit of timing code around it shows that it takes just over 2 seconds out of the ~12 seconds that omegatest now takes to run on my dev machine - that's a lot of overhead for testing a single feature.

We should be able to perform all the testing in one run of omega - `$map{$range{1,100},X}` will execute `X` 100 times and give us a list with all the results, so we can make `X` generate a single random value, range-test it, and give an empty string if all is well and a non-empty string otherwise, then we can concatenate all the entries in the resulting list (with `$list{MAPCOMMAND,}`) and the result will be empty if all 100 trials were good and non-empty otherwise (the `$or` below will give us "true" for each trial which failed the range check).

Putting that together we get something like this, which should output "OK" if the test passed (if an exception is thrown or omega crashes we'll get no output, so I've deliberately avoided having empty output mean "success"):

```
$and{$list{$map{$range{1,100},$set{num,$random{$cgi{x}}}$or{$lt{$opt{num},0},$gt{$opt{num},$cgi{x}}}},},OK}
```

And that takes about 0.022 seconds to run, reducing omegatest's runtime to about 10 seconds.

Don't worry about applying that - I'll merge your branch and add in that change as well shortly.

xapian-applications/omega/query.cc
```diff
@@ -2051,6 +2060,16 @@ eval(const string &fmt, const vector<string> &param)
...
 		break;
 	    }
+	    case CMD_random: {
+		if (args.size() == 1) {
+		    uniform_int_distribution<int>
+			distr(0, string_to_int(args[0]));
+		    value = str(distr(rng));
+		} else {
+		    throw "$random requires exactly 1 argument";
+		}
```
ojwb: You don't need to check the number of arguments here - `T(random,	   1, 1, N, 0), // return a random number` means that `$random` will only be passed one argument (the "1, 1" means "minimum 1, maximum 1").

xapian-applications/omega/omegatest
ojwb: > Don't worry about applying that - I'll merge your branch and add in that change as well shortly.

Oh, except I then spotted some other minor niggles.

xapian-applications/omega/query.cc
```diff
@@ -1228,6 +1233,10 @@ get_subdbs()
     return subdbs;
 }
 
+// mersenne twister for RNG
+random_device rd;
+mt19937 rng(rd());
```
ojwb: It's cleaner to make file-scope objects such as these `static` (unless they actually need to be visible to other source files).

What this code does (as you're probably aware) is to generate a random number using `std::random_device` and use that to seed the MT19037 pseudo-random number generator.  This will happen early in program initialisation, before `main()` is called.

In most cases, `std::random_device` will use the system's hardware entropy source, so doing this on every run of Omega will use up a few bytes of entropy for each run (even when using templates which never use `$random` or always call `$srandom` before calling `$random`).  Servers can sometimes struggle to generate enough entropy, so consuming it for no reason is potentially anti-social and can cause other programs which need entropy to block waiting for more to be harvested.

So I think we don't want to be seeding from `std::random_device` here.  Either we track whether `$srandom` has been called and if not seed from `std::random_device` the first time `$random` is called, or just allow `$srandom` with no argument to mean "seed from `std::random_device`" and then leave it up to the template author to decide if and how to seed the the PRNG.

tanmaysachan: Well, I clearly didn't think that far. I implemented the first option that you suggested. Thanks for the input!

xapian-applications/omega/query.cc
tanmaysachan: Fixed in the latest commit.

tanmaysachan:
Ah sorry, I accidentally committed some file that I wasn't supposed to push.

xapian-applications/omega/query.cc
```diff
@@ -1228,6 +1233,10 @@ get_subdbs()
     return subdbs;
 }
 
+// mersenne twister for RNG
+static mt19937 rng;
+static bool seed_set = 0;
```
ojwb: It's clearer to use `false` and `true` as boolean literals rather than `0` and `1` (here and below).

I only noticed after merging, so I've just fixed this in a follow-up commit.

ojwb:
Thanks for the updates.

Squashed, rebased and pushed to master as 1c25347edb6c76e9659fd882ddcd6f4174026ed7 with 2e6bbee7390143f61d9925e586b1351bf20f7c7d to adjust the testcase to be more efficient and 045d44fda818afed431c6c69e3e1a935953f212e to use false/true not 0/1 for bool literals.

PR #223, opened 2019-03-25
Implement LD_PRELOADable library to log IO.

tanmaysachan:
Addresses https://trac.xapian.org/ticket/390.

I have not formatted the logged calls to be exactly the same as strace, just enough to be compatible with strace-analyse, as strace-analyse does not do any operations on the buffer being used by read(), or write() calls.

### DIFF
--- a/xapian-maintainer-tools/profiling/profiler.c
+++ b/xapian-maintainer-tools/profiling/profiler.c
@@ -62,6 +62,29 @@ int open(const char *pathname, int flags, ...)
     return fd;
 }
 
+// wrapper for open64()
+
+typedef int (*real_open64_t)(const char *, int, ...);
+
+int open64(const char *pathname, int flags, ...)
+{
+    int fd;
+    if (flags & (O_CREAT | O_TMPFILE)) {
+	va_list args_ptr;
+	va_start(args_ptr, flags);
+	fd =
+	    ((real_open_t)dlsym(RTLD_NEXT, "open64"))
+	    (pathname, flags, va_arg(args_ptr, mode_t));
+	va_end(args_ptr);
+    } else {
+	fd = ((real_open_t)dlsym(RTLD_NEXT, "open64"))(pathname, flags);
+    }
+    char *abspath = realpath(pathname, NULL);
+    logcall("open(\"%s\",,) = %d\n", abspath, fd);
+    free(abspath);
+    return fd;
+}
+
 // wrapper for close()
 
 typedef int (*real_close_t)(int);
@@ -110,13 +133,13 @@ ssize_t pread(int fd, void *buf, size_t count, off_t offset)
 
 // wrapper for pread64()
 
-typedef ssize_t (*real_pread64_t)(int, void *, size_t, off_t);
+typedef ssize_t (*real_pread64_t)(int, void *, size_t, off64_t);
 
-ssize_t pread64(int fd, void *buf, size_t count, off_t offset)
+ssize_t pread64(int fd, void *buf, size_t count, off64_t offset)
 {
     ssize_t return_val =
 	((real_pread64_t)dlsym(RTLD_NEXT, "pread64"))(fd, buf, count, offset);
-    logcall("pread64(%d, \"\", %lu, %ld) = %ld\n",
+    logcall("pread(%d, \"\", %lu, %ld) = %ld\n",
 	    fd, count, offset, return_val);
     return return_val;
 }
@@ -136,13 +159,13 @@ ssize_t pwrite(int fd, const void *buf, size_t count, off_t offset)
 
 // wrapper for pwrite64()
 
-typedef ssize_t (*real_pwrite64_t)(int, const void *, size_t, off_t);
+typedef ssize_t (*real_pwrite64_t)(int, const void *, size_t, off64_t);
 
-ssize_t pwrite64(int fd, const void *buf, size_t count, off_t offset)
+ssize_t pwrite64(int fd, const void *buf, size_t count, off64_t offset)
 {
     ssize_t return_val =
 	((real_pwrite64_t)dlsym(RTLD_NEXT, "pwrite64"))(fd, buf, count, offset);
-    logcall("pwrite64(%d, \"\", %lu, %ld) = %ld\n",
+    logcall("pwrite(%d, \"\", %lu, %ld) = %ld\n",
 	    fd, count, offset, return_val);
     return return_val;
 }
--- a/xapian-maintainer-tools/profiling/profiler.c
+++ b/xapian-maintainer-tools/profiling/profiler.c
@@ -41,7 +41,7 @@ void logcall(const char *call)
 
 // wrapper for open()
 
-typedef int (*real_open_t)(const char *, int, ...);
+typedef int(*real_open_t)(const char *, int, ...);
 
 int open(const char *pathname, int flags, ...)
 {
@@ -65,7 +65,7 @@ int open(const char *pathname, int flags, ...)
 
 // wrapper for close()
 
-typedef int (*real_close_t)(int);
+typedef int(*real_close_t)(int);
 
 int close(int fd)
 {
@@ -78,7 +78,7 @@ int close(int fd)
 
 // wrapper for fdatasync()
 
-typedef ssize_t (*real_fdatasync_t)(int);
+typedef ssize_t(*real_fdatasync_t)(int);
 
 ssize_t fdatasync(int fd)
 {
@@ -91,7 +91,7 @@ ssize_t fdatasync(int fd)
 
 // wrapper for fsync()
 
-typedef ssize_t (*real_fsync_t)(int);
+typedef ssize_t(*real_fsync_t)(int);
 
 ssize_t fsync(int fd)
 {
@@ -104,7 +104,7 @@ ssize_t fsync(int fd)
 
 // wrapper for pread()
 
-typedef ssize_t (*real_pread_t)(int, void *, size_t, off_t);
+typedef ssize_t(*real_pread_t)(int, void *, size_t, off_t);
 
 ssize_t pread(int fd, void *buf, size_t count, off_t offset)
 {
@@ -112,7 +112,7 @@ ssize_t pread(int fd, void *buf, size_t count, off_t offset)
 	((real_pread_t)dlsym(RTLD_NEXT, "pread"))(fd, buf, count, offset);
     char call[MAX_CALL_LEN];
     char *new_buf = (char *)buf;
-    for (size_t i = 0; i < count; i++) {
+    for (size_t i = 0; i < count; ++i) {
 	if (new_buf[i] == '\n') {
 	    new_buf[i] = '\0';
 	}
@@ -125,7 +125,7 @@ ssize_t pread(int fd, void *buf, size_t count, off_t offset)
 
 // wrapper for pread64()
 
-typedef ssize_t (*real_pread64_t)(int, void *, size_t, off_t);
+typedef ssize_t(*real_pread64_t)(int, void *, size_t, off_t);
 
 ssize_t pread64(int fd, void *buf, size_t count, off_t offset)
 {
@@ -133,7 +133,7 @@ ssize_t pread64(int fd, void *buf, size_t count, off_t offset)
 	((real_pread64_t)dlsym(RTLD_NEXT, "pread64"))(fd, buf, count, offset);
     char call[MAX_CALL_LEN];
     char *new_buf = (char *)buf;
-    for (size_t i = 0; i < count; i++) {
+    for (size_t i = 0; i < count; ++i) {
 	if (new_buf[i] == '\n') {
 	    new_buf[i] = '\0';
 	}
@@ -146,7 +146,7 @@ ssize_t pread64(int fd, void *buf, size_t count, off_t offset)
 
 // wrapper for pwrite()
 
-typedef ssize_t (*real_pwrite_t)(int, const void *, size_t, off_t);
+typedef ssize_t(*real_pwrite_t)(int, const void *, size_t, off_t);
 
 ssize_t pwrite(int fd, const void *buf, size_t count, off_t offset)
 {
@@ -154,7 +154,7 @@ ssize_t pwrite(int fd, const void *buf, size_t count, off_t offset)
 	((real_pwrite_t)dlsym(RTLD_NEXT, "pwrite"))(fd, buf, count, offset);
     char call[MAX_CALL_LEN];
     char *new_buf = (char *)buf;
-    for (size_t i = 0; i < count; i++) {
+    for (size_t i = 0; i < count; ++i) {
 	if (new_buf[i] == '\n') {
 	    new_buf[i] = '\0';
 	}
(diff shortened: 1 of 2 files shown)


xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -65,7 +65,7 @@ int open(const char *pathname, int flags, ...)
 
 // wrapper for close()
 
-typedef int (*real_close_t)(int);
+typedef int(*real_close_t)(int);
```
ojwb: Hmm, are these changes in response to complaints from `xapian-check-patch`?

If so, I think it's `xapian-check-patch` that is buggy here and your original is better.

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,189 @@
...
+
+int open(const char *pathname, int flags, ...)
+{
+    int fd;
+    if (flags & O_CREAT || flags & O_TMPFILE) {
+	va_list args_ptr;
+	va_start(args_ptr, flags);
+	fd =
+	    ((real_open_t)dlsym(RTLD_NEXT, "open"))
+	    (pathname, flags, va_arg(args_ptr, mode_t));
```
ojwb: From `man va_start`:

```
   va_end()
       Each  invocation  of va_start() must be matched by a corresponding invocation of va_end() in the same function.  
```

You need to call `va_end()` here...

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// wrapper for open()
+
+typedef int(*real_open_t)(const char *, int, ...);
+
+int open(const char *pathname, int flags, ...)
+{
+    int fd;
+    if (flags & O_CREAT || flags & O_TMPFILE) {
```
ojwb: You can check both bits in one operation with `flags & (O_CREAT | O_TMPFILE)`, which seems more readable too.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: Yes, these were being rejected by `xapian-check-patch`.
I updated the PR, now it will probably be failing both the checks.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: I apologize. Updated accordingly.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: Updated accordingly.

xapian-maintainer-tools/profiling/profiler.c
ojwb: > Yes, these were being rejected by `xapian-check-patch`.

OK, I'll fix that.

> I updated the PR, now it will probably be failing both the checks.

Yes, though it's also failing due to an indentation issue.

ojwb:
> I have not formatted the logged calls to be exactly the same as strace, just enough to be compatible with strace-analyse, as strace-analyse does not do any operations on the buffer being used by read(), or write() calls.

Agreed there's no point trying to log the buffer contents - we don't currently use it and I can't see that we would (with strace we suggest using `-s0` which stops it logging the buffer contents).

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,190 @@
...
+{
+    ssize_t return_val =
+	((real_pread_t)dlsym(RTLD_NEXT, "pread"))(fd, buf, count, offset);
+    char call[MAX_CALL_LEN];
+    char *new_buf = (char *)buf;
+    for (size_t i = 0; i < count; ++i) {
+	if (new_buf[i] == '\n') {
+	    new_buf[i] = '\0';
+	}
+    }
```
ojwb: This is just wrong - you're modifying the data returned by `pread()` which will break things.

I guess you're trying to clean up the returned data to log it, but `strace-analyse` isn't interested in that - we can just emulate what strace logs with `-s0`.

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,190 @@
...
+    FILE *file_ptr;
+    if (!reset) {
+	file_ptr = fopen("./strace.log", "w");
+	reset = 1;
+    } else {
+	file_ptr = fopen("./strace.log", "a");
+    }
+    fprintf(file_ptr, "%s", call);
+    fclose(file_ptr);
+}
```
ojwb: You're opening and closing the log file for each call logged, which will slow things down unnecessarily.

Also if the implementation of `fopen()` calls `open()` and we intercept that we'll get an infinite recursion.

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,190 @@
...
+// wrapper for close()
+
+typedef int (*real_close_t)(int);
+
+int close(int fd)
+{
+    int return_val = ((real_close_t)dlsym(RTLD_NEXT, "close"))(fd);
+    char call[MAX_CALL_LEN];
+    snprintf(call, MAX_CALL_LEN, "close(%d) = %d\n", fd, return_val);
+    logcall(call);
```
ojwb: Printing the log message to a temporary buffer and then writing the buffer means that we have to impose a fixed limit of the length of log messages.

If you make the `logcall` function take varargs you can just call `logcall("close(%d) = %d\n", fd, return_val);` here, and then in the implementation of `logcall` call `vfprintf()` to pass on the parameters.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: I apologize, I can't believe how I messed that up. I'll fix it.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: How do you propose I close the strace file? Using __attribute___((destructor))?

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: I'll fix it.

xapian-maintainer-tools/profiling/profiler.c
ojwb: There's no need to close it explicitly - when a process exits any files it has open will be closed automatically.

ojwb:
You said on IRC that analysing the log from this with the `strace-analyse` script failed to give a path for `fsync`/`fdatasync`.

I tried this out on x86 Linux (so 32-bit), testing with:

```
cd xapian-core/tests
make remove-cached-databases
LD_PRELOAD=/home/olly/git/xapian/xapian-maintainer-tools/profiling/profiler.so ./apitest -b glass zerodocid1
```

At least there the issue is that `open64` is used by libxapian to open file, but that isn't intercepted so the opening of files doesn't get logged.  I haven't tested on x86-64 yet.

I also found that I got a corrupted database.  The issue there seems to be that the `pwrite64` and `pread64` wrappers use `off_t` when they ought to use `off64_t`.

But aside from that this seems to actually work - adding a wrapper for `open64` (and just logged it the same as `open`) and fixing the `off_t` issues:

```
olly@pippikin:~/git/xapian/xapian-core/tests$ ~/git/xapian/xapian-maintainer-tools/profiling/strace-analyse < strace.log 
sync /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/v.tmp
write 0 to /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/postlist.glass
write 0 to /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/position.glass
write 0 to /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/termlist.glass
write 0 to /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/docdata.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/postlist.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/position.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/termlist.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/docdata.glass
sync /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/postlist.glass
sync /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/position.glass
sync /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/termlist.glass
sync /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/docdata.glass
sync /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc.tmp/v.tmp
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/docdata.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/termlist.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/position.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/postlist.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/docdata.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/termlist.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/position.glass
read 0 from /home/olly/git/xapian/xapian-core/tests/.glass/db__apitest_onedoc/postlist.glass
```

tanmaysachan:
Added wrapper for open64 and used off64_t inplace of off_t for pread64 and pwrite64.

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,171 @@
...
+    static FILE *file_ptr;
+    if (!is_reset) {
+	file_ptr = fopen("./strace.log", "w");
+	is_reset = 1;
+    }
+    va_list args_ptr;
+    va_start(args_ptr, format);
+    vfprintf(file_ptr, format, args_ptr);
+    va_end(args_ptr);
+}
```
ojwb: There are a couple of minor problems here.

* If the logfile can't be opened, `file_ptr` will be NULL and the call to `vfprintf()` will likely segfault.  Probably you'd want to skip logging in that case.

* Some of these functions can potentially change `errno` - if the intercepted function we're logging the result of failed and set an `errno` value, and then the logging code changes that to a different one, the calling code will see the intercepted function fail with the wrong `errno` value.  The simplest fix is to wrap `int saved_errno = errno;` ... `errno = saved_errno;` around the code here function.

Another thing I notice if that it's a bit unhelpful to have a fixed log file name, but hard to provide a nice way to specify it here.  We could pass in a filename via an environment variable, but then we don't have a good way to report a failure to open it.

We could instead allow specifying a file descriptor to log to via an environment variable (and default to 2 for `stderr` output by default, which is what tools like `strace` and `ltrace` default to).  Then a `xapian-io-profile` wrapper shell script could open a high file descriptor onto the log file and pass that in, e.g. something like (untested):

```
#!/bin/bash
set -e
case $1 in
  --log=*)
    log_file=`${1/--log=/}`
    shift
    exec 99>"$log_file"
    export XAPIAN_IO_PROFILE_LOG_FD=99
    ;;
esac
LD_PRELOAD=/path/to/profiler.so exec "$@"
```

Then you could profile a command which uses xapian like so, which is a bit neater than having to set `LD_PRELOAD`:

```
xapian-io-profile --log=my-io.log xapian-compact in.db out.db
```

Then if the log file fails to open, the wrapper script reports the error (and the command shouldn't even get run).

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,171 @@
...
+
+// wrapper for pread()
+
+typedef ssize_t (*real_pread_t)(int, void *, size_t, off_t);
+
+ssize_t pread(int fd, void *buf, size_t count, off_t offset)
+{
+    ssize_t return_val =
+	((real_pread_t)dlsym(RTLD_NEXT, "pread"))(fd, buf, count, offset);
```
ojwb: I think we probably want to call `dlsym()` just once for each different function name and store the returned function pointers in variables.  Resolve a function name to a function pointer is inevitably going to be slower than just loading a function pointer from a variable.

In eatmydata they use the GCC extension (which clang likely also supports) `__attribute__ ((constructor))` on `eatmydata_init()` to have it run really early and then that calls `dlsym()`, but then they also have explicit checks which result in `eatmydata_init()` being called if it hasn't already been - probably because some of the intercepted libc functions can get called before the constructor function does.  That seems a bit clumsy overall.

Probably OK to leave this as-is for now and think about how best to achieve caching of these.

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -0,0 +1,171 @@
...
+	fd =
+	    ((real_open_t)dlsym(RTLD_NEXT, "open64"))
+	    (pathname, flags, va_arg(args_ptr, mode_t));
+	va_end(args_ptr);
+    } else {
+	fd = ((real_open_t)dlsym(RTLD_NEXT, "open64"))(pathname, flags);
+    }
+    char *abspath = realpath(pathname, NULL);
+    logcall("open(\"%s\",,) = %d\n", abspath, fd);
+    free(abspath);
```
ojwb: Here `realpath()` could set `errno` and return `NULL`, which would result in us overwriting the value of `errno` returned to the caller and trying to format `NULL` as `%s` (glibc supports that as an extension, but on many platforms it will segfault).  Even when it works I think it'd be better to show the value of `pathname` instead of something like `<NULL>` in that case.

xapian-maintainer-tools/profiling/profiler.c
ojwb: On IRC you asked:
> wouldn't the script that you wrote(on my PR) require recompilation of profiler.c each time?

No - `XAPIAN_IO_PROFILE_LOG_FD` is an environment variable (which can be read using `getenv()`), not a compiler preprocessor macro.

tanmaysachan: Yes, understood. I implemented it.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: Saving and restoring errno around realpath in the latest commit.

ojwb:
Thanks for the updates.

There are still some things to sort out here, but this is working at least for x86-64 Linux so I've squashed and rebase and pushed as 3c3c09048e13.  We can iterate on portability, etc.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: We can call `dlsym()` only when needed to, and not create some init function which goes around and calls `dlsym()` for all the calls. That way we won't need these checks, no point in using `__attribute__((constructor))`.

ojwb: Oh yes, that's cleaner.

tanmaysachan: Opened https://github.com/xapian/xapian/pull/236 regarding the same.

PR #224, opened 2019-03-25
Replaces atoi() and atol() with strtol(), strtoul() where necessary

arvinddeshraj:
Replacing atoi and atol with strtol/strtoul where appropriate


ojwb:
Thanks for the patch.

The main issue here is that `atoi()` doesn't provide any way to usefully check for errors (like the input not being a number, or having trailing characters after such as `123abc`, or overflowing the type).  Eliminating use of `atoi()` by switching to `strtoul()`, etc but without actually adding any error checking is not really a useful step forwards.

Also someone else is already working on this so probably better to pick something else to work on.  Sorry, we should have removed this from the list - I'll do that now.

PR #225, opened 2019-03-26
Replace atoi with parse_unsigned in omega

VaibhavKansagara:
Please review this PR when you are free.


xapian-applications/omega/query.cc
```diff
@@ -1444,7 +1452,11 @@ eval(const string &fmt, const vector<string> &param)
 		value = args[0];
 		if (!value.empty()) {
 		    char buf[64] = "";
-		    time_t date = string_to_int(value);
+		    unsigned int temp;
+		    if (!parse_unsigned(value.c_str(), temp)) {
+			throw "Date(in secs) for command date should be >= 0";
+		    }
```
ojwb: Dates before 1970 have negative `time_t` values (and currently work).

I wonder if we want a `parse_signed()` function and to use that to read directly into a `time_t` (which POSIX requires to be a signed integer type IIRC).  Using `int` here means that dates after somewhere in 2038 don't work.

VaibhavKansagara: > I wonder if we want a `parse_signed()` function 
>
yeah that would be a good idea, It would lead to a much clearer and less code too.  I will write one for that. I think long long should solve the date after 2038 problem.

> Dates before 1970 have negative `time_t` values (and currently work).
> 
Didn't know that, will fix it.

ojwb: > I think long long should solve the date after 2038 problem.

I don't think using a wider type really helps (and if you wanted a 64-bit, use `int64_t` - `long long` is only guaranteed to be **at least** a 64 bit type - it could be wider).

But if you parse the value into `long long`, then you'll have to decide what to do if the parsed integer value doesn't fit in a `time_t` - if you just assign it without checking then it's likely to just mask off the upper bits and dates after the highest representable time in 2038 suddenly come out as in 1901 which is just not helpful.

We could throw an exception, but then searching the same database can throw an exception on a platform with 32-bit `time_t` when it works on a platform with 64-bit `time_t`.  But if you wanted to do that, then trying to convert to `time_t` directly seems the best way to do it.

Probably the best alternative is for `$date{X}` to return an empty string if the date can't be represented on the current platform.  But then it seems more consistent to do that for **any** value which doesn't fit in a `time_t`, so converting via `int64_t` or `long long` just makes things more complicated - the proposed `parse_signed()` will tell us if the value doesn't fit in a `time_t`, and then we don't try to convert it further.

VaibhavKansagara: understood.

VaibhavKansagara:
@ojwb Is the given implementation for parse_signed function acceptable?.
 ```cpp
template<typename T>
bool parse_signed(const char* p, T& res)
{
    typedef typename std::make_unsigned<T>::type unsigned_type;
    unsigned_type temp = 0;
    if (*p == '-') {
    	if (!parse_unsigned(++p,temp)){
    		cout << "hello";
    	    return false;
    	} else {
    	    res = -temp;
    	    return true;
    	}
    } else {
    	if (!parse_unsigned(p,temp)) {
    	    return false;
    	} else {
    	    res = temp;
    	    return true;
    	}
    }
} 
```

**Comment by ojwb** [comment]
That looks good for valid values and for very invalid values, but it goes a bit wrong if the numeric part fits in `unsigned_type` but the overall value doesn't fit in `T` (e.g. if `T is signed char` and we try to convert the string `200`).  I think we need to check `temp > unsigned_type(std::numeric_limits<T>::max())` for the positive case, and something similar for the negative case.

Probably best to do that as a separate PR which we can merge first - we'll need to update the commit hashes for common in `bootstrap` before we can use it from omega.

File: xapian-applications/omega/date.cc [hunk 2]
```diff
@@ -191,7 +192,13 @@ date_range_filter(const string & date_start, const string & date_end,
...
     int y1, m1, d1, y2, m2, d2;
     if (!date_span.empty()) {
-	time_t secs = atoi(date_span.c_str()) * (24 * 60 * 60);
+	time_t secs;
+	unsigned int temp;
+	if (!parse_unsigned(date_span.c_str(), temp)) {
+	    throw "Datespan value must be >= 0";
+	} else {
+	    secs = temp * (24 * 60 * 60);
+	}
```
ojwb: You could lose the `else` (the `if` block will always transfer control away) and then you can move the declaration of `secs` down so it gets initialised at the point of declaration.  Being able to declare and initialise in one go is cleaner - it makes it completely obvious to the reader that the variable can't be used uninitialised.

Also, a more descriptive name than `temp` would make the code easier to understand - e.g. `days`.

xapian-applications/omega/datevalue.cc
```diff
@@ -30,6 +30,7 @@
 
 #include "timegm.h"
 #include "values.h"
+#include "parseint.h"
```
ojwb: In general, when adding something to a list in the code which is currently in sorted order you should add the new entry to preserve that property.

xapian-applications/omega/datevalue.cc
```diff
@@ -224,7 +225,13 @@ date_value_range(bool as_time_t,
...
 
     if (!date_span.empty()) {
-	time_t span = atoi(date_span.c_str()) * (24 * 60 * 60) - 1;
+	time_t span;
+	unsigned int temp;
+	if (!parse_unsigned(date_span.c_str(), temp)) {
+	    throw "Datespan value must be >= 0";
+	} else {
+	    span = temp * (24 * 60 * 60) - 1;
+	}
```
ojwb: Same comments as above.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -847,9 +847,15 @@ $weight
...
+For Numeric Operators we allow trailing characters and also allow non-number arguments.
+Reason for this behaviour is that it makes things robust if some of the parameters come in as CGI parameters.
+Ex:- if you had a CGI param at the end of a URL that was supposed to be a number, and sent it in an email
+or a message, it's possible for the person receiving to end up with a URL with a dot or semicolon at the
+end (from punctuation in your message).
+
 $add{...}
 	add arguments together (if called with one argument, this will convert
-	it to a string and back, which ensures it is an integer).
+	it to a int and back, which ensures it is an integer).
```
ojwb: "a**n** int" (or perhaps better "an integer")

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -875,6 +881,10 @@ $sub{A,B}
 Logical Operators:
 ==================
 
+For Logical Operators we allow trailing characters,non-number arguments and empty arguments.
```
ojwb: Put a space after a comma in prose.

xapian-applications/omega/docs/omegascript.rst
```diff
@@ -875,6 +881,10 @@ $sub{A,B}
...
+Reason is that logical operators compare their arguments based on whether they are empty or not.
+Ex: not returns true for empty string and false otherwise.
```
ojwb: I'm not really sure this explanation make things much clearer.  These operators only care about whether the value is an empty string or not an empty string.  So there aren't "trailing characters", and talking about "non-number arguments" risks suggesting that numeric arguments are in some way meaningful here.

Adding a note here is probably useful, but I think it just wants to explain that in this context OmegaScript treats an empty string as a "false" logical value and any non-empty string as "true".

xapian-applications/omega/query.cc
```diff
@@ -1420,9 +1423,14 @@ eval(const string &fmt, const vector<string> &param)
...
 		break;
 	    }
-	    case CMD_chr:
-		Xapian::Unicode::append_utf8(value, string_to_int(args[0]));
+	    case CMD_chr: {
+		unsigned int temp;
+		if (!parse_unsigned(args[0].c_str(), temp)) {
+		    throw "Unicode codepoint for command chr should be >= 0";
+		}
+		Xapian::Unicode::append_utf8(value, temp);
```
ojwb: You really do like your `temp` variable names, don't you?

Here e.g. `codepoint` would be a better name than `temp`.

xapian-applications/omega/query.cc
```diff
@@ -1459,7 +1467,11 @@ eval(const string &fmt, const vector<string> &param)
 		value = args[0];
 		if (!value.empty()) {
 		    char buf[64] = "";
-		    time_t date = string_to_int(value);
+		    time_t date;
+		    if (!parse_signed(value.c_str(), date)) {
+			throw "Date(in secs) for command date should"
```
ojwb: Should have a space before the `(` in the message.

ojwb: Also needs a space after `should` - currently this message will read:

```
Date(in secs) for command date shouldbe integer
```

xapian-applications/omega/query.cc
```diff
@@ -2022,9 +2040,15 @@ eval(const string &fmt, const vector<string> &param)
 		}
 		break;
 	    }
-	    case CMD_pack:
-		value = int_to_binary_string(string_to_int(args[0]));
+	    case CMD_pack: {
+		int temp;
+		if (!parse_signed(args[0].c_str(), temp)) {
+		    throw "NUMBER parameter for pack command"
+			  " must be an integer";
```
ojwb: Pretty much universally in existing Xapian code, when splitting string like this the space goes in the first half not the second.  I don't think it's a documented rule, but I think it makes sense to follow that - when there's a pattern it's easier to spot an exception to it - e.g. when the space gets omitted, or included twice.

xapian-applications/omega/scriptindex.cc
```diff
@@ -495,7 +496,11 @@ parse_index_script(const string &filename)
 			// We don't push an Action for WEIGHT - instead we
 			// store it ready to use in the INDEX and INDEXNOPOS
 			// Actions.
-			weight = atoi(val.c_str());
+			if (!parse_signed(val.c_str(), weight)) {
+			    report_location(DIAG_WARN, filename, line_no);
+			    cerr << "Index action 'weight'"
+				    "takes an integer argument" << endl;
```
ojwb: Missing space at the split in the message.

xapian-applications/omega/scriptindex.cc
```diff
@@ -495,7 +496,11 @@ parse_index_script(const string &filename)
 			// We don't push an Action for WEIGHT - instead we
 			// store it ready to use in the INDEX and INDEXNOPOS
 			// Actions.
-			weight = atoi(val.c_str());
+			if (!parse_signed(val.c_str(), weight)) {
```
ojwb: Here weight should really be non-negative (it's how much the `wdf` is increased by), though the current docs do just say "an integer".  It would be a bit odd to specify zero, but I can see there might be uses for indexing text without increasing wdf.

xapian-applications/omega/scriptindex.cc
```diff
@@ -1002,7 +1007,13 @@ run_actions(vector<Action>::const_iterator action_it,
 		const string & type = action.get_string_arg();
 		string yyyymmdd;
 		if (type == "unix") {
-		    time_t t = atoi(value.c_str());
+		    time_t t;
+		    if (!parse_signed(value.c_str(), t)) {
+			report_location(DIAG_ERROR, fname, line_no);
+			cerr << "Date value(in secs) for action DATE"
```
ojwb: Missing space before `(` again.

xapian-applications/omega/scriptindex.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-applications/omega/query.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-applications/omega/docs/omegascript.rst
VaibhavKansagara: fixed in a fixup commit.

xapian-applications/omega/date.cc
VaibhavKansagara: fixed in a fixup commit.

ojwb:
Thanks for the updates.

I think this is good to go now, but I should look it over a final time when I'm less tired as it changes a number lot of things we don't have test coverage for.

ojwb:
Hmm, rebase is conflicting in unhelpful ways, I guess due to the merges from master, so I can't easily squash the fixup commits.  Maybe I'll have to cherry-pick the commits onto master.

ojwb:
Aha, your fixup commit f95841ae0d735a07401de3a4078dfa1402ab8b4d actually contains fixups for both the commit it targets and the other commit, so once `git rebase -i --autosquash` reorders the commits, that fixup tries to make adjustments on top of changes that haven't yet been applied.

I'll try to split that commit and then rebase, but please bear in mind that you need to ensure fixups are against the right commit when there are multiple non-fixup commits in the branch.

VaibhavKansagara:
Sorry didn't realise that, will keep that in mind.

ojwb:
While I was splitting up commits, I also split your "Update documentation for Numerical and Logical Operators and correct typo in add parameter" commit into two since it's really doing two different things (if you find you're writing "do A and B" or "X.  Also Y" in a commit description, it's worth thinking if the commit is actually trying to do two separate things).

"Update documentation for Numerical and Logical Operators and correct typo in add parameter" is also much too long for a commit summary - the short summary of the first line should ideally be 50 characters or fewer so it formats nicely in the output of common git commands.  This is a limit that it's probably OK to push a little if you really have to (better for the summary to actually be meaningful) but keep it short and to the point and use the longer description for more detail.  If you're writing commit messages on the command line with `git commit -m 'blah blah'` then that's a habit you really want to shake off.

See https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages for more about writing good commit messages.

Anyway, as for the actual changes, have read over them again most are good but a few contradict statements in the documentation.  For example, the docs for scriptindex's `date` action explicitly say (my emphasis) "Unknown formats, and **invalid values are ignored** at present." but after this patch invalid values are give a hard error which stops the indexer.  Overall I think it's best at this point to push the merged patch and then work through and address these.

ojwb:
Merged in c701e8039224fd7d6034705b45c48eb2f063006c, 6d1b40705f3af3b0d38bb6f899f30a6aa7831bf8, 0df56ba087896dfac16d751813011d5d4e5dd87d, and adjusted a few cases in the commits following those.

PR #226, opened 2019-04-01
Add Unittestcases for features in xapian-letor

VaibhavKansagara:
Please review this PR when you have time.

### DIFF
--- a/xapian-letor/tests/api_letor.cc
+++ b/xapian-letor/tests/api_letor.cc
@@ -805,7 +805,7 @@ DEFINE_TESTCASE(err_scorer, !backend)
     return true;
 }
 
-//Test for TfDoclenFeature
+// Test for TfDoclenFeature
 DEFINE_TESTCASE(tfdoclenfeature, generated)
 {
     string db_path = get_database_path("apitest_tfdoclenfeature",
@@ -814,15 +814,25 @@ DEFINE_TESTCASE(tfdoclenfeature, generated)
     Xapian::Document doc = db.get_document(1);
     std::map<std::string,Xapian::termcount> len;
     Xapian::TermIterator dt = doc.termlist_begin();
-    
+
     len["title"] = 4;
     Xapian::termcount whole_len = db.get_doclength(1);
     len["whole"] = whole_len;
-    TEST_EQUAL(whole_len,186);
+    TEST_EQUAL(whole_len, 186);
     len["body"] = whole_len - len["title"];
 
     Xapian::TfDoclenFeature feature;
     feature.set_doc_length(len);
+
+    Xapian::QueryParser queryparser;
+    queryparser.set_stemmer(Xapian::Stem("en"));
+    queryparser.set_stemming_strategy(queryparser.STEM_ALL_Z);
+    queryparser.add_prefix("title", "S");
+    queryparser.add_prefix("description", "XD");
+
+    string querystring = "title:tigers description:tigers tigers";
+    Xapian::Query query = queryparser.parse_query(querystring);
+    feature.set_query(query);
 
     std::map<std::string, Xapian::termcount> tf;
     tf["ZStiger"] = 1;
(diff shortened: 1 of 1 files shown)


xapian-letor/tests/api_letor.cc
```diff
@@ -804,3 +804,42 @@ DEFINE_TESTCASE(err_scorer, !backend)
 
     return true;
 }
+
+//Test for TfDoclenFeature
+DEFINE_TESTCASE(tfdoclenfeature, generated)
+{
+    string db_path = get_database_path("apitest_tfdoclenfeature",
+				       db_index_one_document);
```
ojwb: This copies a pattern from the existing code, but that existing pattern isn't really right.

We're generating databases with `db_index_one_document` but in each testcase the generated database seems to be given a different name.  However these databases will all be identical, and if we use the same name when the same generator function is used (and with the same optional parameter if that's specified which it isn't here) then we can share a single generated database between all the testcases which use it.

I tend to think it's a flaw in the current harness that you have to explicitly give names to generated databases, but I haven't come up with a good way to name them automatically.

Given this is an existing issue, it probably makes sense to leave this as is, and fix the generated databases to be shareable as a separate change.

xapian-letor/tests/api_letor.cc
```diff
@@ -804,3 +804,42 @@ DEFINE_TESTCASE(err_scorer, !backend)
...
+    TEST_EQUAL(res.size(), 3);
+
+    //test for title
+    TEST_EQUAL(res[0], log10(1 + (1 / (1 + len["title"]))));
+
+    //test for body
+    TEST_EQUAL(res[1], log10(1 + (18 / (1 + len["body"]))));
+
+    //test for whole
+    TEST_EQUAL(res[2], log10(1 + (19 / (1 + len["whole"]))));
```
ojwb: The calculation of the expected answers here will use integer division, so e.g. the first is `log10(1 + (1 / (1 + len["title"]))` which is `log10(1 + (1 / (1 + 4))` but `1 / (1 + 4)` is zero in integer division so that's `log10(1 + 0)` which is just `0`.

The simplest way to fix this is to use `1.0 + len["title"]` since `1.0` is a `double` constant, and then everything gets calculated as floating point.

So if this test passes currently then there's something wrong somewhere!

I'd also tend to suggest stylistically that `log10(1 + 1 / (1 + len["title"])))` is a little clearer than `log10(1 + (1 / (1 + len["title"]))))` - you can reasonably assume that someone reading the code understands the relative precedence of addition/subtraction vs multiplication/division (it's the same in almost all programming languages, and matches the precedence in mathematics), but adding extra parentheses makes it harder for someone reading the code to mentally pair them up.

ojwb:
The CI run of `xapian-check-patch` is moaning about the following:

```
xapian-letor/tests/api_letor.cc:817: error: added/changed line has trailing whitespace: +    
xapian-letor/tests/api_letor.cc:821: error: Should have exactly one space after ',': +    TEST_EQUAL(whole_len,186);
xapian-letor/tests/api_letor.cc:836: error: Missing space between comment characters and comment text: +    //test for title
xapian-letor/tests/api_letor.cc:839: error: Missing space between comment characters and comment text: +    //test for body
xapian-letor/tests/api_letor.cc:842: error: Missing space between comment characters and comment text: +    //test for whole
```

You can run this locally (it's just a script in the source tree in `xapian-maintainer-tools/xapian-check-patch`) before pushing to get faster feedback on such code style issues.

xapian-letor/tests/api_letor.cc
VaibhavKansagara: I will try to fix that issue.

xapian-letor/tests/api_letor.cc
VaibhavKansagara: > So if this test passes currently then there's something wrong somewhere!
>
sorry, I forgot to set the query for the feature. I will fix that.
> The calculation of the expected answers here will use integer division
> 
I missed that, will fix it.

VaibhavKansagara:
> You can run this locally (it's just a script in the source tree in `xapian-maintainer-tools/xapian-check-patch`) before pushing to get faster feedback on such code style issues.
>
I used to run those scripts locally as you said for a brief time and it worked fine but due to some issue it stopped working but i have finally figured out the solution for the issue. thanks

xapian-letor/tests/api_letor.cc
ojwb: > sorry, I forgot to set the query for the feature. I will fix that.

Aha, that would explain it.

ojwb:
> I used to run those scripts locally as you said and it worked fine but due to some issue it stopped working but i have finally figured out the solution for the issue.

If that was an issue we can make changes to avoid or could usefully document then let us know.

xapian-letor/tests/api_letor.cc
```diff
@@ -804,3 +804,60 @@ DEFINE_TESTCASE(err_scorer, !backend)
...
+
+    vector<double> res = feature.get_values();
+    TEST_EQUAL(res.size(), 3);
+
+    // test for title
+    TEST_EQUAL(res[0], log10(1 + 1 / (1.0 + len["title"])));
+
+    // test for body
+    double temp = log10(1 + 6 / (1.0 + len["body"])) +
+			log10(1 + 2 / (1.0 + len["body"]));
```
ojwb: It would be clearer to align the `log10`s here - as written this reads as if it's `log10(... + log10(...))`.  I've just squashed that tweak on top of your changes and merged.

ojwb:
Squashed and rebased onto master as f2d1d29dae8203a25f5c155085363ba137f4ae97.

VaibhavKansagara:
> If that was an issue we can make changes to avoid or could usefully document then let us know.
>
I was making a stupid mistake but the instructions were pretty clear. thanks.

PR #227, opened 2019-04-06
delay libmagic checks

barufa:
Trying to solve (omindex: delay libmagic checks ([ticket #743](https://trac.xapian.org/ticket/743)))

Delaying the call to libmagic and the creation of the document.
There is also commented code related to ticket 632 that is not finished (Just to know if my idea was correct).

### DIFF
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -50,6 +50,7 @@
 #include "atomparse.h"
 #include "diritor.h"
 #include "failed.h"
+#include "hashterm.h"
 #include "md5wrap.h"
 #include "metaxmlparse.h"
 #include "mimemap.h"
@@ -69,7 +70,6 @@
 #include "xmlparse.h"
 #include "xlsxparse.h"
 #include "xpsxmlparse.h"
-#include "hashterm.h"
 
 using namespace std;
 
@@ -533,11 +533,14 @@ index_add_document(const string & urlterm, time_t last_altered,
 void
 index_mimetype(const string & file, const string & urlterm, const string & url,
 	       const string & ext,
-	       string &mimetype, DirectoryIterator &d,
-	       string & path_term,
+	       const string & mime_type,
+	       DirectoryIterator & d,
+	       const string & path_term,
 	       string record)
 {
     string context(file, root.size(), string::npos);
+    string mimetype = mime_type;
+    string pathterm = path_term;
 
     // FIXME: We could be cleverer here and check mtime too when use_ctime is
     // set - if the ctime has changed but the mtime is unchanged, we can just
@@ -573,7 +576,7 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	mimetype = d.get_magic_mimetype();
 	if (mimetype.empty()) {
 	    skip(urlterm, file.substr(root.size()),
-                 "Unknown extension and unrecognised format",
+		 "Unknown extension and unrecognised format",
 		 d.get_size(), d.get_mtime(), SKIP_SHOW_FILENAME);
 	    return;
 	}
@@ -588,13 +591,13 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
     // `/home/olly/public_html/foo/bar`.
     Xapian::Document newdocument;
     size_t j;
-    while ((j = path_term.rfind('/')) > 1 && j != string::npos) {
-	path_term.resize(j);
-	if (path_term.length() > MAX_SAFE_TERM_LENGTH) {
-            string term_hash = hash_long_term(path_term, MAX_SAFE_TERM_LENGTH);
+    while ((j = pathterm.rfind('/')) > 1 && j != string::npos) {
+	pathterm.resize(j);
+	if (pathterm.length() > MAX_SAFE_TERM_LENGTH) {
+	    string term_hash = hash_long_term(pathterm, MAX_SAFE_TERM_LENGTH);
 	    newdocument.add_boolean_term(term_hash);
 	} else {
-	    newdocument.add_boolean_term(path_term);
+	    newdocument.add_boolean_term(pathterm);
 	}
     }
 
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -533,7 +533,7 @@ index_add_document(const string & urlterm, time_t last_altered,
 void
 index_mimetype(const string & file, const string & urlterm, const string & url,
 	       const string & ext,
-	       const string &mimetype, DirectoryIterator &d,
+	       string &mimetype, DirectoryIterator &d,
 	       string & path_term,
 	       string record)
 {
@@ -567,7 +567,21 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	}
     }
 
-    if (verbose) cout << flush;
+    // If we didn't get the mime type from the extension, call libmagic to get
+    // it.
+    if (mimetype.empty()) {
+	mimetype = d.get_magic_mimetype();
+	if (mimetype.empty()) {
+	    skip(urlterm, file.substr(root.size()), "Unknown extension and unrecognised format",
+		 d.get_size(), d.get_mtime(), SKIP_SHOW_FILENAME);
+	    return;
+	}
+    }
+
+    if (verbose)
+	cout << "Indexing \"" << file.substr(root.size()) << "\" as "
+	     << mimetype << " ... " << flush;
+
     // Use `file` as the basis, as we don't want URL encoding in these terms,
     // but need to switch over the initial part so we get `/~olly/foo/bar` not
     // `/home/olly/public_html/foo/bar`.
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -533,14 +533,12 @@ index_add_document(const string & urlterm, time_t last_altered,
 void
 index_mimetype(const string & file, const string & urlterm, const string & url,
 	       const string & ext,
-	       const string & mime_type,
+	       string mimetype,
 	       DirectoryIterator & d,
-	       const string & path_term,
+	       string pathterm,
 	       string record)
 {
     string context(file, root.size(), string::npos);
-    string mimetype = mime_type;
-    string pathterm = path_term;
 
     // FIXME: We could be cleverer here and check mtime too when use_ctime is
     // set - if the ctime has changed but the mtime is unchanged, we can just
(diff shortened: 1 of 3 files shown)


xapian-applications/omega/index_file.cc
```diff
@@ -531,8 +531,7 @@ index_add_document(const string & urlterm, time_t last_altered,
 
 void
 index_mimetype(const string & file, const string & urlterm, const string & url,
-	       const string & ext,
-	       const string &mimetype, DirectoryIterator &d,
+	       const string & ext, string &mimetype, DirectoryIterator &d,
```
jaylett: This change is obscure, and it'd be clearer what you've done if you preserved the line breaks where they were. (As you did in the header file.)

xapian-applications/omega/index_file.cc
```diff
@@ -551,44 +552,45 @@ bool index_update_entry(const string & urlterm,DirectoryIterator &d){
...
+	       const string & ext, string &mimetype, string path_term,
+               DirectoryIterator &d, string record)
 {
     string context(file, root.size(), string::npos);
 
     time_t last_altered = use_ctime ? d.get_ctime() : d.get_mtime();
 
     Xapian::docid did = 0;
     if (index_check_existing(urlterm, last_altered, did))
-	return;
```
jaylett: This (and a lot of the later line changes) seem to be whitespace only — suggesting your editor has reformatted them for some reason. These changes probably shouldn't be part of this commit. (If the formatting is incorrect, a separate reformatting commit is much clearer.)

xapian-applications/omega/index_file.cc
```diff
@@ -551,44 +552,45 @@ bool index_update_entry(const string & urlterm,DirectoryIterator &d){
...
     }
 
+    // if the ctime has changed but the mtime is unchanged, we can just
+    // update the existing Document and avoid having to re-extract text, etc.
     if(use_ctime && d.get_ctime()!=d.get_mtime()){
-        // if the ctime has changed but the mtime is unchanged, we can just
-        // update the existing Document and avoid having to re-extract text, etc.
-        // if(index_update_entry(urlterm,d))
-            // return;
+        if(index_update_entry(urlterm,d))
```
jaylett: Again, this doesn't belong in this PR so hopefully you can pull it out.

xapian-applications/omega/index_file.cc
```diff
@@ -551,44 +552,45 @@ bool index_update_entry(const string & urlterm,DirectoryIterator &d){
...
+            return;
     }
 
+    // If we didn't get the mime type from the extension, call libmagic to get
```
jaylett: This comment should really have been in the other commit, I think.

xapian-applications/omega/index_file.cc
```diff
@@ -541,8 +542,8 @@ bool index_update_entry(const string & urlterm,DirectoryIterator &d){
             time_t doc_last_altered = binary_string_to_int(value);
             if(last_modified <= doc_last_altered && doc_last_altered <= last_altered){
                 if (verbose)
-        	    cout << "Updating entry" << endl;
-                // db.replace_document(did, doc);//UPDATE EXISTING DOCUMENT?
+        	    cout << "Updating entry without extract text" << endl;
+                db.replace_document(did, doc);//UPDATE EXISTING DOCUMENT
```
jaylett: It's not clear to me why this change is in the commit. If it's needed, then it'd be helpful for the commit message to explain why.

jaylett:
You shouldn't commit code related to another change (and committing commented-out code is if anything more confusing to other people!). If you're using a graphical git tool you should be able to select which lines to include in a commit, and avoid this kind of thing. If you're using the command line, then `git add -p` allows you to select which changes to include.

Also, remember that you can (and where helpful, should) add more detail into commit messages to explain why a change is desirable. (Although we aim for our ticket tracker to live a long time, git commit messages live effectively forever, and are often easier to search through!)

xapian-applications/omega/index_file.cc
barufa: I will try to avoid this kind of changes.

xapian-applications/omega/index_file.cc
barufa: I will try to avoid this kind of changes.

xapian-applications/omega/index_file.cc
barufa: You are right, I'm going to delete this modification and see if I can move it to another pull request related to ticket 632.

xapian-applications/omega/index_file.cc
barufa: It was my mistake. It is related to another issue.
I will try to make some changes.

barufa:
Try to modify the comits according to the required changes.
On my computer it passes all the make check tests.

ojwb:
The coding style checking script is reporting two problems:

```
xapian-applications/omega/index_file.cc:591: error: added/changed line uses spaces for indentation rather than tab: +                 "Unknown extension and unrecognised format",
xapian-applications/omega/index_file.cc:609: error: added/changed line uses spaces for indentation rather than tab: +            string term_hash = hash_long_term(path_term, MAX_SAFE_TERM_LENGTH);
```

You can run this script locally before pushing (it's `xapian-maintainer-tools/xapian-check-patch` in the tree).

xapian-applications/omega/index_file.cc
```diff
@@ -69,6 +69,7 @@
 #include "xmlparse.h"
 #include "xlsxparse.h"
 #include "xpsxmlparse.h"
+#include "hashterm.h"
```
ojwb: When adding something to a list that's already in sorted order, add it in the correct place.

xapian-applications/omega/index_file.cc
```diff
@@ -532,8 +533,8 @@ index_add_document(const string & urlterm, time_t last_altered,
 void
 index_mimetype(const string & file, const string & urlterm, const string & url,
 	       const string & ext,
-	       const string &mimetype, DirectoryIterator &d,
-	       Xapian::Document & newdocument,
+	       string &mimetype, DirectoryIterator &d,
```
ojwb: This function can now modify the passed `mimetype` value, but the only caller doesn't require that - it doesn't use `mimetype` at all after calling `index_mimetype()`.  Passing by non-`const` reference tends to lead to surprises so doing it when it isn't necessary is better avoided.

xapian-applications/omega/index_file.cc
```diff
@@ -532,8 +533,8 @@ index_add_document(const string & urlterm, time_t last_altered,
...
+	       string & path_term,
```
ojwb: Same goes for `path_term`.

xapian-applications/omega/index_file.cc
barufa: I had not noticed the alphabetical order, I will fix it.

xapian-applications/omega/index_file.cc
barufa: Okay, I understand that. I write it as 'non-const' because I need to make changes on them(these changes were previously made on index_file function), but if we use this function in another part of the code i could produce bugs. I will fix it.

xapian-applications/omega/index_file.cc
barufa: I will try to solve it in the same way.

xapian-applications/omega/index_file.cc
```diff
@@ -532,11 +533,14 @@ index_add_document(const string & urlterm, time_t last_altered,
...
-	       const string &mimetype, DirectoryIterator &d,
-	       Xapian::Document & newdocument,
+	       const string & mime_type,
+	       DirectoryIterator & d,
+	       const string & path_term,
 	       string record)
 {
     string context(file, root.size(), string::npos);
+    string mimetype = mime_type;
+    string pathterm = path_term;
```
ojwb: You can handle this more simply by making the parameter type `string` (like the existing `record` parameter).

That wouldn't be suitable for a public API as we'd need to keep the function signature compatible, and your approach here would be necessary.  But here we can just recompile the code which calls this.

ojwb:
In general don't force push branches that are under review.  It makes it much harder for reviewers to incrementally review, and they'll probably have to instead read through all the changes, which will tend to slow down getting your patch applied.

Sometimes it is the best thing to do, but I'd avoid it unless it's clear everyone involved agrees.

xapian-applications/omega/index_file.cc
barufa: Yes, you are right. I don't know why I do it that way. I will make that change.

xapian-applications/omega/index_file.cc
ojwb: Indeed.  Or if code is added after the existing call which uses the value of `mimetype`.

ojwb:
Thanks for the updates.

This looks OK to me now.  I'll just give @jaylett a chance to say if there's more he thinks needs doing (I think his comments have been addressed, but I'd rather he confirmed).

jaylett:
I'm happy with the changes, but the first commit message has a long first line rather than a short subject and a longer explanation in the body. We have [some guidance](https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages) on good commit messages which you should read and think about before your next PR. (You can rewrite commit messages after the fact, but once you push to a PR you get into the review cycle and you shouldn't rewrite them at that point.)

For this PR, it may be easier to rewrite as we merge. (Although it's probably @ojwb doing the merging, so he should weigh in on that!)

ojwb:
I can adjust the commit message before I merge.

The fixup commits here don't squash properly - it looks to me like you've targetted all the fixup commits onto the first commit, but some of them actually fix problems with the second commit.  In such a situation, you need to commit the changes which apply to each commit separately and specify the appropriate commit to the `commit --fixup` command for the fixups to actually be useful.

While the split between "delay document creation" and "delay libmagic checks" is good, I don't think it's worth the extra effort to try to rebuild two separate commits at this point, so I'll just squash everything into one.

But please try to make the fixup commits actually work next time.  You can test out the rebase without affecting your brach by creating a temporary branch that's a copy of your branch and testing on that:

```
git checkout -b temp-branch
git rebase -i --autosquash
```

And to discard:

```
git checkout master # or whatever your branch is called
git branch -D temp-branch # -D needed to delete a branch git thinks has unmerged changes
```

ojwb:
Squashed and merge to git master as d260b7d94c2f3e6e87e713d29038ef03a026b127.

When a ticket is being addressed the commit message should reference the ticket number for the benefit of people in the future trying to work out why a change was made (e.g. "Fixes #743" or if it's to do with a ticket but doesn't completely solve it "See #743" or similar).

Also, generally you want the commit message to focus more on WHY rather than WHAT.  It's not very useful to describe things that are basically paraphrasing what can just be seen from the patch, like "moving the creation of the document from omindex.cc to index_file.cc" does.  At best that's rather redundant and at worst the comment disagrees with what the patch shows and misleads or confuses the reader.

Or as https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages puts it: "Describe the effect, not the code."

PR #228, opened 2019-04-06
Trim HACKING and update README

jaylett:
A lot of the contents of xapian-core/HACKING has been moved into
the Xapian developer guide, so remove that (to avoid the two copies
getting out of sync) and add a pointer to the guide.

Mention both developer and user guide in README, since they're
probably the most useful places to get started now.


ojwb:
I assume all the deleted stuff is relocated (or no longer relevant), in which case this looks good to me.

If you're wanting me to check that we're still covering everything removed that's still relevant, that'll have to wait until after GSoC student applications close.

jaylett:
I'm pretty sure I haven't axed anything that hasn't made its way across, so I don't think that needs checking.

ojwb:
Sorry, i didn't mean to close this one.

ojwb:
Hmm, I seem to have again managed to close this without merging it.  Meanwhile I went through and removed the redundant stuff from HACKING separately.  Restoring the branch so I can apply the `README` changes, and see if I missed anything in my update.

ojwb:
Ah, I realised why I probably appear to keep closing this and deleting the branch - I bet it's the automatic sync from the xapian master git repo causing it.

I think we actually concluded this before, and that you shouldn't create github-only branches in the xapian repo, but instead work in your own clone for PRs.

Looking at the two HACKING updates, you seem to have removed the wrong section in one place - the release stuff is still there (but has been re-homed) while parts about tools needed for the build, etc is gone but doesn't seem to be in the dev guide yet.  I'll cherry-pick your commit onto current master and fix up that.

ojwb:
(That delete was indeed caused by me simply pushing changes to the git.xapian.org repo and them getting mirrored.)

jaylett:
I think this got more confused by two things: the branch deletion (I'll use the other approach in future) and my getting pulled away from this partway through (meaning I think there's a PR I intended to open which I didn't get round to). I have an entirely-consistent system on a machine thats graphics card seems to have failed, so it's going to be difficult for me to figure things out in the short term :-(

Removing the tools section was however intended — the tools description in the dev guide should be the same information compactly, with no options, to make things less confusing and avoid people getting partway through then things failing and having to ask for help because (inevitably) they don't read the entire guide. I think the only thing I kept separate was the PDF docs building.

PR #229, opened 2019-04-06
Update Vagrantfile to Ubuntu 1804

jaylett:
In the process, update to the latest package names for things like
PHP and Java. Also install the requirements for building PDFs of the
documentation, since it's a one-off cost and may make people's lives
easier (it certainly makes it easier to get people going to not have
to say "oh and if you want X then also install...", and matches the
approach taken in the developer guide more closely).


ojwb:
LGTM.  Nicer to be using a standard Ubuntu image too.

It'd be good to note "Reported by jugnu on #xapian" or similar in the commit message.

PR #231, opened 2019-04-14
Fix tostring method when a minimum representable value of a particular type is passed

VaibhavKansagara:
Currently, tostring method fails when a minimum lvalue of a particular signed type is passed. This PR intends to fix that.

### DIFF
--- a/xapian-core/common/str.cc
+++ b/xapian-core/common/str.cc
@@ -60,16 +60,23 @@ tostring(T value)
     if (value < 10 && value >= 0) return string(1, '0' + char(value));
 
     bool negative = (value < 0);
-    if (negative) value = -value;
 
-    char buf[(sizeof(T) * 5 + 1) / 2 + 1];
+    typedef typename std::make_unsigned<T>::type unsigned_type;
+    unsigned_type val;
+    if (negative) {
+	val = -value;
+    } else {
+	val = value;
+    }
+
+    char buf[(sizeof(unsigned_type) * 5 + 1) / 2 + 1];
     char * p = buf + sizeof(buf);
     do {
 	AssertRel(p,>,buf);
-	char ch = static_cast<char>(value % 10);
-	value /= 10;
+	char ch = static_cast<char>(val % 10);
+	val /= 10;
 	*(--p) = ch + '0';
-    } while (value);
+    } while (val);
 
     if (negative) {
 	AssertRel(p,>,buf);
--- a/xapian-core/common/str.cc
+++ b/xapian-core/common/str.cc
@@ -62,11 +62,9 @@ tostring(T value)
     bool negative = (value < 0);
 
     typedef typename std::make_unsigned<T>::type unsigned_type;
-    unsigned_type val;
+    unsigned_type val(value);
     if (negative) {
-	val = -value;
-    } else {
-	val = value;
+	val = -val;
     }
 
     char buf[(sizeof(unsigned_type) * 5 + 1) / 2 + 1];
--- a/xapian-core/tests/unittest.cc
+++ b/xapian-core/tests/unittest.cc
@@ -536,10 +536,22 @@ static bool test_tostring1()
     TEST_EQUAL(str(-1), "-1");
     TEST_EQUAL(str(-9), "-9");
     TEST_EQUAL(str(-10), "-10");
+    TEST_EQUAL(str(0x7f), "127");
+    TEST_EQUAL(str(-0x80), "-128");
+    TEST_EQUAL(str(0x7fff), "32767");
+    // test for lvalue with minimum value for short type.
+    short temp = -32768;
+    TEST_EQUAL(str(temp), "-32768");
+
     TEST_EQUAL(str(0xffffffff), "4294967295");
     TEST_EQUAL(str(0x7fffffff), "2147483647");
     TEST_EQUAL(str(0x7fffffffu), "2147483647");
-    TEST_EQUAL(str(-0x7fffffff), "-2147483647");
+
+    // test for lvalue with minimum value for int type.
+    int temp1 = -2147483648;
+    TEST_EQUAL(str(temp1), "-2147483648");
+
+    TEST_EQUAL(str(0x7fffffffffffffff), "9223372036854775807");
 
 #ifdef __WIN32__
     /* Test the 64 bit integer conversion to string.
--- a/xapian-core/tests/unittest.cc
+++ b/xapian-core/tests/unittest.cc
@@ -539,14 +539,14 @@ static bool test_tostring1()
     TEST_EQUAL(str(0x7f), "127");
     TEST_EQUAL(str(-0x80), "-128");
     TEST_EQUAL(str(0x7fff), "32767");
-    TEST_EQUAL(str(short(-32768)), "-32768");
+    TEST_EQUAL(str(short(-0x8000)), "-32768");
 
     TEST_EQUAL(str(0xffffffff), "4294967295");
     TEST_EQUAL(str(0x7fffffff), "2147483647");
     TEST_EQUAL(str(0x7fffffffu), "2147483647");
     TEST_EQUAL(str(-0x7fffffff), "-2147483647");
 
-    TEST_EQUAL(str(int(-2147483648)), "-2147483648");
+    TEST_EQUAL(str(int(-0x80000000)), "-2147483648");
 
     TEST_EQUAL(str(0x7fffffffffffffff), "9223372036854775807");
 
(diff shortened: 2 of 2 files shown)


xapian-core/tests/unittest.cc
```diff
@@ -536,10 +536,22 @@ static bool test_tostring1()
     TEST_EQUAL(str(-1), "-1");
     TEST_EQUAL(str(-9), "-9");
     TEST_EQUAL(str(-10), "-10");
+    TEST_EQUAL(str(0x7f), "127");
+    TEST_EQUAL(str(-0x80), "-128");
+    TEST_EQUAL(str(0x7fff), "32767");
+    // test for lvalue with minimum value for short type.
+    short temp = -32768;
+    TEST_EQUAL(str(temp), "-32768");
```
ojwb: This isn't necessarily the minimum value for `short` - the minimum must be <= -32767, but `short` could be wider (an implementation where sizeof(short) > 2 is allowed by the standard, and there have been such implementations).  All modern systems seem to use twos' complement, but on anything that doesn't -32768 may not be representable.

The minimum value for `short` is available using `std::numeric_limits<short>::min()` (which is also self-documenting).  Similarly for other standard types.

xapian-core/tests/unittest.cc
```diff
@@ -536,10 +536,22 @@ static bool test_tostring1()
...
+
     TEST_EQUAL(str(0xffffffff), "4294967295");
     TEST_EQUAL(str(0x7fffffff), "2147483647");
     TEST_EQUAL(str(0x7fffffffu), "2147483647");
-    TEST_EQUAL(str(-0x7fffffff), "-2147483647");
+
+    // test for lvalue with minimum value for int type.
+    int temp1 = -2147483648;
+    TEST_EQUAL(str(temp1), "-2147483648");
```
ojwb: I don't think it's the lvalue-ness that's the issue here - rather a slightly surprising behaviour of the types of integer constants - consider:

```
olly@gemse:~$ cat constant-type-probes.cc 
#include <iostream>
using namespace std;

#define DEF_F(T) void f(T v) { cout << #T "(" << v << ")\n"; }

DEF_F(short)
DEF_F(int)
DEF_F(long)
DEF_F(long long)

int main() {
    f(32767);
    f(-32767);
    f(-32768);
    f(2147483647);
    f(-2147483647);
    f(-2147483648);
    f(int(-2147483648));
}
olly@gemse:~$ g++ -Wall -W -O2 constant-type-probes.cc
olly@gemse:~$ ./a.out
int(32767)
int(-32767)
int(-32768)
int(2147483647)
int(-2147483647)
long(-2147483648)
int(-2147483648)
```

The slightly odd thing here is that `-2147483648` is of type `long` when the value actually fits in an `int`.  I think this is because it's actually parsed as `-(2147483648)` and `2147483648` doesn't fit in an `int`, so has type `long`.  Then that's negated, and the negation preserves the type.

But rather than using temporary variables, we can just write such testcases with an explicit conversion to the type we want to test, e.g.:

```
TEST_EQUAL(str(int(-2147483648)), "-2147483648");
```

xapian-core/common/str.cc
```diff
@@ -60,16 +60,23 @@ tostring(T value)
     if (value < 10 && value >= 0) return string(1, '0' + char(value));
 
     bool negative = (value < 0);
-    if (negative) value = -value;
 
-    char buf[(sizeof(T) * 5 + 1) / 2 + 1];
+    typedef typename std::make_unsigned<T>::type unsigned_type;
+    unsigned_type val;
+    if (negative) {
+	val = -value;
```
ojwb: If passed `int(-2147483648)` this will still negate `-2147483648` as an `int` (and only then convert the result to `unsigned int`).

Running `unittest` when in a build using ubsan I see:

```
Running test: tostring1...../common/str.cc:67:8: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../common/str.cc:67:8 in 
```

I think the message there is suggesting doing this, which indeed seems to work:

```
    unsigned_type val(value);
    if (negative) {
       val = -val;
    }
```

https://en.cppreference.com/w/cpp/language/operator_arithmetic#Unary_arithmetic_operators explains the semantics of unary minus on an unsigned type.

xapian-core/tests/unittest.cc
```diff
@@ -536,10 +536,22 @@ static bool test_tostring1()
...
+    TEST_EQUAL(str(-0x80), "-128");
+    TEST_EQUAL(str(0x7fff), "32767");
+    // test for lvalue with minimum value for short type.
+    short temp = -32768;
+    TEST_EQUAL(str(temp), "-32768");
+
     TEST_EQUAL(str(0xffffffff), "4294967295");
     TEST_EQUAL(str(0x7fffffff), "2147483647");
     TEST_EQUAL(str(0x7fffffffu), "2147483647");
-    TEST_EQUAL(str(-0x7fffffff), "-2147483647");
```
ojwb: This removes an existing testcase - I think it would be better to just add new testcases since each `str()` call is very quick.

xapian-core/tests/unittest.cc
VaibhavKansagara: oh you mean the comment?.

ojwb: The minimum value that a `short` can represent is a compile-time constant, but the value of that constant is not the same for every compiler.  And somewhat unhelpfully for us here the standard allows an implementation where `-32768` can't be represented as a `short`.  In practice I suspect there may already be some twos' complement assumptions in Xapian, but I feel it's a bad idea to knowingly add more such assumptions.

Taking a step back, fundamentally what we want to be adding testing for here is the edge cases where the input value is the maximum and minimum values of each integer type.  The `parseunsigned1` testcase I added recently does that quite nicely for `parse_unsigned()` - we just instantiate a template function for each type we want to test and then call it.

If we try to follow that model here, the only tricky bit seems to be generating the expected value to test against - we can't write out the expected answers like we currently do, and converting the result of `str()` back to an integer means we're testing each function in terms of the other and also we wouldn't spot some potential bugs - e.g. if `str(0)` -> `00` that would convert back to integer `0`.  But we can generate the expected strings with e.g. `std::ostringstream` - that's best avoided in production code as it is rather slow, but here can serve as our "ground truth" for converting integers to strings.

VaibhavKansagara: thanks will change it.

xapian-core/tests/unittest.cc
```diff
@@ -536,11 +536,30 @@ static bool test_tostring1()
...
     TEST_EQUAL(str(-0x7fffffff), "-2147483647");
 
+    const int min_int = std::numeric_limits<int>::min();
+    oss << min_int;
+    TEST_EQUAL(str(min_int), oss.str());
+    oss.str("");
+    oss.clear();
+
+    TEST_EQUAL(str(0x7fffffffffffffff), "9223372036854775807");
+
```
ojwb: On IRC you asked:
> are you happy with the changes i made in my PR. I know we shouldn't use temporary variables but it would get messy if i don't.

My suggestion was to wrap this up in a template function which tests the min and max values for a given type - then the testcase itself can instantiate and call that for each type we want to test - much like the existing testcase here:

```
static bool test_parseunsigned1()
{
    parseunsigned_helper<unsigned char>();
    parseunsigned_helper<unsigned short>();
    parseunsigned_helper<unsigned>();
    parseunsigned_helper<unsigned long>();
    parseunsigned_helper<unsigned long long>();

    return true;
}
```

Then the gory details of the temporary variable and the use of `ostringstream` are encapsulated.

> Also what do you think of using std::to_string instead of ostringstream for testing.? It would result in a much cleaner code and wouldn't need clearing the buffer everytime we use ostringstream.

As I said on IRC `std::to_string()` isn't currently portable enough.  `HACKING` has more details:

> `std::to_string()` - this is completely missing on current versions of mingw and cygwin - in the library, you can `#include "str.h"` and then use the `str()` function instead for most cases.  This is also usually faster than `std::to_string()`.

Of course here we can't use `str()` because we need something to compare the results of `str()` with.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52015 seems to be the GCC bug about this.

VaibhavKansagara: fixed in  a fixup commit.

ojwb:
Thanks, squashed and pushed to master as ff11c68a4981f98f1095a0c19318e020f73390c3.

PR #232, opened 2019-04-20
Wrap C++ iterators better for Ruby

tson1111:
Complete the bite size project ["Wrap C++ iterators better for Ruby"](https://trac.xapian.org/wiki/ProjectIdeas#BiteSize). Supporting block-driven iteration for Ruby would be nicer and more efficient than turning iterators into arrays of values.


ojwb:
This causes widespread testsuite failures for the Ruby bindings.

It looks to me like you have a plausible implementation of Ruby block-driven iterators, but you've simply replaced the existing iterators with this.  I don't think we can do that as it means any existing Ruby code using iterators will no longer work (as all the test failures demonstrate).  Either the block-driven part needs to only activate when used in that context, or we need to provide the block-driven versions via separate methods to the existing iterators.

You really should make sure the test suite passes locally (with `make check`) before opening a PR.  The CI checks builds on more platforms and in some special configurations, so it's possible to have the CI fail when the build was clean locally, but running the tests locally should have caught this.

ojwb:
I've just pushed 4b6a37cd71cce8c5acecb716bec0f6c511947399 to git master which supports block iteration when a block is given, and returns an Array otherwise so existing code will continue to work.

PR #233, opened 2019-04-20
 #771 Handling iWork directory as a single document 

tejasvi:
So far most straightforward fix. Leafnames are checked against iWork 'extensions' and d.get_type() returns as files on matching.

### DIFF
--- a/xapian-applications/omega/daflist.h
+++ b/xapian-applications/omega/daflist.h
@@ -1,0 +1,8 @@
+#include <string>
+#include <vector>
+
+std::vector<std::string> dasfile { 
+".key", 
+".numbers", 
+".pages" 
+};
--- a/xapian-applications/omega/daflist.h
+++ b/xapian-applications/omega/daflist.h
@@ -1,3 +1,7 @@
+/** @file daflist.h
+ * @brief 'Extensions' of directories to be indexed as files.
+ */
+
 #include <string>
 #include <vector>
 
--- a/xapian-applications/omega/daflist.h
+++ b/xapian-applications/omega/daflist.h
@@ -1,8 +1,4 @@
 #include <string>
 #include <vector>
 
-std::vector<std::string> dasfile { 
-".key", 
-".numbers", 
-".pages" 
-};
+std::vector<std::string> dasfile {".key", ".numbers", ".pages"};
--- a/xapian-applications/omega/diritor.h
+++ b/xapian-applications/omega/diritor.h
@@ -148,7 +148,7 @@ class DirectoryIterator {
 	    case DT_DIR: {
 			char * dot_ptr = strrchr(entry->d_name, '.');
 			std::vector<std::string> dasfile {".key", ".numbers", ".pages", "\0"};
-			if(std::find(dasfile.begin(), dasfile.end(), dot_ptr) != array.end())
+			if(std::find(dasfile.begin(), dasfile.end(), dot_ptr) != dasfile.end())
 				return REGULAR_FILE;
 			else return DIRECTORY;
 		}
--- a/xapian-applications/omega/diritor.h
+++ b/xapian-applications/omega/diritor.h
@@ -145,8 +145,13 @@ class DirectoryIterator {
 		break;
 	    case DT_REG:
 		return REGULAR_FILE;
-	    case DT_DIR:
-		return DIRECTORY;
+	    case DT_DIR: {
+			char * dot_ptr = strrchr(entry->d_name, '.');
+			std::vector<std::string> dasfile {".key", ".numbers", ".pages", "\0"};
+			if(std::find(dasfile.begin(), dasfile.end(), dot_ptr) != array.end())
+				return REGULAR_FILE;
+			else return DIRECTORY;
+		}
 #ifdef HAVE_LSTAT
 	    case DT_LNK:
 		if (follow_symlinks) break;
--- a/xapian-applications/omega/diritor.h
+++ b/xapian-applications/omega/diritor.h
@@ -145,13 +145,8 @@ class DirectoryIterator {
 		break;
 	    case DT_REG:
 		return REGULAR_FILE;
-	    case DT_DIR: {
-			char * dot_ptr = strrchr(entry->d_name, '.');
-			std::vector<std::string> dasfile {".key", ".numbers", ".pages", "\0"};
-			if(std::find(dasfile.begin(), dasfile.end(), dot_ptr) != dasfile.end())
-				return REGULAR_FILE;
-			else return DIRECTORY;
-		}
+	    case DT_DIR:
+		return DIRECTORY;
 #ifdef HAVE_LSTAT
 	    case DT_LNK:
 		if (follow_symlinks) break;
--- a/xapian-applications/omega/omindex.cc
+++ b/xapian-applications/omega/omindex.cc
@@ -196,7 +196,7 @@ index_directory(const string &path, const string &url_, size_t depth_limit,
 			const char * pt = strrchr(d.leafname(), '.');
 			if (pt)
 			if (binary_search(dasfile.begin(), dasfile.end(), pt)) {
-            	index_file(file, url, d, mime_map);
+            			index_file(file, url, d, mime_map);
 				break;
 			}
 			size_t new_limit = depth_limit;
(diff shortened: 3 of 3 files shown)


ojwb:
The CI builds are failing, e.g. with:

```
In file included from diritor.cc:23:0:
diritor.h: In member function ‘DirectoryIterator::type DirectoryIterator::get_type()’:
diritor.h:149:47: error: ‘strrchr’ was not declared in this scope
    char * dot_ptr = strrchr(entry->d_name, '.');
                                               ^
diritor.h:150:4: error: ‘vector’ is not a member of ‘std’
    std::vector<std::string> dasfile {".key", ".numbers", ".pages", "\0"};
    ^
diritor.h:150:27: error: expected primary-expression before ‘>’ token
    std::vector<std::string> dasfile {".key", ".numbers", ".pages", "\0"};
                           ^
diritor.h:150:29: error: ‘dasfile’ was not declared in this scope
    std::vector<std::string> dasfile {".key", ".numbers", ".pages", "\0"};
                             ^
diritor.h:151:7: error: ‘find’ is not a member of ‘std’
    if(std::find(dasfile.begin(), dasfile.end(), dot_ptr) != dasfile.end())
       ^
```

Most of that looks like you aren't including the necessary standard C++ library headers.  Some compilers implicitly include some headers from other headers, so you may have managed to build this locally, but for portability you should explicitly include the headers for things you use.

xapian-applications/omega/diritor.h
```diff
@@ -145,8 +145,13 @@ class DirectoryIterator {
...
 		return REGULAR_FILE;
-	    case DT_DIR:
-		return DIRECTORY;
+	    case DT_DIR: {
+			char * dot_ptr = strrchr(entry->d_name, '.');
+			std::vector<std::string> dasfile {".key", ".numbers", ".pages", "\0"};
+			if(std::find(dasfile.begin(), dasfile.end(), dot_ptr) != dasfile.end())
+				return REGULAR_FILE;
+			else return DIRECTORY;
+		}
```
ojwb: The indentation here doesn't match the surrounding code.

`if(...` doesn't match the rest of the code (Xapian coding style is to write `if (` with a space between).

It's cleaner to have the statement after the `else` on its own line when the statement after `if` is.

If a directory name doesn't contain a `.` (which is probably the most common case) then `dot_ptr` will be `NULL` and you'll end up implicitly trying to convert that to `std::string`, which I think is undefined behaviour.

Why the `"\0"` entry in `dasfile`?  You're including the `.` in the string searched, so I can't see how that can ever match.  A filename can't contain a zero byte so even if the `.` weren't included I can't see how it could validly match.

This also hard-codes the extensions which get this special handling, which seems undesirable.  Really this should be configurable in a similar way to how indexing of files is configured.

The `vector` here will get created each time a directory is encountered, which is a bit inefficient.  It also requires a linear search, which is OK when it's hard-coded and has 3 (or 4) entries, but if this were actually configurable it would scale unnecessarily poorly as the number of such extensions grew large.

This code won't kick in for filing systems which set `entry->d_type` to `DT_UNKNOWN`.  There we need to call `stat()` to find out if this is a directory or a file or whatever.  The most commonly used Linux filing systems set `d_type` usefully (and that means we can sometimes avoid the call to `stat()` which is a small optimisation) but we can't rely on that.

I'm a bit unsure about the approach of faking these directories as `REGULAR_FILE` at this level.  The `DirectoryIterator` class is conceptually a mostly generic class for iterating a directory, so it seems a bad fit for it to have to care about what we're doing with the contents of the directory - that your patch hard codes a list of extensions into it is an indication of that.  That knowledge really belongs at a higher level where similar knowledge about files already is.

So to me it seems more natural to patch this handling in at the point where `DirectoryIterator::DIRECTORY` is handled.  As a bonus, there's only one place that consumes `DirectoryIterator::DIRECTORY` but more than one place where that's returned (so a complete version of your approach would need to add the same handling to multiple places).

xapian-applications/omega/omindex.cc
```diff
@@ -193,7 +193,7 @@ index_directory(const string &path, const string &url_, size_t depth_limit,
 	    try {
 		switch (d.get_type()) {
 		    case DirectoryIterator::DIRECTORY: {
-			char * dot_ptr = strrchr(d.leafname, '.');	
+			char * dot_ptr = strrchr(d.leafname(), '.');
```
ojwb: You need to check your code compiles locally at the very, very minimum before pushing changes (if calling `strrchr()` on a function pointer compiled locally, there's something very wrong with your compiler).  You should really be checking that the appropriate test suite passes too, since a patch which causes tests to fail can't be merged.

There's little point us reviewing code when it won't even compile - you're wasting our limited patch reviewing time.  Also every change you push to a PR against Xapian currently triggers 12 different builds, which are all on donated VMs - by triggering pointless builds like these you delay other useful builds.

tejasvi: Oh, I really apologize. I didn't released it. Since my my laptop isn't repaired yet, I don't have a dedicated system to work on. I thought to save the work on repo itself each time. I'll use some other way now on.

tejasvi: Turns out I can deploy Travis CI and AppVeyor in my repo with few clicks :)

xapian-applications/omega/daflist.h
```diff
@@ -0,0 +1,8 @@
+/** @file daflist.h
+ * @brief 'Extensions' of directories to be indexed as files.
```
ojwb: > `Extensions`

Why the quotes?

xapian-applications/omega/daflist.h
```diff
@@ -0,0 +1,8 @@
...
+ */
+
+#include <string>
+#include <vector>
+
+std::vector<std::string> dasfile {".key", ".numbers", ".pages"};
```
ojwb: You're still hard-coding a list of extensions in the code.  As I said before this should be configurable in a similar way to how indexing of files is configured.  The user can then configure their own custom filters for these special directories.

For example, to handle `.xcodeproj` directories on macOS they could write a script `xcodeproj2txt` which takes a path to such a directory and extracts text from some files inside, and then run omindex with options `--mime-type=xcodeproj/:application/x-apple-xcode-project --filter=application/x-apple-xcode-project:xcodeproj2txt`

xapian-applications/omega/omindex.cc
```diff
@@ -192,6 +193,12 @@ index_directory(const string &path, const string &url_, size_t depth_limit,
 	    try {
 		switch (d.get_type()) {
 		    case DirectoryIterator::DIRECTORY: {
+			const char * pt = strrchr(d.leafname(), '.');
+			if (pt)
+			if (binary_search(dasfile.begin(), dasfile.end(), pt)) {
```
ojwb: Indentation error.

xapian-applications/omega/omindex.cc
```diff
@@ -768,6 +775,8 @@ main(int argc, char **argv)
 	url_encode_path(baseurl, start_url);
     }
 
+	sort(dasfile.begin(), dasfile.end());
```
ojwb: Indentation error.

xapian-applications/omega/daflist.h
tejasvi: I always thought _extension_ is meant for files only. Though directories are files in linux.

xapian-applications/omega/omindex.cc
ojwb: That avoids us getting an email about code that isn't useful to review, but it still means running 12 builds on the same donated infrastructure, and that means less build capacity is available to others using travis and appveyor.

It's really not sensible to use CI as a replacement for having your own a machine to build on.

xapian-applications/omega/daflist.h
ojwb: There's not really a formal standard definition of "extension", but I think it's reasonable to think of it as part of a pathname whatever that pathname refers to.

E.g. https://en.wikipedia.org/wiki/Bundle_(macOS) says:
> Application bundles are directory hierarchies, with the top-level directory having a name that ends with a `.app` extension.

The use of extensions on the names of directories seems to be a mac thing - at least I can't think of examples that originate on other platforms.  That's certainly where `.key`, etc originate.

ojwb:
Thanks for your work on this.  However this PR seems to be unfinished and hasn't been worked on for 10 months so it looks like it has been abandoned.

Therefore I'm going to close it.  If you want to continue with it, please just click "reopen".

PR #234, opened 2019-04-20
Fix re-extracting text without modifications

barufa:
If ominex is using ctime to check modifications on files, it is possible 
to update the document without re-extracting its text (using mtime).
Trying to solve FIXME comment a line 558 in index_file.cc

### DIFF
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -545,6 +545,77 @@ index_add_document(const string & urlterm, time_t last_altered,
     }
 }
 
+bool
+index_update_entry(const string & urlterm,
+		   DirectoryIterator & d,
+		   Xapian::docid & did)
+{
+    if (!did) {
+	Xapian::PostingIterator p = db.postlist_begin(urlterm);
+	if (p != db.postlist_end(urlterm))
+	    did = *p;
+    }
+
+    if (did) {
+	time_t mtime = d.get_mtime();
+	time_t ctime = d.get_ctime();
+	Xapian::Document doc = db.get_document(did);
+	string value_ctime = doc.get_value(VALUE_CTIME);
+	time_t doc_ctime = binary_string_to_int(value_ctime);
+	string value_mtime = doc.get_value(VALUE_LASTMOD);
+	time_t doc_mtime = binary_string_to_int(value_mtime);
+
+	if (doc_mtime == mtime && doc_ctime <= ctime) {
+	    if (verbose)
+		cout << "Updating entry without re-extracting text" << endl;
+	    // Remove terms from the document
+	    Xapian::TermIterator term_iterator = doc.termlist_begin();
+	    while (term_iterator != doc.termlist_end()) {
+		const string& term = *term_iterator;
+		bool rem_term = startswith(term, "I") ||
+				startswith(term, "O");
+		// I: boolean filter term for "can see" permission
+		// O: Owner
+		if (rem_term) {
+		    doc.remove_term(term);
+		}
+		term_iterator++;
+	    }
+	    // Updating document terms
+	    bool inc_tag_added = false;
+	    if (d.is_other_readable()) {
+		inc_tag_added = true;
+		doc.add_boolean_term("I*");
+	    } else if (d.is_group_readable()) {
+		const char * group = d.get_group();
+		if (group) {
+		    doc.add_boolean_term(string("I#") + group);
+		}
+	    }
+
+	    const char * owner = d.get_owner();
+	    if (owner) {
+		doc.add_boolean_term(string("O") + owner);
+		if (!inc_tag_added && d.is_owner_readable())
+		    doc.add_boolean_term(string("I@") + owner);
+	    }
+
+	    // Update document mtime and ctime
+	    doc.add_value(VALUE_LASTMOD,
+			  int_to_binary_string(uint32_t(mtime)));
+	    doc.add_value(VALUE_CTIME,
+			  int_to_binary_string(uint32_t(ctime)));
+
+	    db.replace_document(did, doc);
+	    mark_as_seen(did);
+	    return true;
+	}
+    }
+
+    return false;
+}
+
+
 void
 index_mimetype(const string & file, const string & urlterm, const string & url,
 	       const string & ext,
@@ -555,9 +626,6 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 {
     string context(file, root.size(), string::npos);
 
-    // FIXME: We could be cleverer here and check mtime too when use_ctime is
-    // set - if the ctime has changed but the mtime is unchanged, we can just
-    // update the existing Document and avoid having to re-extract text, etc.
     time_t last_altered = use_ctime ? d.get_ctime() : d.get_mtime();
 
     Xapian::docid did = 0;
@@ -581,6 +649,13 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	    // again on this attempt, we'll add a new one.
 	    failed.del(urlterm);
 	}
+    }
+
+    // if the ctime has changed but the mtime is unchanged, we can just
+    // update the existing Document and avoid having to re-extract text, etc.
+    if (use_ctime && d.get_ctime() != d.get_mtime()) {
+	if (index_update_entry(urlterm, d, did))
+	    return;
     }
 
     // If we didn't get the mime type from the extension, call libmagic to get
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -597,7 +597,7 @@ index_update_entry(const string& urlterm,
     time_t lower_ctime = binary_string_to_int(value_lower_ctime);
     time_t upper_mtime = binary_string_to_int(value_upper_mtime);
 
-    if(upper_mtime < mtime || ctime < lower_ctime)
+    if (upper_mtime < mtime || ctime < lower_ctime)
 	return false;
 
     Xapian::Document doc = db.get_document(did);
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -546,7 +546,7 @@ index_add_document(const string & urlterm, time_t last_altered,
 }
 
 static void
-index_ctime_terms(Xapian::Document & doc, DirectoryIterator & d,
+index_ctime_terms(Xapian::Document& doc, DirectoryIterator& d,
 		  time_t mtime, time_t ctime)
 {
     bool inc_tag_added = false;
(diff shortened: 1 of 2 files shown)


barufa:
This code has not been tested. Work in progress.

xapian-applications/omega/index_file.cc
```diff
@@ -545,6 +545,77 @@ index_add_document(const string & urlterm, time_t last_altered,
...
+		   DirectoryIterator & d,
+		   Xapian::docid & did)
+{
+    if (!did) {
+	Xapian::PostingIterator p = db.postlist_begin(urlterm);
+	if (p != db.postlist_end(urlterm))
+	    did = *p;
+    }
+
+    if (did) {
```
ojwb: I would suggest dealing with the "no docid" case returning `false` here - then you can avoid having the whole of the actual implementation indented by an extra level.

xapian-applications/omega/index_file.cc
```diff
@@ -545,6 +545,77 @@ index_add_document(const string & urlterm, time_t last_altered,
...
+	time_t mtime = d.get_mtime();
+	time_t ctime = d.get_ctime();
+	Xapian::Document doc = db.get_document(did);
+	string value_ctime = doc.get_value(VALUE_CTIME);
+	time_t doc_ctime = binary_string_to_int(value_ctime);
+	string value_mtime = doc.get_value(VALUE_LASTMOD);
+	time_t doc_mtime = binary_string_to_int(value_mtime);
+
+	if (doc_mtime == mtime && doc_ctime <= ctime) {
```
ojwb: While this looks correct to me, we can potentially optimise this check to avoid having to look up the values (or to only have to look up one in some cases).  This function potentially gets called for a lot of documents, so if we can eliminate many of them with a cheap check that's good.

We know upper and lower bounds on the values in a given slot, which translates to having bounds on the values of mtime that occur in the database (and similarly for ctime).  So for example we can cheaply check if the file's `mtime` is after any in the database, in which case we know that `doc_mtime == mtime` can't be true without having to even call `get_document()`.  Similarly we may also be able to tell that `doc_ctime <= ctime` must be true just from comparing `ctime` with the upper bound on ctime in the database.

xapian-applications/omega/index_file.cc
```diff
@@ -545,6 +545,77 @@ index_add_document(const string & urlterm, time_t last_altered,
...
+		const string& term = *term_iterator;
+		bool rem_term = startswith(term, "I") ||
+				startswith(term, "O");
+		// I: boolean filter term for "can see" permission
+		// O: Owner
+		if (rem_term) {
+		    doc.remove_term(term);
+		}
+		term_iterator++;
+	    }
```
ojwb: Again this looks correct, but there's scope to optimise as we can avoid having to iterate all the terms.  skip_to("I") will advance us to the first term >= "I", then we can `skip_to("O")` once we've handled all the `I` terms, then we can stop once we've handled all the `O` terms.

ojwb:
BTW don't worry about the current Appveyor CI failures - those are due to a bug in a recent change in master, which is already fixed (by 6d8ce77340078f688458ef0d89f08c1c52709f78).

xapian-applications/omega/index_file.cc
```diff
@@ -545,6 +545,77 @@ index_add_document(const string & urlterm, time_t last_altered,
...
+		    doc.add_boolean_term(string("I#") + group);
+		}
+	    }
+
+	    const char * owner = d.get_owner();
+	    if (owner) {
+		doc.add_boolean_term(string("O") + owner);
+		if (!inc_tag_added && d.is_owner_readable())
+		    doc.add_boolean_term(string("I@") + owner);
+	    }
```
ojwb: This looks like a cut-and-paste of ~18 lines of existing code from `index_mimetype()` - probably better to split this out into a new function and call it from here and `index_mimetype()`.

xapian-applications/omega/index_file.cc
barufa: Yes, you are are right. I didn't realize of that. I will change it

xapian-applications/omega/index_file.cc
barufa: Okay, I will try to use it

xapian-applications/omega/index_file.cc
barufa: I was not sure about creating a new function. I'm going to add it.

xapian-applications/omega/index_file.cc
barufa: I am going to make a little research about it and try to implement it.

barufa:
I think I found a **problem** with this modifications.
Suppose we have only two files in the database (A.txt and B.txt) with different contents and we are working on ubuntu. If I execute the command `mv A.txt B.txt`, it will changes the content of B.txt but the mtime is the same(mv only change the ctime). 
So omindex will remove A.txt and update B.txt without re-extracting its text, and this would be inconsistent.

ojwb:
The `mtime` of the new `A.txt` will be the `mtime` of `B.txt`, so there are really 3 cases before the `mv` to consider:

* `mtime("B.txt") > mtime("A.txt")`: in this case it just looks to omindex like the file's mtime got newer, which it will handle as if the file was updated, which is exactly what we want (unless we hope to detect renames which is probably a whole new can of worms)

* `mtime("B.txt") < mtime("A.txt")`: if the ctime got newer but the mtime got older then we can probably deduce that we need to reindex the file contents (it's always safe from a correctness point of view to do that, and doing this doesn't seem to mean unnecessary reindexing in any case I can think of)

* `mtime("B.txt") == mtime("A.txt")`: this is the tricky case as we can't tell it from a meta-data change.  It might seem like it would be rare, but `st_mtime` has a one second resolution, so any process which creates a lot of files within the same second potentially sets us up for this to happen.  A fairly cheap check which would help in many cases is the file size - if that changes, we need to reindex.

  The inode value (`st_ino` in `struct stat`) could potentially be stored and compared, but not all filing systems have inode values with suitable properties - e.g. if you index files on a CIFS mount then in some configurations the reported inode values are randomly created by the client and not persistent.  It's also only unique within a mount, though that might be OK here since you probably can't rename a file in this problematic way across mounts.  To get a globally unique file ID we'd have to qualify the inode number with `st_dev` field but I suspect the later is not necessarily preserved across an unmount and remount for all filing systems.  Overall this seems doesn't seem like it's a good general purpose solution.

  We could (as you or maybe someone else suggested on IRC) check the MD5 checksum - if we go on to index the file successfully then we'll want this anyway as we store it to allow collapsing duplicate files in search results, and if we check the mtime and file size are the same before calculating the checksum that should significantly cut down the number of cases where we calculate the checksum only to discover the files are the same.

xapian-applications/omega/index_file.cc
```diff
@@ -545,6 +545,97 @@ index_add_document(const string & urlterm, time_t last_altered,
     }
 }
 
+void
+index_ctime_terms(Xapian::Document & doc, DirectoryIterator & d,
+		  const time_t & mtime, const time_t & ctime)
```
ojwb: `time_t` is just a signed integer type (at least on POSIX systems; if memory serves ISO C allows `double` but I don't think anything actually uses that).  There's no reason to pass it by const reference - just pass it by value instead.  We use `const reference` when copying is potentially expensive - e.g. `std::vector<std::string>` is O(n) in the number of entries to pass by value, but by const reference it's O(1).

xapian-applications/omega/index_file.h
```diff
@@ -126,6 +126,16 @@ void
 index_add_document(const std::string & urlterm, time_t last_altered,
 		   Xapian::docid did, const Xapian::Document & doc);
 
+void
+index_ctime_terms(Xapian::Document & doc, DirectoryIterator & d,
+		  const time_t & mtime, const time_t & ctime);
```
ojwb: This new function is only used in `index_file.cc`.  Considering what it does, I think it makes more sense to treat it as an internal implementation detail (so add `static` to the definition in the `.cc`, and don't declare it here in the header).

xapian-applications/omega/index_file.h
```diff
@@ -126,6 +126,16 @@ void
...
+
+/// Update a document without re-extracting text
+bool
+index_update_entry(const std::string & urlterm,
+		   DirectoryIterator & d,
+		   Xapian::docid & did);
```
ojwb: Perhaps the same applies here too?  Less sure about this one, but it's not currently used outside of `index_file.cc`.

xapian-applications/omega/index_file.cc
```diff
@@ -583,6 +671,13 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	}
     }
 
+    // if the ctime has changed but the mtime is unchanged, we can just
+    // update the existing Document and avoid having to re-extract text, etc.
+    if (use_ctime && d.get_ctime() != d.get_mtime()) {
```
ojwb: Comparing the current `ctime` and `mtime` seems here odd - shouldn't we be comparing each separately against the previously value we stored for it?  Or am I missing some trick?

barufa: Actually, at this point I know that I have to update the database. So if the current `mtime` and `ctime` are equal I know that the content of the file has changed (and I can avoid calling `index_update_entry` to check this).

xapian-applications/omega/index_file.cc
barufa: Yes, you are right. I am going to change it.

xapian-applications/omega/index_file.h
barufa: Okay, It makes sense. I will do it.

xapian-applications/omega/index_file.h
barufa: It is only used on `index_file.cc`. I will add static to the definition and remove it from the header.

xapian-applications/omega/index_file.cc
```diff
@@ -673,8 +692,9 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 
     // if the ctime has changed but the mtime is unchanged, we can just
     // update the existing Document and avoid having to re-extract text, etc.
+    string md5 = "";
```
ojwb: If you just write `string md5;` the string is empty by default - there's no need to assign '""` to achieve that, and at least with with some compilers it's a little more efficient not to.

xapian-applications/omega/index_file.cc
```diff
@@ -910,12 +929,14 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	    sample = p.sample;
 	    author = p.author;
 	    created = p.created;
-	    md5_string(text, md5);
+	    if (md5 == "")
```
ojwb: I'd use `md5.empty()` to test for an empty string in preference to this.

barufa: I am going to change it.

barufa:
Hello @ojwb  @jaylett, this PR seems finished. Can you take a look and tell me if there is anything left to do?(when you have time).

PR #235, opened 2019-04-21
Enhancement #780 : added default support for .epub

PrithiPal:
I only changed the index_file.cc. The actual documentation on adding new format : https://trac.xapian.org/wiki/FAQ/OmegaNewFileFormat, also suggests to change the documentation in omega/docs/overview.rst and inc/mimetypes.rst. I will add that later on. 

Link to ticket : https://trac.xapian.org/ticket/780.

### DIFF
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -1097,13 +1097,6 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	    // FIXME: topic = atomparser.topic;
 	    author = atomparser.author;
 	}
-   else if ( mimetype=="application/zip"){
-
-
-
-
-
-   }
 
    else {
 	    // Don't know how to index this type.
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -238,6 +238,10 @@ index_add_default_filters()
 		  Filter("numbers2text", SEEK_DEV_STDIN));
     index_command("application/vnd.apply.pages",
 		  Filter("pages2text", SEEK_DEV_STDIN));
+
+    index_command("application/epub+zip",
+      Filter("unzip -p %f '*/*.*htm*'", "text/html", "iso-8859-1",SEEK_DEV_STDIN));
+
 }
 
 void
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -238,10 +238,10 @@ index_add_default_filters()
 		  Filter("numbers2text", SEEK_DEV_STDIN));
     index_command("application/vnd.apply.pages",
 		  Filter("pages2text", SEEK_DEV_STDIN));
-
     index_command("application/epub+zip",
-      Filter("unzip -p %f '*/*.*htm*'", "text/html", "iso-8859-1",SEEK_DEV_STDIN));
-
+      Filter("unzip -p %f '**.*htm*'", "text/html", "iso-8859-1",SEEK_DEV_STDIN));
+    index_command("application/zip",
+      Filter("unzip -p %f '**.*htm*'", "text/html", "iso-8859-1",SEEK_DEV_STDIN));
 }
 
 void
@@ -1029,6 +1029,9 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 				d.get_size(), d.get_mtime());
 		return;
 	    }
+
+
+
 	} else if (mimetype == "text/csv") {
 	    // Currently we assume that text files are UTF-8 unless they have a
 	    // byte-order mark.
@@ -1093,7 +1096,16 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	    keywords = atomparser.keywords;
 	    // FIXME: topic = atomparser.topic;
 	    author = atomparser.author;
-	} else {
+	}
+   else if ( mimetype=="application/zip"){
+
+
+
+
+
+   }
+
+   else {
 	    // Don't know how to index this type.
 	    skip_unknown_mimetype(urlterm, context, mimetype,
 				  d.get_size(), d.get_mtime());


xapian-applications/omega/index_file.cc
```diff
@@ -238,6 +238,10 @@ index_add_default_filters()
 		  Filter("numbers2text", SEEK_DEV_STDIN));
     index_command("application/vnd.apply.pages",
 		  Filter("pages2text", SEEK_DEV_STDIN));
+
+    index_command("application/epub+zip",
+      Filter("unzip -p %f '*/*.*htm*'", "text/html", "iso-8859-1",SEEK_DEV_STDIN));
+
 }
```
ojwb: Some style issues here:

* The wrapped argument should be aligned under the first argument, like the lines above (so `F` under the first `"`)
* Missing space after the last `,`
* There's an extra blank line before the closing `}` of the function

As an almost universal rule when modifying existing code, you want to make your change code match the code around it.

ojwb:
As I mentioned a couple of times on IRC already, the problem with this approach is that most EPUB files contain multiple HTML files and we'll concatenate their content in an essentially arbitrary order (probably the order they appear in the `.zip` file directory), rather than the order they should appear in the book.

Another problem is it assumes that the HTML files are all one level deep inside the `.zip` container, which isn't true for all EPUB files - e.g.:

```
$ unzip -p /usr/share/doc/debian-policy/policy.epub '*/*.*htm*'
caution: filename not matched:  */*.*htm*
```

In this EPUB, the HTML files are all at the top level.

xapian-applications/omega/index_file.cc
PrithiPal: I see. I downloaded few more including the debianpolicy epub which you used for demonstration. Their mime-types are like : 

`john.epub:   application/epub+zip
kadre.epub:  application/epub+zip
matt.epub:   application/epub+zip
policy.epub: application/zip`

I guess, this means one more mime-type to identify. However I am still thinking how much variation can standard epub files have. Also the application/zip have a lot of xhtml files inside. 

This suggests to me that I have to generalize the epub format. I will add the code for application/zip type, but also would like to ask you what other variations can be observed.

PrithiPal:
I see. I downloaded few more including the debianpolicy epub which you used for demonstration. Their mime-types are like : 

`john.epub:   application/epub+zip
kadre.epub:  application/epub+zip
matt.epub:   application/epub+zip
policy.epub: application/zip`

I guess, this means one more mime-type to identify. However I am still thinking how much variation can standard epub files have. Also the application/zip have a lot of xhtml files inside. 

This suggests to me that I have to generalize the epub format. I will add the code for application/zip type, but also would like to ask you what other variations can be observed.

xapian-applications/omega/index_file.cc
PrithiPal: I will take care of coding format.

PrithiPal: This will find the upper as well as the sub-level html files. I also added one more mime-type for epub : appilcation/zip.

ojwb: `appilcation/zip` is a ZIP archive. We definitely **don't** want to treat all files identified as `application/zip` as EPUB.

 EPUB uses the ZIP format as a container (which quite a lot of modern file formats do).  What's presumably happening here is libmagic's test for EPUB is a bit too strict, which results in it failing to spot some EPUB files and so it gives the more generic `application/zip` instead of `application/epub+zip`.

The best way to address this is to simply tell omindex to map files with the extension `.epub` to mime-type `application/epub+zip`, and then libmagic won't get used on such files (which will also be also faster).  The built-in extension to mime-type mappings are in `mimemap.tokens`.

xapian-applications/omega/index_file.cc
```diff
@@ -1029,6 +1029,9 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 				d.get_size(), d.get_mtime());
 		return;
 	    }
+
+
+
```
ojwb: Don't randomly add (or remove) blank lines as part of a patch.  If whitespace reformatting changes to existing code are needed (which these aren't) make them as part of a separate patch which only does such reformatting.

xapian-applications/omega/index_file.cc
```diff
@@ -1093,7 +1096,16 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
...
+	}
+   else if ( mimetype=="application/zip"){
+
+
+
+
+
+   }
+
+   else {
```
ojwb: This is poorly formatted, but a special case for `application/zip` here shouldn't be needed anyway -

ojwb:
> This suggests to me that I have to generalize the epub format. I will add the code for application/zip type, but also would like to ask you what other variations can be observed.

You should really start from the spec for the EPUB format rather than poking inside sample files and making assumptions based on what you see.  That spec seems to live here:

https://www.w3.org/publishing/groups/epub3-cg/

xapian-applications/omega/index_file.cc
PrithiPal: I did added the epub to the /etc/mimemap.tokens. I guess it has to be added to the xapian version in xapian-applications/omega/mimemap.tokens.

ojwb: Do you mean `/etc/mime.types`?  The `mimemap.tokens` file is a source file for xapian-omega and doesn't get installed - instead code is generated from it.

We don't look at `/etc/mime.types` - in some cases it'll likely contain mappings for files where we really want to look at the file to determine its true type (e.g. some `.doc` files are actually RTF, so we deliberately don't have a mapping for `.doc` so we'll use libmagic to check what such a file actually is).

kelson42:
@PrithiPal This is a quite sad this PR has not been brought to an end :(

PrithiPal:
Thanks, @kelson42 for the reminder. I will be back with the development soon.

ojwb:
While it would be great to have EPUB support, I'm going to close this as it unfortunately seems to be abandoned - the last code push was April 2019.

The only change here which is perhaps useful is:

```
   index_command("application/epub+zip",
      Filter("unzip -p %f '**.*htm*'", "text/html", "iso-8859-1",SEEK_DEV_STDIN));
```

However, that's not a great way to extract content from an EPUB file as it ignores the specified reading order, so page order will be whatever order the HTML files were added to the zip container by whatever tool wrote it.  That might match the reading order if we're lucky, but it's not going to be reliable.

Also we now have a libarchive-based worker subprocess to handle ZIP container formats, so this would be better handled there, or by using a dedicated EPUB parsing library if a good one exists (there's libepub from ebook-tools but that doesn't appear to be actively maintained).

PR #237, opened 2019-04-25
Close RemoteDatabase connection on exceptions

Kronuz:
If an exception was being thrown in RemoteDatabase constructor, the connection is never closed. This fixes ticket #781


ojwb:
Thanks for the bug report and patch.

I've gone for a different approach to fixing this by adding an OwnedRemoteConnection subclass which takes care of releasing the fds and then using it in the appropriate cases.  That should eliminate any other similar cases which might be lurking in dark corners.

PR #238, opened 2019-04-25
Add caching to function pointers and fix comments

tanmaysachan:
Function pointers are now cached, to prevent
calling dlsym() repeatedly, improving performance.
Comments now refer to the log_file instead of strace.log.

Re-opened https://github.com/xapian/xapian/pull/236.

### DIFF
--- a/xapian-maintainer-tools/profiling/profiler.c
+++ b/xapian-maintainer-tools/profiling/profiler.c
@@ -58,7 +58,7 @@ int open(const char *pathname, int flags, ...)
 {
     static real_open_t real_open = NULL;
     if (!real_open) {
-        real_open = (real_open_t)dlsym(RTLD_NEXT, "open");
+	real_open = (real_open_t)dlsym(RTLD_NEXT, "open");
     }
     int fd;
     if (flags & (O_CREAT | O_TMPFILE)) {
@@ -91,7 +91,7 @@ int open64(const char *pathname, int flags, ...)
 {
     static real_open64_t real_open64 = NULL;
     if (!real_open64) {
-        real_open64 = (real_open64_t)dlsym(RTLD_NEXT, "open64");
+	real_open64 = (real_open64_t)dlsym(RTLD_NEXT, "open64");
     }
     int fd;
     if (flags & (O_CREAT | O_TMPFILE)) {
@@ -124,7 +124,7 @@ int close(int fd)
 {
     static real_close_t real_close = NULL;
     if (!real_close) {
-        real_close = (real_close_t)dlsym(RTLD_NEXT, "close");
+	real_close = (real_close_t)dlsym(RTLD_NEXT, "close");
     }
     int return_val = real_close(fd);
     logcall("close(%d) = %d\n", fd, return_val);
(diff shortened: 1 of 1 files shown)


xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -8,9 +8,9 @@
  * xapian-io-profile script, then use:
  * ./xapian-io-profile --log=log_file_name ./executable
  *
- * Running this library will produce an strace.log file,
+ * Running this library will produce a log file,
  * which can be fed to strace-analyse to produce a log, by running:
- * ./strace-analyse strace.log
+ * ./strace-analyse log_file_name
```
ojwb: You shouldn't make two different changes in one commit - the comment updates are useful, but aren't part of implementing the caching of function pointers.  For more advice on structuring commits, see:

https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commits

xapian-maintainer-tools/profiling/profiler.c
```diff
@@ -57,15 +72,17 @@ typedef int (*real_open_t)(const char *, int, ...);
 int open(const char *pathname, int flags, ...)
 {
     int fd;
+    if (!cached_function_ptrs[CALL_OPEN]) {
+	cached_function_ptrs[CALL_OPEN] = dlsym(RTLD_NEXT, "open");
+    }
```
ojwb: I think this would be cleaner if you just use a function-scope static variable for each cached pointer, i.e. something like:

```
static real_open_t real_open = NULL;
if (!real_open) {
    real_open = (real_open_t)dlsym(RTLD_NEXT, "open");
}
```

(If this was C++, you could just write: `static real_open_t real_open = (real_open_t)dlsym(RTLD_NEXT, "open");` as C++ allows arbitrary expressions for initialising a static variable - the expression is evaluation the first time (and only the first time) that control flow reaches the statement - see: https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables.  But prob ably it's better to stick to C for this code.)

With this approach the function pointer has the appropriate type so there's no need for the visually cumbersome cast at the places where it is called - you can just write `fd = real_open(pathname, flags);` etc.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: Force pushed into 2 separate commits.

xapian-maintainer-tools/profiling/profiler.c
tanmaysachan: Changed accordingly.

ojwb:
Thanks for the changes.  I've rebased onto master and pushed.

There was one minor issue - your fixup commit "fixup! profiler.c: Fix comments" should actually have been "fixup! profiler.c: Add caching for function pointers" so that `git rebase -i --autosquash` tried to apply it to the correct commit.  Otherwise a commit which claims to "Fix comments" would also end up fixing the whitespace for changes made the previous commit.  I adjusted this by hand during the rebase, so this is really just a note to make sure to check you specify the appropriate commit when using `git commit --fixup=...`

ojwb:
(On master as 70ede98ab318 and parent)

PR #239, opened 2019-04-25
Optimize encode_length

Kronuz:
Benchmark results are:
```
-----------------------------------------------------------------
Benchmark                          Time           CPU Iterations
-----------------------------------------------------------------
BM_EncodeLength_Original        1684 ns       1681 ns     414721
BM_EncodeLength_Optimized        481 ns        480 ns    1421164
```

### DIFF
--- a/xapian-core/net/length.h
+++ b/xapian-core/net/length.h
@@ -35,23 +35,21 @@ template<class T>
 std::string
 encode_length(T len)
 {
-    std::string result;
-    if (len < 255) {
-	result += static_cast<unsigned char>(len);
-    } else {
-	result += '\xff';
+    char result[12];
+    char* end = result;
+    unsigned char b = static_cast<unsigned char>(len);
+    if (len >= 255) {
+	b = '\xff';
 	len -= 255;
-	while (true) {
-	    unsigned char b = static_cast<unsigned char>(len & 0x7f);
+	do {
+	    *end++ = b;
+	    b = static_cast<unsigned char>(len & 0x7f);
 	    len >>= 7;
-	    if (!len) {
-		result += char(b | static_cast<unsigned char>(0x80));
-		break;
-	    }
-	    result += b;
-	}
+	} while (len);
+	b |= static_cast<unsigned char>(0x80);
     }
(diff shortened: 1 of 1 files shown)


ojwb:
That sounds like a nice improvement, but what mix of values are you benchmarking on here?

I'd expect values which fit in a byte to dominate in practice because term lengths should always actually fit in a byte and most wdf values probably will too, but for example timing something like `for (unsigned v = 0; v < UINT_MAX; ++v) encode_length(v);` will only test 255 values which fit in a byte and more than 16 million times as many which don't.

If you're able to, sharing the benchmark code would be helpful.  It would allow checking that we're not optimising this for one compiler at the expense of others, and allow others to more easily optimise further if they want to.

I've also wondered if the remote backend should just use `pack_uint()` as the on-disk formats mostly do.  This uses a slightly simpler encoding, which is 2 bytes instead of one for values 128-254 but is a byte shorter for some larger values.  So it'd probably increase the message sizes a little, but probably reduce CPU.  If we had a benchmark based on some representative data we could check how the size and speed compared.

Kronuz:
I’m pretty sure I added the benchmarking code in the ticket i opened about this...

Here’s the ticket with the benchmarking code: https://trac.xapian.org/ticket/782

ojwb:
Ah sorry, i missed that there was a ticket for this.  Thanks for the link.

So it seems the benchmark uses a logarithmically skewed set of values, which probably isn't too far from reality (on a typical platform ~1/8 of those values will fit in a byte, which is likely too low, but probably only a few times too low).

ojwb:
After much further analysis and discussion in https://trac.xapian.org/ticket/782 we've decided to switch to using `pack_uint()` and `pack_unit_last()` instead, which are faster still, already used extensively in the library, and should typically result in smaller serialisations.

So I'm going to close this, but thanks for the patch, and for raising the issue and setting up the benchmarks of it.

ojwb:
db790e9e12bb9b3ebeaf916ac0acdea9a7ab0dd1 implements the change to use `pack_uint()`, etc.

PR #240, opened 2019-05-01
Add parse_signed template fuction for parsing signed types

VaibhavKansagara:


xapian-core/tests/unittest.cc
```diff
@@ -881,6 +881,49 @@ static bool test_parseunsigned1()
...
+    TEST(!parse_signed("0a", val));
+    TEST(!parse_signed("-99a", val));
+    TEST(!parse_signed("-a99", val));
+    TEST(!parse_signed("--99", val));
+
+    unsigned long long temp = max_val + 1ull;
+    TEST(!parse_signed(str(temp).c_str(), val));
+
+    if (sizeof(S) < sizeof(long long)) {
+	long long temp1 = static_cast<long long>(min_val) - 1LL;
```
ojwb: There are warnings in many of the builds due to this expression overflowing, which the compiler is able to spot that at compile time.  We avoid executing it which avoids triggering the undefined behaviour, but the compiler warnings need to be avoided even when they don't actually indicate a genuine problem (because if the build is warning clean we can enable `-Werror` in developer builds to stop warnings creeping in).

It occurs to me we can convert the negated value and then prepend `-` as a string op, which would avoid such warnings, and also would allow this test to be done for any type.  Something like:

```
unsigned long long temp1 = -min_val + 1ull;
TEST(!parse_signed(("-" + str(temp1)).c_str(), val));
```

The names `temp` and `temp1` aren't usefully descriptive - something like `one_too_large` and `one_too_small` (or perhaps `one_too_small_negated` for my tweaked version) would make it clearer to the reader what this code is testing.

xapian-core/tests/unittest.cc
```diff
@@ -881,6 +881,49 @@ static bool test_parseunsigned1()
...
+	TEST(!parse_signed(str(temp1).c_str(), val));
+    }
+}
+
+static bool test_parsesigned1()
+{
+    parsesigned_helper<char>();
```
ojwb: Implementations are allow to make `char` signed or unsigned by default (it's signed on x86 Linux, but e.g. on arm Linux it is unsigned - https://wiki.debian.org/ArchitectureSpecificsMemo#Summary has a table for Debian architectures).  To get type which is always signed you need to use `signed char`.

xapian-core/common/parseint.h
```diff
@@ -37,4 +38,21 @@ bool parse_unsigned(const char* p, T& res)
...
+bool parse_signed(const char* p, T& res)
+{
+    typedef typename std::make_unsigned<T>::type unsigned_type;
+    unsigned_type temp = 0;
+    if (*p == '-' && parse_unsigned(++p, temp) &&
+	temp <= unsigned_type(std::numeric_limits<T>::min())) {
+	res = -temp;
+	return true;
+    } else if (parse_unsigned(p, temp) &&
+	temp <= std::numeric_limits<T>::max()) {
```
ojwb: The `t` of `temp` should be aligned under `p` of `parse_unsigned` here.

xapian-core/common/parseint.h
```diff
@@ -37,4 +38,21 @@ bool parse_unsigned(const char* p, T& res)
     return true;
 }
 
+template<typename T>
+bool parse_signed(const char* p, T& res)
+{
+    typedef typename std::make_unsigned<T>::type unsigned_type;
+    unsigned_type temp = 0;
+    if (*p == '-' && parse_unsigned(++p, temp) &&
+	temp <= unsigned_type(std::numeric_limits<T>::min())) {
```
ojwb: Here `T` is a signed type so `std::numeric_limits<T>::min()` is negative.  Considering the case when T is `signed char`, `min()` gives -128, and the cast to unsigned_type should give the twos' complement bit pattern for -128 which is `0x80`, and that's 128 as an unsigned value.

So it looks like the test is correct (as things will work similarly for other signed types), but it relies on the fact that casting the min signed value to unsigned gives us its absolute value which seems a little non-obvious (I initially thought there must be a bug here when I spotted that a negative value was getting cast).

Not sure I see a better way to write this though - something like `temp <= unsigned_type(T(-1) - std::numeric_limits<T>::min()) + 1` would work I think, but doesn't seem any clearer.

Perhaps worth noting in a comment what `unsigned_type(std::numeric_limits<T>::min())` gives us though.

xapian-core/tests/unittest.cc
VaibhavKansagara: fixed in a fixup commit

xapian-core/tests/unittest.cc
VaibhavKansagara: fixed in a fixup commit

xapian-core/common/parseint.h
VaibhavKansagara: fixed in a fixup commit

xapian-core/common/parseint.h
VaibhavKansagara: fixed in a fixup commit

ojwb:
Thanks - rebased, squashed and merged to git master as a2c64b588cfba42fe43db536df1a372f9842095e.

PR #241, opened 2019-05-10
Initialize max_fuzzy_expansion

Kronuz:
QueryParser's `max_fuzzy_expansion` was never initialized unless it was manually initialized using `set_max_expansion()`.


ojwb:
Thanks for spotting.

We to have a regression test for this, so I came up with a tweak to qp_flag_fuzzy3 which fails under valgrind (but oddly not ubsan).

Your patch fails the coding standards check due to an overlong line, so I fixed it slightly differently.  Using C++11 member initialisers seems a better option here as if we'd been using them for the existing `max_wildcard_expansion`, etc that would have probably avoided this bug occurring in the first place.

PR #242, opened 2019-05-19
Add testcases for feature related api's

VaibhavKansagara:


xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -94,7 +94,7 @@ CollTfCollLenFeature::get_values() const
     }
     values.push_back(value);
     value = 0;
-    coll_len_iterator = collection_length.find("body");
+    coll_len_iterator = collection_length.find("whole");
```
jaylett: If you've made changes to more than just the tests, then the commit message should reflect that you've also fixed a bug / changed behaviour. Even better would be to separate that out — add the tests that pass without this change as one commit, then have a separate commit for this change and the test(s) that exercise it. (The commit message can then explain in more detail what's going on — as it stands it's very hard for anyone to review this, because there's no indication why this change is appropriate.)

xapian-letor/tests/api_letor.cc
```diff
@@ -25,7 +25,7 @@
 #include <cstdlib>
 #include <fstream>
 #include <sstream>
-
+#include <iostream>
```
ojwb: It looks to me like this isn't actually used, though testcases really shouldn't be using `cout`, etc directly anyway.

xapian-letor/feature/colltfcolllenfeature.cc
VaibhavKansagara: sorry, will create new commits for different changes.

xapian-letor/tests/api_letor.cc
VaibhavKansagara: yeah will fix it.

VaibhavKansagara:
@olly,jaylett I am closing this PR and opening a new one as I have seriously screwed up in this PR.

ojwb:
There's no need to start a new PR for that - you rebuild the changes on a new branch and then force-that to the existing PR branch on the remote.  That would at least keep the discussion in one place.

PR #244, opened 2019-05-21
Replace atoi with parse_signed and parse_unsigned wherever appropriate in core and letor modules.

VaibhavKansagara:


xapian-core/backends/glass/glass_changes.cc
```diff
@@ -64,7 +65,9 @@ GlassChanges::start(glass_revision_number_t old_rev,
     // Always check max_changesets for modification since last revision.
     const char *p = getenv("XAPIAN_MAX_CHANGESETS");
     if (p) {
-	max_changesets = atoi(p);
+	if (!parse_signed(p, max_changesets)) {
+	    throw "XAPIAN_MAX_CHANGESETS must be a positive integer";
+	}
```
ojwb: Currently setting `XAPIAN_MAX_CHANGESETS` to an empty value has the same effect as not setting it, which seems desirable behaviour to keep - an environment variable being empty works the same as it being unset in many contexts, so it's something people might reasonably be relying on.  Making the condition `(p && *p)` should work.

ojwb: I think this should use `parse_unsigned` so negative values are actually rejected (they don't make sense, and the error message implies they aren't allowed).  `0` is OK though (it's the default after all) so the message should say `non-negative` for clarity.

xapian-core/backends/glass/glass_database.cc
```diff
@@ -1049,8 +1050,11 @@ GlassWritableDatabase::GlassWritableDatabase(const string &dir, int flags,
     LOGCALL_CTOR(DB, "GlassWritableDatabase", dir | flags | block_size);
 
     const char *p = getenv("XAPIAN_FLUSH_THRESHOLD");
-    if (p)
-	flush_threshold = atoi(p);
+    if (p) {
+	if (!parse_signed(p, flush_threshold)) {
```
ojwb: Again, I think an empty environment variable should continue to work the same as not setting it.

And a negative threshold doesn't make much sense, so I think this should use `parse_unsigned`.

xapian-core/bin/xapian-progsrv.cc
```diff
@@ -23,7 +23,7 @@
 #include "net/remoteserver.h"
 
 #include "gnu_getopt.h"
-
+#include "parseint.h"
```
ojwb: The blank line you've removed here was separating two different groups of headers (standard C++ headers below vs in-tree headers above).

xapian-core/bin/xapian-progsrv.cc
```diff
@@ -71,7 +71,13 @@ int main(int argc, char **argv)
 		cout << PROG_NAME " - " PACKAGE_STRING << endl;
 		exit(0);
 	    case 't':
-		timeout = atoi(optarg) * 1e-3;
+		unsigned int temp;
+		if (!parse_unsigned(optarg, temp)) {
+		    cout << "timeout must be a positive integer" << endl;
```
ojwb: Again, "non-negative" would be clearer.

xapian-core/bin/xapian-progsrv.cc
```diff
@@ -71,7 +71,13 @@ int main(int argc, char **argv)
...
+		    show_usage();
+		    exit(0);
```
ojwb: We should `exit(1)` here (the `--help` handling above does `show_usage()` then `exit(0)` because that's not an error case).

xapian-core/bin/xapian-replicate-server.cc
```diff
@@ -26,7 +26,7 @@
 #include <xapian.h>
 
 #include "gnu_getopt.h"
-
+#include "parseint.h"
 #include <cstdlib>
 #include <iostream>
```
ojwb: Another unhelpful removal of a blank line between groups of headers (and more below).

xapian-core/bin/xapian-replicate-server.cc
```diff
@@ -73,7 +73,12 @@ main(int argc, char **argv)
 		host.assign(optarg);
 		break;
 	    case 'p':
-		port = atoi(optarg);
+		if (!parse_signed(optarg, port) ||
+		   (port <= 0 || port >= 65536)) {
```
ojwb: When wrapping an expression at an operator, align the two operands vertically (so the `(` here should align under the `!` on the line above).

And a minor cosmetic point, but writing the range check as `port < 1 || port > 65535` would mean the values here match those in the error message, which seems slightly cleaner.

ojwb: (Ah, I see that range check was actually copied from elsewhere)

xapian-core/examples/simpleexpand.cc
```diff
@@ -26,7 +26,7 @@
 
 #include <iostream>
 #include <string>
-
+#include <parseint.h>
```
ojwb: An in-tree header should be included as `"parseint.h"` not `<parseint.h>` (`""` vs `<>` affects the header search path).

But I wonder if we're actually better off just leaving the "simple" examples alone.  Currently they can be built stand-alone using an installed xapian-core, but this change means that other Xapian source files would be required.

xapian-core/tests/harness/testsuite.cc
```diff
@@ -881,7 +881,12 @@ test_driver::parse_command_line(int argc, char **argv)
     if (verbose == 0) {
 	const char *p = getenv("VERBOSE");
 	if (p != NULL) {
-	    verbose = atoi(p);
+	    if (!parse_signed(p, verbose) || (verbose != 1 && verbose != 2)) {
+		throw "Verbose must be 1 or 2"
+		      "a value of 1 will display detailed information about"
+		      "failures, and a value of 2 will display further"
+		      "information about the progress of tests.";
+	    }
```
ojwb: `VERBOSE=0` currently has the same effect as an empty value or not setting `VERBOSE`.

I tend to think it's better allow any non-negative integer value here - at some point increasing the value won't actually make things more verbose but it's handy to be able to type `VERBOSE=999` to get maximum verbosity rather than have to remember (or look up) what the current maximum value actually is.  And the Python bindings testsuite is more verbose for `VERBOSE=3` than `VERBOSE=2` and so running `VERBOSE=3 make check` from the top-level really ought to work.

Also you're missing a space at each break point so the joined string will say "Verbose must be 1 or 2a value", etc.

xapian-core/backends/glass/glass_changes.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-core/backends/glass/glass_database.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-core/bin/xapian-progsrv.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-core/tests/harness/testsuite.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/bin/xapian-prepare-trainingfile.cc
```diff
@@ -74,7 +75,11 @@ try {
 		have_database = true;
 		break;
 	    case 'm':
-		msize = atoi(optarg);
+		unsigned int size;
+		if (!parse_unsigned(optarg, size)) {
+		    throw "Mset size must be >= 0";
+		}
+		msize = size;
```
ojwb: Currently `msize` is of type `int` [*], so considering when `int` is 32 bits, specifying `-m3000000000`, size is `3000000000` and `msize` is `-1294967296`.

[*] This seems wrong to me - it ought to be `Xapian::doccount`, since that's the what the `msetsize` parameter `prepare_training_file()` takes is, and that's the appropriate type for a count of Xapian documents.  If we change `msize` to be `Xapian::doccount` then we can just pass that to `parse_unsigned()`, eliminate `size` and this problem goes away.

I'm not sure throwing a `const char*` error message is a great approach here - it reduces flexibility since the error message really needs to be a constant (or else you have to mess around with allocating buffers or static `std::string`).  It seems simpler to just report the error and exit in place of the `throw`, and that's what we generally do in similar situations.  It's not like our caller might want to handle the exception in an alternative way, because we're parsing command line options in `main()` here.

Also the error message should really go to `cerr` not `cout` (the `--help` output isn't an error, so that rightly goes to `cout`).

xapian-letor/bin/xapian-rank.cc
```diff
@@ -101,7 +102,11 @@ try {
 		have_database = true;
 		break;
 	    case 'm':
-		msize = atoi(optarg);
+		unsigned int size;
+		if (!parse_unsigned(optarg, size)) {
+		    throw "Mset size must be >= 0";
+		}
+		msize = size;
```
ojwb: The comments from `xapian-prepare-trainingfile.cc` apply here too.

ojwb:
Thanks for the updates.  I noticed another small thing, but otherwise this looks good to go now.

xapian-letor/bin/xapian-prepare-trainingfile.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/bin/xapian-rank.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/bin/xapian-prepare-trainingfile.cc
```diff
@@ -75,11 +76,10 @@ try {
...
 		break;
 	    case 'm':
-		unsigned int size;
-		if (!parse_unsigned(optarg, size)) {
-		    throw "Mset size must be >= 0";
+		if (!parse_unsigned(optarg, msize)) {
+		    cerr << "Mset size must be >= 0" << endl;
 		}
-		msize = size;
+		exit(1);
```
ojwb: This will always exit if `-m` is used, not only when there's an error parsing the argument.  Same problem below too.

VaibhavKansagara: oh sorry stupid mistake.

xapian-core/backends/glass/glass_changes.cc
```diff
@@ -63,8 +64,10 @@ GlassChanges::start(glass_revision_number_t old_rev,
 
     // Always check max_changesets for modification since last revision.
     const char *p = getenv("XAPIAN_MAX_CHANGESETS");
-    if (p) {
-	max_changesets = atoi(p);
+    if (p && *p) {
+	if (!parse_unsigned(p, max_changesets)) {
+	    throw "XAPIAN_MAX_CHANGESETS must be a non-negative integer";
```
ojwb: I did some final testing, and this is wrong:

```
$ XAPIAN_MAX_CHANGESETS=-1 examples/simpleindex tmp.db
terminate called after throwing an instance of 'char const*'
Aborted
```

We shouldn't be throwing `const char*` exceptions from the library - probably `Xapian::InvalidArgumentError` is the most appropriate option here.

xapian-core/backends/glass/glass_database.cc
```diff
@@ -1049,8 +1050,11 @@ GlassWritableDatabase::GlassWritableDatabase(const string &dir, int flags,
     LOGCALL_CTOR(DB, "GlassWritableDatabase", dir | flags | block_size);
 
     const char *p = getenv("XAPIAN_FLUSH_THRESHOLD");
-    if (p)
-	flush_threshold = atoi(p);
+    if (p && *p) {
+	if (!parse_unsigned(p, flush_threshold)) {
+	    throw "XAPIAN_FLUSH_THRESHOLD must be a non-negative integer";
```
ojwb: Similarly:

```
$ XAPIAN_FLUSH_THRESHOLD=-1 examples/simpleindex tmp.db
terminate called after throwing an instance of 'char const*'
Aborted
```

xapian-core/backends/glass/glass_changes.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-core/backends/glass/glass_database.cc
VaibhavKansagara: fixed in a fixup commit.

ojwb:
Thanks.  Rebased, squashed and merged as 60bd2fb4e2a559c214f3569d1f5f64f85ba1766f.

PR #246, opened 2019-05-27
 Add testcases for feature related api's. 

VaibhavKansagara:
This PR adds testcases for feature related api's, fixes some of the bugs and removes redundant code.

### DIFF
--- a/xapian-letor/feature/feature.cc
+++ b/xapian-letor/feature/feature.cc
@@ -37,4 +37,39 @@ Feature::~Feature() {
     LOGCALL_DTOR(API, "Feature");
 }
 
+Xapian::termcount
+Feature::get_termfreq(const std::string& term) const
+{
+    LOGCALL(API, Xapian::termcount, "Feature::get_termfreq", term);
+    return internal->get_termfreq(term);
+}
+
+double
+Feature::get_inverse_doc_freq(const std::string& term) const
+{
+    LOGCALL(API, double, "Feature::get_inverse_doc_freq", term);
+    return internal->get_inverse_doc_freq(term);
+}
+
+Xapian::termcount
+Feature::get_doc_length(const std::string& term) const
+{
+    LOGCALL(API, Xapian::termcount, "Feature::get_doc_length", term);
+    return internal->get_doc_length(term);
+}
+
+Xapian::termcount
+Feature::get_collection_length(const std::string& term) const
+{
+    LOGCALL(API, Xapian::termcount, "Feature::get_collection_length", term);
+    return internal->get_collection_length(term);
+}
+
+Xapian::termcount
+Feature::get_collection_termfreq(const std::string& term) const
+{
+    LOGCALL(API, Xapian::termcount, "Feature::get_collection_termfreq", term);
+    return internal->get_collection_termfreq(term);
+}
+
 }
--- a/xapian-letor/tests/api_letor.cc
+++ b/xapian-letor/tests/api_letor.cc
@@ -1414,16 +1414,15 @@ DEFINE_TESTCASE(createfeaturevector_tfdoclencolllfcolllen, generated)
 
     return true;
 }
-namespace Xapian {
 
-class XAPIAN_VISIBILITY_DEFAULT CustomFeature : public Feature {
+class CustomFeature : public Xapian::Feature {
   public:
     CustomFeature() {
-	need_stat(TERM_FREQUENCY);
-	need_stat(DOCUMENT_LENGTH);
-	need_stat(COLLECTION_TERM_FREQ);
-	need_stat(COLLECTION_LENGTH);
-	need_stat(INVERSE_DOCUMENT_FREQUENCY);
+	need_stat(Xapian::Feature::TERM_FREQUENCY);
+	need_stat(Xapian::Feature::DOCUMENT_LENGTH);
+	need_stat(Xapian::Feature::COLLECTION_TERM_FREQ);
+	need_stat(Xapian::Feature::COLLECTION_LENGTH);
+	need_stat(Xapian::Feature::INVERSE_DOCUMENT_FREQUENCY);
     }
     std::vector<double> get_values() const {
 	return vector<double>();
@@ -1434,8 +1433,8 @@ class XAPIAN_VISIBILITY_DEFAULT CustomFeature : public Feature {
     void test_stats() {
 	// test for term frequency
 	TEST_EQUAL(get_termfreq("ZStiger"), 1);
-	TEST_EQUAL(this->get_termfreq("ZXDtiger"), 6);
-	TEST_EQUAL(this->get_termfreq("Ztiger"), 2);
+	TEST_EQUAL(get_termfreq("ZXDtiger"), 6);
+	TEST_EQUAL(get_termfreq("Ztiger"), 2);
 
 	// test for inverse document frequency
 	TEST_EQUAL_DOUBLE(get_inverse_doc_freq("ZStiger"), 0.176091259055681);
@@ -1459,12 +1458,10 @@ class XAPIAN_VISIBILITY_DEFAULT CustomFeature : public Feature {
     }
 };
 
-}
-
 DEFINE_TESTCASE(populatefeature, generated) {
     vector<Xapian::Feature*> f;
-    Xapian::CustomFeature* f1 = new Xapian::CustomFeature();
-    f.push_back(f1);
+    CustomFeature* custom_feature = new CustomFeature();
+    f.push_back(custom_feature);
 
     Xapian::FeatureList fl(f);
     Xapian::Database db = get_database("db_index_three_documents_no_common",
@@ -1488,6 +1485,6 @@ DEFINE_TESTCASE(populatefeature, generated) {
     auto fv = fl.create_feature_vectors(mset, query, db);
     TEST_EQUAL(fv.size(), 1);
 
-    f1->test_stats();
+    custom_feature->test_stats();
     return true;
 }
--- a/xapian-letor/tests/api_letor.cc
+++ b/xapian-letor/tests/api_letor.cc
@@ -1414,3 +1414,80 @@ DEFINE_TESTCASE(createfeaturevector_tfdoclencolllfcolllen, generated)
 
     return true;
 }
+namespace Xapian {
+
+class XAPIAN_VISIBILITY_DEFAULT CustomFeature : public Feature {
+  public:
+    CustomFeature() {
+	need_stat(TERM_FREQUENCY);
+	need_stat(DOCUMENT_LENGTH);
+	need_stat(COLLECTION_TERM_FREQ);
+	need_stat(COLLECTION_LENGTH);
+	need_stat(INVERSE_DOCUMENT_FREQUENCY);
+    }
+    std::vector<double> get_values() const {
+	return vector<double>();
+    }
+    std::string name() const {
+	return "CustomFeature";
+    }
+    void test_stats() {
+	// test for term frequency
+	TEST_EQUAL(get_termfreq("ZStiger"), 1);
+	TEST_EQUAL(this->get_termfreq("ZXDtiger"), 6);
+	TEST_EQUAL(this->get_termfreq("Ztiger"), 2);
+
+	// test for inverse document frequency
+	TEST_EQUAL_DOUBLE(get_inverse_doc_freq("ZStiger"), 0.176091259055681);
+	TEST_EQUAL_DOUBLE(get_inverse_doc_freq("ZXDtiger"), 0.176091259055681);
+	TEST_EQUAL_DOUBLE(get_inverse_doc_freq("Ztiger"), 0.176091259055681);
+
+	// test for document length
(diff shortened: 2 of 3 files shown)


jaylett:
The getters need some discussion, because they return STL containers. I think in all cases you could return a Xapian or primitive value by taking a `std::string` as an argument — which might make a little more sense.

I think part of the problem here is that the _setters_ API is a bit "internal" — it doesn't make much sense to a user of the `Feature`, only to something that sets it up. I'm not sure how practical it is to change that, but perhaps there's an approach that makes sense in terms of how `FeatureList::Internal::populate_feature()` works. (Or maybe that approach needs changing; it seems to be doing things a not particularly straightforward way to achieve some potential caching improvements.)

ojwb:
It does feel like we're only adding these getters so we can write tests, not because they're a useful public API.

And I agree returning an STL container by value is bad - it makes looking up a single value automatically at least `O(n)` in the size of the container because you have to copy the whole container in order to look at it.

As James said on IRC, this is harder than necessary to review because it mixes several different (though related) changes.  For example, I can see from looking at the surrounding code that the fix in 4bda855 makes sense, but I can't easily tell if there's actually a regression test for it somewhere in the testsuite changes.

xapian-letor/feature/feature.cc
```diff
@@ -94,4 +95,44 @@ Feature::set_collection_termfreq(const std::map<std::string,
     collection_termfreq = collection_tf;
 }
 
+std::map<std::string, Xapian::termcount>
+Feature::get_termfreq() const
+{
+    LOGCALL(API, (std::map<std::string, Xapian::termcount>),
+	    "Feature::get_termfreq", NO_ARGS);
```
ojwb: There's no need to wrap long `LOGCALL` lines - they're not really intended for humans to read, and `xapian-check-patch` has an exception in its line length check for them.

(Ideally we'd have a script to automatically add them so they were complete and automatically ensured to be up-to-date - we quite often find someone failed to update the `LOGCALL` when a method signature changes - often that means a build error in a logging build, but it's possible for it to introduce subtle bugs when logging is enabled.)

xapian-letor/tests/api_letor.cc
```diff
@@ -192,40 +241,106 @@ DEFINE_TESTCASE(createfeaturevectoronevector, generated)
 DEFINE_TESTCASE(createfeaturevectoronevector_wrongquery, generated)
 {
     Xapian::FeatureList fl;
-    Xapian::Database db = get_database("apitest_ranker3",
+    Xapian::Database db = get_database("apitest_wrongquery",
 				       db_index_one_document);
```
ojwb: It's good to clean up these database names, but for generated databases it's better to align the names with the generator functions rather than with the testcase names.

For example, currently the generator function `db_index_one_document` is used to build 11 differently named read-only databases, but this generator function is not parameterised so all these databases will have identical contents.  If we used the same name in each of the 11 testcases, then only one database would need to be built for the first testcase to run, and that would then be reused for the other 10 testcases.

ojwb:
> I can't easily tell if there's actually a regression test for it somewhere in the testsuite changes.

If I `git revert` that commit locally then one of the new tests fails, so it seems there is effectively a regression test in there, which is good.

For now I've cherry-picked "Remove redundant code." (0b25805) to git master and pushed that, as that commit stands by itself, and means there's one fewer thing in here to worry about.

(BTW, don't put a full stop on the end of the summary line in the commit message - kind of like how you wouldn't in an email subject line or the title of a book).

I'm inclined to suggest pulling out the fixes and the test coverage for them and getting those merged, and then we can think about what to do about how `Feature` handle stats.  It looks to me like the whole way these are handled would benefit from a rethink - currently it looks like all these maps get copied into every feature, which means O(n*m) space and time (where n is the number of terms and m the number of features).

xapian-letor/tests/api_letor.cc
VaibhavKansagara: ok I will revert this change.

jaylett:
>>> And I agree returning an STL container by value is bad - it makes looking up a single value automatically at least O(n) in the size of the container because you have to copy the whole container in order to look at it.

I'm pretty sure we can refactor the way that `FeatureList` prepares the `Feature` objects to change the setter approach quite significantly. It's been an incremental thing gradually figuring out (over two or three GSoC projects!) what this could look like, if I remember correctly…

xapian-letor/tests/api_letor.cc
ojwb: My point is not that the existing `apitest_ranker3` is right, rather that a name based on the generator function would be better - e.g. `apitest_index_one_document`.  If each testcase which uses this generator function uses that same name, then we get the speed and disk space savings from reusing the cached database.

ojwb:
> I'm pretty sure we can refactor the way that `FeatureList` prepares the `Feature` objects to change the setter approach quite significantly. It's been an incremental thing gradually figuring out (over two or three GSoC projects!) what this could look like, if I remember correctly…

I wonder if it would work for `Feature` to have a ref-counted pointer to a `Feature::Internal` class which had all these maps as members.  Then all the features for a given search could share a single `Feature::Internal`.

As for how to write testcases to validate what's in the maps, in core for `Xapian::Weight` stats we create a custom `Weight` subclass for this purpose.  Are user sub-classes of `Feature` supported?  If so, perhaps a similar approach could be used here.

xapian-letor/tests/api_letor.cc
VaibhavKansagara: got it.

jaylett:
User sub-classes of `Feature` certainly should be supported — that's the point of having the `FeatureList`. That should be a clean way of testing that `FeatureList` is driving `Feature` objects correctly.

I think in a previous year we discussed the idea of the `Feature` objects having a reference back to `FeatureList` (or `FeatureList::Internal` perhaps) rather than `FeatureList` poking state in every document pass (even using `swap()` to avoid copying it'd be faster if a `Feature` could just access the stats it needs directly). I can't remember why we didn't go that way previously, so it might have got complex around visibility with subclasses. It sounds like it should be feasible, however.

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -94,7 +94,7 @@ CollTfCollLenFeature::get_values() const
     }
     values.push_back(value);
     value = 0;
-    coll_len_iterator = collection_length.find("body");
+    coll_len_iterator = collection_length.find("whole");
```
ojwb: This fixed by e94ccffe3ec5994ff20843dc9a84a1ed8890711e.

xapian-letor/feature/feature.cc
```diff
@@ -94,4 +95,44 @@ Feature::set_collection_termfreq(const std::map<std::string,
...
+}
+
+std::map<std::string, Xapian::termcount>
+Feature::get_collection_termfreq() const
+{
+    LOGCALL(API, (std::map<std::string, Xapian::termcount>),
+	    "Feature::get_collection_termfreq", NO_ARGS);
+    return collection_termfreq;
+}
+
```
ojwb: I think this part is superseded by the changes in #255.

xapian-letor/tests/api_letor.cc
```diff
@@ -861,3 +980,1088 @@ DEFINE_TESTCASE(tfdoclenfeature, generated)
...
+    // test for whole in normalized form
+    TEST_EQUAL_DOUBLE(fvals_doc1[2], test_vals_doc1[2] / max_val[2]);
+    TEST_EQUAL_DOUBLE(fvals_doc2[2], test_vals_doc2[2] / max_val[2]);
+
+    // test for weight in normalized form
+    TEST_EQUAL_DOUBLE(fvals_doc1[3], test_vals_doc1[3] / max_val[3]);
+    TEST_EQUAL_DOUBLE(fvals_doc2[3], test_vals_doc2[3] / max_val[3]);
+
+    return true;
+}
```
ojwb: But presumably these new testcases are still useful?

ojwb: Oh, though many of these seem to be resubmitted as #252.

The ones that haven't seem to be colltfcolllenfeature (which was merged already in e94ccffe3ec5994ff20843dc9a84a1ed8890711e) and the populate*feature testcases which currently rely on the getters.  The latter would be useful to rework to use a user-created Feature subclass mechanism, but I think that would need implementing first.

ojwb: BTW, most English style guides seem to frown on using an apostrophe for the plural of an abbreviation (based on a web search for `abbreviation plural apostrophe`).  I'd suggest using `APIs` rather than `api's`.

VaibhavKansagara: >  The latter would be useful to rework to use a user-created Feature subclass mechanism, but I think that would need implementing first.
>
Sure i will try that out.

VaibhavKansagara: > But presumably these new testcases are still useful?

I think they are.

VaibhavKansagara: > The latter would be useful to rework to use a user-created Feature subclass mechanism, but I think that would need implementing first.
> 
I don't see how we can do that without a public api connecting to the internal maps and I also don't understand the need to create a user-created Feature subclass.

ojwb: If we're aiming for a Feature subclassing API which is flexible enough to implement equivalents to the built-in features then it would be need at least a `protected` API.  But we'd be providing it for the purpose of allowing users to create their own features, rather than just exposing internal details purely for access by the test suite.

User-created features are useful for at least two categories of things:

* general purpose features which xapian-letor doesn't yet implement but would be candidates for merging into a future version.  It's useful (but not vital) to be able to implement these without having to patch and rebuild xapian-letor.  In some cases these would ideally need access to more stats so would need changes to xapian-letor anyway (e.g. if I wanted to add a feature which used the number of unique terms in the document).

* features which are specific to a particular application - for example if I'm implementing letor for search in an online store I might want to try adding features for price, number of sales, physical size of item, etc.

Many of the second category of features could actually be implemented without access to the stats (they'd probably either pull data from value slots in the Xapian DB or by querying some external data store).

xapian-letor/tests/api_letor.cc
```diff
@@ -1414,3 +1414,80 @@ DEFINE_TESTCASE(createfeaturevector_tfdoclencolllfcolllen, generated)
...
+	return vector<double>();
+    }
+    std::string name() const {
+	return "CustomFeature";
+    }
+    void test_stats() {
+	// test for term frequency
+	TEST_EQUAL(get_termfreq("ZStiger"), 1);
+	TEST_EQUAL(this->get_termfreq("ZXDtiger"), 6);
+	TEST_EQUAL(this->get_termfreq("Ztiger"), 2);
```
ojwb: The `this->` qualification here isn't necessary (nor consistently applied!)

Some coding standards do say to use `this->` in some circumstances to make it clear the thing referred to is a member variable, but Xapian's doesn't.

xapian-letor/tests/api_letor.cc
```diff
@@ -1414,3 +1414,80 @@ DEFINE_TESTCASE(createfeaturevector_tfdoclencolllfcolllen, generated)
 
     return true;
 }
+namespace Xapian {
+
+class XAPIAN_VISIBILITY_DEFAULT CustomFeature : public Feature {
```
ojwb: This class is only used within this one file in the testsuite, so it doesn't need to be in namespace `Xapian` (and indeed it's better really not to put it there).

It also doesn't need visibility annotations (those are there to allow us to restrict the symbols which are exported from the libxapianletor library to just those needed for the public API).

xapian-letor/tests/api_letor.cc
```diff
@@ -1414,3 +1414,80 @@ DEFINE_TESTCASE(createfeaturevector_tfdoclencolllfcolllen, generated)
...
+	TEST_EQUAL(get_collection_termfreq("Ztiger"), 2);
+    }
+};
+
+}
+
+DEFINE_TESTCASE(populatefeature, generated) {
+    vector<Xapian::Feature*> f;
+    Xapian::CustomFeature* f1 = new Xapian::CustomFeature();
+    f.push_back(f1);
```
ojwb: I'd suggest just using `f.push_back(new Xapian::CustomFeature());` - it's if anything slightly clearer than using the temporary variable, and avoids the potentially confusing presence of variables `f1` and `fl` in the same function (they look quite similar in some fonts).

xapian-letor/tests/api_letor.cc
VaibhavKansagara: Oh, I forgot to remove them. fixed in a fixup commit.

xapian-letor/tests/api_letor.cc

xapian-letor/tests/api_letor.cc
VaibhavKansagara: We can't do that since I am using it to call test_stats() below but I have changed the name so it doesn't sound confusing anymore.

ojwb:
Thanks - rebased, squashed and merged as 8a904c70053dae410aefdeed1da20782de0b3efc.

PR #247, opened 2019-05-31
Omindex Text-Extraction Libraries

barufa:
Adding support for text-extraction libraries. 
Currently, omindex use external filter from extracting text from different files but it is possible to use libraries and isolate library bugs in a sub-process. 
This code is related to the GSoC project [https://trac.xapian.org/wiki/GSoCProjectIdeas#Project:Text-ExtractionLibraries](url).

It also add a library (Poppler) for extract text from a pdf.


xapian-applications/omega/Makefile.am
```diff
@@ -164,12 +166,16 @@ omindex_SOURCES = omindex.cc myhtmlparse.cc htmlparse.cc index_file.cc mime.cc\
...
+ datetime.cc common/closefrom.cc worker.cc worker_comms.cc
 if NEED_MKDTEMP
 omindex_SOURCES += portability/mkdtemp.cc
 endif
 omindex_LDADD = $(MAGIC_LIBS) $(XAPIAN_LIBS) $(ZLIB_LIBS) libutf8convert.la
 
+omindex_pdf_SOURCES = assistant.cc worker_comms.cc handler_pdf.cc
+omindex_pdf_LDADD = -lpoppler-cpp $(XAPIAN_LIBS)
+omindex_pdf_LDFLAGS = -pthread
+
```
barufa: Here I don't know how to add the -pthread flag in a proper way (e.g. add it just if it is necessary).

xapian-applications/omega/assistant.cc
```diff
@@ -0,0 +1,127 @@
...
+
+#include "worker_comms.h"
+#include "handler_pdf.h"
+#include "safeunistd.h"
+
+using namespace std;
+
+const int FD = 3;
+const int time_limit = 300;
+
```
barufa: I need to use an asynchronous timer to detect possible infinite loops. It is easy to use SIGALARM to implement it, but this is not going to work if we are not on a Unix system.
An alternative (much more expensive) is to use threads, but it is necessary create a new thread every time. So we are going to create as many threads as files for indexing.
Of course, it is possible just to avoid using it. However, without a timer we cannot detect infinite loops on a particular library.

xapian-applications/omega/assistant.cc
```diff
@@ -0,0 +1,127 @@
...
+static void
+stop_timeout() {
+    // Set SIGALRM to be ignore
+    signal(SIGALRM, SIG_IGN);
+    // A pending alarm request, if any, is cancelled by calling alarm(0)
+    alarm(0);
+}
+
+// #elif defined PTHREAD?
+/* If I can use threads (second option)
```
barufa: It is possible to use conditional compilation. But I am not sure about which variables I should use, or how to create new ones.

xapian-applications/omega/handler_pdf.cc
```diff
@@ -0,0 +1,83 @@
...
+	    page *p(doc->create_page(i));
+	    if (!p) {
+		cerr << "Poppler Error: Failed to create page" << endl;
+		return false;
+	    }
+	    dump += clear_text(p->text());
+	}
+    } catch (...) {
+	cerr << "Poppler threw an exception" << endl;
+	return false;
```
barufa: Maybe I can handle some errors in a better way (or give more information about them).

xapian-applications/omega/worker_comms.cc
```diff
@@ -0,0 +1,104 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+
```
barufa: I am not sure about what to put in the license comments.

xapian-applications/omega/index_file.cc
```diff
@@ -101,6 +101,7 @@ static string site_term, host_term;
 static Failed failed;
 
 map<string, Filter> commands;
+map<string, Worker *> workers;
```
barufa: I try to handle Workers in a similar way to Filters. Is that okay?

xapian-applications/omega/handler_pdf.cc
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+#include <config.h>
+
+#include <xapian.h>
+#include "handler_pdf.h"
+#include <iostream>
+#include <poppler-document.h>
+#include <poppler-page.h>
+
```
barufa: Do I have to add poppler as a dependency? In that case, How I do it?

xapian-applications/omega/assistant.cc
ojwb: It seems reasonable to use `alarm()` and just not have such a timeout for platforms which don't provide that - it's specified by POSIX after all.  That makes for a simpler and lower overhead solution for most platforms, and indexing will still work on other platforms, just without a timeout.  People who care about other platforms can always contribute a patch to address this.

Another approach would be for the main omindex process to implement the timeout by using `poll()` or `select()` on the fd for the pipe to the subprocess.  Then if the timeout passes before it receives data it can just kill the subprocess (which should be a rare occurrence so the overhead of having to set up the subprocess again isn't a big concern).  Unfortunately Microsoft's socket implementation stupidly makes a socket a different thing to a file descriptor, and `select()` only works on sockets and not arbitrary file descriptors, so it'd probably still need special code (or to use sockets to communicate with the child process instead of a pipe).

In general I wouldn't worry too much about porting to Microsoft Windows for Omega code.  It's important that xapian-core works there as fully as possible, but I don't think I've heard of anyone using Omega on that platform recently.  So long as it builds and works (with slightly degraded functionality in some places), better to focus energy on getting more done on POSIX platforms.  You can always come back later and try to address the missing functionality.

xapian-applications/omega/assistant.cc
ojwb: I guess (as above) you probably can skip this threaded code at least for now, but to answer the question:

With autotools, there are build-time checks in `configure.ac` to see what features are available, and this arranges for `HAVE_xxx` macros to be defined or not (or in a few cases to be defined to `0` or `1`) to indicate the results of these checks.

If you need threads, we require at least C++11 now which includes thread support so that's an option too.

xapian-applications/omega/assistant.cc
```diff
@@ -0,0 +1,127 @@
...
+
+const int FD = 3;
+const int time_limit = 300;
+
+#if defined SIGALRM
+// IF I have SIGALRM (Best option)
+
+static void
+timeout_handler(int n) {
+    exit(n);
```
ojwb: It's not safe to use `exit()` in a signal handler (but `_exit()` or `_Exit()` are OK) - see `man 7 signal-safety` for a list.

`_Exit()` is the ISO C was (since C99) so probably the more portable choice:

https://en.cppreference.com/w/c/program/_Exit

barufa: I didn't know that. I am going to chage it.

xapian-applications/omega/assistant.cc
barufa: So, I just delete the threads option?

xapian-applications/omega/assistant.cc
barufa: Okay, I will take a look at it and try to add a variable for the timer.

xapian-applications/omega/assistant.cc
ojwb: I think so.  It's certainly good to be thinking about portability though.

xapian-applications/omega/handler_pdf.cc
```diff
@@ -0,0 +1,93 @@
...
+	    title = clear_text(doc->get_title());
+	    keywords = clear_text(doc->get_keywords());
+	    cout << "author: <<" << author << ">>" << endl;
+	    cout << "title: <<" << author << ">>" << endl;
+	    cout << "keywords: <<" << author << ">>" << endl;
+#else
+	    author = "";
+	    title = "";
+	    keywords = "";
+#endif
```
barufa: I am having problems with Poppler version (packages.ubuntu.com/xenial/libpoppler-cpp-dev) and Travis.
Travis uses version 0.41 (Feb 16, 2016), but I need at least 0.46 (Jul 25, 2016) to be able to use method get_author, get_title and get_keywords.
Xenial use 0.41, but Bionic and newer uses poppler version >=0.62.

However, in 0.41 I have a method metadata (it returns a `ustring` with xml format) from which I can extract some information (title, author) but it doesn't as well as the other method. In some cases metadata is empty and get_author and get_title works ok.

I also found this _ppa:opencpu/poppler_ [https://launchpad.net/~opencpu/+archive/ubuntu/poppler](url). I am not sure if it is useful.

jaylett: Can you get the author &c information using `::info_key()`? It looks from the source code like this will work, and is supported back to 0.41 or earlier.

barufa: Yes, you are right. I didn't notice that

barufa:
I am having problems with travis and Poppler. It seems that Travis uses clang with libc++ , but I need to change it because Poopler is linked against GNU stdlibc++.
Can I change that or I must try to solve the problem in another way?

xapian-applications/omega/handler_pdf.cc
```diff
@@ -0,0 +1,116 @@
...
+		cerr << "Poppler Error: Failed to create page" << endl;
+		return false;
+	    }
+	    cstr = poppler_page_get_text(pg);
+	    if (cstr) {
+		dump += clear_text(cstr);
+		g_free(cstr);
+	    }
+	    g_object_unref(pg);
+	}
```
barufa: I have been having problems this library, in particular with some memory leaks. So I try to free as much memory as possible.
Now it seems to work well.

xapian-applications/omega/handler_pdf.cc
barufa: I just change travis file.

xapian-applications/omega/configure.ac
```diff
@@ -246,6 +246,16 @@ AC_SEARCH_LIBS([socketpair], [xnet],
...
 
+dnl check if poppler is available.
+am_save_LIBS="$LIBS"
+AC_LIB_APPENDTOVAR([LIBS], [-lpoppler-cpp])
+AC_TRY_LINK([#include <poppler/cpp/poppler-version.h>],
+            [poppler::version_string()],
+            [AC_DEFINE(HAVE_POPPLER, 1, [Define HAVE_POPPLER if the labrary is available]), lib_poppler=true],
+	    [lib_poppler=false])
+AM_CONDITIONAL([OMINDEX_PDF], [test x$lib_poppler != xfalse])
+LIBS="$am_save_LIBS"
```
barufa: I have not much experience with this kind of tools, Did I modify it correctly?
It seems to compile on Travis. I also add a case to travis (similar to the first one, in a personal repository) where poppler is not available and it passed.

.gitignore
```diff
@@ -1,5 +1,7 @@
 *.o
 
+xapian-applications/omega/omindex_*
+
```
barufa: The idea is to use similar names in next libraries

ojwb: OK, but really this would be better put in `xapian-applications/omega/.gitignore` - otherwise the ignore rules specific to that directory are split up.

.travis.yml
```diff
@@ -88,6 +91,7 @@ matrix:
             - python-docutils
             - python-pygments
             - pngcrush
+            - libpoppler-cpp-dev
       language: node_js
       node_js:
         - node
```
ojwb: We don't build omega in this configuration - see a little below:

```
        # Bootstrap only xapian-core for emscripten build.
        - ./bootstrap xapian-core
```

xapian-applications/omega/Makefile.am
```diff
@@ -76,10 +76,12 @@ AM_CPPFLAGS = \
 	-I$(top_srcdir)/common \
 	-DCONFIGFILE_SYSTEM=\"$(sysconfdir)/omega.conf\" \
 	-DPKGLIBBINDIR=\"$(pkglibbindir)\"
+
 AM_CXXFLAGS += $(XAPIAN_CXXFLAGS)
 
 pkglibbindir = $(pkglibdir)/bin
 pkglibbin_PROGRAMS = omega
+
 dist_pkglibbin_SCRIPTS = mhtml2html outlookmsg2html rfc822tohtml vcard2text
```
ojwb: I guess these apparently randomly added blank lines are a result of adding and then removing things during development.  It's good practice to review the diff and eliminate extraneous changes like these.

xapian-applications/omega/Makefile.am
```diff
@@ -164,12 +167,19 @@ omindex_SOURCES = omindex.cc myhtmlparse.cc htmlparse.cc index_file.cc mime.cc\
...
 omindex_SOURCES += portability/mkdtemp.cc
 endif
 omindex_LDADD = $(MAGIC_LIBS) $(XAPIAN_LIBS) $(ZLIB_LIBS) libutf8convert.la
 
+if OMINDEX_PDF
+pkglibbin_PROGRAMS += omindex_pdf
+noinst_HEADERS += handler_pdf.h
+omindex_pdf_SOURCES = assistant.cc worker_comms.cc handler_pdf.cc
+omindex_pdf_LDADD = -lpoppler-cpp $(XAPIAN_LIBS)
+endif
```
ojwb: This works for building, but the downside is that source files specific to this program will only be included in the tarball built by `make dist` if `OMINDEX_PDF` is true in the tree where `make dist` is run.  That's problematic as it makes it all too easy to miss source files from a release.

The better way to do this is to have all these lines unconditional except `pkglibbin_PROGRAMS += omindex_pdf`, and then list `omindex_pdf` in `EXTRA_PROGRAMS` - this tells automake that these programs may or may not be built, but it then knows they exist, and will include them during `make dist`.

Or scrap the `OMINDEX_PDF` conditional entirely and just `AC_SUBST` a variable which lists the modules to build, then you can substitute that variable in `pkglibbin_PROGRAMS` directly.

xapian-applications/omega/configure.ac
```diff
@@ -246,6 +246,16 @@ AC_SEARCH_LIBS([socketpair], [xnet],
 dnl fnmatch() is used to match ignore pattern.
 AC_CHECK_FUNCS([fnmatch])
 
+dnl check if poppler is available.
+am_save_LIBS="$LIBS"
+AC_LIB_APPENDTOVAR([LIBS], [-lpoppler-cpp])
+AC_TRY_LINK([#include <poppler/cpp/poppler-version.h>],
+            [poppler::version_string()],
+            [AC_DEFINE(HAVE_POPPLER, 1, [Define HAVE_POPPLER if the labrary is available]), lib_poppler=true],
```
ojwb: "labrary" should be "library"

ojwb: And that `,` before `lib_poppler=true` is wrong - that should be a newline or a `;` - this expands to shell code.

xapian-applications/omega/configure.ac
```diff
@@ -246,6 +246,16 @@ AC_SEARCH_LIBS([socketpair], [xnet],
 dnl fnmatch() is used to match ignore pattern.
 AC_CHECK_FUNCS([fnmatch])
 
+dnl check if poppler is available.
+am_save_LIBS="$LIBS"
```
ojwb: I think `save_LIBS` would be a better name (which is the convention we use for saving and restoring variables in other `configure.ac` files).  I think `am_save_LIBS` is rather treading on automake's namespace (I notice `xapian-applications/omega/m4/iconv.m4` uses `am_save_LIBS`, but that's a third party file so we try to avoid modifying it if we can, so as to make importing new versions easier).

xapian-applications/omega/handler_pdf.cc
```diff
@@ -0,0 +1,88 @@
...
+	string & dump,
+	string & title,
+	string & keywords,
+	string & author,
+	string & pages)
+{
+
+    try {
+
+	document * doc = document::load_from_file(filename);
```
ojwb: I would lose these blank lines at the start of blocks - we don't do that in existing code.

.gitignore
barufa: My mistake, I'm going to change it.

.travis.yml
barufa: I didn't realize, I am going to remove it.

xapian-applications/omega/handler_pdf.cc
barufa: Okay, I am going to change it

xapian-applications/omega/configure.ac
barufa: Okay, I am going to fix it.

xapian-applications/omega/Makefile.am
barufa: Yes, I didn't see that line.

xapian-applications/omega/configure.ac
barufa: Okay, I do not have much experience with this type of tool. I'm afraid I might make these kinds of mistakes.

xapian-applications/omega/Makefile.am
barufa: I made some changes, but I am not sure if it is okay.

ojwb:
> barufa force-pushed the barufa:omindex_text-extraction_libraries branch from 2af94a8 to f3bbf7e 2 days ago

You shouldn't need to keep force-pushing the branch.  The reason to use `fixup` commits is to avoid needing to.  Force-pushing makes review harder, because it rewrites the history of the changes - without this the reviewer can usually just look at the new commits to check what's changed since the previous round of review.

There are sometimes circumstances when force-pushing is the best answer, but it isn't something you should do lightly on a branch that you're already collaborating with others on.  A good rule of thumb is to only do it if everyone involved agrees.

xapian-applications/omega/configure.ac
```diff
@@ -247,14 +247,15 @@ dnl fnmatch() is used to match ignore pattern.
 AC_CHECK_FUNCS([fnmatch])
 
 dnl check if poppler is available.
-am_save_LIBS="$LIBS"
+save_LIBS="$LIBS"
 AC_LIB_APPENDTOVAR([LIBS], [-lpoppler-cpp])
```
ojwb: Sorry, I missed some issues here before.

Portable build systems seems to be inherently complex, and there's quite a lot of new stuff to learn here too.

Despite `AC_LIB_APPENDTOVAR` being in the `AC_` namespace, this isn't actually a standard autoconf macro, but one that `iconv.m4` pulls in.

I think I'd rather avoid introducing uses of it outside of `iconv.m4` anyway, but in particular it doesn't use `pkg-config` to find the appropriate flags to use, which we really should be doing when that's the preferred way to find the flags for a library, as it is for libpoppler, e.g.:

```
$ pkg-config poppler-cpp --libs-only-l
-lpoppler-cpp
```

That way things should just work with a non-system install of poppler, or on a platform where poppler needs special compiler flags to be used.

There are some autoconf macros for `pkg-config` which you'd use approximately like so:

```
PKG_CHECK_MODULES([POPPLER_CPP], [poppler-cpp], [
    save_LIBS=$LIBS
    LIBS="$LIBS $POPPLER_LIBS"
    ... do the AC_TRY_LINK here so we catch the libc++/libstdc++ incompatibility situation and auto-disable for that ...
    LIBS=$save_LIBS
    ], [lib_poppler=false])
```

But there's more, and I'm starting to wonder if it would work better to find a time to (virtually) sit down together and work through this interactively.

barufa:
I made a commit by mistake (I thought that I was in another branch). So I made a force push to remove it.

jaylett:
@barufa in future you can make a revert as a fixup (do `git revert --no-commit <unwanted commit>` then `git commit --fixup <unwanted commit>`), and when the branch is subsequently squashed the pair of change-then-removed will disappear (because when rebasing, by default empty commits are ignored).

xapian-applications/omega/worker_comms.h
```diff
@@ -22,6 +22,8 @@
 #include <cstdio>
 #include <string>
 
+#define ASSISTANT_ERROR "XAPIAN - ASSISTANT EXTRACTION FALSE"
+
```
barufa: I wasn't sure about what to put in the error string.

xapian-applications/omega/worker.cc
```diff
@@ -155,17 +155,23 @@ Worker::extract(const std::string & filename,
...
-	read_string(sockt, title) &&
-	read_string(sockt, keywords) &&
-	read_string(sockt, author) &&
-	read_string(sockt, strpage)) {
-	    if (strpage.empty())
+    // If there is an error, the assistant process notify it through dump.
+    if (write_string(sockt, filename) && read_string(sockt, dump)) {
+	// Checks if the assistant succeded
+	if (dump == ASSISTANT_ERROR)
+	    return false;
```
ojwb: Using a magic value for a field like seems a rather awkward design as it can't distinguish an actual failure from a file containing the text `XAPIAN - ASSISTANT EXTRACTION FALSE`.  While that's not a big practical problem, it would be cleaner to simply change the protocol.  There's not compatibility worry at this point.

However, let's not try to address this in this PR and just continue to exit on error for now.

It's better to focus on getting what's here merged rather than try to include more changes to address other things which will just make the whole review process drag out even longer.  Several small PRs are typically easier to review than one PR including all the same changes.

xapian-applications/omega/handler_poppler.cc
```diff
@@ -0,0 +1,86 @@
...
+	    cerr << "Poppler Error: Failed to read pdf file" << endl;
+	    return false;
+	}
+
+	int npages = doc->pages();
+	// Extracting PDF metadata
+	author = clear_text(doc->info_key("Author"));
+	title = clear_text(doc->info_key("Title"));
+	keywords = clear_text(doc->info_key("Keywords"));
+	pages = to_string(npages);
```
ojwb: Don't use `std::to_string()` - as the dev guide says:

> Please avoid the following C++11 features for the time being:
>
> * `std::to_string()` - this is completely missing on current versions of mingw and cygwin - in the library, you can `#include "str.h"` and then use the `str()` function instead for most cases.  This is also usually faster than `std::to_string()`.

(I'm not sure why it says "in the library" - Omega uses `str()` extensively too.)

xapian-applications/omega/assistant.cc
```diff
@@ -0,0 +1,99 @@
...
+// FIXME: Restart filter every N files processed?
+
+int main() {
+
+    string filename;
+    FILE * sockt = fdopen(FD, "r+");
+
+    // Just in case (if the pid of this new process is equal
+    // to an old assistant pid it is possible to have problems)
+    stop_timeout();
```
ojwb: What's the potential problem here?

xapian-applications/omega/configure.ac
```diff
@@ -246,6 +246,23 @@ AC_SEARCH_LIBS([socketpair], [xnet],
...
+     save_LIBS=$LIBS
+     LIBS="$LIBS $POPPLER_LIBS"
+     AC_LINK_IFELSE([#include <poppler/cpp/poppler-version.h>
+                     int main(){
+                         poppler::version_string();
+                         return 0;
+                     }],
+                    [AC_DEFINE(HAVE_POPPLER, 1, [Define HAVE_POPPLER if the library is available])
+                     OMINDEX_MODULES="$OMINDEX_MODULES omindex_pdf"])
+     LIBS=$save_LIBS],[echo 'Disabled worker process related to Poppler'])
```
ojwb: You shouldn't use `echo` here.

The correct way to handle this is to use something like `AC_MSG_CHECKING([if libpoppler is usable])` before your test (you probably want it just before the `save_LIBS=...` here, and then `AC_MSG_RESULT([yes])` and `AC_MSG_RESULT([no])` in the _ACTION_IF_TRUE_ and _ACTION_IF_FALSE_ for the `AC_LINK_IFELSE`.

xapian-applications/omega/worker.cc
barufa: Okay, so Should I revert this change in a fix-up commit?

xapian-applications/omega/handler.h
```diff
@@ -0,0 +1,51 @@
...
+ *  @param title      String which will hold the title of the file.
+ *  @param keywords   String where the keywords will be stored.
+ *  @param author     String where the author will be stored.
+ *  @param pages      Here the number of pages will be stored.
+ *
+ * Note: This function should only be used by an assistant process.
+ *
+ * See Worker::extract() for more details.
+ *
+*/
```
ojwb: Some minor nits:

* Don't put "empty" `*` lines at the end of a multi-line comment
* The closing `*/` should be indented one column so the `*` lines up with those above
* Don't put a blank line between the documentation comment and the function it documents
* The `@internal` marker probably isn't useful here - there's no external API to omega, so all functions are internal.

xapian-applications/omega/handler_poppler.cc
barufa: Sorry, I forget that. I am going to change it.

xapian-applications/omega/assistant.cc
barufa: It may be a very extreme case, but I was not sure what could happen if an assistant process started with the pid of an old assistant process that had ended due to an error (such as SEGFAULT). If this were the case, the new assistant could fail for the old timeout (Assuming that the operating system does not eliminate the signal when a process ends). So I cancel any request just in case.
However, I didn't realize that the alarm would be ignore as the handler has not been set yet.
This line of code is unnecessary. I am going to remove it in the next commit.

xapian-applications/omega/index_file.cc
```diff
@@ -635,7 +646,16 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	}
     }
     try {
-	if (cmd_it != commands.end()) {
+	if (wrk_it != workers.end()) {
+	    // Just use the worker process to extrac the content
```
ojwb: Typo: "extrac"

xapian-applications/omega/index_file.h
```diff
@@ -29,6 +29,7 @@
 #include <map>
 #include <string>
 #include <xapian.h>
+#include "worker.h"
```
ojwb: You don't need to include this here - only a forward declaration (i.e. `class Worker;`) is needed to be able to work with pointers or references to a type (if you dereference then a full declaration is needed).

The benefit of avoiding including headers when they aren't really needed is that fewer things need to rebuilt when that header is modified.  That's especially beneficial for an include in another header -

xapian-applications/omega/worker.cc
```diff
@@ -0,0 +1,185 @@
...
+	if (devnull > 3) close(devnull);
+
+	// FIXME: For filters which support a file descriptor as input, we
+	// could open the file here, and pass the file descriptor across the
+	// socket to the worker process using sendmsg().  Then the worker
+	// process could be chroot()-ed to a sandbox directory, which means
+	// we're reasonably protected from security bugs in the filter.
+
+	const char * mod;
+	if (filter_module.find('/') == string::npos && false) {
```
ojwb: `&& false`?

ojwb: If that was for testing in an uninstalled tree, you can probably achieve what you want using `export XAPIAN_OMEGA_PKGLIBBINDIR=$PWD` in the shell you're testing from - this overrides what `get_pkglibbindir()` returns.

xapian-applications/omega/configure.ac
barufa: Actually, I need to do something if PKG_CHECK_MODULES cannot find poppler-cpp. If I remove this 'echo' and poppler-cpp is not available, PKG_CHECK_MODULES will end the execution with an error for not having found the dependency.
Probably I should remove echo and use something else.
Maybe I can use AC_MSG_CHECKING before PKG_CHECK_MODULES and AC_MSG_RESULT([no]) if PKG_CHECK_MODULES or AC_LINK_IFELSE fail.

xapian-applications/omega/handler.h
barufa: Okay, I am going to change that in the next commit

xapian-applications/omega/configure.ac
ojwb: > Actually, I need to do something if PKG_CHECK_MODULES cannot find poppler-cpp. If I remove this 'echo' and poppler-cpp is not available, PKG_CHECK_MODULES will end the execution with an error for not having found the dependency.
> Probably I should remove echo and use something else.

You should just be able to specify `[ ]` there - then the action is a single space and the default won't be used.  `PKG_CHECK_MODULES` will report that poppler wasn't found for you.

(The `[` and `]` are m4 quotes.)

> Maybe I can use AC_MSG_CHECKING before PKG_CHECK_MODULES and AC_MSG_RESULT([no]) if PKG_CHECK_MODULES or AC_LINK_IFELSE fail.

That won't work well - `PKG_CHECK_MODULES` itself uses `AC_MSG_CHECKING` and `AC_MSG_RESULT` around its check, so the messages will appear interleaved.

xapian-applications/omega/worker.cc
```diff
@@ -0,0 +1,185 @@
...
+	    mod = full_path.c_str();
+	} else {
+	    mod = filter_module.c_str();
+	}
+
+#ifdef HAVE_SETRLIMIT
+#if defined RLIMIT_AS || defined RLIMIT_VMEM || defined RLIMIT_DATA
+	// Set a memory limit if it is possible
+	long mem = get_free_physical_memory();
+	if (mem>0) {
```
ojwb: There should be spaces around `>` here.

xapian-applications/omega/worker.h
```diff
@@ -0,0 +1,87 @@
...
+    static bool ignoring_sigpipe;
+    /// @private PID of the assistant process.
+    pid_t child;
+    /** @private Socket for supporting communication between the worker
+     *	and its assistant.
+    */
+    std::FILE * sockt;
+    /// @private Name of the assistant program.
+    std::string filter_module;
+    /// @private This method create the assistant subprocess.
```
ojwb: You don't need `@private` here - these are `private` (that's the default for `class` before any access specifier) and doxygen knows that.  You only need to say `@private` is you want to tell doxygen to treat something as having a different access to what it actually has (e.g. in the xapian-core API we sometimes make the `internal` member public to avoid having to forward declare internal classes in the public API header, but we want users to treat it as if it is private so we use `@private` so doxygen doesn't show it as being public.)

xapian-applications/omega/worker_comms.cc
ojwb: If you've made significant changes add a "Copyright (C) 2019 Bruno Baruffaldi" after the existing "Copyright ..." lines.

The exact criteria for copyrightability are a bit complex, but the work you're doing during GSoC should definitely count.

xapian-applications/omega/worker.cc
barufa: Okay, I am going to change that.

xapian-applications/omega/configure.ac
barufa: I didn't realize of that. I remove the echo message.

xapian-applications/omega/index_file.cc
barufa: Sorry, I didn't see it.

xapian-applications/omega/index_file.h
barufa: Okay, I am going to change it.

xapian-applications/omega/worker.cc
barufa: Sorry, I forget about that. I will try to be more careful.

xapian-applications/omega/worker.h
barufa: Okay, I am going to fix it. I believed that I need to told doxygen that it was private.

ojwb: Actually you do for some languages (which don't have the concept in the language itself), but not for C++:

http://www.doxygen.nl/manual/commands.html#cmdprivate says:

> Note that doxygen automatically detects the protection level of members in object-oriented languages. This command is intended for use only when the language does not support the concept of protection level natively (e.g. C, PHP 4).

ojwb:
In case you didn't see it, `xapian-check-patch` is flagging some style errors (I recently added some additional checks to it) - you can see them here:

https://travis-ci.org/xapian/xapian/jobs/551982422#L197

It can also be run locally which it's good to get into the habit of doing before pushing.

barufa:
Yes, I know that. But it seems that I was using an old version of xapian-check-patch.
I am going to fix these errors.

xapian-applications/omega/assistant.cc
ojwb: I'd expect the operating system to clear such state when a process exits.

If it didn't, you'd see problems more widely as processes would get apparently random SIGALRM signals, and the default handling for SIGALRM is to terminate the process.

Good to be thinking about how things can go wrong though.

ojwb:
The version of `xapian-check-patch` on your branch would be older.

But you can grab the latest version from master and save it somewhere handy to use instead - e.g. if you have a `~/bin` directory:

```
wget -O ~/bin/xapian-check-patch 'https://git.xapian.org/?p=xapian;a=blob_plain;f=xapian-maintainer-tools/xapian-check-patch;hb=HEAD'
```

And then use it like:

```
git diff master.. | ~/bin/xapian-check-patch
```

For a PR, CI (at least for travis, not sure about appveyor) gets run on the PR branch merged with the branch the PR targets (usually master, as here).  So that's why travis runs the latest `xapian-check-patch` from master.

xapian-applications/omega/configure.ac
barufa: This is solved. We discuss it through IRC channel.

ojwb:
Rebased, auto-squashed and merged to master as c47a51a031b66c2eadde2cfaf1946262fcbe5c57.

On IRC you said:
> I find an error at the first commit, its message is wrong (I am using 'Adding' instead of 'Add')

I fixed that and reformatted the commit message (there should be a blank line between the summary and the more detailed explanation) before merging it.

PR #248, opened 2019-06-03
bootstrap: handle being a git submodule

brong:
This commit gets applied for the Cyruslibs build as part of the cyrusimap project because we pull xapian in as a git submodule.  It just replaces all use of `.git` in the bootstrap script with $gitdir, which it gets from `git rev-parse --git-dir`.

```
       --git-dir
           Show $GIT_DIR if defined. Otherwise show the path to the .git
           directory. The path shown, when relative, is relative to the
           current working directory.

           If $GIT_DIR is not defined and the current directory is not
           detected to lie in a Git repository or work tree print a message to
           stderr and exit with nonzero status.
```

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -407,9 +407,11 @@ swig_git_commit_hash=1e36f51346d95f8b9848e682c2eb986e9cb9b4f4
 omega_common_commit_hash=c65fd56b6ddd11894253b16bc1d0701ff6835672
 letor_common_commit_hash=c65fd56b6ddd11894253b16bc1d0701ff6835672
 
-if [ ! -r .git ] ; then
-  echo "$0: '.git' not found - this script should be run from a git repo cloned"
-  echo "from git://git.xapian.org/xapian or a mirror of it"
+gitdir=`git rev-parse --git-dir`
+
+if [ ! -d "$gitdir" ] ; then
+  echo "$0: No '.git' directory found - this script should be run from a"
+  echo "git repo cloned from git://git.xapian.org/xapian or a mirror of it"
   exit 1
 fi
 
@@ -424,25 +426,21 @@ for emptydir in xapian-applications/omega/m4 xapian-bindings/m4 xapian-letor/m4 
   fi
 done
 
-if [ -f .git ] ; then
-  : # Looks like we're in a git worktree.
+if [ -f "$gitdir/info/exclude" ] ; then
+  sed '/^\(swig\|xapian-applications\/omega\/common$\)/d' "$gitdir/info/exclude" > "$gitdir/info/exclude~"
 else
-  if [ -f .git/info/exclude ] ; then
-    sed '/^\(swig\|xapian-applications\/omega\/common$\)/d' .git/info/exclude > .git/info/exclude~
-  else
-    [ -d .git/info ] || mkdir .git/info
-  fi
-  cat <<END >> .git/info/exclude~
+  [ -d "$gitdir/info" ] || mkdir "$gitdir/info"
+fi
+cat <<END >> "$gitdir/info/exclude~"
 swig
 xapian-applications/omega/common
 xapian-letor/common
 END
-  if [ -f .git/info/exclude ] &&
-     cmp -s .git/info/exclude~ .git/info/exclude ; then
-    rm .git/info/exclude~
-  else
-    mv .git/info/exclude~ .git/info/exclude
-  fi
+if [ -f "$gitdir/info/exclude" ] &&
+   cmp -s "$gitdir/info/exclude~" "$gitdir/info/exclude" ; then
+  rm "$gitdir/info/exclude~"
+else
+  mv "$gitdir/info/exclude~" "$gitdir/info/exclude"
 fi
 
 # If this tree is checked out from the github mirror, use the same access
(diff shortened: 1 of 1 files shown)


bootstrap
```diff
@@ -190,6 +190,10 @@ lazy_build() {
     # then we'll skip the worktree handling because main_worktree_dir will
     # be empty.
     main_worktree_dir=`git worktree list --porcelain 2>/dev/null|sed 's/^worktree //;q'`
+    # If gitdir isn't the default value, assume that we are in a submodule.
+    if [ "$gitdir" != ".git" ] ; then
+      main_worktree_dir=`echo $gitdir|sed 's/\/.git\/modules\//\//;q'`
+    fi
```
ojwb: I'm a bit unsure about this part.  What does `git worktree list --porcelain` report in your situation?

I worry that `$gitdir` can be something other than `.git` in other situations.  For example, if `$GIT_DIR` is set I get a full path on it:

```
olly@gemse:~/git/xapian$ GIT_DIR=/home/olly/git/xapian/.git git rev-parse --git-dir
/home/olly/git/xapian/.git
```

Perhaps it should be testing for the magic substring instead, which is easy to do with `case`:

```
case $gitdir in
*/.git/modules/*)
    main_worktree_dir=`...` ;;
esac
```

I guess this won't handle using submodules and worktrees together.  It'd be nicer if that worked, though if you don't use worktrees I guess that can be left for someone who wants both to take care of.

Also, I'm wondering why there's a `;q` in the `sed` expression here, since I'd have thought this would always be a single line (unless the path actually contains a newline in which case attempting the replace on all lines is more likely to do the right thing) - I'm suspecting it was just copied from the `sed` command just above, where we are explicitly wanting to cull all lines except the first - the output there looks like this:

```
$ git worktree list --porcelain 
worktree /home/olly/git/xapian
HEAD def7279ac6884346e746d063641f13b9fd5fbf35
branch refs/heads/master

worktree /home/olly/git/xapian-1.4
HEAD eea3d616a7371ca97f8615f0d226697fe2f94608
branch refs/heads/RELEASE/1.4
```

I'd also pick a different delimiter to avoid [leaning toothpick syndrome](https://en.wikipedia.org/wiki/Leaning_toothpick_syndrome), so something like: `sed 's|/.git/modules/|/|'`

bootstrap
```diff
@@ -406,7 +410,8 @@ swig_git_commit_hash=613ff08150e6264a5a4a964f2badeb6bba221398
 omega_common_commit_hash=f5921b74bc4bf899d9a043ee3ecdbcbfd4bacb7b
 letor_common_commit_hash=f5921b74bc4bf899d9a043ee3ecdbcbfd4bacb7b
 
-if [ ! -r .git ] ; then
+gitdir=`git rev-parse --git-dir`
+if [ ! -r $gitdir ] ; then
```
ojwb: `$gitdir` should really be `"$gitdir"` everywhere in case it contains spaces.

bootstrap
```diff
@@ -406,7 +410,8 @@ swig_git_commit_hash=613ff08150e6264a5a4a964f2badeb6bba221398
...
   echo "$0: '.git' not found - this script should be run from a git repo cloned"
```
ojwb: The error message should probably use `$gitdir` too.

ojwb: Oh, except that plays out unhelpfully if we're not in a checkout:

```
$ git rev-parse --git-dir
fatal: not a git repository (or any of the parent directories): .git
```

So the user would then see:

```
$ ./bootstrap
fatal: not a git repository (or any of the parent directories): .git
./bootstrap: '' not found - this script should be run from a git repo cloned
```

I think we should now to able to rely on `git rev-parse --git-dir` to take care of checking we're actually in a git repo and to report the issue if not, and so we just need to check `$gitdir` isn't empty, and we can drop the `'.git' not found -` part from this error message.  Does that seem good to you?

ojwb:
Also, are you OK with the patch licensing policy: https://xapian-developer-guide.readthedocs.io/en/latest/contributing/licensing.html

ojwb:
@brong Ping?

I can't do anything with this without an agreement to license it suitably.

Also I don't have a repo using ours as a submodule to test with to try to answer the questions above for myself.  I guess I could try to create one, but I don't have any experience of using submodules (I looked at them and concluded they didn't work in a way which was useful for Xapian's needs).

brong:
Oh - I'm sorry I missed these comments.  Yes, I'm happy with the licensing under GPL2 and MIT, as is my employer.

You can see where it's being used as a submodule in the github.com/cyrusimap/cyruslibs respository.

bootstrap
brong: Yep, that seems fine.  I'm happy to update this patch and resend it, though it will have to wait for it not to be nearly midnight for me :)

FYI: I also ran into an issue where as a submodule it's not finding the right path to download autoconf and other modules, so I'll do some more testing with that too - though this bit is still clearly better than it was before for the submodule case.

bootstrap
brong: yes, good point!  I'll definitely fix that up.

bootstrap
brong: This I'll have to test some more.  Let me get back to you on that.

Sorry for not replying when these came through - I was traveling when the initial reviews happened and clearly scrolled past these when I got back!

bootstrap
ojwb: >  I'm happy to update this patch and resend it, though it will have to wait for it not to be nearly midnight for me :)

I guess this dropped off your radar.

I think you're better placed to finish this off than me as you actually use this configuration, so if you're still happy to update it that would be great.

brong:
Coming back to this after a long time!

I've gone through and put double quotes around every usage of $gitdir.  I think that's the only thing that matters.

ojwb:
Thanks, merged via git.xapian.org.

PR #249, opened 2019-06-04
Add lcov support for letor

VaibhavKansagara:


xapian-letor/Makefile.am
```diff
@@ -37,6 +37,30 @@ EXTRA_PROGRAMS =
 check-@PERCENT@:
 	cd tests && $(MAKE) $(AM_MAKEFLAGS) $@
 
+## --disable-shared as shared builds fail for me on x86-64 Debian etch.
```
jaylett: Debian _etch_? That was discontinued over 9 years ago! We should reassess this and update both here and in `-core` (or at least update the comment).

ojwb:
See https://xapian-developer-guide.readthedocs.io/en/latest/conventions/portability.html?highlight=.PHONY for details as to why.

xapian-letor/Makefile.am
ojwb: And that comment originates in 06453ade9f9954fa826d4b8299b87d425fa24cbf over 10 years ago.

We should certainly reassess as a lot of time has passed, though I'd be OK to merge with the copied comment and then address this as a follow-on patch (the comment isn't actually wrong, just about really old versions).  Once these targets are available I can add letor coverage reports to lcov.xapian.org.

ojwb: I kicked off a xapian-core coverage build on Debian unstable with `--disable-shared` removed and it completed with a plausible coverage report generated.  It might be it fails on older platforms which are still in use though, so wider testing is probably needed.

Perhaps we should refactor to allow specifying arbitrary extra configure options, e.g.:

```
make coverage-reconfigure COVERAGE_CONFIGURE_ARGS=--disable-shared
```

ojwb:
The CI build with STL debugging enabled shows a bit more about one of the failures:

https://travis-ci.org/xapian/xapian/jobs/541366835#L9144

We can decode the type given like so:

```
$ c++filt _ZNSt7__debug6vectorISt4pairIN6Xapian4MSetEjESaIS4_EEE
std::__debug::vector<std::pair<Xapian::MSet, unsigned int>, std::allocator<std::pair<Xapian::MSet, unsigned int> > >
```

And `git grep 'pair<.*MSet'` finds the code in `matcher/matcher.cc` - looks like the loop to merge MSet objects doesn't handle the case where there are no MSet objects to merge (which it seems can happen when they were all empty).  I'll test this patch:

```
diff --git a/xapian-core/matcher/matcher.cc b/xapian-core/matcher/matcher.cc
index 4ba4f25c3a39..795d6df2fe73 100644
--- a/xapian-core/matcher/matcher.cc
+++ b/xapian-core/matcher/matcher.cc
@@ -644,7 +644,7 @@ Matcher::get_mset(Xapian::doccount first,
 
     CollapserLite collapser(collapse_max);
     merged_mset.internal->first = first;
-    while (merged_mset.size() != maxitems) {
+    while (!msets.empty() && merged_mset.size() != maxitems) {
 	auto& front = msets.front();
 	auto& result = front.first.internal->items[front.second];
 	if (percent_threshold) {
@@ -665,8 +665,6 @@ Matcher::get_mset(Xapian::doccount first,
 	if (n == msets.front().first.size()) {
 	    Heap::pop(msets.begin(), msets.end(), heap_cmp);
 	    msets.resize(msets.size() - 1);
-	    if (msets.empty())
-		break;
 	} else {
 	    msets.front().second = n;
 	    Heap::replace(msets.begin(), msets.end(), heap_cmp);
```

ojwb:
Autosquashed, rebased and pushed as 5329d3966a159b4f952b06d8e827ec6b5798c960.

xapian-letor/Makefile.am
ojwb: The VM which hosts lcov.xapian.org runs Debian stretch and that works without `--disable-shared` so I think removing it by default makes sense.  Generating coverage reports is really a developer thing, and developers should be running relatively up-to-date systems.  I'll push that change.

ojwb: Done, and also updated the developer guide.

PR #250, opened 2019-06-04
Add remoteprog support to BackendManagerMulti

tanmaysachan:
BackendManagerMulti now supports a remoteprog db
as an alternate sub manager alongside glass.
[WIP]

### DIFF
--- a/xapian-core/tests/api_backend.cc
+++ b/xapian-core/tests/api_backend.cc
@@ -169,7 +169,7 @@ DEFINE_TESTCASE(dbstats1, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote") && !startswith(get_dbtype(), "multi_glass_remote")) {
+    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")) {
 	TEST_EQUAL(db.get_wdf_upper_bound("the"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
@@ -213,7 +213,7 @@ DEFINE_TESTCASE(dbstats2, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote") && !startswith(get_dbtype(), "multi_glass_remote")) {
+    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")) {
 	TEST_EQUAL(db.get_wdf_upper_bound("word"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
--- a/xapian-core/tests/api_backend.cc
+++ b/xapian-core/tests/api_backend.cc
@@ -169,7 +169,7 @@ DEFINE_TESTCASE(dbstats1, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")) {
+    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote") && !startswith(get_dbtype(), "multi_glass_remote")) {
 	TEST_EQUAL(db.get_wdf_upper_bound("the"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
@@ -213,7 +213,7 @@ DEFINE_TESTCASE(dbstats2, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")) {
+    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote") && !startswith(get_dbtype(), "multi_glass_remote")) {
 	TEST_EQUAL(db.get_wdf_upper_bound("word"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
--- a/xapian-core/tests/api_backend.cc
+++ b/xapian-core/tests/api_backend.cc
@@ -169,7 +169,8 @@ DEFINE_TESTCASE(dbstats1, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")) {
+    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")
+        && !endswith(get_dbtype(), "remoteprog_glass")) {
 	TEST_EQUAL(db.get_wdf_upper_bound("the"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
@@ -213,7 +214,8 @@ DEFINE_TESTCASE(dbstats2, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")) {
+    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")
+        && !endswith(get_dbtype(), "remoteprog_glass")) {
 	TEST_EQUAL(db.get_wdf_upper_bound("word"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
--- a/xapian-core/tests/api_backend.cc
+++ b/xapian-core/tests/api_backend.cc
@@ -169,8 +169,8 @@ DEFINE_TESTCASE(dbstats1, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")
-        && !endswith(get_dbtype(), "remoteprog_glass")) {
+    if (get_dbtype() != "inmemory" &&
+	get_dbtype().find("remote") == string::npos) {
 	TEST_EQUAL(db.get_wdf_upper_bound("the"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
@@ -214,8 +214,8 @@ DEFINE_TESTCASE(dbstats2, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")
-        && !endswith(get_dbtype(), "remoteprog_glass")) {
+    if (get_dbtype() != "inmemory" &&
+	get_dbtype().find("remote") == string::npos) {
 	TEST_EQUAL(db.get_wdf_upper_bound("word"), max_wdf);
     } else {
 	// For inmemory and remote backends, we usually give rather loose
--- a/xapian-core/tests/harness/backendmanager.h
+++ b/xapian-core/tests/harness/backendmanager.h
@@ -139,6 +139,13 @@ class BackendManager {
     /// Get a remote database instance with the specified timeout.
     virtual Xapian::Database get_remote_database(const std::vector<std::string> & files, unsigned int timeout);
 
+     /*
+     * Get the args for opening a writable remote database with the
+     * specified timeout.
+     */
+    virtual std::string get_writable_database_args(const std::string& path,
+						   unsigned int timeout);
+
     /// Create a Database object for the last opened WritableDatabase.
     virtual Xapian::Database get_writable_database_as_database();
 
--- a/xapian-core/tests/harness/backendmanager.h
+++ b/xapian-core/tests/harness/backendmanager.h
@@ -123,6 +123,10 @@ class BackendManager {
     /// Get a writable database instance.
     virtual Xapian::WritableDatabase get_writable_database(const std::string & name, const std::string & file);
 
+    /// Get a remote Xapian::WritableDatabase instance with specified args.
+    virtual Xapian::WritableDatabase
+    get_remote_writable_database(std::string args);
+
     /// Get the path of a writable database instance, if such a thing exists.
     virtual std::string get_writable_database_path(const std::string & name);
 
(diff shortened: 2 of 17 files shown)


ojwb:
The CI build with STL debugging enabled shows a bit more about one of the failures:

https://travis-ci.org/xapian/xapian/jobs/541366835#L9144

We can decode the type given like so:

```
$ c++filt _ZNSt7__debug6vectorISt4pairIN6Xapian4MSetEjESaIS4_EEE
std::__debug::vector<std::pair<Xapian::MSet, unsigned int>, std::allocator<std::pair<Xapian::MSet, unsigned int> > >
```

And `git grep 'pair<.*MSet'` finds the code in `matcher/matcher.cc` - looks like the loop to merge MSet objects doesn't handle the case where there are no MSet objects to merge (which it seems can happen when they were all empty).  This patch seems to solve that - I'll push it to master shortly (I just want to look over the code to make sure there aren't any other similar problems):

```
diff --git a/xapian-core/matcher/matcher.cc b/xapian-core/matcher/matcher.cc
index 4ba4f25c3a39..795d6df2fe73 100644
--- a/xapian-core/matcher/matcher.cc
+++ b/xapian-core/matcher/matcher.cc
@@ -644,7 +644,7 @@ Matcher::get_mset(Xapian::doccount first,
 
     CollapserLite collapser(collapse_max);
     merged_mset.internal->first = first;
-    while (merged_mset.size() != maxitems) {
+    while (!msets.empty() && merged_mset.size() != maxitems) {
 	auto& front = msets.front();
 	auto& result = front.first.internal->items[front.second];
 	if (percent_threshold) {
@@ -665,8 +665,6 @@ Matcher::get_mset(Xapian::doccount first,
 	if (n == msets.front().first.size()) {
 	    Heap::pop(msets.begin(), msets.end(), heap_cmp);
 	    msets.resize(msets.size() - 1);
-	    if (msets.empty())
-		break;
 	} else {
 	    msets.front().second = n;
 	    Heap::replace(msets.begin(), msets.end(), heap_cmp);
```

ojwb:
Patch looks good after a deeper look so pushed as aabdb132390c39c2ed29089fb0f7bd75e2a5a927.

Merging your branch with master (as the travis CI builds do) I now see the following failures:

```
$ ./runtest ./apitest -v -bmulti_glass_remoteprog_glass
Running test './apitest -v -bmulti_glass_remoteprog_glass' under valgrind
Running tests with backend "multi_glass_remoteprog_glass"...
Running test: msetfirst1... FAILED
harness/testutils.cc:58: mset2.size() >= first2 + count
mset2 is too small: expected at least 3 items, got 0


Running test: eliteset1... FAILED
api_anydb.cc:1309: ((mymset1) == (mymset2))
Expected 'mymset1' and 'mymset2' to be equal: were MSet(matches_lower_bound=2, matches_estimated=2, matches_upper_bound=2, max_possible=1.2414088043957824947, max_attained=1.0464816871772451012, [Result(2, 1.0464816871772451012), Result(4, 0.64098768659591376373)]) and MSet(matches_lower_bound=3, matches_estimated=3, matches_upper_bound=3, max_possible=1.2414088043957824947, max_attained=1.0464816871772451012, [Result(2, 1.0464816871772451012), Result(4, 0.64098768659591376373), Result(5, 0.46338869498075929698)])


Running test: eliteset2... FAILED
api_anydb.cc:1333: ((mymset1) == (mymset2))
Expected 'mymset1' and 'mymset2' to be equal: were MSet(matches_lower_bound=1, matches_estimated=1, matches_upper_bound=1, max_possible=3.7543504109652774403, max_attained=1.9700834242150864206, [Result(2, 1.9700834242150864206)]) and MSet(matches_lower_bound=4, matches_estimated=4, matches_upper_bound=4, max_possible=3.7543504109652774403, max_attained=1.9700834242150864206, [Result(2, 1.9700834242150864206), Result(1, 0.051103097360122355652), Result(3, 0.043131803408968126534), Result(5, 0.043131803408968126534)])


Running test: eliteset4... FAILED
api_anydb.cc:1408: ((mymset1) == (mymset2))
Expected 'mymset1' and 'mymset2' to be equal: were MSet(matches_lower_bound=1, matches_estimated=1, matches_upper_bound=1, max_possible=2.5129416065694951676, max_attained=1.4848948390060123792, [Result(3, 1.4848948390060123792)]) and MSet(matches_lower_bound=1, matches_estimated=1, matches_upper_bound=1, max_possible=4.9608653679271013104, max_attained=1.4848948390060123792, [Result(3, 1.4848948390060123792)])


Running test: qterminfo1... FAILED
api_anydb.cc:1534: ((mymset1a.get_termweight(term3)) == (0))
Expected 'mymset1a.get_termweight(term3)' and '0' to be equal: were 5.49729 and 0


Running test: matches1... FAILED
api_anydb.cc:1679: ((mymset.get_matches_estimated()) == (2))
Expected 'mymset.get_matches_estimated()' and '2' to be equal: were 1 and 2


Running test: tradweight1... FAILED
api_anydb.cc:2373: ((mset[0].get_weight()) == (mset[5].get_weight()))
Expected 'mset[0].get_weight()' and 'mset[5].get_weight()' to be equal: were 0.0377403 and 0.0377403


Running test: collapsekey5...     
```

... and it seems to hang there...

`tradweight1` looks like some sort of small rounding difference as the reported values are the same.

ojwb:
This is now showing merge conflicts with master, which will be due to #254 getting merged and touching some of the same places in the same files.

The obvious options are to merge master into the branch or rebase the branch onto master.  The rebase option would then require force-pushing which is usually unhelpful, but I think here it's probably the better way to go as we probably want to rebase before merging anyway.

I tried both options and both result in the same conflict near the start of `harness/backendmanager_remote.cc` - the branch adds `BackendManagerRemote::get_writable_database_args_from_path(const std::string & path)` and master adds two different methods in the same place.  However master also adds `BackendManagerRemote::get_remote_database_args(const std::string& path, unsigned int timeout)` lower down which does essentially the same things as `BackendManagerRemote::get_writable_database_args_from_path(const std::string & path)` except the timeout can also be specified.  So I think we want to just take the changes from master to this file, and then fix the branch to use the method added to master.

If you need help with getting this rebased, just ask on IRC.

tanmaysachan:
I made a few changes to the original commit, the `sub_manager_alt` is now of BackendManagerRemote to get access to functions like get_writable_database_args etc.
I can clean the code later, after fixing the failing testcases.

ojwb:
I've pushed a fix for the `collapsekey5` infinite loop to master (74ab8407ee3641ff07196a8ff92af9043fd299ac).

I suggest you cherry-pick that commit to your PR branch - then it should drop out cleanly when I `git rebase` your branch before merging:

```
git checkout RemoveBackendmanager_local
git fetch
git cherry-pick -x 74ab8407ee3641ff07196a8ff92af9043fd299ac
```

(`-x` means the commit hash that was cherry-picked is automatically recorded in the commit message, which is helpful unless you're doing something like cherry-picking from a temporary local branch).

ojwb:
> I made a few changes to the original commit, the `sub_manager_alt` is now of BackendManagerRemote to get access to functions like get_writable_database_args etc.

That seems the wrong approach to me - `BackendManagerMulti` should really be able to work with arbitrary combinations of submanagers (e.g. one glass, one inmemory).  However, this change means that if there are two different submanagers, the second one must be remote.

There shouldn't be code in `BackendManagerMulti` which requires `sub_manager_alt` to be a remote manager - instead code specific to handling a particular backend should really all be in the backend manager subclass for that backend (so remote-specific code should be in `BackendManagerRemote`) with virtual method dispatch used to select the appropriate behaviour.

> I can clean the code later, after fixing the failing testcases.

I think it's good to focus on solving the remaining failures, but we should sort out the above after that.

xapian-core/tests/api_backend.cc
```diff
@@ -169,7 +169,8 @@ DEFINE_TESTCASE(dbstats1, backend) {
 	TEST_REL(db.get_doclength_lower_bound(),<=,min_len);
     }
 
-    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")) {
+    if (get_dbtype() != "inmemory" && !startswith(get_dbtype(), "remote")
+        && !endswith(get_dbtype(), "remoteprog_glass")) {
```
ojwb: Xapian coding style is to wrap an expression after the operator, so the `&&` should be on the end of the first line here, not the start of the second.

But I think probably the test here should be that "remote" is not a substring of the dbtype (so `get_dbtype().find("remote") == string::npos` or something like that).  That should get all the cases it should without any false matches in practice, and avoids a growing list of more precise checks.

xapian-core/tests/harness/backendmanager.h
```diff
@@ -139,6 +139,13 @@ class BackendManager {
     /// Get a remote database instance with the specified timeout.
     virtual Xapian::Database get_remote_database(const std::vector<std::string> & files, unsigned int timeout);
 
+     /*
+     * Get the args for opening a writable remote database with the
+     * specified timeout.
+     */
+    virtual std::string get_writable_database_args(const std::string& path,
```
ojwb: The first line of the comment here is indented by an extra space, so the `*` doesn't align with the following lines.

But also this really should be a doxygen documentation comment which (for a multiline comment) start with `/**`, with the brief description following on the same line.

xapian-core/tests/harness/backendmanager_multi.h
```diff
@@ -23,6 +23,7 @@
 #define XAPIAN_INCLUDED_BACKENDMANAGER_MULTI_H
 
 #include "backendmanager.h"
+#include "backendmanager_remote.h"
```
ojwb: This doesn't seem to actually be needed now.

xapian-core/tests/harness/backendmanager_multi.h
```diff
@@ -39,7 +40,8 @@ class BackendManagerMulti : public BackendManager {
     /// The path of the last writable database used.
     std::string last_wdb_path;
 
-    BackendManager* sub_manager;
+    // vector of sub_managers.
+    std::vector<BackendManager*> sub_managers;
```
ojwb: You should `#include <vector>` if you're using `std::vector` (I guess some header being included is including `<vector>` for you, but you shouldn't rely on such things.

xapian-core/tests/harness/backendmanager_remote.h
```diff
@@ -47,6 +47,12 @@ class BackendManagerRemote : public BackendManager {
     std::string get_writable_database_args(const std::string & name,
 					   const std::string & file);
 
+    /*
+     * Get the args for opening a writable remote database with the
+     * specified timeout.
+     */
```
ojwb: This should also be a documentation comment.

xapian-core/tests/harness/backendmanager_multi.cc
```diff
@@ -91,28 +113,44 @@ BackendManagerMulti::createdb_multi(const string& name,
...
+#ifdef HAVE_VALGRIND
+	    if (RUNNING_ON_VALGRIND) {
+		args.insert(0, XAPIAN_PROGSRV" ");
+		dbs.add_database(Xapian::Remote::open_writable("./runsrv", args));
+		out << "remote :" << args << '\n';
+	    }
+#else
+	    dbs.add_database(Xapian::Remote::open_writable(XAPIAN_PROGSRV, args));
+	    out << "remote :" << XAPIAN_PROGSRV << " " << args << '\n';
+#endif
```
ojwb: This seems like way too much knowledge about remote databases to be here rather than in `BackendManagerRemote`.

xapian-core/tests/harness/backendmanager_multi.cc
```diff
@@ -91,28 +113,44 @@ BackendManagerMulti::createdb_multi(const string& name,
...
-	dbbase += str(n);
-	dbs.add_database(Xapian::WritableDatabase(dbbase, flags));
+	const string &subtype = sub_managers[n]->get_dbtype();
+	int flags = Xapian::DB_CREATE_OR_OVERWRITE;
+	if (subtype == "glass") {
+	    flags |= Xapian::DB_BACKEND_GLASS;
+	    dbbase += str(n);
+	    dbs.add_database(Xapian::WritableDatabase(dbbase, flags));
+	    string line = subtype + " " + dbname + "___" + str(n);
+	    out << line << '\n';
```
ojwb: There's no reason to put the line to output in a string now since we rebuild it each time from scratch anyway - just write the elements to the `out` stream directly.

xapian-core/tests/harness/backendmanager_multi.cc
```diff
@@ -22,6 +22,7 @@
 #include <config.h>
 
 #include "backendmanager_multi.h"
+#include "backendmanager_remote.h"
```
ojwb: This doesn't seem to be used here either now.

xapian-core/tests/harness/backendmanager_multi.cc
```diff
@@ -32,22 +33,43 @@
 #include <cstdio> // For rename().
 #include <cstring>
 
+#ifdef HAVE_VALGRIND
+# include <valgrind/memcheck.h>
+#endif
+
```
ojwb: I don't think this is needed here now.

ojwb:
The style-checking script is complaining about two things:

```
xapian-core/tests/harness/backendmanager.cc:269: error: Line extends beyond column 80 (to column 86): +    string msg = "BackendManager::get_writable_database_args() called for non-remote "
xapian-core/tests/harness/backendmanager_multi.cc:122: error: Preferred style is 'int& x' (not 'int &x' or 'int & x'): +	const string &subtype = sub_managers[n]->get_dbtype();
```

We also need to do something about the new testsuite failures uncovered by these changes before we can merge them.  If they're existing bugs which are just being revealed by the improved coverage, then marking them as `XFAIL` for now seems reasonable.  Doing that would actually make it easier to investigate them.

I think `postingsourceshardindex1` should just be marked as `multi && !remote` though - it would need the `PostingSource` subclass it uses to support serialisation (which is easy) but also it would need to be registered with the remote server, which we don't have a good way to achieve for the testsuite currently.  And it's mostly there as a regression test, so it's not particularly important to run it for more backend combinations.

ojwb:
CI is failing as there are some new generated files which aren't in `.gitignore`:

```
?? xapian-core/tests/.multiglassremoteprog_glass/
?? xapian-core/tests/.multiremoteprog_glass/
The generated files listed above are not in .gitignore
```

It's good to have generated files ignored as then `git status` doesn't show them by default, and it's harder to accidentally commit them with `git add`.

ojwb:
The cache directories for the new backend combinations should be removed by `make remove-cached-databases` - see this in `tests/Makefile.am`:

```
remove-cached-databases:
        rm -rf .glass .honey .multiglass .replicatmp .singlefileglass .stub
```

ojwb:
Now merged, with extra commits to:

* remove two XFAILs which are fixed by changes on master: 72c07f53e39f2c0499a60ce3fa14f22bdb5713c9
* work around appveyor test failures: e2a8011d09fc9a7fe40029ae44f32a8eb780f7fa

Thanks for all your work on this.

PR #251, opened 2019-06-05
Fix bug in colltfcolllenfeature.cc

VaibhavKansagara:
Before this fix CollTfCollLenFeature consisted of title,body and body
feature values instead of title,body and whole feature values.Also
this contains a regression test for the same.

### DIFF
--- a/xapian-letor/tests/api_letor.cc
+++ b/xapian-letor/tests/api_letor.cc
@@ -861,3 +861,65 @@ DEFINE_TESTCASE(tfdoclenfeature, generated)
 
     return true;
 }
+
+// Test for CollTfCollLenFeature
+DEFINE_TESTCASE(colltfcolllenfeature, generated)
+{
+    string db_path = get_database_path("apitest_ranker4",
+				       db_index_three_documents);
+    Xapian::CollTfCollLenFeature feature;
+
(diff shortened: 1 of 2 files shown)


xapian-letor/tests/api_letor.cc
```diff
@@ -861,3 +861,65 @@ DEFINE_TESTCASE(tfdoclenfeature, generated)
...
+    feature.set_database(db);
+
+    Xapian::Document doc = db.get_document(1);
+    feature.set_doc(doc);
+
+    std::map<std::string, Xapian::termcount> collection_length;
+
+    collection_length["title"] = 15;
+    Xapian::termcount whole_len = db.get_doclength(1) + db.get_doclength(2) +
+				  db.get_doclength(3);
```
ojwb: This statistic is actually directly available via `db.get_total_length()`:

```
    /** Get the total length of all the documents in the database.
     *
     *  @since Added in Xapian 1.4.5.
     */
    Xapian::totallength get_total_length() const; 
```

ojwb:
Great, thanks - autosquashed and rebased and pushed to master as e94ccffe3ec5994ff20843dc9a84a1ed8890711e.

PR #253, opened 2019-06-07
testsuite.cc: Add high_resolution_clock to track slow tests

tanmaysachan:
high_resolution_clock track tests taking more than 10000 ms.

### DIFF
--- a/xapian-core/tests/harness/testsuite.cc
+++ b/xapian-core/tests/harness/testsuite.cc
@@ -115,10 +115,10 @@ string test_driver::col_yellow, test_driver::col_reset;
 bool test_driver::use_cr = false;
 
 // time constant in seconds to mark tests as slow or not
-static const float SLOW_TEST_THRESHOLD = 10.00;
+static const double SLOW_TEST_THRESHOLD = 10.00;
 
 // vector to store the slow tests
-static vector<pair<string, float>> slow_tests;
+static vector<pair<string, double>> slow_tests;
 
 void
 test_driver::write_and_clear_tout()
@@ -707,7 +707,7 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
 	    auto starttime = chrono::high_resolution_clock::now();
 	    test_driver::test_result test_res = runtest(test);
 	    auto endtime = chrono::high_resolution_clock::now();
-	    auto test_duration = chrono::duration_cast<chrono::duration<float>>
+	    auto test_duration = chrono::duration_cast<chrono::duration<double>>
 				 (endtime - starttime);
 #ifndef NO_LIBXAPIAN
 	    if (backendmanager)
@@ -718,8 +718,8 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
 		    ++res.succeeded;
 
 		    if (test_duration.count() >= SLOW_TEST_THRESHOLD) {
-			slow_tests.push_back(make_pair(test->name,
-						       test_duration.count()));
+			slow_tests.emplace_back(test->name,
+						test_duration.count());
 		    }
 
 		    if (verbose || !use_cr) {
--- a/xapian-core/tests/harness/testsuite.cc
+++ b/xapian-core/tests/harness/testsuite.cc
@@ -820,8 +820,8 @@ test_driver::report(const test_driver::result &r, const string &desc)
 		}
 	    }
 	    cout << "." << endl;
+	    slow_tests.clear();
 	}
-	slow_tests.clear();
     }
 }
 
--- a/xapian-core/tests/harness/testsuite.cc
+++ b/xapian-core/tests/harness/testsuite.cc
@@ -714,16 +714,18 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
 		case PASS:
 		    ++res.succeeded;
 		    if (verbose || !use_cr) {
-			out << col_green << " ok" << col_reset << endl;
+			out << col_green << " ok" << col_reset;
 
 			if (test_duration.count() >= SLOW_TEST_THRESHOLD) {
 			    out << col_yellow << " SLOW TEST ("
 				<< test_duration.count()
-				<< " ms)" << col_reset << endl;
+				<< " ms)" << col_reset;
 			}
+
+			out << endl;
 		    } else {
 			out << "\r                                        "
-			    << "                                   \r";
+			       "                                       \r";
 		    }
 		    break;
 		case XFAIL:
--- a/xapian-core/tests/harness/testsuite.cc
+++ b/xapian-core/tests/harness/testsuite.cc
@@ -723,7 +723,7 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
 			}
 		    } else {
 			out << "\r                                        "
-			    << "	                               \r";
+			    << "                                   \r";
 		    }
 		    break;
 		case XFAIL:
--- a/xapian-core/tests/harness/testsuite.cc
+++ b/xapian-core/tests/harness/testsuite.cc
@@ -114,6 +114,9 @@ string test_driver::col_red, test_driver::col_green;
 string test_driver::col_yellow, test_driver::col_reset;
 bool test_driver::use_cr = false;
 
+// time constant to mark tests as slow or not
+static const int SLOW_TEST_THRESHOLD = 10000;
+
 void
 test_driver::write_and_clear_tout()
 {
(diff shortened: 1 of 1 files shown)


xapian-core/tests/harness/testsuite.cc
```diff
@@ -706,9 +711,20 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
...
+			    out << col_green << " ok" << col_reset << endl;
+			}
 		    } else {
-			out << "\r                                                                               \r";
+			string fill = "\r";
+			for (int i = 0; i < 90; ++i) {
+			    fill += " ";
+			}
+			fill += "\r";
+			out << fill;
```
jaylett: This change would have been better in its own commit, but actually I think it's a change we don't want in this form — it looks like you're dynamically constructing a `std::string`` one character at a time, which is almost certainly slower than the character constant previously. (You might be able to do better under C++11 using the `string(size_t n, char c)` constructor and then changing the first and last characters.)

xapian-core/tests/harness/testsuite.cc
```diff
@@ -706,9 +711,20 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
 		case PASS:
 		    ++res.succeeded;
 		    if (verbose || !use_cr) {
-			out << col_green << " ok" << col_reset << endl;
```
jaylett: You could just keep everything to `col_reset`, then have the "long run" conditional do `col_yellow` to `col_reset` again (and no else leg), with a final `\n` at the end. I _think_ that would be a little more readable, but I could be wrong.

xapian-core/tests/harness/testsuite.cc
```diff
@@ -706,9 +711,20 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
...
+			if (test_duration.count() >= 10000) {
```
jaylett: This is a little bit of a magic number, so it might be a good idea to drop a comment ahead of it making it clear that 10s is somewhat arbitrary, but "not too long and not too short". (There could be advantage in making it configurable, pulling from an environment variable with a default of 10s — it would mean we could keep an eye out for really long tests, and periodically go after medium-length ones. Not sure it's worth it right now, though, so probably better to keep a note at most.)

xapian-core/tests/harness/testsuite.cc
```diff
@@ -706,9 +711,20 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
...
+			    out << col_green << " ok" << col_yellow <<
+				" SLOW TEST (" << test_duration.count() <<
+				" ms)" << col_reset << endl;
```
ojwb: The convention for Xapian code when iostream output code is to put the `<<` after the inserted newline.

xapian-core/tests/harness/testsuite.cc
ojwb: I agree with James that the current approach seems better.

The current code has 79 spaces - is there a reason to increase that to 90?  On a standard 80 column terminal, 90 will cause the terminal to scroll which will look odd.

xapian-core/tests/harness/testsuite.cc
ojwb: It could at least be pulled out into a `SLOW_TEST_THRESHOLD` constant defined near the start of the source file - the name then documents what we're testing here, and someone looking to adjust it can find it more easily.

xapian-core/tests/harness/testsuite.cc
tanmaysachan: changed accordingly.

xapian-core/tests/harness/testsuite.cc
tanmaysachan: Added a variable.

xapian-core/tests/harness/testsuite.cc
tanmaysachan: changed accordingly.

xapian-core/tests/harness/testsuite.cc
tanmaysachan: removed the change.

xapian-core/tests/harness/testsuite.cc
```diff
@@ -707,8 +715,15 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
...
+
+			if (test_duration.count() >= SLOW_TEST_THRESHOLD) {
+			    out << col_yellow << " SLOW TEST ("
+				<< test_duration.count()
+				<< " ms)" << col_reset << endl;
+			}
 		    } else {
-			out << "\r                                                                               \r";
+			out << "\r                                        "
+			    << "                                   \r";
```
jaylett: We should drop this change when merging, because it doesn't make things better (except that it allows the check script to pass on the diff).

ojwb: I'm happy to ignore the check script when it's not helpful.  The point of it is to automate the checks for things that we see people getting wrong - this takes away some of the duller parts of patch review and the submitter gets initial feedback sooner.

But here if the `<<` between the two string literals is removed then it's totally equivalent to the existing code (adjacent string literals are concatenated early on in parsing), passes the check script, and I think is at least as readable.

tanmaysachan: removed the `<<`.

xapian-core/tests/harness/testsuite.cc
```diff
@@ -711,20 +714,17 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
 		case PASS:
 		    ++res.succeeded;
 		    if (verbose || !use_cr) {
-			if (test_duration.count() >= 10000) {
-			    out << col_green << " ok" << col_yellow <<
-				" SLOW TEST (" << test_duration.count() <<
-				" ms)" << col_reset << endl;
-			} else {
-			    out << col_green << " ok" << col_reset << endl;
+			out << col_green << " ok" << col_reset << endl;
```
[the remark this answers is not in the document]
tanmaysachan: fixed.

ojwb:
I tried this out and it seems to work well.

The presentation could be improved I think:

* Unless you run with `--verbose` (or equivalent) then the `SLOW TEST` messages are overwritten by the next output.  I guess that's why there was the extra newline?

* The slow test messages are not connected to correctness of the test result, so maybe we should use a different colour (e.g. cyan).  I tried this out to see what it looks like so I can easily add a commit for that.

* I wonder if reporting the time in seconds would be clearer - e.g. `Running test: consistency1... ok SLOW TEST (132598 ms)` makes me have to stop and think, whereas if it said `132.598s` that to me is more obviously two and a bit minutes.

Perhaps it would be best to plan to report the slow tests at the end of a run rather than annotating as we go (which would solve the first two points).

tanmaysachan:
> I tried this out and it seems to work well.
> 
> The presentation could be improved I think:
> 
>     * Unless you run with `--verbose` (or equivalent) then the `SLOW TEST` messages are overwritten by the next output.  I guess that's why there was the extra newline?
> 
>     * The slow test messages are not connected to correctness of the test result, so maybe we should use a different colour (e.g. cyan).  I tried this out to see what it looks like so I can easily add a commit for that.
> 
>     * I wonder if reporting the time in seconds would be clearer - e.g. `Running test: consistency1... ok SLOW TEST (132598 ms)` makes me have to stop and think, whereas if it said `132.598s` that to me is more obviously two and a bit minutes.
> 
> 
> Perhaps it would be best to plan to report the slow tests at the end of a run rather than annotating as we go (which would solve the first two points).

I added a commit which makes reporting these slow tests as a part of the report function.

xapian-core/tests/harness/testsuite.cc
```diff
@@ -114,8 +114,11 @@ string test_driver::col_red, test_driver::col_green;
 string test_driver::col_yellow, test_driver::col_reset;
 bool test_driver::use_cr = false;
 
-// time constant to mark tests as slow or not
-static const int SLOW_TEST_THRESHOLD = 10000;
+// time constant in seconds to mark tests as slow or not
+static const float SLOW_TEST_THRESHOLD = 10.00;
```
ojwb: `double` is a better default choice of floating point type than `float`.  The exceptions that come to mind are if you're storing so many values that space becomes a concern (not the case here), or you don't have hardware floating point in which case `float` might be faster.  Otherwise `double` offer a much better precision for a likely very similar speed.

There are in fact currently no uses of the `float` type anywhere in the xapian source tree:

```
olly@gemse:~/git/xapian$ git grep -w float|grep '\.\(cc\|h\):'
xapian-core/ChangeLog:	* api/matchspy.cc: Use <cfloat> and <cmath> rather than <float.h> and
xapian-core/api/queryinternal.cc:	// results to float (which is another approach which works).
xapian-core/api/sortable-serialise.cc:	// INFINITY is C99.  Oddly, it's of type "float" so sanity check in
```

xapian-core/tests/harness/testsuite.cc
```diff
@@ -713,16 +716,14 @@ test_driver::do_run_tests(vector<string>::const_iterator b,
...
-			out << col_green << " ok" << col_reset;
 
-			if (test_duration.count() >= SLOW_TEST_THRESHOLD) {
-			    out << col_yellow << " SLOW TEST ("
-				<< test_duration.count()
-				<< " ms)" << col_reset;
-			}
+		    if (test_duration.count() >= SLOW_TEST_THRESHOLD) {
+			slow_tests.push_back(make_pair(test->name,
+						       test_duration.count()));
```
ojwb: You can use `emplace_back()` and avoid the `make_pair()`.

tanmaysachan: replaced push_back with emplace_back.

xapian-core/tests/harness/testsuite.cc
tanmaysachan: changed float to double.

ojwb:
Thanks for the updates.  Auto-squashed, rebased and pushed to master as 4b731e7897b474213b92f4ca1c237cbc43d05277.

I adjusted the commit message a little - in particular the summary was too long, so I removed the filename reference as "Add high_resolution_clock to track slow tests" is a clear summary by itself.

You don't generally need to list the filenames affected in the commit message - commands like `git show --stat` will show someone the list of affected files.  Sometimes mentioning the filename is an efficient way to add useful context though, like in 6b3ee5381a5517ae6ba41c5ad21d07b503bfbc47 which has summary "xapian-check-patch: Add two more typos" which is much clearer than just "Add two more typos" and shorter than "Check for two more typos in the patch checking script".

PR #254, opened 2019-06-10
Add generated database support to RemoteTcp, RemoteProg and Honey BackendManagers

tanmaysachan:
1) RemoteTcp and RemotProg now support generated databases.
2) Honey now supports generated databases.
[WIP]


xapian-core/tests/harness/backendmanager.cc
```diff
@@ -123,16 +125,32 @@ BackendManager::get_database(const std::string &dbname,
...
+
     if (path.empty()) {
 	// InMemory doesn't have a path but we want to support generated
 	// databases for it.
+	if (is_remote_bm) {
+	    Xapian::WritableDatabase wdb =
+		get_writable_database_from_sub_manager(dbleaf);
+	    gen(wdb, arg);
+	    return get_remote_database_by_name(dbleaf, 30000);
+	}
```
ojwb: Here we're only dealing with the inmemory backend (that's the only one where `path.empty()` will be true, so I don't think this code would ever get used...

ojwb: I think you want to this lower down so we don't pass a remote WritableDatabase to the generator function (which will work, but be unnecessarily slower).

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -42,6 +42,8 @@
 #include "index_utils.h"
 #include "backendmanager.h"
 #include "unixcmds.h"
+#include "common/stringutils.h"
+
 
 using namespace std;
```
ojwb: `common` is in the include path, so that should just be `#include "stringutils.h"`.

Also, you've added an extra blank line after it.

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -123,16 +125,32 @@ BackendManager::get_database(const std::string &dbname,
     string dbleaf = "db__";
     dbleaf += dbname;
     const string& path = get_generated_database_path(dbleaf);
+
+    // checking if a remote backendmanager is being used
+    bool is_remote_bm = false;
+    if (startswith(get_dbtype(), "remote")) {
+	is_remote_bm = true;
+    }
```
ojwb: Checking a boolean return value with `if` only to set a boolean variable based on the result of the test is a bit of an anti-pattern - it's simpler and clearer to just assign the result in such cases:

```
bool is_remote_bm = startswith(get_dbtype(), "remote");
```

xapian-core/tests/harness/backendmanager.cc
tanmaysachan: Oh I misread the code, I apologize for the mistake.
get_writable_database_from_sub_manager's purpose was exactly that what you have written.

xapian-core/tests/harness/backendmanager.cc
tanmaysachan: fixed.

xapian-core/tests/harness/backendmanager.cc
tanmaysachan: changed accordingly.

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -145,14 +155,36 @@ BackendManager::get_database(const std::string &dbname,
...
-	Xapian::WritableDatabase wdb = get_writable_database(tmp_dbleaf,
-							     string());
+	Xapian::WritableDatabase wdb;
+	if (is_remote_bm) {
+	    wdb = get_writable_database_from_sub_manager(tmp_dbleaf);
+	} else if (is_honey_bm) {
+	    wdb = get_alt_writable_database(tmp_dbleaf, string());
+	} else {
+	    wdb = get_writable_database(tmp_dbleaf, string());
+	}
```
ojwb: I think it would be better to implement this dispatch via the polymorphism of the `BackendManager` class hierarchy instead of setting boolean flags based on string operations on the backend name.

The essence of what we're wanting to do here is "get a WritableDatabase for generating" so I'd suggest a new virtual method called say `get_generated_database()`, and then each of these alternatives forms the body of the implementation of that for the respective `BackendManager` subclass, e.g.:

```
Xapian::WritableDatabase
BackendManagerRemote::get_generated_database(const string& path)
{
    return get_writable_database_from_sub_manager(path);
}
```

The `else` case would then become the default implementation of this in the base class.

This also means this backend-specific code is in the appropriate `BackendManager` subclass, which is what we ideally want.

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -145,14 +155,36 @@ BackendManager::get_database(const std::string &dbname,
...
+
+	Xapian::WritableDatabase wdb(path);
+	wdb.compact(tmpfile, Xapian::DB_BACKEND_HONEY);
+	wdb.close();
+
+	rm_rf(path);
+
+	rename(tmpfile.c_str(), path.c_str());
+    }
     return Xapian::Database(path);
```
ojwb: ... and then a `finalise_generated_database()` virtual method could implement this part.

Or have a single method which does both and the `gen()` and `rename()` steps in between get duplicated.

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -207,6 +259,13 @@ BackendManager::get_writable_database(const string &, const string &)
     invalid_operation("Attempted to open a disabled database");
 }
 
+Xapian::WritableDatabase
+BackendManager::get_alt_writable_database(const string &, const string &)
```
ojwb: Putting the `&` with the type seems to be widely preferred as C++ style and we've decided to follow that (a lot of existing Xapian code doesn't conform but please follow this for new code so the problem at least doesn't get worse.  Same for pointer types (`char*` not `char *`).  Currently `xapian-check-patch` doesn't try to check for this.

xapian-core/tests/harness/backendmanager.h
```diff
@@ -120,6 +120,13 @@ class BackendManager {
     /// Get a writable database instance.
     virtual Xapian::WritableDatabase get_writable_database(const std::string & name, const std::string & file);
 
+    /** Get alternate writable database for backends which
+     * don't support writing directly.
+     */
+    virtual Xapian::WritableDatabase get_alt_writable_database
+					(const std::string & name,
+					 const std::string & file);
```
ojwb: I think it's better to wrap after the return type if wrapping the parameter list isn't sufficient (since we wrap after the return type in out-of-line function definitions), like so:

```
    virtual Xapian::WritableDatabase
    get_alt_writable_database(const std::string& name,
			      const std::string& file);
```

xapian-core/tests/harness/backendmanager.cc
tanmaysachan: I tried fixing it a little. I'll fix the code style issues once you're satisfied with the current changes.

xapian-core/tests/harness/backendmanager.cc
tanmaysachan: Added new function.

xapian-core/tests/harness/backendmanager.cc
ojwb: > Currently `xapian-check-patch` doesn't try to check for this.

FYI, added a check for this (at least in common cases) yesterday.

xapian-core/tests/harness/backendmanager_honey.cc
```diff
@@ -88,6 +88,28 @@ BackendManagerHoney::get_alt_writable_database(const string & name,
     return wdb;
 }
 
+Xapian::WritableDatabase
+BackendManagerHoney::get_generated_database(const string& name)
+{
+    return get_alt_writable_database(name, string());
```
ojwb: This is the only call to `get_alt_writable_database()` so we can move the code from that here and eliminate that entirely.

xapian-core/tests/harness/backendmanager_honey.cc
```diff
@@ -88,6 +88,28 @@ BackendManagerHoney::get_alt_writable_database(const string & name,
...
+}
+
+void
+BackendManagerHoney::finalise_generated_database(const string& path)
+{
+    // converting a glass backend to honey
+    string tmpfile = path;
+    tmpfile += ".tmp";
```
ojwb: I think it would be better to pass in `tmpfile` to this method - as this is the knowledge about how the temporary file is named has to be in duplicated in (at least) two different files.

xapian-core/tests/harness/backendmanager_remote.cc
```diff
@@ -29,16 +29,22 @@
...
 std::string
-BackendManagerRemote::get_generated_database_path(const std::string & name) {
+BackendManagerRemote::get_generated_database_path(const std::string& name) {
     return sub_manager->get_writable_database_path(name);
 }
 
+Xapian::WritableDatabase
+BackendManagerRemote::get_generated_database(const std::string& name)
+{
+    return get_writable_database_from_sub_manager(name);
```
ojwb: Similar to `get_alt_writable_database()`, `get_writable_database_from_sub_manager()`'s code can now be moved here and it can be eliminated.

tanmaysachan: eliminated `get_alt_writable_database()`

xapian-core/tests/harness/backendmanager.cc
ojwb: I think the structure is looking better, but there's still some dispatching based on checking the name (`is_remote_bm`).

You've already added a `get_remote_database_by_name()` virtual method, but with a default implementation which throws an exception.  If you instead made the default implementation `return Xapian::Database(path);` (and
added a parameter to pass in `path`) then we wouldn't need `is_remote_bm` here at all (`finalise_generated_database()` does nothing by default, and the remote managers don't override that).  That lets the virtual method do the dispatch for us.

The method name would make rather less sense then - I'd suggest renaming to `get_database_by_name()`.

I'd probably not pass in the timeout from here.  We have to hard-code it somewhere in the backend manager machinery for this case, and doing so in the remote tcp manager at least means less backend-specific stuff here.  

Also the timeouts used for other remote tcp databases in the test harness are all `300000` (5 minutes) not `30000` (30 seconds).  Probably either is long enough to avoid timeouts during runs even under heavy load in practice, but the timeout used by default really should be consistent.

xapian-core/tests/harness/backendmanager_honey.cc
ojwb: As discussed on IRC, I was confusing this temporary file with a different one.  There is a problem I think, but not what I thought it was.

If I follow, for honey we:

* call get_generated_database() writing a glass database to a temporary path
* then we rename that to the final path (this is bad - if the testsuite is interrupted at this point we leave a valid glass database in the honey database cache)
* then we compact that and convert it to honey at a temporary path
* then we remove the glass database from the final path and rename the honey one there instead

We really want to skip the second step and compact and convert to honey from the first temporary path to a second temporary path instead.

An alternative would be to lazily generate the glass database to a temporary path and then rename it to the final name in the **glass** database cache and then compact and convert from there to a temporary path for honey which then gets renamed to the final name in the honey cache.  That would save work as we avoid an extra generation of the glass database corresponding to each generated honey database, but probably means more scheduling restrictions for the parallel stuff later.

tanmaysachan: Made the required changes.

xapian-core/tests/harness/backendmanager.cc
tanmaysachan: Added `get_database_by_path` function, and removed `get_remote_database_by_name`.

tanmaysachan: I've made `finalise_generated_database` take a name instead of a path. If I modify it(basically make `get_database_by_name` a replacement of `finalise_generated_database`) I will have to perform messy string operations inside honey's backendmanager. So I've not removed `finalise_generated_database()`.

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -42,6 +42,7 @@
 #include "index_utils.h"
 #include "backendmanager.h"
 #include "unixcmds.h"
+#include "stringutils.h"
```
ojwb: This header is no longer needed.

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -219,13 +233,25 @@ BackendManager::get_compaction_output_path(const std::string&)
...
 {
     invalid_operation("Generated databases aren't supported for this database "
 		      "type");
 }
 
+void
+BackendManager::finalise_generated_database(const std::string&)
+{
+
+}
```
ojwb: Xapian coding style wouldn't put a blank line in an empty function like this - it doesn't add anything useful.

xapian-core/tests/harness/backendmanager_honey.cc
```diff
@@ -73,12 +73,62 @@ BackendManagerHoney::do_get_database_path(const vector<string> & files)
...
+{
+    // creating generated database inside glass cache
+    // to prevent a valid glass db inside honey cache
+    // if testsuite was interrupted
+    string db_path = ".glass/" + name;
+
+    rm_rf(db_path);
+
+    auto flags = Xapian::DB_CREATE|Xapian::DB_BACKEND_GLASS;
+    Xapian::WritableDatabase wdb(db_path, flags);
```
ojwb: This creates the database at the final path in the glass cache directory, so if creation is interrupted a valid DB with the wrong contents may be left in the cache, and if it is a future run could then use for glass tests.  Depending on the contents and testcases which use the DB, this could causing mysterious test failures, or worse tests to pass when they shouldn't.  For those sort of reasons, we carefully create these cached databases at a temporary path and only rename to the final path once they're complete.

This also blows away any existing cached glass version of this database unnecessarily - if there's a cached glass version but no cached honey version we only need to do the conversion step.

It would also be better to use `BackendManagerGlass` to handle the creation of the glass database to convert from - that class already knows how lazily create the database and to safely create it at a temporary path and then rename.  And really `BackendManagerHoney` shouldn't encode knowledge about glass databases.

xapian-core/tests/harness/backendmanager_honey.cc
```diff
@@ -73,12 +73,62 @@ BackendManagerHoney::do_get_database_path(const vector<string> & files)
...
+string
+BackendManagerHoney::get_generated_database_path(const string& name)
+{
+    // the generated database would go inside the glass cache
+    // if it does not exist yet
+    string path = CACHE_DIRECTORY "/" + name;
+    if (path_exists(path)) {
+        return path;
+    }
+    return ".glass/" + name;
```
ojwb: That's quite cunning, but we shouldn't need this peculiar alternating return value if we delegate the creation of the glass database to `BackendManagerGlass` - we can just pass through the generator function to that and it'll handle creating the glass database and calling the generator function on it (if it doesn't already exist).  Then we just need to convert that.

xapian-core/tests/harness/backendmanager_honey.cc
tanmaysachan: Inside get_database(in backendmanager.cc) the function get_generated_database is being called on a temporary path already(tmp_dbleaf). So the concerns you mentioned won’t be true(or if there’s something else im missing). I agree with the BackendManagerGlass handling the glass part, and would work on its implementation.

xapian-core/tests/harness/backendmanager_remote.h
```diff
@@ -51,11 +51,22 @@ class BackendManagerRemote : public BackendManager {
...
 
     /// Get the args for opening the last opened WritableDatabase again.
     std::string get_writable_database_again_args();
+
+    /// Get generated database path from sub_manager
+    std::string get_generated_database_path(const std::string& name);
+
+    /// Get generated database
+    Xapian::WritableDatabase
+    get_generated_database(const std::string& name);
```
ojwb: There's no need to wrap this line.

xapian-core/tests/harness/backendmanager_honey.h
```diff
@@ -39,21 +39,39 @@ class BackendManagerHoney : public BackendManager {
...
+			BackendManager* generated_sub_manager_)
+	: BackendManager(datadir_),
+	  generated_sub_manager(generated_sub_manager_) { }
 
     /// Return a string representing the current database type.
     std::string get_dbtype() const;
 
+    /// Get generated database for honey
+    Xapian::WritableDatabase
+    get_generated_database(const std::string& name);
```
ojwb: This one doesn't need wrapping either.

xapian-core/tests/harness/backendmanager_honey.cc
```diff
@@ -73,12 +73,50 @@ BackendManagerHoney::do_get_database_path(const vector<string> & files)
     return dbpath;
 }
 
+Xapian::WritableDatabase
+BackendManagerHoney::get_generated_database(const string& name)
+{
+    // creating generated database inside glass cache
+    // to prevent a valid glass db inside honey cache
+    // if testsuite was interrupted
+    return generated_sub_manager->get_generated_database(name);
```
ojwb: "Create ..." would read better and make this a complete sentence.  Similarly "Covert ..." just below.

ojwb:
I think it's probably OK to just mark the new honey compaction failures as XFAIL.  The `FAIL: DatabaseClosedError: Closed!` ones are a bit odd, so I'm going to take a quick look at those.

I'll also take a look at `subdbwithoutpos1`.

Aside from that and the minor nits I just noted, I think this is pretty much there.  Thanks for all your work on it.

xapian-core/tests/harness/backendmanager_honey.cc
tanmaysachan: Changed appropriately.

tanmaysachan:
> I think it's probably OK to just mark the new honey compaction failures as XFAIL. The `FAIL: DatabaseClosedError: Closed!` ones are a bit odd, so I'm going to take a quick look at those.
> 
> I'll also take a look at `subdbwithoutpos1`.
> 
> Aside from that and the minor nits I just noted, I think this is pretty much there. Thanks for all your work on it.

I XFAILed the failing cases in a new commit. `subdpwithoutpos1` needs looking at.

ojwb:
I fixed the `DatabaseClosed` error from two of the generated honey compaction testcases (though they now fail for other reasons) on git master in 9f2e29ad5233.

And I've fixed `subdbwithoutpos1` on git master in 5cea6eb057f1.

The fixup commits didn't all apply properly to the commits they were meant to fixup.  It didn't look like they were just wrongly attributed - I think the issue is some overlap in the changes for remote and honey here.  In hindsight it would probably have been better to address each in turn in separate PRs - we'd have got each merged more quickly.  I looked at trying to tease apart the change but decided in the end to just squash those changes into one commit.

I've rebased that onto current master and pushed as 1bffe5e765b334fbaf7c3ded4c28b7c73865a5c5,

However, I just realised that we're still generating `.glass/db__subdbwithoutpos1~`, etc which won't actually get shared with the glass backend tests, as I mentioned on IRC.  That doesn't actually break anything, but is a missed opportunity for sharing work between testcases and would be good to address.

PR #252, opened 2019-06-05
Add testcases for feature related api's

VaibhavKansagara:
These include unit testcases for the 4 features such as TfFeature,IdfFeature, TfIdfDoclenFeature,TfDoclenCollTfCollLenFeature.Additionally it includes test
for FeatureList class which includes testing an empty featurelist and a
featurelist containing more than it's default features and also for generating
a database with three documents in which one has no common terms with other two.
Also consists of testing the output of create_feature_vetors method for all six
features like TfIdfDoclenFeature etc.

PR #255, opened 2019-06-13
Change feature API design

VaibhavKansagara:
This includes adding PIMPL approach for feature class and changing how
the stats needed for featurelist are populated and computed.After this
design change all the features in the featurelist share a common
Feature::internal object which greatly decreases the memory required as
maps related to statistics do not need to be copied for every feature.


xapian-letor/api/feature_internal.h
```diff
@@ -0,0 +1,147 @@
...
+
+  public:
+    /// get database
+    Database get_database() const;
+
+    /// get query
+    Query get_query() const;
+
+    /// get document
+    Document get_doc() const;
```
ojwb: I think this should be `get_document()` since there's a xapian-core `get_document()` method.  Consistency in naming makes an API easier to work with.

xapian-letor/api/feature_internal.cc
```diff
@@ -0,0 +1,181 @@
...
+#include "debuglog.h"
+
+using namespace std;
+using namespace Xapian;
+
+Database
+Feature::Internal::get_database() const
+{
+    LOGCALL(API, Database, "Feature::Internal::get_database", NO_ARGS);
+    return feature_db;
```
ojwb: I would inline all these trivial getters and setters in the header - that way they're exposed to the compiler to optimise better.  We won't have logging for them, but I've found logging is of limited interest in such methods anyway.

xapian-letor/api/feature_internal.cc
```diff
@@ -0,0 +1,181 @@
...
+Document
+Feature::Internal::get_doc() const
+{
+    LOGCALL(API, Document, "Feature::Internal::get_doc", NO_ARGS);
+    return feature_doc;
+}
+
+bool
+Feature::Internal::get_termfreq(const std::string & term,
+			double & termfreq_) const
```
ojwb: The wrapped line isn't indented enough.

And why is this returning `double`?  `termfreq` should be type `Xapian::doccount`.

Also preferred Xapian style for writing the `&` in reference types (and the `*` in pointer types) is what seems to be the standard C++ style of putting no space before but a space after, i.e. `const std::string& term` and `double& termfreq_` here.  Quite a lot of older code doesn't conform to this - fixing everything by hand is a lot of work, and we've failed to find an automatic code formatting tool which does a good enough job.  But I'd really like any new code to follow this preferred style.

xapian-letor/api/feature_internal.cc
```diff
@@ -0,0 +1,181 @@
...
+}
+
+bool
+Feature::Internal::get_inverse_doc_freq(const std::string & term,
+			double & inverse_doc_freq_) const
+{
+    LOGCALL(API, bool, "Feature::Internal::get_inverse_doc_freq", term | inverse_doc_freq_);
+    auto inverse_doc_freq_iterator = inverse_doc_freq.find(term);
+    if (inverse_doc_freq_iterator != inverse_doc_freq.end()) {
+	inverse_doc_freq_ = (double)inverse_doc_freq_iterator->second;
```
ojwb: This cast shouldn't be needed, as the type of the values in the map should be `double`.

If a cast were needed, please avoid C-style casts - instead use a C++-style cast which explains the intent of the cast so that the compiler can reject conversions other than the intended one (here that would be `static_cast<double>(...)`), or for a case like this a C++ "constructor cast" like `double(...)` is OK.

xapian-letor/api/feature_internal.cc
```diff
@@ -0,0 +1,181 @@
...
+	return true;
+    }
+    return false;
+}
+
+bool
+Feature::Internal::get_doc_length(const std::string & term,
+			double & doc_length_) const
```
ojwb: Document lengths should `Xapian::termcount`.

xapian-letor/api/feature_internal.cc
```diff
@@ -0,0 +1,181 @@
...
+}
+
+bool
+Feature::Internal::get_collection_length(const std::string & term,
+			double & collection_length_)
+{
+    LOGCALL(API, bool, "Feature::Internal::get_collection_length", term | collection_length_);
+    auto collection_length_iterator = collection_length.find(term);
+    if (collection_length_iterator != collection_length.end()) {
+	collection_length_ = (double)collection_length_iterator->second;
```
ojwb: And this shouldn't be `double` either - almost none of these stats should be...

This one is `Xapian::termcount` in the map currently, though maybe `Xapian::totallength` is a more appropriate type as `Xapian::termcount` could be overflowed by the total document length of a large database.

xapian-letor/api/featurelist.cc
```diff
@@ -104,12 +106,15 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
 	Xapian::Document doc = i.get_document();
 	std::vector<double> fvals;
 	internal->set_data(letor_query, letor_db, doc);
+	Feature::Internal *internal_feature = new Feature::Internal();
+	// set all the parameters for internal_feature
+	internal_feature->set_database(letor_db);
+	internal_feature->set_query(letor_query);
+	internal_feature->set_doc(doc);
```
ojwb: We don't ever use these set methods except here - we could get rid of them completely and pass these 3 values as parameters to the constructor.

xapian-letor/tests/api_letor.cc
```diff
@@ -866,10 +869,10 @@ DEFINE_TESTCASE(colltfcolllenfeature, generated)
     Xapian::CollTfCollLenFeature feature;
 
     Xapian::Database db(db_path);
-    feature.set_database(db);
+    feature.internal->set_database(db);
 
     Xapian::Document doc = db.get_document(1);
-    feature.set_doc(doc);
+    feature.internal->set_doc(doc);
```
ojwb: Oh, except those setters are used here...

But I think if we're trying to test a particular feature we should do that via a `FeatureList` holding it, and test the reported values of the feature.

While if we're wanting to test the stats that features see, we should instead create a custom `Feature` subclass for that purpose (like `CheckStatsWeight` in `xapian-core/tests/api_weight.cc` which is used to test the stats provided to `Weight` subclasses), and again test via a `FeatureList` holding it.

Currently it looks to me like we seem to be trying to do both at the same time here.

xapian-letor/feature/tfdoclencolltfcolllenfeature.cc
```diff
@@ -53,94 +55,65 @@ TfDoclenCollTfCollLenFeature::get_values() const
...
-	    auto tf_iterator = termfreq.find(*qt);
-	    auto coll_tf_iterator = collection_termfreq.find(*qt);
-	    if (tf_iterator != termfreq.end())
-		tf = (double)tf_iterator->second;
-	    else
+	    double tf = 0;
+	    double coll_tf = 0;
+	    if (!internal->get_termfreq(*qt, tf))
 		tf = 0;
-	    if (coll_tf_iterator != collection_termfreq.end())
```
ojwb: The pattern "return `bool` indicating success, and the actual value found by reference" makes for somewhat clumsy code here.  It's also what's presumably led you to an API where everything's returned as a `double` regardless of what type is actually is.

Under what circumstances would we actually not find an entry when looking up in these maps?

If it should never happen, then I don't think it makes sense to complicate the API here to allow for it - we can either `Assert(false);` and `return 0;` if it's an error in internal state, or if it can be caused by bad parameters passed from user code throw `Xapian::InvalidArgumentError`.

If it can happen, but there's a sensible default (it looks like we actually always just use zero if the term is not found) then we can just make `get_termfreq()` etc return that default if they don't find the answer.

xapian-letor/tests/api_letor.cc
VaibhavKansagara: > But I think if we're trying to test a particular feature we should do that via a `FeatureList` holding it, and test the reported values of the feature.
> 
This tests are unittests which means we can't use any other api's for testing them and I have already written tests which check a particular feature using FeatureList in #252(checkcreatefeaturevector_(feature_name)).
> While if we're wanting to test the stats that features see, we should instead create a custom `Feature` subclass for that purpose (like `CheckStatsWeight` in `xapian-core/tests/api_weight.cc` which is used to test the stats provided to `Weight` subclasses), and again test via a `FeatureList` holding it.
>
I think there is no need for this as in case of `Weight` class the statistics were protected and hence there was a need for testing them via custom subclass but here internal is public and we can easily access the maps via getter and setter methods.

xapian-letor/api/feature_internal.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/api/feature_internal.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/api/feature_internal.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/api/feature_internal.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/api/feature_internal.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/api/feature_internal.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/feature/tfdoclencolltfcolllenfeature.cc
VaibhavKansagara: > Under what circumstances would we actually not find an entry when looking up in these maps?
> 
It can happen when a feature query contains terms which are not in the feature database.
I am going with the latter option you suggested. Correct me if I am wrong.

xapian-letor/api/featurelist_internal.cc
```diff
@@ -162,33 +164,30 @@ FeatureList::Internal::compute_collection_termfreq()
 }
 
 void
-FeatureList::Internal::populate_feature(Feature *feature_)
+FeatureList::Internal::populate_featurelist(Feature::Internal *internal_feature)
 {
-    stat_flags stats_needed = stat_flags(feature_->get_stats());
+    for (Feature* it : feature) {
+	stats_needed = stat_flags(stats_needed | it->get_stats());
+    }
```
jaylett: I'm pretty sure this can (and should) be done in the constructor, called when the `FeatureList` itself is constructed (which is when the list of `Feature` objects is put together / copied in). As it is, doing it here suggests that it can change every time we populate, which isn't true. (Indeed, it can't be true because you aren't clearing `stats_needed`.)

VaibhavKansagara: You are right. I missed it. fixed in a fixup commit.

xapian-letor/tests/api_letor.cc
jaylett: The distinction between unittest and not is somewhat arbitrary. It's absolutely fine to test via another class — particularly if that's the way they are used in the codebase.

xapian-letor/feature/tfdoclencolltfcolllenfeature.cc
ojwb: I guess it depends what we do for a term which doesn't exist when we fill these maps (I assume we must fill them based on terms from the query - creating them for all terms in the database would be a lot of unnecessary work).

It probably makes more sense not to bother inserting zero (or other default) values into the map for such terms, so defaulting on lookup if the term isn't found sounds good.

xapian-letor/api/featurelist_internal.h
```diff
@@ -178,7 +183,7 @@ class FeatureList::Internal : public Xapian::Internal::intrusive_base {
 			    const Xapian::Document & doc);
 
     /// Computes and populates the stats needed by a Feature.
-    void populate_feature(Feature *feature_);
+    void populate_featurelist(Feature::Internal *internal_feature);
```
jaylett: This should probably be called `populate_feature_internal()` or something now. Also, it's only called from `FeatureList`, which is already a friend class, so does it have to be public?

VaibhavKansagara: I will change that in a fixup commit.

VaibhavKansagara:
> I think my only outstanding concern beyond the name (and possibly visibility) of `populate_featurelist()` is about hiding away those setters (and testing indirectly). That could be left until later to get this merged, though.
>
@jaylett Not changing the test would cause compilation errors. Do you want me to remove those tests for now or do you want to remove them permanently?.

jaylett:
Are the tests actually testing anything that isn't covered elsewhere? If so, may be better to remove them. If not, ideally rewrite them to use only public APIs (so you're testing from the point of view of code that could be written by a Xapian library user). But as I say: probably best to ignore that for the moment to get this merged.

xapian-letor/api/feature_internal.h
```diff
@@ -0,0 +1,166 @@
...
+    std::map<std::string, Xapian::termcount> collection_length;
+
+    /// Frequency of the Query Terms in the whole database
+    std::map<std::string, Xapian::termcount> collection_termfreq;
+
+  public:
+    /// get database
+    inline Database get_database() const {
+	return feature_db;
+    }
```
ojwb: A method defined inside a class like this is implicitly `inline` - there's no need to specify it explicitly.

xapian-letor/api/featurelist.cc
```diff
@@ -41,13 +43,21 @@ FeatureList::FeatureList() : internal(new FeatureList::Internal())
     internal->feature.push_back(new CollTfCollLenFeature());
     internal->feature.push_back(new TfIdfDoclenFeature());
     internal->feature.push_back(new TfDoclenCollTfCollLenFeature());
+    for (Feature* it : internal->feature) {
+	internal->stats_needed = Internal::stat_flags(internal->stats_needed
+					    | it->get_stats());
```
ojwb: The Xapian convention for breaking a line at most binary operators (such as `|` here) is that the operator goes before the break, and the start of the two arguments of the operator are then vertically aligned to make it easier for the reader to see what they are (so the "i" of  `it->...` should line up under the "i" of `internal->...`).

(iostream `<<` is treated differently - while that's actually a binary operator, it really reads more like `<<` applied to the thing to its right).

xapian-letor/api/feature_internal.h
VaibhavKansagara: fixed in a fixup commit

xapian-letor/api/featurelist.cc
VaibhavKansagara: fixed in a fixup commit

xapian-letor/tests/api_letor.cc
ojwb: Very few testcases in our testsuite are unit tests of a single class, but that's indeed really not a problem.

In `core`, `unittest` and `internaltest` unit test things low-level functionality inside the library by linking internal code directly into the test program.  These tests cover things which are hard to reliably test via the public API.  Everything else is tested via the public APIs which users use.

In some cases a public API class can be tested in isolation (e.g. `stemtest` has unit tests of the `Xapian::Stem` API), but generally API classes interact.

Currently `api_letor.cc` has `#include <api/feature_internal.h>` which really isn't something that should be included there.  The `internal` member currently has `public` visibility, but it's definitely not intended for public use.

ojwb:
I'm not sure why the CI run of the patch checking script isn't reporting these, but if I download the patch and run it by hand I get:

```
$ wget https://github.com/xapian/xapian/pull/255.patch
[...]
$ xapian-maintainer-tools/xapian-check-patch 255.patch
xapian-letor/api/feature_internal.cc:177: error: Should have exactly one space after '&': +				 Xapian::termcount> &collection_tf)
xapian-letor/api/feature_internal.cc:177: error: Preferred style is 'int& x' (not 'int &x' or 'int & x'): +				 Xapian::termcount> &collection_tf)
xapian-letor/api/featurelist.cc:109: error: Should have exactly one space after '*': +	Feature::Internal *internal_feature = new Feature::Internal();
xapian-letor/api/featurelist.cc:109: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +	Feature::Internal *internal_feature = new Feature::Internal();
xapian-letor/api/featurelist_internal.cc:167: error: Should have exactly one space after '*': +FeatureList::Internal::populate_featurelist(Feature::Internal *internal_feature)
xapian-letor/api/featurelist_internal.cc:167: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +FeatureList::Internal::populate_featurelist(Feature::Internal *internal_feature)
xapian-letor/api/featurelist_internal.h:186: error: Should have exactly one space after '*': +    void populate_featurelist(Feature::Internal *internal_feature);
xapian-letor/api/featurelist_internal.h:186: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +    void populate_featurelist(Feature::Internal *internal_feature);
xapian-letor/api/feature_internal.h:123: error: Should have exactly one space after '&': +			     Xapian::termcount> &tf) {
xapian-letor/api/feature_internal.h:123: error: Preferred style is 'int& x' (not 'int &x' or 'int & x'): +			     Xapian::termcount> &tf) {
xapian-letor/api/featurelist.cc:117: error: Should have exactly one space after '*': +	Feature::Internal *internal_feature = new Feature::Internal(letor_db,
xapian-letor/api/featurelist.cc:117: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +	Feature::Internal *internal_feature = new Feature::Internal(letor_db,
xapian-letor/api/featurelist_internal.cc:167: error: Line extends beyond column 80 (to column 85): +FeatureList::Internal::populate_feature_internal(Feature::Internal* internal_feature)
error count:	13

Files patched:	40
Lines added:	660
Lines removed:	797
```

xapian-letor/api/feature_internal.h
```diff
@@ -0,0 +1,151 @@
...
+     *
+     */
+    void set_termfreq(const std::map<std::string,
+			     Xapian::termcount>& tf) {
+	termfreq = tf;
+    }
+
+    /** Sets the inverse_doc_freq that is going to be used for
+     *  Feature building.
+     *
```
ojwb: It'd be better to avoid these extra `*` lines (here and below).

xapian-letor/api/feature_internal.h
```diff
@@ -0,0 +1,151 @@
...
+     */
+    void set_inverse_doc_freq(const std::map<std::string,
+				     double>& idf) {
```
ojwb: Wrapped template arguments ought to align vertically just like wrapped function arguments (so `double` under `std::string`).  There seem to be quite a lot of cases of this in this patch.

xapian-letor/api/featurelist.cc
```diff
@@ -104,12 +114,13 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
...
+								    letor_query,
+								    doc);
+	internal->populate_feature_internal(internal_feature);
 	for (Feature* it : internal->feature) {
-	    it->set_database(letor_db);
-	    it->set_query(letor_query);
-	    it->set_doc(doc);
+	    it->internal = internal_feature;
 	    // Computes and populates the Feature with required stats.
-	    internal->populate_feature(it);
```
ojwb: You've removed the code, but left the comment saying what it did.

VaibhavKansagara:
> I'm not sure why the CI run of the patch checking script isn't reporting these, but if I download the patch and run it by hand I get:
> 
> ```
> $ wget https://github.com/xapian/xapian/pull/255.patch
> [...]
> $ xapian-maintainer-tools/xapian-check-patch 255.patch
> xapian-letor/api/feature_internal.cc:177: error: Should have exactly one space after '&': +				 Xapian::termcount> &collection_tf)
> xapian-letor/api/feature_internal.cc:177: error: Preferred style is 'int& x' (not 'int &x' or 'int & x'): +				 Xapian::termcount> &collection_tf)
> xapian-letor/api/featurelist.cc:109: error: Should have exactly one space after '*': +	Feature::Internal *internal_feature = new Feature::Internal();
> xapian-letor/api/featurelist.cc:109: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +	Feature::Internal *internal_feature = new Feature::Internal();
> xapian-letor/api/featurelist_internal.cc:167: error: Should have exactly one space after '*': +FeatureList::Internal::populate_featurelist(Feature::Internal *internal_feature)
> xapian-letor/api/featurelist_internal.cc:167: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +FeatureList::Internal::populate_featurelist(Feature::Internal *internal_feature)
> xapian-letor/api/featurelist_internal.h:186: error: Should have exactly one space after '*': +    void populate_featurelist(Feature::Internal *internal_feature);
> xapian-letor/api/featurelist_internal.h:186: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +    void populate_featurelist(Feature::Internal *internal_feature);
> xapian-letor/api/feature_internal.h:123: error: Should have exactly one space after '&': +			     Xapian::termcount> &tf) {
> xapian-letor/api/feature_internal.h:123: error: Preferred style is 'int& x' (not 'int &x' or 'int & x'): +			     Xapian::termcount> &tf) {
> xapian-letor/api/featurelist.cc:117: error: Should have exactly one space after '*': +	Feature::Internal *internal_feature = new Feature::Internal(letor_db,
> xapian-letor/api/featurelist.cc:117: error: Preferred style is 'int* x' (not 'int *x' or 'int * x'): +	Feature::Internal *internal_feature = new Feature::Internal(letor_db,
> xapian-letor/api/featurelist_internal.cc:167: error: Line extends beyond column 80 (to column 85): +FeatureList::Internal::populate_feature_internal(Feature::Internal* internal_feature)
> error count:	13
> 
> Files patched:	40
> Lines added:	660
> Lines removed:	797
> ```
I don't know how you are getting this but i think it is because of the last commit screw up i did. As you can see there is no line 177 in feature_internal.cc

xapian-letor/feature/colltfcolllenfeature.cc
```diff
@@ -53,62 +55,38 @@ CollTfCollLenFeature::get_values() const
...
-	    auto coll_tf_iterator = collection_termfreq.find(*qt);
-	    if (coll_tf_iterator != collection_termfreq.end()) {
-		value += log10(1 + (coll_len /
-				    (double)(1 + coll_tf_iterator->second)));
-	    } else {
-		value += log10(1 + coll_len);
-	    }
+	    double tf = internal->get_collection_termfreq(*qt);
+	    value += log10(1 + (coll_len /
+				(double)(1 + tf)));
```
ojwb: The cast to `double` here isn't needed now (since `tf` is a `double` already).  Similarly for some cases below.

xapian-letor/api/feature_internal.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/api/feature_internal.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/api/featurelist.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/feature/colltfcolllenfeature.cc
VaibhavKansagara: fixed in a fixup commit.

ojwb:
> I don't know how you are getting this but i think it is because of the last commit screw up i did. As you can see there is no line 177 in feature_internal.cc

I worked out what's going on here - the github `.patch` URL returns a patch series in one file, so the patch checking script is seeing the diff for each of your commits in turn, and is moaning about things fixed by later commits.

If I ask for all the changes in one diff that things look much better:

```
$ wget https://github.com/xapian/xapian/pull/255.diff
[...]
$ xapian-maintainer-tools/xapian-check-patch  255.diff
Files patched:	15
Lines added:	352
Lines removed:	495
```

So the problem is essentially at my end - sorry about the noise.

ojwb:
Thanks.  Squashed, rebased and pushed to master as 95800e3e991fcf5531b95884c2545f533f329434.  I also tweaked the commit message to add a missing space (in "computed.After").

It seemed worth merging what we have here as much better that what was on master, but I think there's scope for further clean-up.

It looks to me like `FeatureList::Internal` doesn't need to have the maps (`termfreq`, etc) as members.  If `compute_termfreq()`, etc returned the computed map and `Feature::Internal::set_termfreq()`, etc took an rvalue reference, then assuming the compiler implements the "Named Return Value Optimisation" (NRVO) we avoid copying the map and simplify the code as well.  I prototyped this for `termfreq` and it seems to work (though I didn't attempt to check if the copy is actually avoided):

https://github.com/ojwb/xapian/commit/f69fb3120d968779e8b7c0912b80cbc3efb2affd

If the equivalent were done for all the maps, then `FeatureList::Internal::clear_stats()` could be eliminated too.

You can read about NRVO and other copy elision stuff here: https://en.cppreference.com/w/cpp/language/copy_elision

The C++ standard doesn't require a compiler to perform NRVO, but it's probably reasonable to assume a decent compiler will in a simple case like this.  If it doesn't we're no worse off in terms of copy operations compare to currently, and we still avoid storing the map in two places.

An alternative approach would be to pass a pointer/reference to the map which is inside the `Feature::Internal` object to `compute_termfreq()` - that would definitely avoid any copy operations.  I didn't try that approach, but it's probably similar in code complexity to the approach I prototyped, so that may be a better option.

ojwb:
clang gives an error (promoted from a warning due to `-Werror` being used) on my prototype:

```
api/featurelist_internal.cc:171:33: error: moving a temporary object prevents
      copy elision [-Werror,-Wpessimizing-move]
        internal_feature->set_termfreq(std::move(compute_termfreq()));
                                       ^
api/featurelist_internal.cc:171:33: note: remove std::move call here
        internal_feature->set_termfreq(std::move(compute_termfreq()));
                                       ^~~~~~~~~~                  ~
```

So actually we need to remove that `std::move()` for NRVO to be used (and the returned type is a prvalue, so it's not needed anyway).

VaibhavKansagara:
> Thanks. Squashed, rebased and pushed to master as [95800e3](https://github.com/xapian/xapian/commit/95800e3e991fcf5531b95884c2545f533f329434). I also tweaked the commit message to add a missing space (in "computed.After").
> 
> It seemed worth merging what we have here as much better that what was on master, but I think there's scope for further clean-up.
> 
Done.

PR #257, opened 2019-06-16
Make letor build under VPATH

jaylett:
* create the include/xapian-letor directory if necessary; otherwise
   we can't generate headers into it
 * ensure the include from our build tree is part of the CPPFLAGS
 * use full xapian-letor/HEADER.h path instead of just HEADER.h;
   this really only matters for generated files (since they'll not
   be in the same directory during VPATH build)
 * in our public headers, include our own headers using <> rather
   than "". This matches what we do in core, and means that
   relative path includes won't look in the directory of the
   current source file -- which due to the previous change we no
   longer need under any circumstance


xapian-letor/configure.ac
```diff
@@ -569,3 +569,12 @@ AC_CONFIG_FILES([
...
 AC_CONFIG_FILES([tests/runtest], [chmod +x tests/runtest])
 AC_OUTPUT
+
+dnl There are no files generated by AC_OUTPUT in the following directories
+dnl and we need to ensure they exist so that the rest of configure or make
+dnl won't fail because they don't exist when srcdir != builddir.
+if test yes = "$vpath_build" ; then
+  for dir in include/xapian-letor ; do
+    AS_MKDIR_P(["$dir"])
+  done
```
ojwb: The loop seems like a bit of premature generality - I wouldn't expect us to grow more such directories.

But wouldn't it make more sense to handle this in `generate_exceptions`, since that's what requires this to exist?  That encapsulates the requirement along with where it's required, and is more robust (the user can even `rm -rf include/xapian-letor` after running `configure` and it should still get recreated), and slightly lazier (since we don't run any extra code in a VPATH build from a tarball, where the generated files are already present at configure time).

ojwb:
Looks good, though see comment above about where to mk the dir.

xapian-letor/configure.ac
jaylett: Yes, that's a good idea — I just copied stuff from -core, which has more directories (hence the loop) but also should move that into `generate_exceptions`.

jaylett: Oh, except in -core it can't because immediately after the loop, it builds `include/xapian/version.h`. I think ensuring that the output directory exists in `generate_exceptions` is correct even there, though. I'll add another commit…

jaylett: That's not going to work, because I didn't test it at my end. Sigh.

ojwb: Ah right - yes, I'd agree ensuring the output directory exists in the script that uses it is the better approach generally.

xapian-core/generate-exceptions
```diff
@@ -21,7 +21,9 @@ use strict;
 use exception_data qw(
     $copyright $generated_warning @baseclasses @classes %classcode
 );
+use File::Path qw(make_path);
 
+make_path("include/xapian");
```
ojwb: The appveyor CI build for mingw fails with:

```
/usr/bin/perl -w -I"." "./generate-exceptions"
"make_path" is not exported by the File::Path module
Can't continue after import errors at ./generate-exceptions line 24
```

It looks like `make_path` is the new name and this used to be called `mkpath` before File::Path 2.00.

According to `corelist -a File::Path`, perl 5.9.4 with File::Path 1.08 is the most recent version without `make_path`.  That's a pretty ancient perl version but I think just changing `make_path` to `mkpath` would allow this to work there (for a single directory argument they should do the same thing) so I guess we should if only to keep the CI working.

ojwb: Actually, I think I misread the docs and I think it's before File::Path 2.08 which means before Perl 5.11.1.

ojwb:
Thanks - mingw build has passed, so rebased and merged as efaaad5e6eeeff0a85a89da4fe63ca960c149836.

PR #258, opened 2019-06-18
Throw DatabaseNotFoundError when DB doesn't exist

Kronuz:
Regarding https://trac.xapian.org/ticket/773, and after having been actively using the `DatabaseNotFoundError` for a while, this is what I've found are the two missing points where Xapian should throw `DatabaseNotFoundError` instead of `DatabaseOpeningError`.

### DIFF
--- a/xapian-core/backends/dbfactory.cc
+++ b/xapian-core/backends/dbfactory.cc
@@ -194,7 +194,11 @@ Database::Database(const string& path, int flags)
 
     struct stat statbuf;
     if (stat(path.c_str(), &statbuf) == -1) {
-	throw DatabaseOpeningError("Couldn't stat '" + path + "'", errno);
+	if (errno == ENOENT) {
+	    throw DatabaseNotFoundError("Couldn't stat '" + path + "'", errno);
+	} else {
+	    throw DatabaseOpeningError("Couldn't stat '" + path + "'", errno);
+	}
     }
 
     if (S_ISREG(statbuf.st_mode)) {
@@ -268,7 +272,7 @@ Database::Database(const string& path, int flags)
 	throw FeatureUnavailableError("Flint backend no longer supported");
     }
 
-    throw DatabaseOpeningError("Couldn't detect type of database");
+    throw DatabaseNotFoundError("Couldn't detect type of database");
 }
 
 /** Helper factory function.


xapian-core/backends/dbfactory.cc
```diff
@@ -268,7 +272,7 @@ Database::Database(const string& path, int flags)
 	throw FeatureUnavailableError("Flint backend no longer supported");
     }
 
-    throw DatabaseOpeningError("Couldn't detect type of database");
+    throw DatabaseNotFoundError("Couldn't detect type of database");
```
ojwb: If we get to here `path` exists and is a directory.

What's you reasoning for this case being "not found"?

And if this is "not found", why not the case around line 229/233 above where `path` exists but isn't a regular file or directory?

ojwb:
The CI failure is unrelated (we see that from time to time - I think it's some sort of test harness race which only manifests on platforms where you can't just delete an open file).

I do think we should have test coverage for this to ensure that the desired exception is thrown in each of the cases we care about, and particularly to make sure we don't regress on that.

xapian-core/backends/dbfactory.cc
Kronuz: The reasoning of when it should return `DatabaseNotFoundError` instead of `DatabaseOpeningError` is, I believe, simple. If you try `DB_OPEN` and it fails, the failure should be:

1. `DatabaseNotFoundError` if passing `DB_CREATE` or `DB_OPEN_OR_CREATE` or `DB_CREATE_OR_OVERWRITE` on the same `path` would succeed.

2. `DatabaseOpeningError` if passing `DB_CREATE` or `DB_OPEN_OR_CREATE` or `DB_CREATE_OR_OVERWRITE` on the same `path` would still fail.

ojwb: That reminds me that we discussed this before, and looking back in addition to the rule above we both thought you should also get `DatabaseNotFoundError` if the parent directories don't all exist: https://github.com/xapian/xapian/pull/213#issuecomment-450450384

So I guess the rule is if `mkdir -p path` followed by opening a `WritableDatabase` on `path` with `DB_CREATE` etc would succeed (or be expected to - I don't think we want to get into checking permissions, etc).

Anyway, this still seems logical (if we make the obvious allowances for read-only backends like current honey).

ojwb:
I've merged and added a regression test (c98b213992b89ead0072a213481c923f505a66b7).

PR #260, opened 2019-06-21
Fix parse_query_string method

VaibhavKansagara:
Previously the query string was broken into tokens seperated by spaces
instead of being parsed as a whole unit.


xapian-letor/ranker/ranker.cc
```diff
@@ -196,12 +196,15 @@ static std::pair<string, string>
...
-	j = query_line.find_first_of(' ', i);
-	token.push_back(query_line.substr(i, j - i));
+    size_t i = 0;
+    size_t j = query_line.find_first_of(' ', i);
+    token.push_back(query_line.substr(i, j - i));
+    if (j != string::npos) {
+	i = query_line.find_first_of("'", j);
+	j = string::npos;
+	if (j > i) {
+	    token.push_back(query_line.substr(i, j - i));
```
jaylett: Use the single parameter version of `substr()` and it'll return the substring to the end of `query_line`.

jaylett:
Oh, and as with any "fix X" kind of change — it'd be good to have a regression test that exercises the different between the old code and this.

xapian-letor/ranker/ranker.cc
```diff
@@ -196,12 +196,15 @@ static std::pair<string, string>
...
     vector<string> token;
-    size_t j = 0;
-    while (j < query_line.size()) {
-	size_t i = query_line.find_first_not_of(' ', j);
-	if (i == string::npos) break;
-	j = query_line.find_first_of(' ', i);
-	token.push_back(query_line.substr(i, j - i));
+    size_t i = 0;
+    size_t j = query_line.find_first_of(' ', i);
+    token.push_back(query_line.substr(i, j - i));
```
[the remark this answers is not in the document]
ojwb: Actually, that'll work as written - `string::npos` is the maximum value of `string::size_type`, so if `j == string::npos` this will ask for a substring which is >= the length of the string and that will just return the rest of the string.  That's perhaps not entirely clear to the reader without some thought though.

Also, `i` is always zero here and only really used when looking for single quotes.

ojwb:
There's a typo in the commit message - "seperated" should be "separated".  I don't know of a way to fix that in the PR without rewriting history, but I can fix when merging.

xapian-letor/ranker/ranker.cc
```diff
@@ -196,12 +196,15 @@ static std::pair<string, string>
...
 {
     vector<string> token;
-    size_t j = 0;
-    while (j < query_line.size()) {
-	size_t i = query_line.find_first_not_of(' ', j);
-	if (i == string::npos) break;
-	j = query_line.find_first_of(' ', i);
-	token.push_back(query_line.substr(i, j - i));
+    size_t i = 0;
+    size_t j = query_line.find_first_of(' ', i);
```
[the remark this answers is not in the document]

VaibhavKansagara:
> I think you should also take advantage of the fact that you're working on this function to:
> 
> 1. Give it some sort of documentation at the top of the function;
> 2. Stop using the `vector<string>` given you immediately pull the two tokens back out and work on them, and then build a `pair<string, string>` out of them;
> 3. You can strip the quotes around the query string while parsing the line, rather than building a `string`, checking the character at each end, and then erasing each character one at a time. This would be a bit more space and processing efficient, and probably be a little clearer (with some suitable comments around the parsing section).
>
fixed in a fixup commit.

xapian-letor/ranker/ranker.cc
```diff
@@ -192,35 +192,31 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
+	j = query_line.length() - 1;
+	// check if the last character is '
+	if (query_line[j] != '\'') {
+	    throw LetorParseError("Could not parse Query file at line:" +
+				   str(line_number));
 	}
+	querystr = query_line.substr(i, j - i);
+    } else {
+	throw LetorParseError("Could not parse Query file at line:" +
+			       str(line_number));
```
ojwb: This exception could be thrown earlier, right after we determine the value of `j`.

Saving a bit of work in an exceptional case generally isn't important (because such cases should happen rarely, and exception handling is slow anyway), but dealing with the exceptional cases promptly often simplifies things because there are then fewer states to consider in the code which follows. E.g. here we wouldn't have to think about what `query_line.substr(i, j - i)` does when `j` is `string::npos` if we dealt with that case right away.

xapian-letor/ranker/ranker.cc
```diff
@@ -192,35 +192,31 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
+    string::size_type i = 0;
+    string::size_type j = query_line.find_first_of(' ', i);
+    string qid = query_line.substr(i, j - i);
+    string querystr;
     if (j != string::npos) {
-	i = query_line.find_first_of("'", j);
-	j = string::npos;
-	if (j > i) {
-	    token.push_back(query_line.substr(i, j - i));
+	i = query_line.find_first_not_of("' ", j);
```
ojwb: This code is a bit odd - we don't actually check for a `'` at the start of the query string, but instead skip arbitrary numbers of spaces and `'`.

So this code won't complain about cases like:

```
1234 no open quote'
```

And cases like this will be mishandled:

```
1234 ''tis a shame'
```

Either `'` in query string is allowed (in which case this goes wrong because it strips any leading `'` in the query string, or it isn't (in which case this should give an error but doesn't), or `'` needs to be escaped in the query string in some way (in which case we need to handle undoing such escaping, and this case also ought to be an error because the `'` isn't escaped).

The format for this file ideally ought to specify if `'` in the query string is allowed and if it needs escaping and how.  If not, we'll need to decide how to resolve that underspecification.

This code also mishandles cases where the query string is empty or consists only of spaces, such as:

```
1234 ''
1234 ' '
```

For these, `i` is `string::npos` and it looks to me like that will cause `query_line.substr(i, j - i)` to throw `std::out_of_range` (see https://en.cppreference.com/w/cpp/string/basic_string/substr)

xapian-letor/ranker/ranker.cc
```diff
@@ -192,35 +192,31 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
+	j = query_line.length() - 1;
+	// check if the last character is '
+	if (query_line[j] != '\'') {
```
ojwb: We need to take care to handle the case where there's a `'` and nothing after that appropriately:

```
1234 '
```

The closing quote must not be at the same string index we interpreted as being the opening quote.

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/ranker.cc
VaibhavKansagara: > Either `'` in query string is allowed (in which case this goes wrong because it strips any leading `'` in the query string, or it isn't (in which case this should give an error but doesn't), or `'` needs to be escaped in the query string in some way (in which case we need to handle undoing such escaping, and this case also ought to be an error because the `'` isn't escaped).
> 
I think it should be allowed. The fixup commit handles all the cases you mentioned and also we won't need to undo the escaping.

xapian-letor/ranker/ranker.cc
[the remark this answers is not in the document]
ojwb: We do currently use `size_t` quite widely in the Xapian code where `size_type` would be more correct, but that is really a bug (albeit a latent one on all platforms I know of) and I'd certainly encourage use of the correct type.

xapian-letor/ranker/ranker.cc
```diff
@@ -198,21 +198,20 @@ parse_query_string(const string & query_line, int line_number)
...
-	querystr = query_line.substr(i, j - i);
-    } else {
+    i = query_line.find_first_of("'", j);
+    j = query_line.length() - 1;
+    // check if the last character is '
+    if (query_line[j] != '\'' || j == i) {
 	throw LetorParseError("Could not parse Query file at line:" +
 			       str(line_number));
     }
+    querystr = query_line.substr(i + 1, j - i - 1);
```
ojwb: A small thing, but it's cleaner to declare at the point of first assignment where possible, and there's no longer a reason we can't for `querystr`.

ojwb:
@jaylett Are you happy your "changes requested" has been addressed?

xapian-letor/ranker/ranker.cc
ojwb: OK.

It would be good to add the cases which were wrongly handled to serve as regression test cases.

For the cases which were wrongly failing, we can just add them to a query file and check it parses without an exception.

For the cases which were wrongly being accepted we can just test that `LetorParseError` is thrown.

The `1234 ''tis a shame'` case seems harder as QueryParser parses that the same without the leading `'`.

VaibhavKansagara: sure will add that in a fixup commit.

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/ranker.cc
```diff
@@ -192,32 +192,29 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
+	throw LetorParseError("Could not parse Query file at line:" +
+			       str(line_number));
     }
-    // Query file is in the format: <qid> <query_string>
-    // Therefore, <qid> goes into token[0] and <query_string> to token[1]
-    // Exceptions for parse errors
-    if (token.size() != 2) {
-	throw LetorParseError("Could not parse Query file at line:" + str(line_number));
+    string qid = query_line.substr(i, j - i);
+    i = query_line.find_first_of("'", j);
```
jaylett: As I think @ojwb noted, you don't actually need `size_type i` until here and could just use `0` (or in `find_first_of()` it's the default for that parameter). That would probably be slightly more readable, but is a very minor point.

xapian-letor/ranker/ranker.cc
```diff
@@ -192,32 +192,29 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
     }
 }
 
+// Query file is in the format: <qid> '<query_string>'
```
jaylett: This isn't quite true. As the parser is written, it will accept arbitrary characters between the first space (`j` up to line 207) and the first single quote (`i` from line 206). This is fine (it can cope with arbitrary numbers of spaces between the two terms) but perhaps needs a little note in the comment.

xapian-letor/ranker/ranker.cc
```diff
@@ -192,32 +192,29 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
 static std::pair<string, string>
 parse_query_string(const string & query_line, int line_number)
 {
```
jaylett: I had to think twice about why this will work (a) when `query_line` is an empty string, and (b) when `i` is `::npos`. I don't think we should change the code, because it's quite tight as it is, but perhaps some more comments would help future readers.

I _do_ think we should have explicit tests for an empty line, a line without a space, and perhaps a line that ends with a space but has no single quote at all. At least that would give someone glancing at this confidence that the code does cope with all the edge cases.

xapian-letor/ranker/ranker.cc
```diff
@@ -192,32 +192,29 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
-    size_t j = 0;
-    while (j < query_line.size()) {
-	size_t i = query_line.find_first_not_of(' ', j);
-	if (i == string::npos) break;
-	j = query_line.find_first_of(' ', i);
-	token.push_back(query_line.substr(i, j - i));
+    string::size_type i = 0;
+    string::size_type j = query_line.find_first_of(' ', i);
+    if (j == string::npos) {
+	throw LetorParseError("Could not parse Query file at line:" +
```
jaylett: I wonder if a more explicit message ("missing space between fields in Query file" or something) might be helpful here.

xapian-letor/tests/api_letor.cc
```diff
@@ -263,6 +263,36 @@ DEFINE_TESTCASE(preparetrainingfileonedb, generated)
...
+    // All those cases which are not valid.
+    string db_path = get_database_path("apitest_listnet_ranker1",
+				       db_index_one_document);
+    string data_directory = test_driver::get_srcdir() + "/testdata/";
+    string query = data_directory + "parse_query_noopenquote.txt";
+    string qrel = data_directory + "qrelone.txt";
+    string training_data = data_directory + "training_output.txt";
+    TEST_EXCEPTION(Xapian::LetorParseError,
+		   Xapian::prepare_training_file(db_path, query, qrel, 10,
+						 "training_output.txt"));
```
jaylett: With three cases (and especially if you add more) I might be inclined to define a `TEST_PARSE_EXCEPTION` macro so each instance becomes a single line. (You could also inline the `query` expressions, since they change for every test.)

xapian-letor/tests/api_letor.cc
```diff
@@ -263,6 +263,36 @@ DEFINE_TESTCASE(preparetrainingfileonedb, generated)
...
 
+DEFINE_TESTCASE(test_parse_querystring, generated)
+{
+    // All those cases which are not valid.
+    string db_path = get_database_path("apitest_listnet_ranker1",
+				       db_index_one_document);
+    string data_directory = test_driver::get_srcdir() + "/testdata/";
+    string query = data_directory + "parse_query_noopenquote.txt";
+    string qrel = data_directory + "qrelone.txt";
+    string training_data = data_directory + "training_output.txt";
```
jaylett: This variable isn't used.

xapian-letor/tests/testdata/queryone.txt
```diff
@@ -1 +1 @@
-110001 'tigers'
+110001 'tigers score'
```
jaylett: This isn't really related to the rest of the changes, so I don't think should be here. More worryingly, a change like this should really have broken one or more of the other tests. I suspect it hasn't because `score` isn't a term in the databases used or something, but can you double check that there's nothing wrong with the other tests? (They could be under-testing the desired output. for instance.)

xapian-letor/tests/api_letor.cc
ojwb: Also, output files should be written to the build tree not the source tree (which are often the same but don't have to be) so `"training_output.txt"` as actually used below is good, and the value computed here wouldn't be.

When building with `srcdir` != `builddir` (also known as a "VPATH build" after the VPATH feature in make used to implement it) one `srcdir` may be shared between several build trees and so generally should be treated as read-only.

xapian-letor/tests/testdata/queryone.txt
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/tests/api_letor.cc

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/ranker.cc
VaibhavKansagara: The case of an empty string will never reach this part of the code as it will be eliminated on line 265 in ranker.cc. As for the remaining cases, I have added corresponding query files.

VaibhavKansagara: >  perhaps a line that ends with a space but has no single quote at all.
>
xapian-check-patch is complaining about trailing whitespace in this case.

jaylett: If it is never supposed to cope with an empty string, then put an assert in there to state that clearly.

jaylett: Oh, and equivalent to a no-quote line ending with a space would be something like:

```
12345 noquote
```

I believe that will test the same code path…

xapian-letor/tests/api_letor.cc
jaylett: I meant go even further, because the invocations are so similar:

```c++
#define TEST_PARSE_EXCEPTION(testfile) TEST_EXCEPTION(Xapian::LetorParseError,\
    Xapian::prepare_training_file(db_path, data_directory + testfile, qrel, 10, "training_output.txt"));
```

or something like that.

VaibhavKansagara: Ah I will change that now

xapian-letor/ranker/ranker.cc

xapian-letor/tests/harness/testutils.h
```diff
@@ -90,4 +90,11 @@ void test_mset_order_equal(const Xapian::MSet &mset1,
 	expected_exception = NULL;\
     } while (0)
 
+/// Check that CODE throws exactly Xapian LetorParseError TYPE.
+#define TEST_PARSE_EXCEPTION(db_path, qrel, data_directory, testfile)\
+	TEST_EXCEPTION(Xapian::LetorParseError,\
+		       Xapian::prepare_training_file(db_path, data_directory +\
+						     testfile, qrel, 10,\
+						     "training_output.txt"));
```
ojwb: It's better not to include the `;` on the end of the definition of a function-like macro like this where a `;` is expected to be added when the macro is used.  People will tend to add a `;` without thinking about it, editors which auto-indent will do the right thing, etc.

In simple cases, the extra `;` you get if the macro includes one and there's also one where it is used makes no difference, but it can - e.g. consider:

```
if (foo)
    TEST_PARSE_EXCEPTION(db_path, qrel, data_directory, "foo.txt");
else
    TEST_PARSE_EXCEPTION(db_path, qrel, data_directory, "bar.txt");
```

The first `TEST_PARSE_EXCEPTION` would currently expand to give a double `;` which means the `if` statement ends there and is followed by an empty statement, and then you'll get `error: ‘else’ without a previous ‘if’`` or similar for the `else`.

I think there are also situations where this can even result in a successful compilation which does the wrong thing but an example doesn't come to mind right now.

xapian-letor/tests/harness/testutils.h
```diff
@@ -90,4 +90,11 @@ void test_mset_order_equal(const Xapian::MSet &mset1,
 	expected_exception = NULL;\
     } while (0)
 
+/// Check that CODE throws exactly Xapian LetorParseError TYPE.
```
ojwb: This macro doesn't take parameters `CODE` or `TYPE` (unlike `TEST_EXCEPTION()`) so the comment doesn't really make sense.

Also, it's conventional to make macro parameters all-caps.

xapian-letor/tests/harness/testutils.h
```diff
@@ -90,4 +90,11 @@ void test_mset_order_equal(const Xapian::MSet &mset1,
...
+#define TEST_PARSE_EXCEPTION(db_path, qrel, data_directory, testfile)\
+	TEST_EXCEPTION(Xapian::LetorParseError,\
+		       Xapian::prepare_training_file(db_path, data_directory +\
+						     testfile, qrel, 10,\
```
ojwb: The wrapping here is unconventional as the arguments of `+` aren't aligned vertically.  Better to wrap after `db_path,` so `data_directory + testfile,` doesn't need to be split (the macro body only needs a 4 column indent rather than the 8 currently used so it should fit).

xapian-letor/tests/harness/testutils.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/tests/harness/testutils.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/tests/harness/testutils.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/ranker.cc
```diff
@@ -192,32 +193,31 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
-    vector<string> token;
-    size_t j = 0;
-    while (j < query_line.size()) {
-	size_t i = query_line.find_first_not_of(' ', j);
-	if (i == string::npos) break;
-	j = query_line.find_first_of(' ', i);
-	token.push_back(query_line.substr(i, j - i));
+    Assert(!query_line.empty());
+    string::size_type j = query_line.find_first_of(' ');
+    if (j == string::npos) {
```
AyushP123: You should also check for `j == 0`. `          '123'` will be a valid query according to this.

xapian-letor/ranker/ranker.cc
```diff
@@ -192,32 +193,31 @@ write_to_file(const std::vector<Xapian::FeatureVector> & list_fvecs, const strin
...
 {
-    vector<string> token;
-    size_t j = 0;
-    while (j < query_line.size()) {
-	size_t i = query_line.find_first_not_of(' ', j);
-	if (i == string::npos) break;
-	j = query_line.find_first_of(' ', i);
-	token.push_back(query_line.substr(i, j - i));
+    Assert(!query_line.empty());
+    string::size_type j = query_line.find_first_of(' ');
```
AyushP123: Also lets make sure that its mentioned in the docs that space is used as a delimiter between query and query id. Right now its quite vague as this is the description ( although the delimiter looks like a space, it would be better to mention it ).

> Query` file: This file has information of queries to be involved in learning and its id. It should be formatted in such a way:

> 2010001 'landslide,malaysia'
> 2010002 'search,engine'
> 2010003 'Monuments,of,India'
> 2010004 'Indian,food'

> where 2010xxx being query-id followed by a comma separated query in single-quotes.

ojwb: I'd agree it would be better to be more precise about what separators are allowed between the fields.

Are commas separating terms required?  The docs currently give examples such as `2010003 'Monuments of India'`, but then just below that say "a comma separated query".  The code seems to parse these strings using `Xapian::QueryParser::parse_query()` which will work with spaces or commas between terms.

If this is an externally-specified file format which says commas are required, then we should probably follow that.  But if this is a xapian-letor-specific format, it seems more helpful to allow query strings as entered by the user.

Looking at what the docs currently say, I also noticed "All the fields are whitespace delimited" about the Qrel file format, which appears to be incorrect - the parsing in `load_relevance()` in `xapian-letor/ranker/ranker.cc` seems to look for spaces, whereas even if "whitespace" is read as "horizontal whitespace" that includes tabs.  Requiring spaces seems OK (unless there's an external specification for this format which says tabs are allowed) - the problem here is the docs not matching the code.

VaibhavKansagara: > If this is an externally-specified file format which says commas are required, then we should probably follow that. But if this is a xapian-letor-specific format, it seems more helpful to allow query strings as entered by the user.
>
I think space separated query are fine as the same format was used in FIRE Dataset.
I have updated the docs accordingly.

VaibhavKansagara:
> A few minor changes from my side. Also while I was going through this PR, I realised that there is no check to ensure that the query ids should be unique. You need to add that check ( Although I am not sure if it should be a part of this PR )
>
I will add that check in another PR.

xapian-letor/ranker/ranker.cc
VaibhavKansagara: fixed in a fixup commit. I have also added a test case regarding the same.

AyushP123: Message should be "Empty query id found " .... Test case should be just a space followed by query

>   'query'

xapian-letor/docs/letor.rst
```diff
@@ -30,14 +30,15 @@ Learning the model
 As mentioned before, this process requires a training file in the above format. xapian-letor API empowers you to generate such training file. But for that you have to supply some information files like:
 
     1. Query file: This file has information of queries to be involved in
-    learning and its id. It should be formatted in such a way::
+    learning and its id. Here space is used as delimiter betweem query id and query string.
```
ojwb: Typo: "betweem"

I'll just fold a fix for this in as I merge.

ojwb:
Rebased, autosquashed and merged to master as 27852165271452921e89b9e30c68425962d81f31.

xapian-letor/docs/letor.rst
```diff
@@ -30,14 +30,15 @@ Learning the model
...
+    It should be formatted in such a way::
 
       2010001 'landslide malaysia'
       2010002 'search engine'
       2010003 'Monuments of India'
       2010004 'Indian food'
 
-    where 2010xxx being query-id followed by a comma separated query in
+    where 2010xxx being query-id followed by a space separated query in
```
AyushP123: It's can be space separated or comma separated query as @ojwb mentioned. "followed by a query in single quotes"

AyushP123: We will fix this and more issues in the documentation in a different PR.

VaibhavKansagara: @AyushP123 Which other issues are you talking about?.

AyushP123: I should have framed it better: We will fix this along with other with other issues in the documentation, which will form as we keep on developing letor, in a separate PR.

PR #261, opened 2019-06-21
Fix score method

VaibhavKansagara:
Add only those FeatureVector which are found in the qrel file.


jaylett:
It's not clear at first blush why this is correct. Is there a test case you can write which fails on the old code and passes on the new? Or if not, I think we need a more detailed commit message to explain the change.

(The test failure is a transient failure to resolve github.com from Travis. Pushing a fixup commit, or force pushing if it's just a new commit message, should get it to pass next time.)

ojwb:
> (The test failure is a transient failure to resolve github.com from Travis. Pushing a fixup commit, or force pushing if it's just a new commit message, should get it to pass next time.)

I've poked it to retry that build.

ojwb:
> [...] their labels remained undefined and as a result gave unexpected values for NDCG scores like nan.

Can we turn that into a testcase which fails before your change and passes after?

If it requires valgrind or ubsan or assertions enabled to fail, that's OK (though the ideal is something that always fails in a standard build).  So if it's not possible to tell via the API and the NDCG score should never be NaN, adding `Assert(!isnan(ndcg_score));` at suitable point or points in the code would achieve this.

VaibhavKansagara:
> 
> Can we turn that into a testcase which fails before your change and passes after?
> 
Initially, I thought it was difficult to construct a test case as the undefined value doesn't necessarily always give nan but I have managed to generate an error message in case of undefined value as 
`conditional jump or move depends on uninitialised value(s)`.

ojwb:
> > Can we turn that into a testcase which fails before your change and passes after?
> 
> Initially, I thought it was difficult to construct a test case as the undefined value doesn't necessarily always give nan but I have managed to generate an error message in case of undefined value as
> `conditional jump or move depends on uninitialised value(s)`. So we wouldn't need to test for nan.

OK - that error is actually from valgrind (the test harness uses some valgrind hooks to collect valgrind errors during running a testcase).

xapian-letor/tests/api_letor.cc
```diff
@@ -918,3 +918,24 @@ DEFINE_TESTCASE(colltfcolllenfeature, generated)
...
+    string data_directory = test_driver::get_srcdir() + "/testdata/";
+    string query = data_directory + "querythree.txt";
+    string qrel = data_directory + "score_qrel.txt";
+    string training_data = data_directory + "training_data_ndcg.txt";
+    ranker.set_database_path(db_path);
+    ranker.set_query(Xapian::Query("score"));
+    ranker.train_model(training_data, "ListNet_Ranker");
+    ranker.rank(mymset, "ListNet_Ranker");
+    ranker.score(query, qrel, "ListNet_Ranker", "ndcg_score_test.txt",
+		 10);
```
ojwb: This line doesn't actually need wrapping.  I'll squash that trivial tweak on top and merge.

ojwb:
Rebase with autosquash and pushed as 283ab90a3010fc6e0a9efcb0aa418aee585b8b63.

ojwb:
Oh, just noticed we aren't shipping the new data files, so `make distcheck` will fail - pushed a fix for that as a3c8fa052a43cecf14a20cd0671d87364e67eebf.

PR #262, opened 2019-06-22
bootstrap: Adding pkg.m4 directory on Mac OS.

barufa:
Currently, if aclocal is installed by bootstrap, it will try to find pkg.m4 at /usr/share/aclocal. However, if we are on Mac OS it is necessary to add /usr/local/share/aclocal and /opt/local/share/aclocal. If pkg-config was installed with homebrew, pkg.m4 would be at /usr/local/share/aclocal. But if it was installed with darwing ports we need /opt/local/share/aclocal.

### DIFF
--- a/xapian-core/HACKING
+++ b/xapian-core/HACKING
@@ -64,7 +64,7 @@ On Mac OS X, if you're using macports you'll want the following:
 
 If you're using homebrew you'll want the following::
 
-    brew install libmagic pcre
+    brew install libmagic pcre pkgconfig
 
 If you're doing much development work, you'll probably also want the following
 tools installed:
--- a/bootstrap
+++ b/bootstrap
@@ -637,6 +637,10 @@ aclocal_system_needed=
 if [ ! -r "$aclocal_ac_dir/pkg.m4" ] ; then
   if [ -r /usr/share/aclocal/pkg.m4 ] ; then
     aclocal_system_needed=/usr/share/aclocal
+  elif [ -r /usr/local/share/aclocal/pkg.m4 ] ; then
+    aclocal_system_needed=/usr/local/share/aclocal
+  elif [ -r /opt/local/share/aclocal/pkg.m4 ] ; then
+    aclocal_system_needed=/opt/local/share/aclocal
   fi
 fi
 


ojwb:
It doesn't seem great to have a growing list of directories to check, but I don't have a much better idea.  We could run `aclocal --print-ac-dir` for the system aclocal, but it seems we'd still need a list of directories in case that's not installed (`pkg.m4` would typically be installed by the package of `pkg-config`, and that could be installed without the package of `automake` being installed).

So overall I think the change itself is OK.

However, the commit message has multiple formatting problems.

https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages has some useful advice:

> Start with a short (50 characters) summary line.
>    git (and github) are designed to work better this way.

Your commit message does start with a < 50 character summary (which it's good), but then continues from that with more detail on the same line (which isn't).  This creates display problems from tools which are built around the standard git commit message formatting assumptions, so such commit messages make the git repo less usable for other developers.  For example, here's what `git log --oneline` output look like in an 80 column terminal:

```
81b8e81712a9 bootstrap: Adding pkg.m4 directory on Mac OS. Currently, if aclocal
 is installed by bootstrap, it will try to find pkg.m4 at /usr/share/aclocal. Ho
wever, if we are on Mac OS it is necessary to add /usr/local/share/aclocal and /
opt/local/share/aclocal. If pkg-config was installed with homebrew, pkg.m4 would
 be at /usr/local/share/aclocal. But if it was installed with darwing ports we n
eed /opt/local/share/aclocal.
4393a20ef133 (master, origin/master) Eliminate use of anonymous struct
83c30bc2e020 [buildbot] Fix get_tarballs.py xz probing
d86f328ff89e xapian-check-patch: Fix false positive
7336aa8bda5f Remove limitations of $subdb and $subid
acdc03528241 xapian-check-patch: Fix reversed conditional in previous change
9fd5f31be651 xapian-check-patch: Improve whitespace checks
0207e7c11379 Make class indentation conform with preferred style
186e5869d590 xapian-check-patch: Check access specifier indents
d9f787a2731a xapian-check-patch: Add $header flag
6fdc157841a5 Remove unused <cerrno>
a63693ab8c94 Add missing <cerrno>
28fcf7f4f6dc Fix signed vs unsigned comparison warning
4b731e7897b4 Add high_resolution_clock to track slow tests
220119d10646 Add Database::size() method
efaaad5e6eee Ensure output directory in generate_exceptions
4134fb6a91b3 Make letor build under VPATH
```

Many of those commits have a more detailed commit message, but `git log --oneline` just shows the summary (which it assumes is up to the first blank line).  This assumption is used throughout git - e.g. `git format-patch` generates email with the summary as the subject and the rest in the message body.

So there should be a blank line (so two linefeeds) between the commit summary and the more detailed description.

> The summary should be in the imperative (“Fix bug on OS X” rather than “Fixed bug on OS X”). This matches git’s automatic messages around merges, reverts and so on.

So "Add ..." would be better than "Adding ..." here.

We're adding two directories, so "Add pkg.m4 directories ..." would be more accurate.

Also, don't put a full stop/period after the summary (like you wouldn't for a document title or an email subject.)

>    Follow that with more detail as needed, wrapping long lines at 72 characters (one exception is that long URLs are best not wrapped).

Your commit message is just one long line, not wrapped at all.

>    Describe the effect, not the code. The important thing is for people to be able to read the commit message and understand what you were trying to achieve when you made those changes. That way, if someone needs to work on that part of the code in future, they can understand the purpose of it, and not accidentally remove some useful functionality. (Tests help here, but the commit message is very important.)

The actual content of your commit message is good (except for a minor typo: "darwing"), the problems are essentially formatting it to match git conventions.

In this case the changes are reviewed OK, and the only way to rewrite the commit message is to force-push the branch, so if you do that we can get this merged.

barufa:
Okay, I will make a force-push with the changes.

barufa:
I also add a modification to HACKING. It seems that `pkgconfig` is not a dependency of `automake`.
Maybe I can try to use the command `find` to avoid having a growing list of directories. Is it possible?

xapian-core/HACKING
```diff
@@ -64,7 +64,7 @@ On Mac OS X, if you're using macports you'll want the following:
 
 If you're using homebrew you'll want the following::
 
-    brew install libmagic pcre
+    brew install libmagic pcre pkgconfig
```
ojwb: Updating the documentation of what packages to install is useful, but...

* This change doesn't belong in a commit described as "bootstrap: Add pkg.m4 directories on Mac OS".  Each commit should do one thing, and while this is somewhat connected, it's really a separate change - "Add pkg-config to required packages for build" or something like that.

* It's also not the complete change for that - the `pkg-config` should be added to the list of packages needed for Omega on Ubuntu and Debian.

* And `HACKING` is a legacy document we're in the process of getting rid of - the preferred location for listing packages to install is now: https://github.com/xapian/xapian-developer-guide/blob/master/getting-started/index.rst

It's OK to update HACKING, but equivalent changes should be made there too.  And if only one gets updated, it's better for it to be the developer guide.

So I'm going to revert the change to HACKING and get the other part merged.

ojwb:
Using `find` seems problematic - we've no idea how big the filing system is or what might be mounted on it.  So it potentially takes a very long time, and could generate network traffic if there are NFS or SMB mounts.

And hopefully I can get the patches to automake merged and we can stop having to build it here just so we can apply them.  Then for most developers the system automake can be used again.

I've merged without the extra change, so closing.

A more complete version of the extra change as a separate PR would be good.

PR #263, opened 2019-06-27
Make letor releasable

VaibhavKansagara:
This fixes errors while running make dist and make distcheck and thus makes letor officially releasable. This would only work after #259 is merged.


xapian-letor/Makefile.am
```diff
@@ -108,4 +134,4 @@ noinst_HEADERS +=\
 	common/str.h\
 	common/stringutils.h
 
-DISTCHECK_CONFIGURE_FLAGS = "XAPIAN_CONFIG=$(XAPIAN_CONFIG)"
+DISTCHECK_CONFIGURE_FLAGS = "XAPIANLETOR_CONFIG=$(XAPIANLETOR_CONFIG)"
```
ojwb: This change seems wrong - `xapian-letor` doesn't use `xapianletor-config`, but it ought to use the in-tree xapian-config for the nested build when running `make distcheck`.

ojwb:
This seems to include a version of the changes from #259.

Neither merges cleanly with master, so I'm struggling to work out if there's actually anything in this PR which wasn't merged as part of #259.  If there is, please can you rebuild what's left to apply on current master and force push to the PR branch?  If not, please close.

xapian-letor/Makefile.am
```diff
@@ -115,9 +117,15 @@ include feature/Makefile.mk
...
-libgetopt_la_SOURCES = common/getopt.cc
+noinst_LTLIBRARIES += libcommon.la
+
+libcommon_la_SOURCES +=\
+	common/getopt.cc\
+	common/str.cc\
+	common/omassert.cc\
+	common/debuglog.cc\
+	common/errno_to_string.cc\
+	common/fileutils.cc
```
ojwb: I'm rather luke-warm on this part.

I don't think it's good to lump together "everything we use from common" and "GNU getopt (but only for systems where it isn't in libc)" like this.  Pretty much all the command line programs will use getopt, but probably most won't using debug logging or assertions.

Also, how come things build currently if these source files aren't currently getting linked in (except for common/str.cc)?

VaibhavKansagara: @ojwb You are right we don't need any other sources other than str.cc and getopt.cc.I will fix it in a fixup commit.

VaibhavKansagara: fixed in a fixup commit.

xapian-letor/Makefile.am
```diff
@@ -98,7 +98,7 @@ dist_m4data_DATA = m4-macros/xapianletor.m4
 lib_LTLIBRARIES = libxapianletor.la
 libxapianletor_la_SOURCES = $(lib_src)
 libxapianletor_la_LDFLAGS = \
-	$(XAPIAN_LDFLAGS) $(NO_UNDEFINED) -version-info $(LIBRARY_VERSION_INFO)
+	$(XAPIANLETOR_LDFLAGS) $(NO_UNDEFINED) -version-info $(LIBRARY_VERSION_INFO)
```
ojwb: `XAPIAN_LDFLAGS` was wrong here, but nothing sets `XAPIANLETOR_LDFLAGS` currently.

In `xapian-core/configure.ac` `XAPIAN_LDFLAGS` gets `-Bsymbolic-functions` added if that's supported - I think we want the equivalent for xapian-letor (`-Bsymbolic-function` is a linker option which binds references from the shared library to global symbols defined by the same library at library link time, which is more efficient that the default of resolving them when the library is dynamically loaded at runtime - the downside is that then you can't intercept uses of a symbol in the library by defining it in the program, but for a C++ API that is really unlikely to be something you'd want to be able to do).

xapian-letor/tests/Makefile.am
```diff
@@ -81,14 +82,46 @@ EXTRA_DIST = collate-test generate-api_generated api_all.h api_collated.h \
...
+	ndcg_output_ListNet_2.txt \
+	ndcg_output_ListNet_3.txt \
+	ndcg_output_svm_1.txt \
+	ndcg_output_svm_2.txt \
+	ndcg_output_svm_3.txt \
+	ndcg_score_test.txt \
+	training_output_data_one_doc.txt \
+	training_output_empty.txt \
+	training_output_three_correct.txt \
+	training_output1.txt
```
ojwb: These are all output files produced and used within a single testcase aren't they?

If so, I think it would be good for these testcases to at least check that the expected output file is created (ideally the content would be checked too in some appropriate way) and then remove it - e.g. see how I did in 4a27dcf65fcfcc670768dee90396dfca17ed304a for `ndcg_score_test.txt`.  If such a testcase fails the output file will be left behind, so one can still check it in the case where you'd want to.

It's still good to be cleaning up these files here (though I think they want to be in `CLEANFILES` so they are removed on `make clean` not just `make distclean`) in case they're left behind by a failing testcase or by an interrupted testsuite run.

VaibhavKansagara: fixed in a fixup commit.

xapian-letor/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/tests/Makefile.am
```diff
@@ -13,6 +13,7 @@ AM_CPPFLAGS = -I$(top_srcdir) -I$(top_srcdir)/common -I$(top_srcdir)/include \
 AM_CXXFLAGS += $(XAPIAN_CXXFLAGS)
 
 CLEANFILES =
+DISTCLEANFILES =
```
ojwb: There's no need to specify `DISTCLEANFILES` as empty here (I guess this is left over from when some files were being added to it, but they're now in `CLEANFILES` instead).

xapian-letor/tests/Makefile.am
```diff
@@ -81,14 +82,46 @@ EXTRA_DIST = collate-test generate-api_generated api_all.h api_collated.h \
...
+	testdata/training_data_one_document.txt \
+	testdata/training_data_three_correct.txt
+
+CLEANFILES +=\
+	err_output_listmle_1.txt \
+	err_output_listmle_2.txt \
+	err_output_listmle_3.txt \
+	err_output_ListNet_1.txt \
+	err_output_ListNet_2.txt \
+	err_output_ListNet_3.txt \
```
ojwb: Not a problem with this patch as such, but at some point it's probably worth renaming the `*_ListNet_*.txt` files to be `*_listnet_*.txt` so the naming is consistent with all the others files here (and with the naming of files in the codebase in general).

ojwb:
I tried `make distcheck` in the xapian-letor directory, which creates a distribution, then tries to unpack and build it in various ways (see the automake manual section 14.4 for all the details: https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html except that's not working as I write).

This manages to build and run most tests, but there's a failure:

```
Running test: parse_querystring... FileNotFoundError: No Query file found. Check path.
```

I think the paths passed to `TEST_PARSE_EXCEPTION` in `parse_querystring` need to include `data_directory` to work when srcdir != builddir.

xapian-letor/tests/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/tests/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

VaibhavKansagara:
> I think the paths passed to TEST_PARSE_EXCEPTION in parse_querystring need to include data_directory to work when srcdir != builddir.
>
I think it is because I haven't included some files in EXTRA_DIST.
Fixed in a fixup commit.

ojwb:
> > I think the paths passed to TEST_PARSE_EXCEPTION in parse_querystring need to include data_directory to work when srcdir != builddir.
> 
> I think it is because I haven't included some files in EXTRA_DIST.

Ah yes, that would explain it.

This didn't rebase at all cleanly - the fixup commits collide with each other, and the merge from master (which was probably necessary) doesn't help.

I ended up having to rebuild the commits instead.  I did this by cherry-picking the "Rename generated files" and "Improve output file handling for letor testcases" commits and their fixups onto current master and resolving the conflicts, then temporarily reverting `c4b7eb5222db Add testcases for feature related APIs` locally, and the rebuilt "Make letor releasable" commit was then the changes to the xapian-letor directory between that and your PR branch.  I'm confident I got this right, but it's fiddly and potentially error prone, and would have been much harder to do if there had been more overlap with changes to master.

Pushed to master as 8c2dbbcb3e362fd35edf46e526848b5cb61e6ce4, 8f84f7a0e21b561fe01ec18d1e26afa6363ee62c and fc76880b252d464da2eff69229934ab95918be9b.

It's a good idea to test your fixup commits actually rebase successfully before pushing them - you can create a temporary branch to do this:

```
git checkout -b scratch-branch
git rebase -i --autosquash master
```

And see how it goes.  Once done you can switch back and delete the temporary branch:

```
git checkout makedist
git branch -D scratch-branch
```

(`-D` suppresses the usual check to stop you deleting an unmerged branch, so use with care - here it is needed because git can't see that the rebased branch is essentially equivalent to the branch you started from).

PR #264, opened 2019-06-29
Fix memory leak in createfeature_vector method

VaibhavKansagara:
If there are no features then we still create a new Feature::Internal
but we never assign it anywhere so the reference counting doesn't get
a chance to delete it again.

### DIFF
--- a/xapian-letor/api/featurelist.cc
+++ b/xapian-letor/api/featurelist.cc
@@ -50,9 +50,12 @@ FeatureList::FeatureList() : internal(new FeatureList::Internal())
 }
 
 FeatureList::FeatureList(const std::vector<Feature*> & f)
-    : internal(new FeatureList::Internal())
 {
     LOGCALL_CTOR(API, "FeatureList", f);
+    if (f.empty()) {
+	throw InvalidArgumentError("FeatureList cannot be empty");
+    }
+    internal = new Internal();
     internal->feature = f;
     for (Feature* it : internal->feature) {
 	internal->stats_needed = Internal::stat_flags(internal->stats_needed |
@@ -108,11 +111,6 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
     LOGCALL(API, std::vector<FeatureVector>, "FeatureList::create_feature_vectors", mset | letor_query | letor_db);
     if (mset.empty())
 	return vector<FeatureVector>();
-
-    if (!internal->feature.empty()) {
-	throw InvalidArgumentError("FeatureList cannot be empty");
-    }
-
     std::vector<FeatureVector> fvec;
 
     for (Xapian::MSetIterator i = mset.begin(); i != mset.end(); ++i) {
--- a/xapian-letor/api/featurelist.cc
+++ b/xapian-letor/api/featurelist.cc
@@ -28,8 +28,8 @@
 #include "featurelist_internal.h"
 #include "feature_internal.h"
 
-#include "omassert.h"
 #include "debuglog.h"
+#include "omassert.h"
 
 using namespace std;
 
--- a/xapian-letor/api/featurelist.cc
+++ b/xapian-letor/api/featurelist.cc
@@ -108,23 +108,26 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
     LOGCALL(API, std::vector<FeatureVector>, "FeatureList::create_feature_vectors", mset | letor_query | letor_db);
     if (mset.empty())
 	return vector<FeatureVector>();
+
+    if (!internal->feature.empty()) {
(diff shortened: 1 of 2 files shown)


xapian-letor/api/featurelist.cc
```diff
@@ -114,16 +114,17 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
...
-	// Computes and populates the Feature::Internal with required stats.
-	internal->populate_feature_internal(internal_feature);
-	for (Feature* it : internal->feature) {
-	    it->internal = internal_feature;
-	    const vector<double>& values = it->get_values();
-	    // Append feature values
-	    fvals.insert(fvals.end(), values.begin(), values.end());
+	if (!internal->feature.empty()) {
+	    Feature::Internal* internal_feature = new Feature::Internal(
+						  letor_db, letor_query, doc);
```
ojwb: `auto internal_feature = ...` would make the formatting easier here.

ojwb:
Oh, I see what you mean (in a comment elsewhere) about the weight feature meaning we can't just shortcut the method completely.

It occurs to me that maybe this is actually a useless situation - it seems a machine learning reranking system working from one feature (weight) can only really end up returning results in weight order.  Yes there's an obvious thing to do in that situation, but it isn't a very useful thing to do.

So perhaps "no features in the feature list" is best regarded as a configuration error which should be flagged up, in which case throwing `Xapian::InvalidArgumentError` is more helpful.

Thoughts?

VaibhavKansagara:
> It occurs to me that maybe this is actually a useless situation - it seems a machine learning reranking system working from one feature (weight) can only really end up returning results in weight order. Yes there's an obvious thing to do in that situation, but it isn't a very useful thing to do.
> 
Ah I see that indeed appears to be the case.
> So perhaps "no features in the feature list" is best regarded as a configuration error which should be flagged up, in which case throwing Xapian::InvalidArgumentError is more helpful.
>
I think that is a good idea. I will change accordingly.

xapian-letor/api/featurelist.cc
```diff
@@ -108,23 +108,26 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
     LOGCALL(API, std::vector<FeatureVector>, "FeatureList::create_feature_vectors", mset | letor_query | letor_db);
     if (mset.empty())
 	return vector<FeatureVector>();
+
+    if (!internal->feature.empty()) {
```
jaylett: The sense of this conditional is wrong, surely? We want to throw the exception if `internal->feature` _is_ empty. It looks like the test below should catch this — are you running the tests locally before pushing?

@ojwb when we squash / merge this, we should probably rewrite the commit message to lead with "Raise exception on empty FeatureList", then explain it fixes memory leaks in the commit body.

However this makes me think: shouldn't we actually check this on the `FeatureList` constructor, and then turn this here into an assertion?

ojwb: > It looks like the test below should catch this

CI logs indeed show the new test (and a lot of existing tests) failing due to this.

Checking in the constructor probably makes more sense.  It needs a little care as the destructor won't be called if the constructor throws an exception - probably cleanest to delay the allocation of `internal` until after the check.

VaibhavKansagara: > It looks like the test below should catch this — are you running the tests locally before pushing?
>
Ah I was so sure that I didn't check it. Fixed in a fixup commit.

xapian-letor/tests/api_letor.cc
```diff
@@ -225,21 +225,7 @@ DEFINE_TESTCASE(createfeaturevectorthree, generated)
...
-    enquire.set_query(Xapian::Query("tigers"));
-    Xapian::MSet mset;
-    mset = enquire.get_mset(0, 10);
-
-    TEST(!mset.empty());
-
-    TEST_EXCEPTION(Xapian::InvalidArgumentError,
-		   fl.create_feature_vectors(mset, Xapian::Query("tigers"),
-					     db));
+    TEST_EXCEPTION(Xapian::InvalidArgumentError, Xapian::FeatureList fl(f));
```
ojwb: This no longer uses any database, let alone a `generated` one, so the testcase condition is now wrong - we use `!backend` for such tests (so they're run once in their own pass, rather than pointlessly run once per backend the test harness is testing over).

ojwb: I'd also suggest not including `check` (or `test`, etc) in the testcase name - it doesn't really add useful information and tends to result in rather long test case names.

ojwb:
Thanks for the updates.

Autosquashed, rebased and merged to master with the adjusted commit message James suggested as b336f99539222eb922c8ae6051889b069d863060.

PR #267, opened 2019-07-03
Add e-book support to Omega

barufa:
At the moment, Omega cannot index e-book formats. We can use libe-book
to add support for formats fb2, fb2.zip, lrf, tcr, pdb and jar.
Also, I modify assistant to try to keep it alive in case of the extraction function return false.

### DIFF
--- a/xapian-applications/omega/Makefile.am
+++ b/xapian-applications/omega/Makefile.am
@@ -44,7 +44,7 @@ BUILT_SOURCES = extra/omegascript.vim my-html-tok.h mimemap.h namedents.h
 
 MAINTAINERCLEANFILES = my-html-tok.h mimemap.h mimemap.stamp namedents.h
 
-EXTRA_PROGRAMS = omindex_pdf
+EXTRA_PROGRAMS = omindex_pdf omindex_ebook
 
 EXTRA_DIST =\
 	clickmodel/testdata/test1.log\
@@ -177,6 +177,10 @@ omindex_LDADD = $(MAGIC_LIBS) $(XAPIAN_LIBS) $(ZLIB_LIBS) libutf8convert.la
 
 omindex_pdf_SOURCES = assistant.cc worker_comms.cc common/str.cc handler_poppler.cc
 omindex_pdf_LDADD = $(POPPLER_LIBS) $(XAPIAN_LIBS)
+
+omindex_ebook_SOURCES = assistant.cc worker_comms.cc handler_libebook.cc
+omindex_ebook_LDADD = $(LIBEBOOK_LIBS) $(XAPIAN_LIBS)
+omindex_ebook_CPPFLAGS = $(AM_CPPFLAGS) $(LIBEBOOK_CFLAGS)
 
 scriptindex_SOURCES = scriptindex.cc myhtmlparse.cc htmlparse.cc\
  common/getopt.cc common/str.cc commonhelp.cc utils.cc hashterm.cc loadfile.cc\
--- a/xapian-applications/omega/Makefile.am
+++ b/xapian-applications/omega/Makefile.am
@@ -179,7 +179,7 @@ omindex_pdf_SOURCES = assistant.cc worker_comms.cc common/str.cc handler_poppler
 omindex_pdf_LDADD = $(POPPLER_LIBS) $(XAPIAN_LIBS)
 
 omindex_ebook_SOURCES = assistant.cc worker_comms.cc handler_libebook.cc
-omindex_ebook_LDADD = $(LIBEBOOK_LIBS) $(XAPIAN_LIBS)
+omindex_ebook_LDADD = $(LIBEBOOK_LIBS)
 omindex_ebook_CXXFLAGS = $(LIBEBOOK_CFLAGS) $(XAPIAN_CXXFLAGS)
 
 scriptindex_SOURCES = scriptindex.cc myhtmlparse.cc htmlparse.cc\
--- a/xapian-applications/omega/Makefile.am
+++ b/xapian-applications/omega/Makefile.am
@@ -180,7 +180,7 @@ omindex_pdf_LDADD = $(POPPLER_LIBS) $(XAPIAN_LIBS)
 
 omindex_ebook_SOURCES = assistant.cc worker_comms.cc handler_libebook.cc
 omindex_ebook_LDADD = $(LIBEBOOK_LIBS) $(XAPIAN_LIBS)
-omindex_ebook_CPPFLAGS = $(AM_CPPFLAGS) $(LIBEBOOK_CFLAGS)
+omindex_ebook_CXXFLAGS = $(LIBEBOOK_CFLAGS) $(XAPIAN_CXXFLAGS)
 
 scriptindex_SOURCES = scriptindex.cc myhtmlparse.cc htmlparse.cc\
  common/getopt.cc common/str.cc commonhelp.cc utils.cc hashterm.cc loadfile.cc\
--- a/xapian-applications/omega/configure.ac
+++ b/xapian-applications/omega/configure.ac
@@ -261,6 +261,17 @@ PKG_CHECK_MODULES([POPPLER], [poppler-cpp], [
                      OMINDEX_MODULES="$OMINDEX_MODULES omindex_pdf"])
      LIBS=$save_LIBS],[ ])
 
+dnl check if libe-book is available.
+PKG_CHECK_MODULES([LIBEBOOK], [libe-book-0.1,librevenge-0.0,librevenge-generators-0.0,librevenge-stream-0.0], [
+     save_LIBS=$LIBS
+     LIBS="$LIBS $LIBEBOOK_CFLAGS $LIBEBOOK_LIBS"
+     AC_CHECK_HEADERS([librevenge-0.0/librevenge-generators/librevenge-generators.h \
+                       librevenge-0.0/librevenge-stream/librevenge-stream.h \
+                       libe-book/libe-book.h],
+                       [AC_DEFINE(HAVE_LIBEBOOK, 1, [Define HAVE_LIBEBOOK if the library is available])
+                        OMINDEX_MODULES="$OMINDEX_MODULES omindex_ebook"])
+     LIBS=$save_LIBS],[ ])
+
 AC_SUBST([OMINDEX_MODULES])
 
 dnl Check that snprintf actually works as it's meant to.
--- a/xapian-applications/omega/configure.ac
+++ b/xapian-applications/omega/configure.ac
@@ -263,14 +263,8 @@ PKG_CHECK_MODULES([POPPLER], [poppler-cpp], [
 
 dnl check if libe-book is available.
 PKG_CHECK_MODULES([LIBEBOOK], [libe-book-0.1,librevenge-0.0,librevenge-generators-0.0,librevenge-stream-0.0], [
-     save_LIBS=$LIBS
-     LIBS="$LIBS $LIBEBOOK_CFLAGS $LIBEBOOK_LIBS"
-     AC_CHECK_HEADERS([librevenge-0.0/librevenge-generators/librevenge-generators.h \
-                       librevenge-0.0/librevenge-stream/librevenge-stream.h \
-                       libe-book/libe-book.h],
-                       [AC_DEFINE(HAVE_LIBEBOOK, 1, [Define HAVE_LIBEBOOK if the library is available])
-                        OMINDEX_MODULES="$OMINDEX_MODULES omindex_ebook"])
-     LIBS=$save_LIBS],[ ])
+     AC_DEFINE(HAVE_LIBEBOOK, 1, [Define HAVE_LIBEBOOK if the library is available])
+     OMINDEX_MODULES="$OMINDEX_MODULES omindex_ebook"],[ ])
 
 AC_SUBST([OMINDEX_MODULES])
 
--- a/xapian-applications/omega/docs/overview.rst
+++ b/xapian-applications/omega/docs/overview.rst
@@ -310,7 +310,7 @@ other filters too - see below):
 * MIME email messages (.eml) and USENET articles if perl with MIME::Tools and
   HTML::Parser is available
 * vCard files (.vcf, .vcard) if perl with Text::vCard is available
-* FictionBook v.2 files (.fb2, .fb2.zip) if libe-book is available
+* FictionBook v.2 files (.fb2) if libe-book is available
 * QiOO (mobile format, for java-enabled cellphones) files (.jar) if libe-book is available
 * TCR (simple compressed text format) files (.tcr) if libe-book is available
 * eReader files (.pdb) if libe-book is available
--- a/xapian-applications/omega/docs/overview.rst
+++ b/xapian-applications/omega/docs/overview.rst
@@ -310,6 +310,11 @@ other filters too - see below):
 * MIME email messages (.eml) and USENET articles if perl with MIME::Tools and
   HTML::Parser is available
 * vCard files (.vcf, .vcard) if perl with Text::vCard is available
+* FictionBook v.2 files (.fb2, .fb2.zip) if libe-book is available
+* QiOO (mobile format, for java-enabled cellphones) files (.jar) if libe-book is available
+* TCR (simple compressed text format) files (.tcr) if libe-book is available
+* eReader files (.pdb) if libe-book is available
+* Sony eBook files (.lrf) if libe-book is available
 
 If you have additional extensions that represent one of these types, you can
 add an additional MIME mapping using the ``--mime-type`` option.  For
--- a/xapian-applications/omega/handler_libebook.cc
+++ b/xapian-applications/omega/handler_libebook.cc
@@ -18,6 +18,7 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
  * USA
  */
+#include <config.h>
 #include "handler.h"
 #include "stringutils.h"
 
@@ -42,7 +43,7 @@ parse_metadata_field(const char* start,
 		     size_t len,
 		     string& out)
 {
-    if (size_t(end - start) > len && memcmp(start,field,len) == 0) {
+    if (size_t(end - start) > len && memcmp(start, field, len) == 0) {
 	start += len;
 	while (start != end && isspace(*start)) start++;
 	if (start != end && (end[-1] != '\r' || --end != start)) {
@@ -55,12 +56,13 @@ parse_metadata_field(const char* start,
 
 static void
 parse_metadata(const char* data,
+	       size_t len,
 	       string& author,
 	       string& title,
 	       string& keywords)
 {
     const char* p = data;
-    const char* end = p + strlen(data);
+    const char* end = p + len;
 
     while (p != end) {
 	const char* start = p;
@@ -70,7 +72,7 @@ parse_metadata(const char* data,
 	    eol = p++;
 	else
 	    p = eol = end;
-	if ((end - start) > 5 && memcmp(start,"meta:",5) == 0) {
+	if ((end - start) > 5 && memcmp(start, "meta:", 5) == 0) {
 	    start += 5;
 	    switch (*start) {
 		case 'i': {
@@ -83,7 +85,7 @@ parse_metadata(const char* data,
 		    break;
 		}
 	    }
-	} else if ((end - start) > 3 && memcmp(start,"dc:",3) == 0) {
+	} else if ((end - start) > 3 && memcmp(start, "dc:", 3) == 0) {
 	    start += 3;
 	    switch (*start) {
 		case 'c': {
@@ -111,7 +113,7 @@ parse_metadata(const char* data,
 static void
 clear_text(string& str, const char* text)
 {
-    for (int i = 0; text[i]!='\0'; ++i)
+    for (int i = 0; text[i] != '\0'; ++i)
 	if (!isspace(text[i]) || (i && !isspace(text[i - 1])))
 	    str.push_back(text[i]);
 }
@@ -154,7 +156,9 @@ extract(const string& filename,
 	RVNGTextTextGenerator metadata(metadata_dump, true);
 	if (EBOOKDocument::RESULT_OK ==
 	    EBOOKDocument::parse(input.get(), &metadata, type)) {
-	    parse_metadata(metadata_dump.cstr(), author, title, keywords);
+	    const char* metadata = metadata_dump.cstr();
+	    size_t len = metadata_dump.size();
+	    parse_metadata(metadata, len, author, title, keywords);
 	    f_meta = true;
 	} else {
 	    cerr << "Libe-book Error: Fail to extract metadata" << endl;
--- a/xapian-applications/omega/handler_libebook.cc
+++ b/xapian-applications/omega/handler_libebook.cc
@@ -18,35 +18,100 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
  * USA
  */
-#include <config.h>
 #include "handler.h"
+#include "stringutils.h"
 
-#include <xapian.h>
-#include <string.h>
-#include <unordered_map>
 #include <memory>
 #include <iostream>
 
-#include <librevenge-0.0/librevenge-generators/librevenge-generators.h>
-#include <librevenge-0.0/librevenge-stream/librevenge-stream.h>
+#include <librevenge-generators/librevenge-generators.h>
+#include <librevenge-stream/librevenge-stream.h>
 #include <libe-book/libe-book.h>
+
+#define PARSE_FIELD(START, END, FIELD, OUT) \
+   parse_metadata_field((START), (END), (FIELD), (CONST_STRLEN(FIELD)), (OUT))
 
 using libebook::EBOOKDocument;
 using namespace std;
 using namespace librevenge;
 
-static
-char* str_copy(const char* text) {
-    int len = strlen(text);
-    char* line = new char[len];
-    strcpy(line, text);
-    return line;
+static void
+parse_metadata_field(const char* start,
+		     const char* end,
+		     const char* field,
+		     size_t len,
+		     string& out)
+{
+    if (size_t(end - start) > len && memcmp(start,field,len) == 0) {
+	start += len;
+	while (start != end && isspace(*start)) start++;
+	if (start != end && (end[-1] != '\r' || --end != start)) {
+	    if (!out.empty())
+		out.push_back(' ');
+		out.append(start, end - start);
+	}
+    }
 }
 
-static
-void clear_text(string& str,const char* text) {
-    int len = strlen(text);
-    for (int i = 0; i < len; ++i)
+static void
+parse_metadata(const char* data,
+	       string& author,
+	       string& title,
+	       string& keywords)
+{
+    const char* p = data;
+    const char* end = p + strlen(data);
+
+    while (p != end) {
+	const char* start = p;
+	p = static_cast<const char*>(memchr(p, '\n', end - start));
+	const char* eol;
+	if (p)
+	    eol = p++;
+	else
+	    p = eol = end;
+	if ((end - start) > 5 && memcmp(start,"meta:",5) == 0) {
+	    start += 5;
+	    switch (*start) {
+		case 'i': {
+		    if (author.empty())
+			PARSE_FIELD(start, eol, "initial-creator", author);
+		    break;
+		}
+		case 'k': {
+		    PARSE_FIELD(start, eol, "keyword", keywords);
+		    break;
+		}
+	    }
+	} else if ((end - start) > 3 && memcmp(start,"dc:",3) == 0) {
+	    start += 3;
+	    switch (*start) {
+		case 'c': {
+		    if (!author.empty())
+			author.clear();
+		    PARSE_FIELD(start, eol, "creator", author);
+		    break;
+		}
+		case 's': {
+		    PARSE_FIELD(start, eol, "subject", keywords);
+		    break;
+		}
+		case 't': {
+		    PARSE_FIELD(start, eol, "title", title);
+		    break;
+		}
+	    }
+	} else if ((end - start) > 8 && memcmp(start, "dcterms:", 8) == 0) {
+	    start += 8;
+	    PARSE_FIELD(start, eol, "available", keywords);
+	}
+    }
+}
+
+static void
+clear_text(string& str, const char* text)
+{
+    for (int i = 0; text[i]!='\0'; ++i)
 	if (!isspace(text[i]) || (i && !isspace(text[i - 1])))
 	    str.push_back(text[i]);
 }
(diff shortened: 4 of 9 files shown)


xapian-applications/omega/mimemap.tokens
```diff
@@ -193,6 +193,13 @@ rpm		application/x-redhat-package-manager
 # Atom feeds:
 atom		application/atom+xml
 
+# Ebooks:
+fb2		application/x-fictionbook+xml
+fb2.zip		application/x-zip-compressed-fb2
+lrf		application/x-sony-bbeb
+tcr		application/x-tcr-ebook
+pdb		application/vnd.palm
+
```
barufa: Should I associate `application/java-archive` to `jar`?

ojwb: Is `.jar` really an e-book format?  It's mainly a way to package up Java `.class` files.

`.jar` files are actually ZIP archives with a different extension and a particular layout inside, and ZIP containers are also quite popular for modern file formats, so I'm wondering if libmagic is getting confused and reporting `application/java-archive` for some e-book format.

xapian-applications/omega/docs/overview.rst
```diff
@@ -310,6 +310,11 @@ other filters too - see below):
 * MIME email messages (.eml) and USENET articles if perl with MIME::Tools and
   HTML::Parser is available
 * vCard files (.vcf, .vcard) if perl with Text::vCard is available
+* FictionBook v.2 files (.fb2, .fb2.zip) if libe-book is available
+* QiOO (mobile format, for java-enabled cellphones) files (.jar) if libe-book is available
+* TCR (simple compressed text format) files (.tcr) if libe-book is available
+* eReader files (.pdb) if libe-book is available
+* Sony eBook files (.lrf) if libe-book is available
```
ojwb: I'd expect EPUB files to be handled (https://en.wikipedia.org/wiki/EPUB) - they're one of the more popular ebook formats, and there appears to be a parser for them in the library: https://sources.debian.org/src/libe-book/0.1.3-1/src/lib/EPubParser.h/

xapian-applications/omega/mimemap.tokens
ojwb: Oh I see, it's used by "QiOO".  It looks like that's an old format which never really caught on which actually shipped the Java software to read the book along with the book (I think I can see why that never caught on).

So probably better to leave it off the default list.  If someone has a collection of these they want to index, they can always run `omindex` with `-Mjar:application/java-archive` to override the default of ignoring `.jar` files.

xapian-applications/omega/mimemap.tokens
```diff
@@ -193,6 +193,13 @@ rpm		application/x-redhat-package-manager
 # Atom feeds:
 atom		application/atom+xml
 
+# Ebooks:
+fb2		application/x-fictionbook+xml
+fb2.zip		application/x-zip-compressed-fb2
```
ojwb: I don't think that will work - we take everything after the final `.` as the extension to look up, so `foo.fb2.zip` would be looked up as extension `zip`.

It's probably worth addressing that, but I'd probably make a note and come back to it once the basic e-book support is done.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
+/** @file handler_libebook.cc
+ * @brief @brief Extract text and metadata using libe-book.
```
ojwb: Double `@brief`.

I see this is was also present in the merge changes for poppler, so I pushed a fix for those: a600cbb9db8424e29d5db2b91a1e5885329127ef

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+#include "handler.h"
+
+#include <xapian.h>
+#include <string.h>
+#include <unordered_map>
+#include <memory>
+
+#include <librevenge-0.0/librevenge-generators/librevenge-generators.h>
+#include <librevenge-0.0/librevenge-stream/librevenge-stream.h>
+#include <iostream>
```
ojwb: `<iostream>` belongs with the group of C++ standard library headers above.

xapian-applications/omega/docs/overview.rst
ojwb: The list of types defined in the header is quite a bit longer than that in the README:

https://sources.debian.org/src/libe-book/0.1.3-1/inc/libe-book/EBOOKDocument.h/#L67

Some don't look useful (e.g. we handle HTML and XHTML already) but others do, assuming they are actually supported.

barufa: Actually, a lot of that formats are not supported. Others (TYPE_EPUB, TYPE_HTML, TYPE_HTMLHELP, TYPE_OPENEBOOK, TYPE_ROCKETEBOOK, TYPE_XHTML) are only supported in an experimental way and it is probably that they appear in future versions.
[Here](https://sourceforge.net/projects/libebook/files/) is a list of formats.

Maybe I can add epub with libepub, but I am not sure if this is the best option.

xapian-applications/omega/mimemap.tokens
barufa: Okay, I am going to remove it from the list and add a note [here](https://trac.xapian.org/wiki/GSoC2019/TextExtraction/Notes)

xapian-applications/omega/handler_libebook.cc
barufa: Sorry, I didn't see that.

xapian-applications/omega/handler_libebook.cc
barufa: Sorry, I will move it.

xapian-applications/omega/docs/overview.rst
ojwb: > Here is a list of formats.

That says "Source: README, updated 2013-11-16" which made me wonder if it was actually up to date, but looking at the git commits since then I don't see anything obviously either adding or firming up support for more formats.

The command-line tools from the library aren't packaged for Debian, but I tried loading several EPUB files into Libreoffice and it didn't handle them, so it does indeed seem EPUB isn't usefully supported.  That's rather a shame, since it seems to be the dominant format at this point.

> Maybe I can add epub with libepub, but I am not sure if this is the best option.

Another potential option is to make the EPUB support in libe-book work (at least well enough to get the text out without formatting, since that's what we care about here).  There is a parser class for it, so presumably the framework exists and it's a matter of implementing some missing methods within that:

https://sources.debian.org/src/libe-book/0.1.3-1/src/lib/EPubParser.cpp/

But let's get this merged before worrying about EPUB.

xapian-applications/omega/index_file.cc
```diff
@@ -153,8 +153,19 @@ void
...
+    index_library("application/pdf", omindex_pdf);
 #endif
+#if defined HAVE_LIBEBOOK
+    Worker* omindex_ebook = new Worker("omindex_ebook");
+    index_library("application/vnd.palm", omindex_ebook);
+    index_library("application/x-fictionbook+xml", omindex_ebook);
+    index_library("application/x-zip-compressed-fb2", omindex_ebook);
+    index_library("application/x-sony-bbeb", omindex_ebook);
+    index_library("application/x-tcr-ebook", omindex_ebook);
+    index_library("application/java-archive", omindex_ebook);
```
barufa: Should I remove `application/java-archive` or `application/x-zip-compressed-fb2` from here?

xapian-applications/omega/configure.ac
```diff
@@ -261,6 +261,17 @@ PKG_CHECK_MODULES([POPPLER], [poppler-cpp], [
                      OMINDEX_MODULES="$OMINDEX_MODULES omindex_pdf"])
      LIBS=$save_LIBS],[ ])
 
+dnl check if libe-book is available.
+PKG_CHECK_MODULES([LIBEBOOK], [libe-book-0.1,librevenge-0.0,librevenge-generators-0.0,librevenge-stream-0.0], [
+     save_LIBS=$LIBS
+     LIBS="$LIBS $LIBEBOOK_CFLAGS $LIBEBOOK_LIBS"
+     AC_CHECK_HEADERS([librevenge-0.0/librevenge-generators/librevenge-generators.h \
```
barufa: Is it okay to add LIBEBOOK_CFLAGS to LIBS?

xapian-applications/omega/index_file.cc
ojwb: Listing `application/x-zip-compressed-fb2` here seems OK - that will mean we index such files if libmagic identifies that type from the contents or the user passes `--mime-type-match='*.fb2.zip':application/x-zip-compressed-fb2` (we could support via that mechanism internally, but I think it makes more sense to add support for multi-level extensions).

Having a rule for `application/java-archive` seems more problematic since libmagic will likely report that for many files which aren't QiOO format (because it's the type for Java JAR files).  Probably the best way to handle this is to make up a mime type for this format in the `x-` space which is reserved for private use (e.g. `application/x-qioo`) and use that here.  Then a user with such files can use `--mime-type=jar:application/x-qioo` to enable indexing of them.

xapian-applications/omega/configure.ac
ojwb: No, `CFLAGS` are used for compiling, `LIBS` for linking.

xapian-applications/omega/Makefile.am
```diff
@@ -178,6 +178,10 @@ omindex_LDADD = $(MAGIC_LIBS) $(XAPIAN_LIBS) $(ZLIB_LIBS) libutf8convert.la
 omindex_pdf_SOURCES = assistant.cc worker_comms.cc common/str.cc handler_poppler.cc
 omindex_pdf_LDADD = $(POPPLER_LIBS) $(XAPIAN_LIBS)
 
+omindex_ebook_SOURCES = assistant.cc worker_comms.cc handler_libebook.cc
+omindex_ebook_LDADD = $(LIBEBOOK_LIBS) $(XAPIAN_LIBS)
+omindex_ebook_CPPFLAGS = $(AM_CPPFLAGS) $(LIBEBOOK_CFLAGS)
```
ojwb: This is wrong - `CPPFLAGS` are for the preprocessor, `CFLAGS` for the compiler.

This should be:

```
omindex_ebook_CXXFLAGS = $(LIBEBOOK_CFLAGS)
```

(We need to put them in CXXFLAGS so they're used for C++ compilation - this does kind of assume the C and C++ compilers are from the same toolchain, or at least compatible in the flags they take, but that's hard to avoid with `pkg-config`'s model of the world).

It looks like you're also missing the equivalent above:

```
omindex_pdf_CXXFLAGS = $(POPPLER_CFLAGS)
```

xapian-applications/omega/configure.ac
```diff
@@ -261,6 +261,17 @@ PKG_CHECK_MODULES([POPPLER], [poppler-cpp], [
...
+                       librevenge-0.0/librevenge-stream/librevenge-stream.h \
+                       libe-book/libe-book.h],
+                       [AC_DEFINE(HAVE_LIBEBOOK, 1, [Define HAVE_LIBEBOOK if the library is available])
+                        OMINDEX_MODULES="$OMINDEX_MODULES omindex_ebook"])
```
ojwb: I don't think we need this `AC_CHECK_HEADERS` check if we're using pkg-config to report the flags to use.

For poppler we needed an additional test to catch the libstdc++ vs libc++ compatibility issue, because poppler provides a C++ API.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+#include "handler.h"
+
+#include <xapian.h>
+#include <string.h>
+#include <unordered_map>
+#include <memory>
+#include <iostream>
+
+#include <librevenge-0.0/librevenge-generators/librevenge-generators.h>
+#include <librevenge-0.0/librevenge-stream/librevenge-stream.h>
```
ojwb: I don't think you should have `librevenge-0.0/` explicitly here.

```
$ pkg-config libe-book-0.1 --cflags
-I/usr/include/libe-book-0.1 -I/usr/include/libxml2 -I/usr/include/librevenge-0.0
```

So if you've hooked up those flags correctly in `Makefile.am` then the compiler should look in `/usr/include/librevenge-0.0`.  By hard coding this here, you'll stop things working if that changes to `librevenge-0.1` for example.

xapian-applications/omega/index_file.cc
barufa: Okay, I am going to change it.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+#include <libe-book/libe-book.h>
+
+using libebook::EBOOKDocument;
+using namespace std;
+using namespace librevenge;
+
+static
+char* str_copy(const char* text) {
```
ojwb: Xapian coding conventions would format this as:

```
static char*
str_copy(const char* text)
{
```

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+    int len = strlen(text);
+    char* line = new char[len];
+    strcpy(line, text);
+    return line;
+}
+
+static
+void clear_text(string& str,const char* text) {
```
ojwb: And similarly here, plus there's a missing space after the `,` (which `xapian-check-patch` isn't spotting for some reason).

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+    int len = strlen(text);
+    for (int i = 0; i < len; ++i)
+	if (!isspace(text[i]) || (i && !isspace(text[i - 1])))
+	    str.push_back(text[i]);
+}
```
ojwb: Is it actually necessary to run every string through this function?  Unless the output we're getting has a lot of sequences of multiple spaces, any saving we get from stripping them out here is going to be cancelled out by the cost of scanning and copying these strings like this.

If we do need to do this, we should at least avoid the `strlen(text)` which unnecessary scans the string yet another time just to find its length - we can make the `for` loop's end condition check for the zero byte instead.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+	    pch = strtok(NULL, "\n");
+	}
+	delete[] str;
+	unordered_map<string,char*>::const_iterator it;
+
+	// Extract Author if it possible
+	if ((it = hash.find("dc:creator")) != hash.end())
+	    clear_text(author, it->second);
+	else if ((it = hash.find("meta:initial-creator")) != hash.end())
+	    clear_text(author, it->second);
```
ojwb: This first copies `metadata_dump` to get a modifiable copy to be able to use `strtok()`, then copies everything again into an `unordered_map`, then looks up stuff in that map.  That's a lot of extra copying that could be avoided.

Take a look at `parse_pdf_metainfo()` in `index_file.cc` which does a very similar job to this code here, but performs a single scan over `const` data without needing to create a temporary data structure.  The format being parsed isn't exactly the same, but the basic approach should work.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+	    clear_text(keywords, it->second);
+	pages = "";
+	hash.clear();
+
+	// Extract Dump if it possible
+	RVNGTextTextGenerator content(content_dump, false);
+
+	if (EBOOKDocument::RESULT_OK !=
+	    EBOOKDocument::parse(input.get(), &content, type))
+	    return false;
```
ojwb: If you need to wrap an `if` expression, it's probably clearer to make the action a block even if there's only one statement.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+	// Extract Title if it possible
+	if ((it = hash.find("dc:title")) != hash.end())
+	    clear_text(title, it->second);
+
+	// Extract Keywords if it possible
+	if ((it = hash.find("dc:subject")) != hash.end())
+	    clear_text(keywords, it->second);
+	if ((it = hash.find("dcterms:available")) != hash.end())
+	    clear_text(keywords, it->second);
+	pages = "";
```
ojwb: I think we should just rely on the caller arranging for these variables to be empty before calling us (as the current implementation does).  That way we avoid having to have code in each module to clear any fields which aren't handled - instead that's all dealt with once in one place in the calling code in `assistant.cc`.

xapian-applications/omega/handler_libebook.cc
barufa: Okay, I am going to try to optimize it.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,133 @@
...
+	while (pch != NULL) {
+	    char* sep = strpbrk(pch, " ");
+	    *sep = '\0';
+	    hash[pch] = sep + 1;
+	    pch = strtok(NULL, "\n");
+	}
+	delete[] str;
+	unordered_map<string,char*>::const_iterator it;
+
+	// Extract Author if it possible
```
ojwb: `... if possible` would be better English (also below)

xapian-applications/omega/handler_libebook.cc
barufa: Actually, if I don't use an argument I get a compiler error. That is why I assign an empty string to pages.

xapian-applications/omega/handler_libebook.cc
barufa: I will change it.

xapian-applications/omega/handler_libebook.cc
barufa: Okay, I will remove strlen.
I think I can use the function clean_text only in dump. Searching time increases a lot if I didn't use this function (at least with poppler). I will try to find some ebooks to test it (with more than 500 pages).

ojwb: PDF is a very different format to these ebooks though.

Also I'm surprised to hear there's any search-time impact - the whitespace should be skipped when splitting the text into terms, and the stored sample gets whitespace cleaned by `omindex`.  I'd expect any saving you get from cleaning whitespace here to be down to the reduced traffic across the pipe between omindex and the worker process.  The one exception is `author` which is currently just stored as-is (that's really a bug, but I suspect in practice that author metadata doesn't have a lot of extra whitespace).

xapian-applications/omega/handler_libebook.cc
ojwb: What was the compiler error?

If it's "unused parameter" the answer is to tell the compiler that's OK.  There are various ways to do this:

* remove the parameter name from the method definition (so it says `..., string& author, string&)`) - this is portable, but only works for C++ (C doesn't allow omitting parameter names in function definitions).  This has the downside that it's not clear what the parameter is without looking at the header, though you can comment it out instead of removing it like so: `string& /*pages*/`

* add a dummy use via `(void)pages;` (conventionally at the start of the function) which is a way to tell the compiler that it's OK that the parameter is otherwise unused.  This is portable and most compilers understand it as meaning "pretend `pages` is used".  This is the probably the best option in portable C code.

* C++17 added a standard attribute (https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) but currently Xapian only requires a C++11 compiler.

* There are also compiler-specific ways to specify this (e.g. `__attribute__((unused))` for GCC and clang), but those are inherently less portable.

So here I'd go for the `(void)` cast or commenting out the parameter name.

xapian-applications/omega/configure.ac
barufa: I didn't realize that, sorry. I am going to fix it.

xapian-applications/omega/handler_libebook.cc
barufa: Okay, i didn't notice that I was hard coding that. I will remove it.

xapian-applications/omega/handler_libebook.cc
barufa: I will fix handler_poppler after get this merged.

xapian-applications/omega/handler_libebook.cc
barufa: Sorry, I didn't know that. I am going to fix it.

xapian-applications/omega/handler_libebook.cc
barufa: Okay, I am going to change it.

xapian-applications/omega/configure.ac
ojwb: > because poppler provides a C++ API

Actually, both libe-book and librevenge are C++ APIs too.  However, there's a probably important difference - they don't use anything from the C++ standard library from their public API headers.  That means they may well avoid the problems we had with poppler.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,172 @@
...
+
+static void
+parse_metadata(const char* data,
+	       string& author,
+	       string& title,
+	       string& keywords)
+{
+    const char* p = data;
+    const char* end = p + strlen(data);
+
```
barufa: I am not sure if it is possible to get rid of this call to strlen.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,177 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+#include "handler.h"
```
ojwb: This is a `.cc` file, so `#include <config.h>` should be included before anything else.

xapian-applications/omega/handler_libebook.cc
ojwb: It is - `metadata_dump.cstr()` is passed for `data`, so the corresponding length is available in the caller as `metadata_dump.size()` (see `RVNGString` in `/usr/include/librevenge-0.0/librevenge/RVNGString.h`).

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,177 @@
...
+
+    while (p != end) {
+	const char* start = p;
+	p = static_cast<const char*>(memchr(p, '\n', end - start));
+	const char* eol;
+	if (p)
+	    eol = p++;
+	else
+	    p = eol = end;
+	if ((end - start) > 5 && memcmp(start,"meta:",5) == 0) {
```
ojwb: Missing space after `,`.  Repeated below.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,177 @@
...
+	    start += 8;
+	    PARSE_FIELD(start, eol, "available", keywords);
+	}
+    }
+}
+
+static void
+clear_text(string& str, const char* text)
+{
+    for (int i = 0; text[i]!='\0'; ++i)
```
ojwb: Missing whitespace around `!=`.

barufa: I am going to fix it

xapian-applications/omega/Makefile.am
```diff
@@ -178,6 +178,10 @@ omindex_LDADD = $(MAGIC_LIBS) $(XAPIAN_LIBS) $(ZLIB_LIBS) libutf8convert.la
 omindex_pdf_SOURCES = assistant.cc worker_comms.cc common/str.cc handler_poppler.cc
 omindex_pdf_LDADD = $(POPPLER_LIBS) $(XAPIAN_LIBS)
 
+omindex_ebook_SOURCES = assistant.cc worker_comms.cc handler_libebook.cc
+omindex_ebook_LDADD = $(LIBEBOOK_LIBS) $(XAPIAN_LIBS)
+omindex_ebook_CXXFLAGS = $(LIBEBOOK_CFLAGS) $(XAPIAN_CXXFLAGS)
+
```
barufa: As I am using stringutils.h, I add XAPIAN_CXXFLAGS. Is it okay?

ojwb: You shouldn't need it (not `XAPIAN_LIBS`).

Most of `stringutils.h` is macros and inline functions, which don't require anything to be linked.

If you were using `C_tolower()`, `C_isdigit()` (which are like the standard library versions except they always work in the "C" locale rather than returning results which depend on the current locale), etc then you would need to link the library, as they use a data table that is provided by it.  But you aren't using any of these that I can see.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,181 @@
...
+	// Extract Dump if possible
+	RVNGTextTextGenerator content(content_dump, false);
+	if (EBOOKDocument::RESULT_OK ==
+	    EBOOKDocument::parse(input.get(), &content, type)) {
+	    clear_text(dump, content_dump.cstr());
+	    f_txt = true;
+	} else {
+	    cerr << "Libe-book Error: Fail to extract text" << endl;
+	}
+	return f_txt || f_meta;
```
ojwb: Is this actually the check you intended?  This will return "ok" if there's at most one error.

Looking at `enum Result` in `EBOOKDocument.h`, the results other than `RESULT_OK` look to be problems with accessing the file or with its contents (or encryption/password issues that we ought not hit because the confidence check above should have already caught those cases).  Those seem like problems that are better handled by giving up.

But if that is actually the check you wanted, why use two flag variables?  You'd only need a single `bool extracted_something = false;` and set that to `true` if anything works.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,181 @@
...
+	bool f_txt = false, f_meta = false;
+
+	if (RVNGDirectoryStream::isDirectory(file))
+	    input.reset(new RVNGDirectoryStream(file));
+	else
+	    input.reset(new RVNGFileStream(file));
+
+	EBOOKDocument::Type type = EBOOKDocument::TYPE_UNKNOWN;
+	EBOOKDocument::Confidence confidence =
+				EBOOKDocument::isSupported(input.get(), &type);
```
ojwb: This wrapped line seems rather over-indented - the Xapian convention for this case would be to just indent by 4 columns more.

But you could just use `auto` to avoid the long type name and the need to wrap the line at all:

```
        auto confidence = EBOOKDocument::isSupported(input.get(), &type);
```

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,181 @@
...
+
+	if (EBOOKDocument::CONFIDENCE_SUPPORTED_PART == confidence) {
+	    input.reset(RVNGDirectoryStream::createForParent(file));
+	    confidence = EBOOKDocument::isSupported(input.get(), &type);
+	}
```
ojwb: Are you sure we want this?

If I follow what this does, it means that if we try to index a file `/home/olly/Documents/foo.fb2` libe-book could tell us `CONFIDENCE_SUPPORTED_PART` and then we'd ask libe-book to try to extract the directory `/home/olly/Documents` instead.  Given how omindex works by recursing a directory structure, this doesn't seem like it's ever going to be a good idea.

xapian-applications/omega/handler_libebook.cc
barufa: I am going to change that.

xapian-applications/omega/handler_libebook.cc
barufa: I did not think of ​using 'auto'. I'm going to change it.

xapian-applications/omega/handler_libebook.cc
barufa: Sorry, misunderstand the 'createForParent' function. I am going to remove it.

ojwb: I think this mechanism is probably there for the case of things like Apple keynote where the document can be a directory with a `.key` extension.  If you ask libetonyek to open a file within that such as `/path/to/foo.key/preview.jpg` it might return `CONFIDENCE_SUPPORTED_PART` and `createForParent` would then open `/path/to/foo.key`.

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,177 @@
...
+bool
+extract(const string& filename,
+	string& dump,
+	string& title,
+	string& keywords,
+	string& author,
+	string& pages)
+{
+    try {
+	shared_ptr<RVNGInputStream> input;
```
ojwb: We don't actually share `input` (we always call `input.get()` and pass the raw pointer) so `unique_ptr` would be a better choice here (it expresses the situation more clearly and has a simpler implementation).

xapian-applications/omega/handler_libebook.cc
```diff
@@ -0,0 +1,177 @@
...
+	if (RVNGDirectoryStream::isDirectory(file))
+	    input.reset(new RVNGDirectoryStream(file));
+	else
+	    input.reset(new RVNGFileStream(file));
+
+	EBOOKDocument::Type type = EBOOKDocument::TYPE_UNKNOWN;
+	auto confidence = EBOOKDocument::isSupported(input.get(), &type);
+
+	if ((EBOOKDocument::CONFIDENCE_EXCELLENT != confidence) &&
+	    (EBOOKDocument::CONFIDENCE_WEAK != confidence)) {
```
ojwb: Some people advocate this `if (CONSTANT == variable)` style of testing, but I think it's less readable than `if (variable == CONSTANT)`.  Also we don't use it elsewhere in the codebase, so for consistency it's probably better to avoid it.

The supposed benefit of it is that if you accidentally write `=` instead of `==` the code won't compile (because you can't assign to a constant).

I think that argument may have been valid a few decades ago, but modern compilers will warn about `=` in a condition (for example in `if` or `while`) so such cases will be caught anyway.  And the compiler warning will also spot `if (foo = bar)` when both are variables.

xapian-applications/omega/index_file.cc
```diff
@@ -153,8 +153,19 @@ void
...
+    Worker* omindex_ebook = new Worker("omindex_ebook");
+    index_library("application/vnd.palm", omindex_ebook);
+    index_library("application/x-fictionbook+xml", omindex_ebook);
+    index_library("application/x-zip-compressed-fb2", omindex_ebook);
+    index_library("application/x-sony-bbeb", omindex_ebook);
+    index_library("application/x-tcr-ebook", omindex_ebook);
+    index_library("application/x-qioo-ebook", omindex_ebook);
+#endif
+
 }
```
ojwb: Better to avoid blank lines at the start and end of blocks - they don't help readability, just consume vertical space.

xapian-applications/omega/handler_libebook.cc
barufa: Okay, I am going to change it.
Should I do it with the if below? (line 153 and 165)

xapian-applications/omega/index_file.cc
barufa: I will remove that line.

xapian-applications/omega/handler_libebook.cc
barufa: I am going to change it.

xapian-applications/omega/handler_libebook.cc
ojwb: Yes - I'd avoid this construct completely (in Xapian code at least - if you were contributing a patch to a different codebase where that was the preferred coding style, you ought to follow that).

xapian-applications/omega/Makefile.am
ojwb: Oh i see, `XAPIAN_CXXFLAGS` is needed so that `xapian/constinfo.h` can be included by `stringutils.h`.

Probably we should split `stringutils.h` into two headers to avoid that being necessary, but for now keeping this is OK.

ojwb:
Looking good - squashed and rebased onto master and pushed as eb39b7646ed2202c30ff62d4cc3080d2eb3d8f10.

PR #268, opened 2019-07-04
Fix Apple iWork document mimetypes

barufa:
It seems that there is a little error on the Apple iWork document 
mimetypes. It is defined “application/vnd.apple.keynote” at 
mimemap.tokens. However we use “application/vnd.apply.keynote” in 
index_file.cc.


ojwb:
Oops, well spotted.  Merged to master as f61dd99454f1dd4717cfd53c51dc4d4c41054875.

PR #270, opened 2019-07-10
Add letor bindings of python3 to xapian-bindings

VaibhavKansagara:
This also makes stat_flags(xapian-letor/feature.h) public as swig is
not able to wrap protected members properly.


xapian-bindings/configure.ac
```diff
@@ -11,6 +11,7 @@ dnl See xapian-core's HACKING document for details of the reasons for required
 dnl versions.
 AM_INIT_AUTOMAKE([1.13 parallel-tests -Wportability subdir-objects tar-ustar no-dist-gzip dist-xz])
 AC_CONFIG_SRCDIR([xapian-headers.i])
+AC_CONFIG_SRCDIR([xapianletor-headers.i])
```
ojwb: Don't do this.  `AC_CONFIG_SRCDIR` just needs to be invoked once, with a relative filename that should exist in the source tree but is unlikely to exist in a random other directory (so the macro can sanity check that we're in the right source tree or something like that).

VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/configure.ac
```diff
@@ -94,6 +94,9 @@ XO_LIB_XAPIAN([], [],
 			  [^\([0-9]*\.[0-9]*[13579]\)\..*$], [-\1]))
 XO_REQUIRE([1.4.0])
 
+dnl Check for xapian-letor.
+XO_LIB_XAPIANLETOR([], [], [xapianletor-config])
```
jaylett: I think we need a different `ACTION-IF-NOT-FOUND` to the default, because we don't want bindings to not build when letor isn't present; we just want a notification (probably `AC_MSG_WARN`) that we aren't going to build the letor bindings, and then we keep going — various things (principally in `Makefile.am`) would then need to be conditional so we don't try to build the letor bindings in the absence of the underlying library.

(Currently, it looks like `XO_LIB_XAPIANLETOR` supports `ACTION-IF-FOUND` but not `ACTION-IF-NOT-FOUND`, so that would need implementing.)

This _could_ be left to a separate PR, but it would mean that the bindings aren't fully releasable until done.

xapian-bindings/python3/Makefile.am
```diff
@@ -30,51 +30,78 @@ RST_DOCS = \
...
 	run-python-test \
 	sphinx-build \
 	testsuite.py \
 	test_xapian_star.py \
+	test_xapian_letor_star.py \
 	$(TESTS) $(BUILT_SOURCES) $(RST_DOCS)
 
 pkgpylibdir = @PYTHON3_LIB@/xapian
 
+letorpkgpylibdir = @PYTHON3_LIB@/xapianletor
```
jaylett: I'd probably keep all the `letorpkgpylib` stuff together, rather than interspersed with the core bindings as currently.

xapian-bindings/python3/Makefile.am
```diff
@@ -30,51 +30,78 @@ RST_DOCS = \
...
 
 # Because we don't install the .la file, "make uninstall" doesn't work and
 # we need to remove the file ourselves.
 uninstall-local:
 	rm -f $(DESTDIR)$(pkgpylibdir)/_xapian$(PYTHON3_SO)
+	rm -f $(DESTDIR)$(letorpkgpylibdir)/_xapianletor$(PYTHON3_SO)
 
 AM_CPPFLAGS = -I$(PYTHON3_INC)
-AM_CXXFLAGS = $(SWIG_CXXFLAGS) $(XAPIAN_CXXFLAGS)
+AM_CXXFLAGS = $(SWIG_CXXFLAGS) $(XAPIAN_CXXFLAGS) $(XAPIANLETOR_CXXFLAGS)
```
jaylett: I think this now should be target-specific, meaning that both `XAPIAN_CXXFLAGS` and `XAPIANLETOR_CXXFLAGS` move out of `AM_CXXFLAGS` into [per-object variables](https://www.gnu.org/software/automake/manual/html_node/Per_002dObject-Flags.html). I _think_ they belong in `xapianletor_la_CXXFLAGS`. I think also that `SWIG_CXXFLAGS` belongs with them, because the `AM_CXXFLAGS` [won't be included automatically](https://www.gnu.org/software/automake/manual/html_node/Program-and-Library-Variables.html#index-per_002dtarget-compilation-flags_002c-defined).

(@ojwb should jump in if I've got the automake details wrong!)

xapian-bindings/python3/Makefile.am
```diff
@@ -99,24 +126,59 @@ xapian/_xapian$(PYTHON3_SO): _xapian.la
...
     xapian/__init__.py \
     xapian/__pycache__/__init__.@PYTHON3_CACHE_TAG@.pyc \
     xapian/__pycache__/__init__.@PYTHON3_CACHE_OPT1_EXT@ \
+    xapianletor/_xapianletor$(PYTHON3_SO) \
+    xapianletor/__init__.py \
+    xapianletor/__pycache__/__init__.@PYTHON3_CACHE_TAG@.pyc \
+    xapianletor/__pycache__/__init__.@PYTHON3_CACHE_OPT1_EXT@ \
+    err_output_ListNet_2.txt \
+    ndcg_output_ListNet_2.txt \
+    training_output1.txt \
```
jaylett: This isn't for now, but the fact that we need to do this suggests that the API is slightly too high-level around training and scoring. There should be an API (which is the one we actually test) which returns the data, rather than writing out files. (There are various situations where you don't want to or even can't write to your file system but can store output elsewhere.)

jaylett (changes requested):
While this will work, at some point we should see if there's a suitable way of making the library `import xapian.letor`, which feels more natural from a python point of view.

xapian-bindings/configure.ac
VaibhavKansagara: sure I would create a new PR for that after this is merged

xapian-bindings/python3/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: I will try to change the API to incorporate the changes you suggested.

jaylett: Don't jump on it right away. Add it to your list of changes (either for during the project or after). It's always good to spot future useful work, but there's an important discipline in not getting distracted by it!

.travis.yml
```diff
@@ -70,6 +72,7 @@ matrix:
             - python-docutils
             - pngcrush
             - python-sphinx
+	    - python3-sphinx
```
jaylett: While this will probably work, the whitespace should be consistent.

VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/configure.ac
ojwb: > (Currently, it looks like `XO_LIB_XAPIANLETOR` supports `ACTION-IF-FOUND` but not `ACTION-IF-NOT-FOUND`, so that would need implementing.)

Hmm, it documents support but comparing with `xapian-1.5.m4` it seems the equivalent of `if test -z "$XAPIAN_CONFIG"; then` ... has been omitted for some reason.

That really needs fixing, and I think it makes much more sense to sort this out before we merge this patch.  Otherwise until this gets addressed xapian-letor becomes a hard requirement for xapian-bindings for anybody working on git master, which really isn't helpful.

xapian-bindings/python3/Makefile.am
```diff
@@ -149,7 +212,23 @@ $(stamp): except.i doccomments.i fixup-swig-py3-wrapper
 
 -include xapian_wrap.d
 
-CLEANFILES += xapian_wrap.d $(stamp)
```
ojwb: It seems cleaner to leave this alone and just add a separate `CLEANFILES += ...` for the letor files below - that way all the core stuff is together followed by all the letor stuff.

xapian-bindings/python3/pythonletortest.py
```diff
@@ -0,0 +1,184 @@
...
+if have_threads:
+    print("Running tests with threads")
+
+    # This testcase seems to just block when run in a thread under Python 3
+    # on some plaforms.  It fails with 3.2.3 on Debian wheezy, but passes
+    # with the exact same package version on Debian unstable not long after
+    # the jessie release.  The issue it's actually serving to regression
+    # test for is covered by running it without threads, so just disable it
+    # rather than risk test failures that don't seem to indicate a problem
+    # in Xapian.
```
ojwb: Just copying this comment over results in it making some rather misleading claims - I doubt test_import_letor_star has really been tested on Debian wheezy or jessie at all.

xapian-bindings/python3/utilletor.i
```diff
@@ -0,0 +1,165 @@
...
+	    Py_DECREF(result);
+	    return NULL;
+	}
+
+	PyList_SET_ITEM(result, i, str);
+    }
+    $input = result;
+}
+
+/* vim:set syntax=cpp:set noexpandtab: */
```
ojwb: This looks like it's just a copy of `util.i`, complete with several typemaps letor doesn't currently need.

Perhaps it would be better to simply include `util.i` directly rather than forking our own code needlessly and creating extra maintenance work for our future selves?

xapian-bindings/xapianletor-head.i
```diff
@@ -0,0 +1,69 @@
...
+ * USA
+ */
+
+// Disable any deprecation warnings for Xapian methods/functions/classes.
+#define XAPIAN_DEPRECATED(D) D
+
+#include <xapian-letor.h>
+
+#include <fstream>
+#include <iostream>
```
ojwb: Neither of these seem to be used by the letor bindings (they are for the core ones).

xapian-bindings/xapianletor-head.i
```diff
@@ -0,0 +1,69 @@
...
+
+%}
+
+using namespace std;
+
+%include exception.i
+%include stl.i
+
+// Disable errors about not including headers individually.
+#define XAPIAN_IN_XAPIAN_H
```
ojwb: This is specific to xapian-core.

xapian-bindings/xapianletor-headers.i
```diff
@@ -0,0 +1,88 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+%}
+
+/* Ignore these functions: */
+%ignore Xapian::iterator_rewind;
+%ignore Xapian::iterator_valid;
```
ojwb: These are in the core API...

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -72,7 +72,6 @@ class XAPIAN_VISIBILITY_DEFAULT Feature {
     /// A bitmask of the statistics this Feature needs.
     stat_flags stats_needed;
 
-  public:
```
ojwb: I think we want to mark the above as `@protected` in the doxygen comments - we're only making them `public` because of limitations from SWIG, not because we want people to treat them as `public` in general.

Also, does all of the above need exposing?  I'd think only the `enum` does.

(And not directly related to this patch, but `stats_needed` should probably be `private` not `protected` - it's readable via public `get_stats()` and can be added to via protected `need_stat()` so nothing should need to directly access it.)

xapian-bindings/python3/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/pythonletortest.py
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/utilletor.i
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/xapianletor-head.i
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/xapianletor-head.i
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/xapianletor-headers.i
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/configure.ac
VaibhavKansagara: I have fixed all those issues in this PR.

xapian-letor/include/xapian-letor/feature.h
VaibhavKansagara: > Also, does all of the above need exposing? I'd think only the enum does.
>
I think yes because feature default constructor initializes stats_needed so it has to be public
and making need_stat() protected or private gives swig errors.

VaibhavKansagara: > Also, does all of the above need exposing? I'd think only the enum does.
>
I think yes because feature default constructor initializes stats_needed so it has to be public
and making need_stat() protected or private gives swig errors.

ojwb: The `Feature` default constructor is a method of the `Feature` class, so should be able to access all members.  The question is whether any SWIG generated code which isn't in a subclass of `Feature` needs to access these things.  In the case of `stats_needed` I can't see why SWIG needs access it directly at all.

VaibhavKansagara: > The Feature default constructor is a method of the Feature class, so should be able to access all members. 
>
Oh sorry that's not what I meant. I meant that because Feature default constructor is public, we need to make stats_needed public as not doing so, compiler will complain of not seeing stats_needed as it will be defined in private block below, so this is an ordering issue. Also we can't put private parts first as stat_flags need to be defined first in order to define stats_needed.

VaibhavKansagara: > In the case of stats_needed I can't see why SWIG needs access it directly at all.
>
I think it doesn't need direct access but as I said above we can't make stats_needed private or protected if even that doesn't cause swig errors.

xapian-bindings/python3/docs/examples/xapian-train.py
```diff
@@ -0,0 +1,49 @@
...
+import xapianletor
+
+# We require at least two command line arguments.
+if len(sys.argv) < 3:
+    print("Usage: PATH_TO_DATABASE PATH_TO_TRAINING_FILE",
+          sep='\n', file=sys.stderr)
+    sys.exit(1)
+
+try:
+    # Open the database for searching.
```
jaylett: Again, this isn't what the following lines do.

jaylett (changes requested):
*Future thought*, although it'd be good to decide if this is a short-term thing or a longer-term thing (because it will change the API).

Now the API is coming together and I'm seeing more example code driving it, I'm wondering at all the times we're passing around a `db_path` instead of a suitable `Database` object. There's some subtlety here when we're training (because we need a `WritableDatabase`), but at least for searching I wonder if we can pass in a database to the ranker constructor.

Otherwise, we make some more complex uses harder. For instance, you'd have to create a stub file to reference a remote. This isn't really a problem, but it's a little inelegant, and is contrary to the rest of Xapian.

VaibhavKansagara:
> _Future thought_, although it'd be good to decide if this is a short-term thing or a longer-term thing (because it will change the API).
> 
> Now the API is coming together and I'm seeing more example code driving it, I'm wondering at all the times we're passing around a `db_path` instead of a suitable `Database` object. There's some subtlety here when we're training (because we need a `WritableDatabase`), but at least for searching I wonder if we can pass in a database to the ranker constructor.
> 
> Otherwise, we make some more complex uses harder. For instance, you'd have to create a stub file to reference a remote. This isn't really a problem, but it's a little inelegant, and is contrary to the rest of Xapian.
>
Changes made in a fixup commit.

xapian-bindings/python3/docs/examples/xapian-prepare-trainingfile.py
```diff
@@ -0,0 +1,45 @@
...
+import xapianletor
+
+# We require at least five command line arguments.
+if len(sys.argv) < 6:
+    print("Usage: PATH_TO_DATABASE PATH_TO_QUERY_FILE"
+          "PATH_TO_QREL_FILE PATH_TO_TRAINING_FILE MSIZE", sep='\n', file=sys.stderr)
+    sys.exit(1)
+
+try:
+    # Open the database for searching.
```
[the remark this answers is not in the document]
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/examples/xapian-train.py
VaibhavKansagara: fixed in a fixup commit.

jaylett:
> > _Future thought_… I'm wondering at all the times we're passing around a `db_path` instead of a suitable `Database` object.
> 
> Changes made.

I'm unconvinced that the change you've made actually encompasses what I was talking about, which is why I said very clearly that this was a _future thought_. This is emphasised perhaps most clearly by the fact that I made the comment after looking at the python examples, which you've left unchanged.

I think you should revert that commit and then we can discuss how the API should change.

VaibhavKansagara:
> > > _Future thought_… I'm wondering at all the times we're passing around a `db_path` instead of a suitable `Database` object.
> > 
> > 
> > Changes made.
> 
> I'm unconvinced that the change you've made actually encompasses what I was talking about, which is why I said very clearly that this was a _future thought_. This is emphasised perhaps most clearly by the fact that I made the comment after looking at the python examples, which you've left unchanged.
> 
> I think you should revert that commit and then we can discuss how the API should change.

Ah sorry didn't realise that. Will revert the change.

xapian-letor/include/xapian-letor/feature.h
ojwb: > this is an ordering issue

In a C++ class definition you can use a member before it is declared/defined.

Indeed, look at the current code on the branch and you'll see `stats_needed` is **already** defined after the `need_stat()` function refers to it:

```
    void need_stat(stat_flags flag) { 
        stats_needed = stat_flags(stats_needed | flag);
    }                         
                                 
    /// @protected A bitmask of the statistics this Feature needs.
    stat_flags stats_needed;
```

If I move `stats_needed` down to the `private` section and adjust the member initialisation order in the constructor then the code compiles fine.

(It's also possible to use `private:` multiple times, though generally better not to unless there's a good reason to).

VaibhavKansagara: Ah I see thanks.

xapian-bindings/configure.ac
```diff
@@ -131,6 +149,10 @@ dnl CXXFLAGS for the duration of configure.
 save_CXXFLAGS=$CXXFLAGS
 CXXFLAGS="$CXXFLAGS $XAPIAN_CXXFLAGS"
 
+if test yes = "$enable_letor"; then
+    CXXFLAGS+="$XAPIANLETOR_CXXFLAGS"
```
ojwb: `+=` isn't portable shell - use the same idiom as just above to append to a variable instead.

ojwb: Also this lack a space before the appended value, so would result in `CXXFLAGS` with a value like `-I/home/olly/git/xapian/xapian-core/include-I/home/olly/git/xapian/xapian-letor/include`...

xapian-bindings/python3/Makefile.am
```diff
@@ -39,8 +39,7 @@ EXTRA_DIST = python.i util.i extra.i extracomments.i except.i \
 	run-python-test \
 	sphinx-build \
 	testsuite.py \
-	test_xapian_star.py \
-	$(TESTS) $(BUILT_SOURCES) $(RST_DOCS)
```
ojwb: I don't see any need to move these from here - make assignments aren't processed in file order (unless you use `:=` or similar) so it doesn't matter that the values of these are appended to below.

xapian-bindings/python3/docs/examples/xapian-prepare-trainingfile.py
```diff
@@ -0,0 +1,44 @@
...
+# USA
+
+import sys
+import xapian
+import xapianletor
+
+# We require at least five command line arguments.
+if len(sys.argv) < 6:
+    print("Usage: PATH_TO_DATABASE PATH_TO_QUERY_FILE"
+          "PATH_TO_QREL_FILE PATH_TO_TRAINING_FILE MSIZE", sep='\n', file=sys.stderr)
```
ojwb: The name of the program is missing after `Usage: `.  It's common to do something like `"Usage: %s ...." % sys.argv[0]`.

The output is be missing a space:

```
Usage: PATH_TO_DATABASE PATH_TO_QUERY_FILEPATH_TO_QREL_FILE PATH_TO_TRAINING_FILE MSIZE
```

Also the `PATH_TO_` on the filenames seems a bit redundant and makes this harder to read - compare with:

```
Usage: DATABASE QUERY_FILE QREL_FILE TRAINING_FILE MSIZE
```

`sep='\n'` seems redundant when there's only one item listed.

xapian-bindings/python3/docs/examples/xapian-prepare-trainingfile.py
```diff
@@ -0,0 +1,44 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+# USA
+
+import sys
+import xapian
+import xapianletor
+
+# We require at least five command line arguments.
+if len(sys.argv) < 6:
```
ojwb: This allows more than 5 arguments, but ignores the extra ones.  I think you want `!=` here.

xapian-bindings/python3/docs/examples/xapian-rank.py
```diff
@@ -0,0 +1,92 @@
...
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+# USA
+
+import sys
+import xapian
+import xapianletor
+
+# We require at least three command line arguments.
+if len(sys.argv) < 4:
+    print("Usage: PATH_TO_DATABASE MSIZE QUERY",
```
ojwb: Again allows but ignores extra arguments, and the program name is missing.  Please fix any further instances too.

xapian-bindings/python3/docs/introduction.rst
```diff
@@ -13,6 +14,12 @@ them, you'll need to add this to your code::
 
   import xapian
 
+It also includes `xapianletor` bindings module - to use
+them, you'll need to configure xapian-bindings with --enable-letor=yes
```
ojwb: Just `--enable-letor` should work and seems clearer.

It's also helpful to mark up command options so they stand out visually when converted to HTML, PDF, etc - double backticks render their content as a fixed width font, which it like `<code>` or `<tt>` in HTML: ` ``--enable-letor`` `

xapian-bindings/python3/docs/introduction.rst
```diff
@@ -210,8 +223,10 @@ Test Suite
 ==========
 
 The Python bindings come with a test suite, consisting of two test files:
-`smoketest.py` and `pythontest.py`. These are run by the `make check` command,
-or may be run manually.  By default, they will display the names of any tests
+`smoketest.py` and `pythontest.py`. It also includes test suite for letor-bindings
```
ojwb: ... includes **a** test suite ...

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -41,7 +41,7 @@ namespace Xapian {
 
 /// Abstract base class for features in learning to rank
 class XAPIAN_VISIBILITY_DEFAULT Feature {
-  protected:
+  public:
     /// Stats which FeatureList can use.
```
ojwb: The documentation comment for this enum now ought to say `@protected` so that the API docs continue to show it as protected.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: It's because `BUILT_SOURCES` and `TESTS` are not finalized yet, otherwise `EXTRA_DIST` would include repeated files.

xapian-bindings/configure.ac
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/examples/xapian-prepare-trainingfile.py
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/examples/xapian-prepare-trainingfile.py
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/examples/xapian-rank.py
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/introduction.rst
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/introduction.rst
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/include/xapian-letor/feature.h
VaibhavKansagara: fixed in a fixup commit.

ojwb:
> It's because BUILT_SOURCES and TESTS are not finalized yet, otherwise
> EXTRA_DIST would include repeated files.
>
That's not how make works - read about deferred variable expansion:
https://www.gnu.org/software/make/manual/html_node/Reading-Makefiles.html

>

xapian-bindings/configure.ac
```diff
@@ -94,6 +94,24 @@ XO_LIB_XAPIAN([], [],
 			  [^\([0-9]*\.[0-9]*[13579]\)\..*$], [-\1]))
 XO_REQUIRE([1.4.0])
 
+AC_ARG_ENABLE([letor],
+  [AS_HELP_STRING([--enable-letor], [])],
```
ojwb: This has an empty help string, so `configure --help` says nothing about what this option does, unlike all the options around it:

```
  --disable-dependency-tracking
                          speeds up one-time build
  --disable-libtool-lock  avoid locking (might break parallel builds)
  --enable-letor
  --disable-visibility    disable use of GCC visibility
  --enable-maintainer-mode
                          enable make rules and dependencies not useful (and
                          sometimes confusing) to the casual installer
  --enable-documentation  enable make rules to rebuild shipped documentation
                          [default=maintainer-mode]
```
```

File: xapian-bindings/configure.ac [hunk 25]
```diff
@@ -94,6 +94,24 @@ XO_LIB_XAPIAN([], [],
...
+  [case ${enableval} in
+    yes) ;;
+    no)
+      AC_MSG_WARN([xapian-letor is unavailable, hence letor-bindings won't be built])
```
ojwb: This warning doesn't really make sense.  We'll only emit it if the user explicitly ran something like `./configure --disable-letor` and if they did that it shouldn't be a surprise that letor was disabled.  Also, the warning message is wrong - we're disabling it because we were told to - we haven't checked if it was available or not at this point.

xapian-bindings/configure.ac
```diff
@@ -94,6 +94,24 @@ XO_LIB_XAPIAN([], [],
...
+      ;;
+    *) AC_MSG_ERROR([bad value ${enableval} for --enable-letor]) ;;
+  esac],
+  [enable_letor=no])
+
+AM_CONDITIONAL([LETOR_FOUND], [test yes = "$enable_letor"])
+
+if test yes = "$enable_letor"; then
+    dnl Check for xapian-letor.
+    XO_LIB_XAPIANLETOR([], [], [xapianletor-config])
```
ojwb: The existing shell code in this file uses a 2 space indent, but here (and below) you're using 4 spaces.

xapian-bindings/python3/Makefile.am
```diff
@@ -172,3 +174,110 @@ $(sphinxdocs): xapian/__init__.py xapian/_xapian$(PYTHON3_SO) docs/conf.py $(RST
...
+
+xapianletor_wrap.cc xapianletor_wrap.h xapianletor.py xapianletor_wrap.d: $(stampletor)
+	$(make_many_locked)
+$(stampletor): exceptletor.i fixup-swig-py3-wrapper
+	$(multitarget_begin)
+	$(RUN_SWIG_LETOR) $(SWIG_WERROR) -I. -I'$(srcdir)' $(SWIG_FLAGS) $(SWIG_LETOR_FLAGS) -c++ \
+	    -python -threads -shadow -O \
+	    -o xapianletor_wrap.cc '$(srcdir)/'pythonletor.i
+	$(PERL) -i $(srcdir)/fixup-swig-py3-wrapper xapianletor.py
+	$(multitarget_end)
```
ojwb: We ship the SWIG-generated sources, so these rules need to have run before `make dist` works.  That means `make dist` requires configure to have been run with `--enable-letor`.  If it wasn't, these files won't be found, and the rules to create them are disabled.  Or worse, if stale versions of them are present in the tree from a previous build with `--enable-letor` then those could get shipped.

I'm not quite sure how we want this to work, but the current way it does seems a bit problematic.

Not shipping the SWIG-generated sources is one answer, but because we build bindings for N languages we often have quite a tight SWIG version dependency because there's a recent fix for some language or other we really need.  So that would require users to install a particular SWIG build, and we'd probably end up fending off bug reports from users who don't read the instructions.

xapian-bindings/python3/Makefile.am
```diff
@@ -172,3 +174,110 @@ $(sphinxdocs): xapian/__init__.py xapian/_xapian$(PYTHON3_SO) docs/conf.py $(RST
...
+
+CLEANFILES += \
+    xapianletor/_xapianletor$(PYTHON3_SO) \
+    xapianletor/__init__.py \
+    xapianletor/__pycache__/__init__.@PYTHON3_CACHE_TAG@.pyc \
+    xapianletor/__pycache__/__init__.@PYTHON3_CACHE_OPT1_EXT@ \
+    err_output_ListNet_2.txt \
+    ndcg_output_ListNet_2.txt \
+    training_output1.txt \
+    $(srcdir)/*.pyc
```
ojwb: `$(srcdir)/*.pyc` is already in `CLEANFILES` from above.

xapian-bindings/python3/docs/examples.rst
```diff
@@ -41,3 +41,33 @@ simplematchdecider.py
    :language: python
    :emphasize-lines: 12,15-18
    :linenos:
+
+.. _xapian-prepare-trainingfile:
+
+xapian-prepare-trainingfile.py
+=====================
```
ojwb: These `=` lines ought to be the same length as the line above each one (multiple instances).

xapian-bindings/python3/docs/examples.rst
```diff
@@ -41,3 +41,33 @@ simplematchdecider.py
...
+
+.. literalinclude :: examples/xapian-prepare-trainingfile.py
+   :language: python
+   :emphasize-lines: 12,15-18
```
ojwb: These lines are meant to be the lines in the example code which are the interesting part of the code, not just copied verbatim from an existing example (though they also seem to be wrong for all the existing examples - we're currently emphasising some lines in the licensing boilerplate).

ojwb: Having looked at this a bit, I think it's best just to drop this idea - it's only really visible in the output if you know it's there (https://xapian.org/docs/bindings/python3/examples.html) and will require updating each time the example code changes in a way which adds or removes lines.  If it actually improved the output visibly then we could consider how to automate that, but given you can't really see the effect that doesn't seem worth doing.

xapian-bindings/configure.ac
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/configure.ac
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/configure.ac
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/examples.rst
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/docs/examples.rst
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: Understood but It's not exactly clear what you want me to do here.

xapian-bindings/configure.ac
ojwb: @jaylett Are you happy the changes you requested have been made?

xapian-bindings/python3/Makefile.am
ojwb: I'm not quite sure exactly what we want to do here either.

I think the underlying problem is that conditionals in `Makefile.am` can interact with `make dist`, but the distribution tarballs really need to be reproducible, and so should not depend on configure options, etc.

I think automake can "see through" conditionals when determining the sources files to ship, so at least we wouldn't just end up creating a distribution which is missing the letor sources entirely, but `make` needs to be able to see rules to create files we should be shipping and it can't if those rules are inside conditionals, and `automake` can't really solve that.

So perhaps the rule to run swig should be outside the `if LETOR_FOUND` conditional, and similarly for any other rules which generate files to ship.  Just building the non-letor bindings should not trigger such rules, but at least then `make dist` can't end up shipping stale generated files.

This is all a bit fiddly to get right though.

xapian-bindings/python3/.gitignore
```diff
@@ -6,9 +6,13 @@
 /.libs
 /*.pyc
 /_xapian.la
+/_xapianletor.la
 /xapian
+/xapianletor
 /modern
 /except.i
+/exceptletor.i
+/database_two_docs
```
jaylett: Is there a reason not to prefix this `db_` (instead of `database_`) so it matches other test databases? Would also mean it's covered by the local clean rule.

xapian-bindings/configure.ac
```diff
@@ -94,6 +94,21 @@ XO_LIB_XAPIAN([], [],
...
 
+AC_ARG_ENABLE([letor],
+  [AS_HELP_STRING([--enable-letor], [enable make rules to build bindings related to xapian-letor])],
+  [case ${enableval} in
+    yes|no) ;;
+    *) AC_MSG_ERROR([bad value ${enableval} for --enable-letor]) ;;
+  esac],
+  [enable_letor=no])
+
+AM_CONDITIONAL([LETOR_FOUND], [test yes = "$enable_letor"])
```
jaylett: `LETOR_FOUND` isn't a great choice for this variable, given at this point it's "just" about whether it's requested or not. `BUILD_LETOR` or `BUILD_LETOR_BINDINGS` would be clearer.

xapian-bindings/python3/Makefile.am
jaylett: I think the most important thing will be to ensure that any `make dist` that _we_ (as the Xapian project) might do is run with `--enable-letor`. Putting a warning somewhere about this (`INSTALL`, perhaps?) is probably a good idea. Or even, if we can, spitting out a warning on `make dist` without `--enable-letor` in the configure run.

xapian-bindings/xapianletor-head.i
```diff
@@ -0,0 +1,64 @@
...
+#define XAPIAN_DEPRECATED(D) D
+#define XAPIAN_DEPRECATED_EX(D) D
+#define XAPIAN_DEPRECATED_CLASS
+#define XAPIAN_DEPRECATED_CLASS_EX
+#define XAPIAN_VISIBILITY_DEFAULT
+#define XAPIAN_VISIBILITY_INTERNAL
+#define XAPIAN_CONST_FUNCTION
+#define XAPIAN_PURE_FUNCTION
+#define XAPIAN_NOEXCEPT
+#define XAPIAN_NOTHROW(D) D
```
jaylett: Anything we need for both letor and core we should share explicitly (quite a lot of this file) — a `xapian-common.i` that we `%include` in both should be sufficient.

xapian-bindings/xapianletor-headers.i
```diff
@@ -0,0 +1,84 @@
...
+%ignore difference_type;
+%ignore iterator_category;
+%ignore value_type;
+%ignore max_size;
+%ignore swap;
+%ignore iterator;
+%ignore const_iterator;
+%ignore size_type;
+%ignore unserialise(const char **, const char *);
+%ignore release();
```
jaylett: Everything above this seems common with core, and could perhaps be shared?

xapian-bindings/xapianletor-headers.i
```diff
@@ -0,0 +1,84 @@
...
+%include <xapian/types.h>
+
+%include <xapian/visibility.h>
+
+/* The Error subclasses are handled separately for languages where we wrap
+ * them. */
+/* %include <xapian-letor/letor_error.h> */
+
+SUBCLASSABLE(Xapian, Feature)
+STANDARD_IGNORES(Xapian, Feature)
```
jaylett: In core, both of these macros are defined in `xapian-headers.i`, so I'm not quite sure how this works here.

jaylett (changes requested):
Some of these are minor, but some are a bit more involved and perhaps we need to discuss in the next meeting.

xapian-bindings/xapianletor-headers.i
ojwb: Probably a good plan, but perhaps better to address after merging rather than dragging out this PR further.

The current structuring is partly a result of the history - originally we had a copy of the C++ API declarations we wanted to wrap here in the bindings, which was a maintenance headache so we switched over to having SWIG parse the C++ API headers instead. 

We switched class by class with the "wrap via API headers" code being added to `xapian-headers.i` and the old declarations living in `xapian.i` (e.g. see 553db57962ba6d2d80e7518ad53441db5dbb2310 for TermIterator being switched).  Once everything was switched, `xapian.i` went away completely.

`xapian-head.i` is mostly stuff that needs including early on, but also has fallback implementations of the remote backend factory functions which perhaps should live somewhere else.  Arguably that should be handled in xapian-core, though if you're disabling backends you don't need to reduce the library size (e.g. for embedded use) then it's nicer not to have stub implementations.  They could be inlined from the core API headers if disabled though, which should mean no overhead in a program which doesn't call them.

jaylett: Yeah, I'd be happy to punt this to the future.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: Done.

xapian-bindings/python3/.gitignore
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/configure.ac
VaibhavKansagara: fixed in a fixup commit.

xapian-bindings/xapianletor-head.i
VaibhavKansagara: fixed in a fixup commit

xapian-bindings/xapianletor-headers.i
VaibhavKansagara: > In core, both of these macros are defined in `xapian-headers.i`, so I'm not quite sure how this works here.
>
Yes but this file is included in pythonletor.i which includes xapian-headers.i also.

jaylett:
Travis doesn't seem to be actually running the python letor tests, so I assume we need to poke --enable-letor in somewhere in the CI configuration as well.

jaylett (approved):

VaibhavKansagara:
@ojwb @jaylett I don't know why CI is failing.

xapian-bindings/python3/pythonletortest.py
```diff
@@ -98,6 +99,7 @@ def test_preparetrainingfile():
     import filecmp
     expect(filecmp.cmp(training_data, "training_output1.txt"), True)
     os.remove("training_output1.txt")
+    shutil.rmtree(dbpath)
```
jaylett: Presumably you want to clear up `dbpath` no matter what? In which case you should use `try…finally`, which ensures the cleanup happens even with exceptions. (So probably run it after setting up the database, or ensure the `rmtree()` won't fail if the directory is missing.)

xapian-bindings/configure.ac
```diff
@@ -116,15 +116,15 @@ INTREE_DYLD_PATH=
 case $host_os in
 darwin*)
   NEED_JNILIB_EXT=yes
-  dnl If we're building against an uninstalled xapian-core, xapian-config
+  dnl If we're building against an uninstalled xapian-letor, xapianletor-config
   dnl --libs will give an error.
-  if $XAPIAN_CONFIG --libs >/dev/null 2>/dev/null ; then
+  if $XAPIANLETOR_CONFIG --libs >/dev/null 2>/dev/null ; then
```
jaylett: Surely we need to do both? Because we want to be able to work against an uninstalled core _and_ an uninstalled letor.

xapian-bindings/configure.ac
```diff
@@ -131,6 +169,10 @@ dnl CXXFLAGS for the duration of configure.
 save_CXXFLAGS=$CXXFLAGS
 CXXFLAGS="$CXXFLAGS $XAPIAN_CXXFLAGS"
 
+if test yes = "$enable_letor"; then
+    CXXFLAGS="$CXXFLAGS $XAPIANLETOR_CXXFLAGS"
+fi
```
ojwb: Looks like you didn't fix the overindentation here.

xapian-bindings/python3/Makefile.am
```diff
@@ -172,3 +174,109 @@ $(sphinxdocs): xapian/__init__.py xapian/_xapian$(PYTHON3_SO) docs/conf.py $(RST
...
+	. ./libtoolconfig.tmp; cp $$objdir/_xapianletor$(PYTHON3_SO) xapianletor
+	rm -f libtoolconfig.tmp
+
+CLEANFILES += \
+    xapianletor/_xapianletor$(PYTHON3_SO) \
+    xapianletor/__init__.py \
+    xapianletor/__pycache__/__init__.@PYTHON3_CACHE_TAG@.pyc \
+    xapianletor/__pycache__/__init__.@PYTHON3_CACHE_OPT1_EXT@ \
+    err_output_ListNet_2.txt \
+    ndcg_output_ListNet_2.txt \
```
ojwb: In core we have what seem to be equivalent filenames, but with all lowercase names: `err_output_listnet_2.txt` and `ndcg_output_listnet_2.txt` - it seems odd to use a different case for them here (and in general the codebase doesn't use mixed-case names except when it's a requirement of some tool, such as `Makefile` and `Makefile.am`).

xapian-bindings/python3/docs/examples.rst
```diff
@@ -8,7 +8,6 @@ simplesearch.py
 
 .. literalinclude :: examples/simplesearch.py
    :language: python
-   :emphasize-lines: 12,15-18
```
ojwb: Once fixups are collapsed, this change (and the similar ones below) is being made as part of a commit "Add letor bindings of python3 to xapian-bindings", but it has no connection to that change at all.  It is a change we want to make, but it really ought to be done separately (indeed I have a note to address this - I've held off from just doing this as it would probably have conflicted with this PR).

xapian-bindings/python3/docs/introduction.rst
```diff
@@ -28,6 +35,11 @@ There's also
 `simplematchdecider.py <examples/simplematchdecider.py>`_
 which shows how to define a MatchDecider in Python.
 
+Examples specific to letor are:
+`xapian-prepare-trainingfile.py <examples/xapian-prepare-trainingfile.py>`_.,
+`xapian-train.py <examples/xapian-train.py>`_.,
```
ojwb: The two commas above both have an unwanted full stop before them - in the rendered output this looks like:

```
Examples specific to letor are: xapian-prepare-trainingfile.py., xapian-train.py., xapian-rank.py.
```

xapian-bindings/python3/docs/examples.rst
ojwb: Perhaps I should give some practical examples on why lumping something like this into an unrelated commit is unhelpful.

Currently the result of these `:emphasize-lines: 12,15-18` is an odd visual glitch in the HTML output, where those 5 lines get emphasised for no good reason.  That's something which it would be beneficial to fix in 1.4.x too - it's about a safe a change as they come, and improves the documentation a little.

But adding bindings for xapian-letor is not 1.4.x material at all.  Include this change here makes it harder to backport (because I can't just `git cherry-pick` a single commit and have it apply cleanly - I would need to pull out part of a commit, and even part of the changes made to one file by a commit).

It also makes it less obvious there's even something to backport.  I backport based on checking each commit on master and marking it as not to backport or backported - if a commit says "Add letor bindings of python3 to xapian-bindings" then I would likely quickly mark that as not to backport, and a change which ought to be backported gets missed.

Similar to backporting, a packager of xapian-bindings might want to pull out a particular fix and apply it to their package as an extra patch - that's much simpler to do if it's a single clean commit, and simpler to keep track of when the patch is safe to drop.

If on the other hand this extra change turned out to be wrong (which in this case seems very unlikely, but in general can happen) then we'd want to revert the change.  If it's done as a separate commit, then `git revert` will undo it, usually without needing manual intervention.  But if it's lumped in to another commit then again someone has to pick out parts of a commit, but this time to undo them rather than apply them to a different barnch - and that introduces more scope for errors.

xapian-bindings/python3/docs/examples/xapian-rank.py
```diff
@@ -0,0 +1,94 @@
...
+    parser.add_prefix("subject", "S")
+    parser.set_database(db)
+    parser.set_default_op(xapian.Query.OP_OR)
+    parser.set_stemmer(xapian.Stem("en"))
+    parser.set_stemming_strategy(xapian.QueryParser.STEM_SOME)
+
+    query_no_prefix = parser.parse_query(query_string,
+                                         parser.FLAG_DEFAULT|
+                                         parser.FLAG_SPELLING_CORRECTION)
+    # query with title as default prefix
```
ojwb: Passing `FLAG_SPELLING_CORRECTION` (and below) is pointless here as we don't look at the result - it just makes the example a bit less easy to follow.  Without that we can just drop the second argument to `parse_query()` here as the default flags are `FLAG_DEFAULT`.

xapian-bindings/python3/pythonletortest.py
VaibhavKansagara: fixed in  a fixup commit.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: fixed in  a fixup commit.

xapian-bindings/configure.ac
VaibhavKansagara: fixed in  a fixup commit.

xapian-bindings/python3/docs/introduction.rst
VaibhavKansagara: fixed in  a fixup commit.

xapian-bindings/python3/docs/examples/xapian-rank.py
VaibhavKansagara: fixed in  a fixup commit.

xapian-bindings/python3/docs/examples.rst
VaibhavKansagara: I have added a fixup commit which adds the emphasis lines and a new commit which removes it.
I hope that is good enough.

xapian-letor/m4-macros/xapianletor.m4
```diff
@@ -52,56 +52,64 @@ AC_DEFUN([XO_LIB_XAPIANLETOR],
...
-    126)
-    if test -d "$XAPIANLETOR_CONFIG" ; then
-      AC_MSG_ERROR(['$XAPIANLETOR_CONFIG' is a directory; it should be the filename of the xapianletor-config script])
-    fi
-    AC_MSG_ERROR(['$XAPIANLETOR_CONFIG' not executable, aborting])
-    ;;
-    esac
+    AC_MSG_CHECKING([$XAPIANLETOR_CONFIG works])
+    dnl check for --ltlibs but not --libs as "xapianletor-config --libs" will
+    dnl continue if xapian-letor isn't installed...
```
ojwb: In `xapian.m4` the equivalent comment says `... will **fail** if xapian isn't installed".

I'm not sure why you've changed from "fail" to "continue" here - testing in an uninstalled source tree shows it fails for letor just like it does for core:

```
olly@gemse:~/git/xapian/xapian-letor$ ./xapianletor-config --libs
./xapianletor-config: Use libtool and 'xapianletor-config --ltlibs'
to link with an uninstalled libxapianletor.
olly@gemse:~/git/xapian/xapian-letor$ echo $?
1
```

Am I missing something?

xapian-bindings/python3/Makefile.am
ojwb: I tested building snapshot/release tarballs - currently this is the command that we build tarballs with:

```
env LD_LIBRARY_PATH="`pwd`"/xapian-core/.libs make -s distcheck VALGRIND=
```

And the obvious extension to that is:

```
env LD_LIBRARY_PATH="`pwd`"/xapian-core/.libs:"`pwd`"/xapian-letor/.libs make -s distcheck VALGRIND=
```

But this doesn't work because:

```
make[5]: *** No rule to make target 'xapianletor_wrap.cc', needed by 'distdir-am'.  Stop.
```

The error here seems to indicate you haven't implemented what I suggested above - and indeed looking at `python3/Makefile.am` the rules to generate these files are still conditional on `BUILD_LETOR_BINDINGS`.  Changing that is not enough to make this actually work (as it needs to find the right xapian-letor headers but `xapianletor-config` won't be consulted without `--enable-letor`), but as I noted before this change would at least avoid the risk of shipping stale generated files.

Just documenting that you need to do X for Y to work properly is all very well, but do you always read all the documentation for something before using it?  I know I don't, and I doubt most people do either.  If the result of not doing X might be to silently do the wrong thing that's likely to end up biting someone sooner or later, especially for something like `make dist` where no special steps are usually needed.

As for actually making the snapshot and release tarball building process work, I think we probably need to add `--enable-letor` to `DISTCHECK_CONFIGURE_FLAGS` - I tested this patch and it seems to work:

```
diff --git a/xapian-bindings/Makefile.am b/xapian-bindings/Makefile.am
index 45a0e3cb1fcb..d3a9ce41d1de 100644
--- a/xapian-bindings/Makefile.am
+++ b/xapian-bindings/Makefile.am
@@ -28,7 +28,8 @@ doxygen-xml/index.xml: $(XAPIAN_HEADERS)
 endif
 endif
 
-DISTCHECK_CONFIGURE_FLAGS = XAPIAN_CONFIG="$(XAPIAN_CONFIG)" \
+DISTCHECK_CONFIGURE_FLAGS = --enable-letor \
+	XAPIAN_CONFIG="$(XAPIAN_CONFIG)" \
 	XAPIANLETOR_CONFIG="$(XAPIANLETOR_CONFIG)" \
 	PYTHON2_LIB="`pwd`/../_inst/tmp_py2lib" \
 	PYTHON3_LIB="`pwd`/../_inst/tmp_py3lib" \
```

That's kind of ugly, but I don't really see a better approach.

xapian-bindings/python3/Makefile.am
```diff
@@ -172,3 +174,109 @@ $(sphinxdocs): xapian/__init__.py xapian/_xapian$(PYTHON3_SO) docs/conf.py $(RST
...
+	xapianletor_wrap.h
+
+EXTRA_DIST += pythonletor.i \
+	exceptletor.i \
+	generate-letor-python-exceptions \
+	test_xapian_letor_star.py \
+	pythonletortest.py \
+	xapianletor.py \
+	xapianletor_wrap.cc \
+	xapianletor_wrap.h
```
ojwb: If I take the distribution tarball `xapian-bindings-1.5.0.tar.xz` then unpack it, generate a new distribution tarball, and compare the two tarballs:

```
tar xf xapian-bindings-1.5.0.tar.xz
cd xapian-bindings-1.5.0
# Need to specify XAPIAN_CONFIG here only because I haven't installed xapian-core-1.5.0
./configure XAPIAN_CONFIG=~/git/xapian/xapian-core/xapian-config
make dist
cd ..
tardiff xapian-bindings-1.5.0.tar.xz xapian-bindings-1.5.0/xapian-bindings-1.5.0.tar.xz
```

then the `tardiff` output shows that the following files are missing:

```
- python/doccomments.i
- python3/doccomments.i
- python3/exceptletor.i
- python3/generate-letor-python-exceptions
- python3/pythonletor.i
- python3/pythonletortest.py
- python3/test_xapian_letor_star.py
- python3/xapianletor.py
- python3/xapianletor_wrap.h
```

The first two are an existing bug here (which I've just pushed 5198109aacc7cfa2edfb611039fe216003eb37c5 to fix) but the rest indicate that this `EXTRA_DIST += ...` should be outside the conditional.

Also, this highlights that `xapianletor_wrap.cc` and `xapianletor_wrap.h` don't need to be listed here (note that `xapian_wrap.cc` and `xapian_wrap.h` aren't listed in `EXTRA_DIST`).

xapian-bindings/python3/docs/examples.rst
ojwb: Sounds OK.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: I think you forgot to configure with --enable-letor. 
> Also, this highlights that `xapianletor_wrap.cc` and `xapianletor_wrap.h` don't need to be listed here (note that `xapian_wrap.cc` and `xapian_wrap.h` aren't listed in `EXTRA_DIST`).
>
I think `xapian_wrap.h` and `xapian_wrap.cc` are listed in `EXTRA_DIST` as `BUILT_SOURCES`.

xapian-letor/m4-macros/xapianletor.m4
VaibhavKansagara: oh sorry, I don't know why I did it. fixed in a fixup commit.

xapian-bindings/python3/Makefile.am
ojwb: Right, but they aren't listed in `EXTRA_DIST` explicitly like the letor ones currently are.

(And I did check that removing the letor ones works before commenting...)

.travis.yml
```diff
@@ -155,7 +155,7 @@ before_script:
   # Bootstrap everything (including letor, which isn't done
   # by default), then configure using our chosen compiler.
   - travis_retry ./bootstrap xapian-core xapian-applications/omega swig xapian-letor xapian-bindings
-  - ./configure $confargs CC="$USE_CC" CXX="$USE_CXX"
+  - ./configure --enable-letor $confargs CC="$USE_CC" CXX="$USE_CXX"
```
ojwb: I don't think this should be handled in a CI-specific way.

If someone is doing development work on letor then we really want them to be working in a source tree with the bindings to letor also enabled - otherwise they might commit (or submit a PR for) changes which break the letor bindings without knowing it, because `make check` worked fine in their local tree because it wasn't testing the letor bindings.

So I think we want to handle this in `bootstrap` instead - e.g. if `xapian-letor` is being bootstrapped, then add `--enable-letor` to the configure options.

That way it gets enabled for the CI automatically (because that bootstraps `xapian-letor`) but also gets automatically enabled for developers working on `xapian-letor`.

jaylett: Do we have a way in `bootstrap` of adding `--enable-letor` only to configure for the bindings? That would be preferable. Looks like it should be possible around line 900 of bootstrap in the `END_OF_CONFIGURE` heredoc.

xapian-bindings/python3/Makefile.am
ojwb: Sorry, I missed this part of the comment before:

> I think you forgot to configure with --enable-letor.

No, that was a deliberate omission.

My point here is that what `make dist` produces should **not** depend on which configure options were or were not specified - if it does, there's a trap for someone to build an incomplete distribution.  Repacking a distribution like this shouldn't be a lossy process.

.travis.yml
ojwb: Line 900 is for SWIG - line 918 is the case for `xapian-bindings`.

It'd be a bit cleaner to target just `xapian-bindings` but it doesn't really matter if it's specified more broadly though - we already deliberately run `configure` with `--disable-option-checking` which turns off checks for unknown `--enable-*`, etc (so that you can use things like `./configure --enable-log` with the top level configure without getting warnings.)

VaibhavKansagara: Done.

xapian-bindings/python3/Makefile.am
VaibhavKansagara: fixed in a fixup commit.

.travis.yml
jaylett: > Line 900 is for SWIG - line 918 is the case for `xapian-bindings`.

Gosh, my `HEAD` seems to be *old*…

PR #271, opened 2019-07-10
Fix handler_poppler and avoid using clear_text in metadata

barufa:
There are some issues with handler_poppler. The main of them is that we 
can avoid calling the function clear_text for metadata. Beside, it is 
also possible to improve the error messages.

### DIFF
--- a/xapian-applications/omega/handler_poppler.cc
+++ b/xapian-applications/omega/handler_poppler.cc
@@ -18,8 +18,8 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
  * USA
  */
-#include <config.h>
 
+#include <config.h>
 #include "handler.h"
 #include "str.h"
 
@@ -76,7 +76,7 @@ extract(const string& filename,
 	    page *p(doc->create_page(i));
 	    if (!p) {
 		cerr << "Poppler Error: Failed to create page " << i << endl;
-		continue;
+		return false;
 	    }
 	    dump += text_to_utf8(p->text());
 	}
@@ -85,5 +85,6 @@ extract(const string& filename,
 	cerr << "Poppler threw an exception" << endl;
 	return false;
     }
+
     return true;
 }
--- a/xapian-applications/omega/handler_poppler.cc
+++ b/xapian-applications/omega/handler_poppler.cc
@@ -18,12 +18,9 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
  * USA
  */
-
-#include <config.h>
 #include "handler.h"
 #include "str.h"
 
-#include <xapian.h>
 #include <iostream>
 
 #include <poppler/cpp/poppler-document.h>
@@ -33,18 +30,21 @@ using namespace std;
 using namespace poppler;
 
 static string
-clear_text(const ustring& x)
+text_to_utf8(const ustring& x)
 {
     byte_array buf = x.to_utf8();
     string text(buf.data(), buf.size());
-    int i, j, sz = text.length();
+    return text;
+}
 
+static void
+clear_text(string& text)
+{
+    int i, j, sz = text.length();
     for (j = i = 0; i < sz; ++i)
 	if (!isspace(text[i]) || (i && !isspace(text[i - 1])))
 	    text[j++] = text[i];
     text.resize(j);
-
-    return text;
 }
 
 bool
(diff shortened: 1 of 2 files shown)


xapian-applications/omega/handler_poppler.cc
```diff
@@ -18,12 +18,9 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
  * USA
  */
-
-#include <config.h>
```
ojwb: This needs to stay - when using autotools, `<config.h>` should be included by every `.cc` file (and it should be the first header included).

xapian-applications/omega/handler_poppler.cc
```diff
@@ -63,25 +65,25 @@ extract(const string& filename,
...
+	int npages = doc->pages();
 	pages = str(npages);
 	// Extracting text from PDF file
 	for (int i = 0; i < npages; ++i) {
 	    page *p(doc->create_page(i));
 	    if (!p) {
-		cerr << "Poppler Error: Failed to create page" << endl;
-		return false;
+		cerr << "Poppler Error: Failed to create page " << i << endl;
+		continue;
```
ojwb: I'm not really clear about the change from `return false` to `continue` here - the commit message says "improve the error messages", but the actual changes also change the behaviour on error.  That makes me wonder if the commit message is incomplete, or the change to `continue` here an accidental one...

Under what circumstances might `create_page()` fail?

If (say) a page without text fails, then `continue` would definitely make sense.  If it's aiming to robustly handle a broken PDF file, then I guess that's OK, though there's also an argument for flagging up broken input.  Either way (or if this is for some other reason entirely) it'd be good to document the reason for this handling in a comment.

Also, what does pdftotext do in this situation?

xapian-applications/omega/handler_poppler.cc
```diff
@@ -18,12 +18,11 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
  * USA
  */
-
 #include <config.h>
+
```
ojwb: It's better to take care to avoid this sort of arbitrary whitespace change in patches (it can easily result from amending a patch, like here, or from not taking care with what's included in a patch in the first place).  It's a good habit to review the patch against the branch point before pushing (both initially and before pushing fixup commits).

It's distracting noise for the patch reviewers, but also makes it harder to look back through the history of the code - e.g. `git blame` would show this code as having changed.

xapian-applications/omega/handler_poppler.cc
barufa: It seems that `pdftotext` return an error if `create_page()` returns NULL.
Actually, the problem is that if only one page fails and `extract` returns false, we lose all the information (including the metadata). I was trying to avoid that.
Maybe I should make this change in another commit(or in a new PR with a proper name).

xapian-applications/omega/handler_poppler.cc
barufa: It appears with the remove/add of config.h. I didn't notice.
I will try to be more careful with this kind of thinks.

xapian-applications/omega/handler_poppler.cc
ojwb: > Actually, the problem is that if only one page fails and `extract` returns false, we lose all the information (including the metadata). I was trying to avoid that.

But under what circumstances can that actually happen?

The poppler API documentation doesn't say what can cause `create_page()` to return `NULL`.  The handling inside poppler doesn't make it totally easy to see, but it looks to me like the list of pages which the file says it has would need to be inconsistent with the pages which are actually present, which sounds like a bad PDF file to me.

If `pdftotext` errors out in this case then that seems a sensible thing for us to do too, unless we have evidence that doing something else is actually better (such as some PDF files where `create_page()` fails for some pages, but which appear to be functional PDF files which you can view in a PDF viewer, etc).  We've also been feeding PDF files to `pdftotext` for years, so if such files exist we will already have been rejecting them for years.

If we are going to change this, it should really be a separate commit (the commit here already mixes up several changes - "use POPPLER_CFLAGS", "call clear_text() less", "improve error message" should ideally each have been their own commit).

ojwb:
Rebased squashed and pushed to master as e0d9eb97edf623b39046bb83615f6fee627c4df5.

xapian-applications/omega/handler_poppler.cc
```diff
@@ -23,7 +23,6 @@
 #include "handler.h"
 #include "str.h"
 
-#include <xapian.h>
```
ojwb: If xapian-core isn't used here we should also drop `$(XAPIAN_LIBS)` from `omindex_pdf_LDADD` - I've pushed that as 1060cb6ba111ed174006d262c032ce6902269019.

PR #272, opened 2019-07-20
Fix bugs in letor

VaibhavKansagara:


VaibhavKansagara:
@mentors testcase for the second bug is not possible to construct.

ojwb:
"Fix simple bugs in letor" is rather a vague commit message.  It's also better to avoid fixing multiple unconnected bugs in one commit, and better to say what the bug actually fixed was.

So I'd have made two commits in place of the first here, with commit messages something like "Fix misdocumented default learning rate" and "Fix off-by-one counting of iterations".

xapian-letor/tests/api_letor.cc
```diff
@@ -652,6 +652,44 @@ DEFINE_TESTCASE(listnet_ranker_three_correct, generated)
...
+    TEST_EQUAL(ranker.get_database_path(), db_path);
+    ranker.set_query(Xapian::Query("score"));
+    ranker.train_model(training_data);
+    Xapian::docid doc1 = *mymset[0];
+    Xapian::docid doc2 = *mymset[1];
+    ranker.rank(mymset);
+    TEST_EQUAL(doc1, *mymset[1]);
+    TEST_EQUAL(doc2, *mymset[0]);
+    unlink("ndcg_output_listnet_3.txt");
+    ranker.score(query, qrel, "ListNet_Ranker", "ndcg_output_listnet_3.txt",
```
ojwb: It looks like before this patch we use unique output filenames, but this new testcase reuses the filename used by an existing testcase.

I think in practice that isn't currently a problem, but Tanmay is working on making the test harness run testcases in parallel and it would be good not to be added more obstacles to allowing that.

VaibhavKansagara: sure will change the filenames.

ojwb:
Thanks.  Rebase and autosquashed onto master as 011cb929a933a103946833ae19fe801bf72a6301 and the two preceding commits.

PR #273, opened 2019-07-20
Improve Worker Communication

barufa:
Currently, if the assistant fail for a non-fatal error, the process 
ends. We can improve the communication protocol by keeping the assistant 
process alive in these cases.

### DIFF
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -150,14 +150,14 @@ skip_unknown_mimetype(const string & urlterm, const string & context,
 }
 
 void
-index_add_default_libraries()
+index_add_default_libraries(bool vrbs)
 {
 #if defined HAVE_POPPLER
-    Worker* omindex_pdf = new Worker("omindex_pdf");
+    Worker* omindex_pdf = new Worker("omindex_pdf", vrbs);
     index_library("application/pdf", omindex_pdf);
 #endif
 #if defined HAVE_LIBEBOOK
-    Worker* omindex_ebook = new Worker("omindex_ebook");
+    Worker* omindex_ebook = new Worker("omindex_ebook", vrbs);
     index_library("application/vnd.palm", omindex_ebook);
     index_library("application/x-fictionbook+xml", omindex_ebook);
     index_library("application/x-zip-compressed-fb2", omindex_ebook);
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -150,14 +150,14 @@ skip_unknown_mimetype(const string & urlterm, const string & context,
 }
 
 void
-index_add_default_libraries(bool vrbs)
+index_add_default_libraries()
 {
 #if defined HAVE_POPPLER
-    Worker* omindex_pdf = new Worker("omindex_pdf", vrbs);
+    Worker* omindex_pdf = new Worker("omindex_pdf");
     index_library("application/pdf", omindex_pdf);
 #endif
 #if defined HAVE_LIBEBOOK
-    Worker* omindex_ebook = new Worker("omindex_ebook", vrbs);
+    Worker* omindex_ebook = new Worker("omindex_ebook");
     index_library("application/vnd.palm", omindex_ebook);
     index_library("application/x-fictionbook+xml", omindex_ebook);
     index_library("application/x-zip-compressed-fb2", omindex_ebook);
@@ -663,7 +663,9 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	    Worker* wrk = wrk_it->second;
 	    if (!wrk ||
 		!wrk->extract(file, dump, title, keywords, author, pages)) {
-		string msg = "Couldn't extract text from " + file;
+		string msg = wrk->get_error();
+		if (msg.empty())
+		    msg = "Couldn't extract text from " + file;
 		skip(urlterm, context, msg, d.get_size(), d.get_mtime());
 		return;
 	    }
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -38,6 +38,7 @@
 
 #include <sys/types.h>
 #include "safeunistd.h"
+#include <cassert>
 #include <cerrno>
 #include <cstdio>
 #include <cstdlib>
@@ -664,8 +665,7 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	    if (!wrk ||
 		!wrk->extract(file, dump, title, keywords, author, pages)) {
 		string msg = wrk->get_error();
-		if (msg.empty())
-		    msg = "Couldn't extract text from " + file;
+		assert(!msg.empty());
 		skip(urlterm, context, msg, d.get_size(), d.get_mtime());
 		return;
 	    }
--- a/xapian-applications/omega/omindex.cc
+++ b/xapian-applications/omega/omindex.cc
@@ -391,6 +391,9 @@ main(int argc, char **argv)
 
     map<string, string> mime_map;
 
+    index_add_default_filters();
+    index_add_default_libraries();
+
     if (argc == 2 && strcmp(argv[1], "-v") == 0) {
 	// -v was the short option for --version in 1.2.3 and earlier, but
 	// now it is short for --verbose (for consistency with scriptindex)
@@ -696,10 +699,6 @@ main(int argc, char **argv)
 	    return 1;
 	}
     }
-
-    // Indexing default filters and libraries
-    index_add_default_filters();
-    index_add_default_libraries(verbose);
 
     if (dbpath.empty()) {
 	cerr << PROG_NAME": you must specify a database with --db." << endl;
--- a/xapian-applications/omega/omindex.cc
+++ b/xapian-applications/omega/omindex.cc
@@ -391,9 +391,6 @@ main(int argc, char **argv)
 
     map<string, string> mime_map;
 
-    index_add_default_filters();
-    index_add_default_libraries();
-
     if (argc == 2 && strcmp(argv[1], "-v") == 0) {
 	// -v was the short option for --version in 1.2.3 and earlier, but
 	// now it is short for --verbose (for consistency with scriptindex)
@@ -699,6 +696,10 @@ main(int argc, char **argv)
 	    return 1;
 	}
     }
+
+    // Indexing default filters and libraries
+    index_add_default_filters();
+    index_add_default_libraries(verbose);
 
     if (dbpath.empty()) {
 	cerr << PROG_NAME": you must specify a database with --db." << endl;
--- a/xapian-applications/omega/worker.cc
+++ b/xapian-applications/omega/worker.cc
@@ -185,7 +185,7 @@ Worker::extract(const std::string& filename,
 		break;
 	}
     }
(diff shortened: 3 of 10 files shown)


xapian-applications/omega/index_file.cc
```diff
@@ -150,14 +150,14 @@ skip_unknown_mimetype(const string & urlterm, const string & context,
 }
 
 void
-index_add_default_libraries()
+index_add_default_libraries(bool vrbs)
```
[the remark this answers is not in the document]
ojwb: Also `vrbs` is rather cryptic - I didn't guess its meaning until I looked at the header.

It's not like `verbose` is an overly verbose variable name that benefits from shortening.

xapian-applications/omega/omindex.cc
```diff
@@ -391,9 +391,6 @@ main(int argc, char **argv)
 
     map<string, string> mime_map;
 
-    index_add_default_filters();
```
ojwb: The default filters need to be set up before the option parsing loop - otherwise the user can't override the default filters on the command line.

xapian-applications/omega/worker.cc
```diff
@@ -153,21 +154,36 @@ Worker::extract(const std::string& filename,
...
+		    !read_string(sockt, keywords) ||
+		    !read_string(sockt, author) ||
+		    !read_string(sockt, strpage)) {
+		    break;
+		}
+		if (strpage.empty() ||
+		    strpage.find_first_not_of("0123456789") != string::npos) {
+		    pages = 0;
+		} else {
+		    pages = stoi(strpage);
```
ojwb: I know this wasn't added by this commit, but this ought to be fixed - it's silly to use `stoi()` but first have to check what's being parsed consists only of digits to avoid `stoi()` throwing an exception - by doing this you're essentially parsing the number twice.

Better to use the Xapian-provided `parse_unsigned()` helper function instead (see `common/parseint.h`).

xapian-applications/omega/worker.h
```diff
@@ -54,8 +56,8 @@ class Worker {
      *
      *  The assistant will not be started until it is necessary.
      */
-    Worker(const std::string& path)
-	: sockt(NULL), filter_module(path) { }
+    Worker(const std::string& path, bool verb)
+	: verbose(verb), sockt(NULL), filter_module(path) { }
```
ojwb: Don't just arbitrarily abbreviate a parameter name like this - it's especially confusing here since `verb` is also an English word, but means something different to `verbose`.

The Xapian convention for a parameter which initialises a member like this is just to append `_` to the parameter name (so `verbose_` here).

xapian-applications/omega/worker.cc
```diff
@@ -153,21 +154,36 @@ Worker::extract(const std::string& filename,
...
+		}
+		return true;
+	    case MSG_NON_FATAL_ERROR:
+		if (strstate.length() > 1 && verbose) {
+		    cerr << (strstate.c_str() + 1) << endl;
```
ojwb: I don't think this is the right place to be reporting the error - it ought to be returned to the caller so they can report it in the appropriate way, rather than just chucking it at stderr and hoping that's OK.

The current approach won't work well if an assistant process has its own workers, as the errors from those will go to stderr of the assistant rather than stderr of omindex.

That also avoids the `Worker` needing to have a `verbose` flag.

xapian-applications/omega/worker.cc
barufa: Okay, I am going to change it

xapian-applications/omega/index_file.cc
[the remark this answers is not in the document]
barufa: I will try to avoid that kind of names

xapian-applications/omega/worker.cc
barufa: I will try to return the error to ha caller

xapian-applications/omega/index_file.cc
```diff
@@ -663,7 +663,9 @@ index_mimetype(const string & file, const string & urlterm, const string & url,
 	    Worker* wrk = wrk_it->second;
 	    if (!wrk ||
 		!wrk->extract(file, dump, title, keywords, author, pages)) {
-		string msg = "Couldn't extract text from " + file;
+		string msg = wrk->get_error();
+		if (msg.empty())
+		    msg = "Couldn't extract text from " + file;
```
ojwb: Your documentation for `get_error()` says that if it returns an empty string that means "no error", in which case this handling this case isn't necessary (you could reasonably `assert(!msg.empty());` here).

xapian-applications/omega/worker.cc
```diff
@@ -25,15 +25,16 @@
 #include "worker.h"
 #include "worker_comms.h"
 
-#include "pkglibbindir.h"
 #include <csignal>
 #include <cstring>
 #include <cerrno>
-#include <iostream>
-#include "safefcntl.h"
-#include "safeunistd.h"
```
ojwb: Each `"safeXXX.h"` header is essentially a compatibility stand-in for the respective `<XXX.h>` system header, which is why they were in this block of system headers.  That's the convention that's generally used in the code base, and I think it makes sense - it means that someone going to add (say) `<fcntl.h>` is more likely to notice that "safefcntl.h" is already there, and it means we don't have to moved header includes around in a lot of files if a new "safeXXX.h" is introduced, or an existing one eliminated (which would happen if the compatibility handling it was doing is no longer needed for any current platform).

xapian-applications/omega/worker.cc
```diff
@@ -25,15 +25,16 @@
...
 #include <sys/socket.h>
 #include <sys/wait.h>
+
+#include "parseint.h"
+#include "pkglibbindir.h"
+#include "safefcntl.h"
+#include "safeunistd.h"
 #include "closefrom.h"
 #include "freemem.h"
```
ojwb: As above, I don't think the `"safeXXX.h"` headers should be here.

Moving `"pkglibbindir.h"` to this block makes sense, and adding `"parseint.h"` here is good too, but it would be better to maintain the sorted order of the headers in this block of headers.

xapian-applications/omega/worker.cc
```diff
@@ -169,22 +171,19 @@ Worker::extract(const std::string& filename,
...
 		    pages = 0;
-		} else {
-		    pages = stoi(strpage);
-		}
 		return true;
 	    case MSG_NON_FATAL_ERROR:
-		if (strstate.length() > 1 && verbose) {
-		    cerr << (strstate.c_str() + 1) << endl;
+		if (strstate.length() > 1) {
+		    error = (strstate.c_str() + 1);
```
ojwb: This effectively converts the C++ string to a C string and back, which is probably better avoided - the conversion back will take the length by searching for the first zero byte, which means a string with embedded zero bytes will be truncated, and is also a little less efficient than just taking the known length of the C++ string.  The truncation issue isn't likely to be a problem in practice in this situation, but still seems a bit unclean.

You can assign a substring of a C++ string like this:

```
error.assign(strstate, 1, string::npos)
```

(The third argument isn't needed from C++14 on, but currently we aim to support C++11).

xapian-applications/omega/worker.cc
```diff
@@ -169,22 +171,19 @@ Worker::extract(const std::string& filename,
...
 		}
 		return false;
 	    default:
 		break;
 	}
     }
+    error = "The assistant process " + filter_module + " fails";
```
ojwb: The present tense sounds wrong for an error message - the event is in the past by the time the error message is read, so "X failed" sounds much more natural than "X fails".

xapian-applications/omega/worker.cc
```diff
@@ -195,3 +194,9 @@ Worker::extract(const std::string& filename,
 
     return false;
 }
+
+string
+Worker::get_error()
+{
+    return error;
+}
```
ojwb: This should really be a `const` method, since it just inspects state and doesn't change the object it's called on.

Also, better to define a trivial getter like this inline in the class definition - that way the compiler can inline the method call and optimise it to just a string copy in the calling code.

xapian-applications/omega/worker.h
```diff
@@ -84,4 +83,8 @@ class Worker {
 		 std::string& keywords,
 		 std::string& author,
 		 int& pages);
+    /** Returns and error message if the extraction fails, or and empty string
+     *  if everything is okay.
```
ojwb: In both cases, "and" should be "an".

xapian-applications/omega/omindex.cc
barufa: Sorry, I didn't realize.

xapian-applications/omega/worker.cc
barufa: Okay, I am going to change it.

xapian-applications/omega/worker.cc
barufa: Okay, I will fix it.

xapian-applications/omega/worker.h
barufa: Sorry, I will change it.

xapian-applications/omega/index_file.cc
barufa: I am going to fix it.

xapian-applications/omega/worker.h
```diff
@@ -83,8 +83,10 @@ class Worker {
...
 		 std::string& author,
 		 int& pages);
-    /** Returns and error message if the extraction fails, or and empty string
+    /** Returns an error message if the extraction fails, or an empty string
      *  if everything is okay.
      */
-    std::string get_error();
+    inline std::string get_error() const{
+	return error;
+    }
```
ojwb: You don't need to say `inline` here - a method defined within a class like this is implicitly `inline`.

Also, there's a missing space between `const` and `{`.

xapian-applications/omega/worker.cc
```diff
@@ -176,14 +176,16 @@ Worker::extract(const std::string& filename,
...
+		} else {
+		    error = "Couldn't extract text from " + filename;
 		}
 		return false;
 	    default:
 		break;
 	}
     }
-    error = "The assistant process " + filter_module + " fails";
+    error = "Fatal Error: The assistant process " + filter_module + " failed";
```
ojwb: The error is fatal for the assistant, not for the main omindex process.

If we add "Fatal Error" here, when omindex reports the error message to the user it'll look confusing because omindex is continuing to run despite having reported a fatal error.

barufa: Okay, I will remove that.

xapian-applications/omega/worker.h
barufa: I will remove that.

xapian-applications/omega/worker_comms.h
```diff
@@ -22,6 +22,11 @@
 #include <cstdio>
 #include <string>
 
+enum assistant_msg {
+    MSG_OK,
+    MSG_NON_FATAL_ERROR,
+    MSG_FATAL_ERROR
+};
 /// Read a string from the file descriptor @a f and storage it in @a s
```
ojwb: I've added a blank like after the enum definition above (I only noticed this during merging)

ojwb:
Merged to master as 64dfcb1baa5e7604d730e2e2530e713664bc6724.

PR #274, opened 2019-07-24
Add a check for determining unique query ids

VaibhavKansagara:


xapian-letor/ranker/ranker.cc
```diff
@@ -276,6 +277,10 @@ Xapian::prepare_training_file(const string & db_path, const string & queryfile,
 	std::pair<string, string> parsed_query = parse_query_string(str1, query_count);
 	string querystr = parsed_query.first;
 	string qid = parsed_query.second;
+	if (queries.find(qid) != queries.end()) {
+	    throw Xapian::LetorParseError("Query id should be unique");
+	}
+	queries.insert(qid);
```
ojwb: Here the usual case is that `qid` isn't already present.

In that case, `find(qid)` will search the set (which is typically a red-black tree) to find there's no entry.  Then `insert(qid)` repeats essentially the same search to find where to add the new entry.  The overall work isl O(log(n)) for each so still O(log(n)) in total, but it's needlessly repeated work.

Instead we can just call `insert(qid)` and check the return value to see if a value was actually inserted - if not, there was a value there already - the return value for this form is `std::pair<iterator, bool>` where the `bool` is `true` if insertion took place, so we can insert and check in one go with:

```
if (!queries.insert(qid).second) {
    // throw exception
}
```

See https://en.cppreference.com/w/cpp/container/set/insert for details of the return value, etc.

xapian-letor/tests/api_letor.cc
```diff
@@ -359,6 +359,18 @@ DEFINE_TESTCASE(preparetrainingfileonedb, generated)
     return true;
 }
 
+// test whether query id's are unique in queryfile.
```
ojwb: "query ids" is generally preferred by English style guides.

ojwb:
The two CI failures are inherited from master (they're due to `append()` in Ruby code not working in Ruby < 2.5) so they're not related to the changes here.

xapian-letor/ranker/ranker.cc

xapian-letor/tests/api_letor.cc

ojwb:
You shouldn't need to rebase and force-push to pick up a fix from master - travis-ci runs on your branch merged with current master, so the rebuild caused by pushing changes to your branch will automatically pick up such fixes.

ojwb:
Rebased and autosquashed onto master as 5507a80512227f2ca40cc852ad28a7ccd161a1fc.

PR #275, opened 2019-07-31
Fix rfc822tohtml entry in index_file.cc

barufa:
It seems that there is an error with ‘rfc822tohtml’. This script takes a 
file name as an argument, but we are indexing it as if it reads the file 
from the standard input.

### DIFF
--- a/xapian-applications/omega/index_file.cc
+++ b/xapian-applications/omega/index_file.cc
@@ -247,11 +247,9 @@ index_add_default_filters()
 		  Filter(get_pkglibbindir() + "/mhtml2html", "text/html",
 			 PIPE_DEV_STDIN));
     index_command("message/news",
-		  Filter(get_pkglibbindir() + "/rfc822tohtml", "text/html",
-			 PIPE_DEV_STDIN));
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for spotting this, and for the patch.

Your fix works, but I don't think this is the best fix since the `rfc822tohtml` script does support specifying `/dev/stdin` as the filename and piped input (which is what `PIPE_DEV_STDIN` is meant to mean):

```
$ cat /home/olly/git/xapian/xapian-applications/omega/test-data/message.rfc822 | ./rfc822tohtml /dev/stdin | head -n6
<head>
<title>apt-listchanges: news for gemse</title>
<meta name="author" content="root &lt;root@gemse&gt;">
<meta name="created" content="2015-11-29T18:57:04">
</head>
<pre>apt (1.1~exp9) experimental; urgency=medium
```

I think the bug here is actually that `PIPE_DEV_STDIN` isn't handled correctly unless there's an explicit `%f` on the command line.  We should append ` /dev/stdin` to the command in this case, but the code to do that is incorrectly gated.  I'll push an alternative fix which addresses that.

PR #276, opened 2019-07-31
Add libetonyek to Omega

barufa:
At the moment, Omega uses pages2txt and similar filters to extract
information from iWork documents. We can replace them with libetonyek.

### DIFF
--- a/xapian-applications/omega/Makefile.am
+++ b/xapian-applications/omega/Makefile.am
@@ -44,7 +44,7 @@ BUILT_SOURCES = extra/omegascript.vim my-html-tok.h mimemap.h namedents.h
 
 MAINTAINERCLEANFILES = my-html-tok.h mimemap.h mimemap.stamp namedents.h
 
-EXTRA_PROGRAMS = omindex_pdf omindex_ebook
+EXTRA_PROGRAMS = omindex_pdf omindex_ebook omindex_idocs
 
 EXTRA_DIST =\
 	clickmodel/testdata/test1.log\
@@ -184,6 +184,10 @@ omindex_pdf_CXXFLAGS = $(POPPLER_CFLAGS)
 omindex_ebook_SOURCES = assistant.cc worker_comms.cc handler_libebook.cc
 omindex_ebook_LDADD = $(LIBEBOOK_LIBS)
 omindex_ebook_CXXFLAGS = $(LIBEBOOK_CFLAGS) $(XAPIAN_CXXFLAGS)
+
+omindex_idocs_SOURCES = assistant.cc worker_comms.cc handler_libetonyek.cc
+omindex_idocs_LDADD = $(LIBETONYEK_LIBS)
+omindex_idocs_CXXFLAGS = $(LIBETONYEK_CFLAGS) $(XAPIAN_CXXFLAGS)
 
 scriptindex_SOURCES = scriptindex.cc myhtmlparse.cc htmlparse.cc\
  common/getopt.cc common/str.cc commonhelp.cc utils.cc hashterm.cc loadfile.cc\
--- a/xapian-applications/omega/Makefile.am
+++ b/xapian-applications/omega/Makefile.am
@@ -44,7 +44,7 @@ BUILT_SOURCES = extra/omegascript.vim my-html-tok.h mimemap.h namedents.h
 
 MAINTAINERCLEANFILES = my-html-tok.h mimemap.h mimemap.stamp namedents.h
 
-EXTRA_PROGRAMS = omindex_pdf omindex_ebook omindex_libetonyek
+EXTRA_PROGRAMS = omindex_poppler omindex_libebook omindex_libetonyek
 
 EXTRA_DIST =\
 	clickmodel/testdata/test1.log\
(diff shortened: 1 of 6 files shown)


xapian-applications/omega/Makefile.am
```diff
@@ -44,7 +44,7 @@ BUILT_SOURCES = extra/omegascript.vim my-html-tok.h mimemap.h namedents.h
 
 MAINTAINERCLEANFILES = my-html-tok.h mimemap.h mimemap.stamp namedents.h
 
-EXTRA_PROGRAMS = omindex_pdf omindex_ebook
+EXTRA_PROGRAMS = omindex_pdf omindex_ebook omindex_idocs
```
ojwb: I think we want to name these helper binaries after the **library** not the **format** (I'm fairly sure I mentioned this re `omindex_pdf` which I think would be better named `omindex_poppler`, but it seems that never got addressed).

If we name the helper after the format it's harder to cleanly support two helpers which can handle the same format.  E.g. `omindex_wv2` and `omindex_libreofficekit` rather than `omindex_msword` and being stuck for a good way to name the other one or clearly tell them apart.  We really want these helpers to have a stable name so that we can provide a way for users to configure them at run time, and the most stable name is the name of the library they use.

This also works better for handlers which handle more than one format, which is quite common.

xapian-applications/omega/handler_libetonyek.cc
```diff
@@ -0,0 +1,233 @@
...
+    RVNGString text_dump, data_dump;
+    bool succeed = false;
+
+    // Extract text if possible
+    RVNGTextTextGenerator content(text_dump, false);
+    if (EtonyekDocument::parse(input, &content)) {
+	clear_text(dump, text_dump.cstr());
+	succeed = true;
+    } else {
+	 error = "Libetonyek Error: Fail to extract text";
```
ojwb: There's an extra space in the indent here.

barufa: Sorry, I didn't see it.

xapian-applications/omega/Makefile.am
barufa: Yes, I remember that. Actually, it seems that I misunderstand that and only change the name of the file (handler_pdf.cc -> handler_poppler.cc).
I am going to fix it.

barufa:
I am not sure if the last commit belongs to this PR. Should I add it in a new PR?

ojwb:
> I am not sure if the last commit belongs to this PR. Should I add it in a new PR?

It's OK here - it's in a separate commit at least.

xapian-applications/omega/handler_libetonyek.cc
```diff
@@ -0,0 +1,233 @@
...
+#include <librevenge-stream/librevenge-stream.h>
+#include <libetonyek/libetonyek.h>
+
+#define PARSE_FIELD(START, END, FIELD, OUT) \
+   parse_metadata_field((START), (END), (FIELD), (CONST_STRLEN(FIELD)), (OUT))
+
+using std::shared_ptr;
+using libetonyek::EtonyekDocument;
+using namespace librevenge;
+using namespace std;
```
ojwb: Isn't `using std::shared_ptr;` redundant given we then do `using namespace std;`?

ojwb: Also, we don't actually seem to use `shared_ptr` anywhere anyway.

ojwb:
Looks good otherwise.

xapian-applications/omega/handler_libetonyek.cc
```diff
@@ -0,0 +1,233 @@
...
+    }
+    // Extract metadata if possible
+    RVNGTextTextGenerator data(data_dump, true);
+    if (EtonyekDocument::parse(input, &data)) {
+	const char* metadata = data_dump.cstr();
+	size_t len = data_dump.size();
+	parse_metadata(metadata, len, author, title, keywords);
+	succeed = true;
+    } else {
+	if (!error.empty())
```
guruhegde: In the case where text extract - success and metadata extract - fail, function `extract_pages` returns `true`. 
Is that expected behavior?  Although `error` string won't be empty.

xapian-applications/omega/handler_libetonyek.cc
barufa: Sorry, I am going to remove that.

xapian-applications/omega/handler_libetonyek.cc
barufa: Yes, I think that this way the handler is a bit more robust.
Sorry, I forget to clean the `error` string.

Is it okay?

guruhegde: It was ok to keep the `error` string I guess, I was trying to understand the invariant of the function.
This is a minor thing, see if any change necessary w.r.t my comment:

If (text extract fail && metadata extract fail) : `succeed` = `false`, `error` = `non-empty`
All other cases:  `succeed = `true`, error = `empty` (both op success) or `non-empty` (that means one of the op failed, in that case keeping the error indicates that behavior).

By adding `error.clear()` function returns `succeed` = `true`, error = `non-empty`, only when metadata op fail.

barufa: Sorry, you are right. If `succeed` is `true`, then `error` should be `empty`.

ojwb:
Thanks - rebased, autosquashed and pushed to master.

PR #277, opened 2019-08-07
Add Tesseract to Omega

barufa:
Currently, omindex cannot handle image which contain text. It is 
possible to use an OCR(like tesseract) to extract text from images and 
index them.

### DIFF
--- a/xapian-applications/omega/docs/overview.rst
+++ b/xapian-applications/omega/docs/overview.rst
@@ -315,7 +315,8 @@ other filters too - see below):
 * TCR (simple compressed text format) files (.tcr) if libe-book is available
 * eReader files (.pdb) if libe-book is available
 * Sony eBook files (.lrf) if libe-book is available
-* Image files that contain text (.png, .jpg) if libtesseract-dev is available.
+* Image JPEG files that contain text (.png, .jpg, .jpeg, .jfif, .jpe, .webq, .tif, .tiff,
+  .pbm, .gif, .ppm, .pgm) if libtesseract-dev is available.
 
 If you have additional extensions that represent one of these types, you can
 add an additional MIME mapping using the ``--mime-type`` option.  For
--- a/xapian-applications/omega/docs/overview.rst
+++ b/xapian-applications/omega/docs/overview.rst
@@ -315,6 +315,7 @@ other filters too - see below):
 * TCR (simple compressed text format) files (.tcr) if libe-book is available
 * eReader files (.pdb) if libe-book is available
 * Sony eBook files (.lrf) if libe-book is available
+* Image files that contain text (.png, .jpg) if libtesseract-dev is available.
 
 If you have additional extensions that represent one of these types, you can
 add an additional MIME mapping using the ``--mime-type`` option.  For
--- a/xapian-applications/omega/docs/overview.rst
+++ b/xapian-applications/omega/docs/overview.rst
@@ -315,7 +315,7 @@ other filters too - see below):
 * TCR (simple compressed text format) files (.tcr) if libe-book is available
 * eReader files (.pdb) if libe-book is available
 * Sony eBook files (.lrf) if libe-book is available
-* Image JPEG files that contain text (.png, .jpg, .jpeg, .jfif, .jpe, .webq, .tif, .tiff,
+* Image files that contain text (.png, .jpg, .jpeg, .jfif, .jpe, .webp, .tif, .tiff,
   .pbm, .gif, .ppm, .pgm) if libtesseract-dev is available.
 
 If you have additional extensions that represent one of these types, you can
--- a/xapian-applications/omega/handler_tesseract.cc
+++ b/xapian-applications/omega/handler_tesseract.cc
@@ -29,20 +29,16 @@ using namespace tesseract;
 
 static TessBaseAPI* ocr = NULL;
 
-static bool
+static void
 clear_text(string& dump, const char* text)
 {
-    bool k = false;
     if (text) {
 	for (int i = 0; text[i] != '\0'; ++i) {
-	    if (!isspace(text[i]) || (i && !isspace(text[i - 1]))) {
+	    if (!isspace(text[i]) || (i && !isspace(text[i - 1])))
 		dump.push_back(text[i]);
-		k |= !isspace(text[i]);
-	    }
 	}
 	delete [] text;
     }
-    return k;
 }
 
 bool
@@ -65,22 +61,17 @@ extract(const string& filename,
     Pix* image = pixRead(filename.c_str());
 
     if (!image) {
-	error = "Tesseract Error: Error while openning the image";
+	error = "Tesseract Error: Error while opening the image";
 	return false;
     }
 
     ocr->SetImage(image);
 
     // Get OCR result
-    bool has_text = clear_text(dump, ocr->GetUTF8Text());
+    clear_text(dump, ocr->GetUTF8Text());
 
     // Destroy used object and release memory
     pixDestroy(&image);
-
-    if (!has_text) {
-	error = "Tesseract Error: The image does not content text";
-	return false;
-    }
 
     (void)title;
     (void)keywords;
--- a/xapian-applications/omega/handler_tesseract.cc
+++ b/xapian-applications/omega/handler_tesseract.cc
@@ -1,0 +1,85 @@
+/** @file handler_tesseract.cc
+ * @brief Extract text from Images using tesseract.
+ */
+/* Copyright (C) 2019 Bruno Baruffaldi
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+#include <config.h>
+#include "handler.h"
+
+#include <tesseract/baseapi.h>
+#include <leptonica/allheaders.h>
+
+using namespace std;
+using namespace tesseract;
+
+TessBaseAPI* ocr = NULL;
+
+static bool
+clear_text(string& dump, const char* text)
+{
+    bool k = false;
+    if (text) {
+	for (int i = 0; text[i] != '\0'; ++i) {
+	    if (!isspace(text[i]) || (i && !isspace(text[i - 1]))) {
+		dump.push_back(text[i]);
+		k |= isalnum(text[i]);
+	    }
+	}
+	delete [] text;
+    }
+    return k;
+}
+
+bool
+extract(const string& filename,
+	string& dump,
+	string& title,
+	string& keywords,
+	string& author,
+	string& pages,
+	string& error)
+{
+    // Create the ocr if necessary
+    if (!ocr) {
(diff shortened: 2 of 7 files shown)


xapian-applications/omega/handler_tesseract.cc
```diff
@@ -0,0 +1,85 @@
...
+{
+    // Create the ocr if necessary
+    if (!ocr) {
+	ocr = new TessBaseAPI();
+	if (ocr->Init(NULL, NULL))
+	    _Exit(1);
+	ocr->SetPageSegMode(PSM_AUTO_OSD);
+    }
+    // Open Image
+    Pix* image = pixRead(filename.c_str());
```
ojwb: This apparently returns `NULL` on error - shouldn't we check for that?

xapian-applications/omega/handler_tesseract.cc
```diff
@@ -0,0 +1,85 @@
...
+    ocr->SetImage(image);
+
+    // Get OCR result
+    bool has_text = clear_text(dump, ocr->GetUTF8Text());
+
+    // Destroy used object and release memory
+    pixDestroy(&image);
+
+    if (!has_text) {
+	error = "Tesseract Error: The image does not content text\n";
```
ojwb: The error messages returned by other workers don't include a trailing newline, so this shouldn't either.  Otherwise the caller would either have to check if there's a newline and add/remove one depending on whether it wants one or not...

xapian-applications/omega/index_file.cc
```diff
@@ -172,6 +172,11 @@ index_add_default_libraries()
     index_library("application/vnd.apple.pages", omindex_libetonyek);
     index_library("application/vnd.apple.numbers", omindex_libetonyek);
 #endif
+#if defined HAVE_TESSERACT
+    Worker* omindex_tesseract = new Worker("omindex_tesseract");
+    index_library("image/jpeg", omindex_tesseract);
+    index_library("image/png", omindex_tesseract);
```
ojwb: It looks like there are other common image formats supported, though expanding this list could be handled as a follow-on change.

Also, we should have mappings from the extensions used for these formats (`jpeg`, `jpg`, and `png`) to the MIME content-types in `mimemap.tokens` so we don't have to call libmagic to find out what type they are.

xapian-applications/omega/handler_tesseract.cc
```diff
@@ -0,0 +1,85 @@
...
+
+static bool
+clear_text(string& dump, const char* text)
+{
+    bool k = false;
+    if (text) {
+	for (int i = 0; text[i] != '\0'; ++i) {
+	    if (!isspace(text[i]) || (i && !isspace(text[i - 1]))) {
+		dump.push_back(text[i]);
+		k |= isalnum(text[i]);
```
ojwb: This will only fire for ASCII alphanumerics, and so will work badly for text in other alphabets.

xapian-applications/omega/handler_tesseract.cc
```diff
@@ -0,0 +1,85 @@
...
+#include <config.h>
+#include "handler.h"
+
+#include <tesseract/baseapi.h>
+#include <leptonica/allheaders.h>
+
+using namespace std;
+using namespace tesseract;
+
+TessBaseAPI* ocr = NULL;
```
ojwb: It looks like this isn't needed outside this file, so make it `static` - that's cleaner, and may also allow the compiler to optimise better (for example, it can then see that `ocr` can't be modified by a call to a function in a different source file).

barufa: Okay, I am going to fix that.

xapian-applications/omega/handler_tesseract.cc
barufa: Sorry, it shouldn't include a trailing newline. I will remove it.

xapian-applications/omega/index_file.cc
barufa: Yes, [here](https://tpgit.github.io/UnOfficialLeptDocs/leptonica/README.html#supported-image-file-formats) is the "list" of supported formats. I am going to add them to `mimemap.tokens`

xapian-applications/omega/handler_tesseract.cc
barufa: Yes, I didn't notice that. I am going to check that.

xapian-applications/omega/handler_tesseract.cc
barufa: Okay, I didn't realize of that.
I am going to change it for `!isspace`, but I am not sure is that is a good solution.
Here I am trying to detect if tesseract was able to extract some useful text and not just a bunch of spaces and punctuation characters without sense. Maybe we can use [isalnum](https://en.cppreference.com/w/cpp/locale/isalnum) or [iswalnum](https://en.cppreference.com/w/cpp/string/wide/iswalnum).

xapian-applications/omega/handler_tesseract.cc
```diff
@@ -0,0 +1,91 @@
...
+
+    ocr->SetImage(image);
+
+    // Get OCR result
+    bool has_text = clear_text(dump, ocr->GetUTF8Text());
+
+    // Destroy used object and release memory
+    pixDestroy(&image);
+
+    if (!has_text) {
```
AyushP123: I have seen tesseract give an empty output on images containing strings. Its not necessary that
 `no tesseract output == no text in the image`

xapian-applications/omega/handler_tesseract.cc
ojwb: The text will be UTF-8 not wide characters, so `iswalnum()` isn't directly helpful.

I don't think you need to worry about an "all whitespace" document here - omindex will notice that without your help and handle it appropriately.

If tesseract commonly returns some non-alphanumeric junk text for certain images then that would be good to handle, but I think it might be better not to try to overthink this up front, but instead get this working without such special handling and then test it on a variety of images which aren't scans and see what you get.  I tested on a handful here, and I seem to get either SPACE LINEFEED FORMFEED, or something which contains at least one ASCII letter.

xapian-applications/omega/mimemap.tokens
```diff
@@ -199,6 +199,20 @@ lrf		application/x-sony-bbeb
...
+jpe		image/jpeg
+jpeg		image/jpeg
+jpg		image/jpeg
+pbm		image/x-pbm
+pgm		image/x-pgm
+png		image/png
+ppm		image/x-portable-pixmap
+tif		image/tiff
+tiff		image/tiff
+webq		image/webp
```
ojwb: Shouldn't that be `webp`?

xapian-applications/omega/mimemap.tokens
```diff
@@ -199,6 +199,20 @@ lrf		application/x-sony-bbeb
...
 pdb		application/vnd.palm
 
+# Image files:
+gif		image/gif
+jfif		image/jpeg
+jpe		image/jpeg
+jpeg		image/jpeg
+jpg		image/jpeg
+pbm		image/x-pbm
+pgm		image/x-pgm
```
ojwb: libmagic seems to give `image/x-portable-pixmap` and `image/x-portable-greymap` (and that's what https://en.wikipedia.org/wiki/Netpbm_format lists too).  It makes sense to align with libmagic since then we'll correctly handle a file which doesn't have an extension listed here but which libmagic identifies (for example `.pnm` means "one of PBM, PGM or PPM").

Also, perhaps we should have an entry here for `pnm`, since we just pass all these formats to tesseract and don't actually care which of them it is internally.

xapian-applications/omega/index_file.cc
ojwb: You've added all the listed entries from there except BMP - was there a reason to exclude that?

(I get `image/x-ms-bmp` from libmagic for it.)

xapian-applications/omega/handler_tesseract.cc
ojwb: Are you suggesting that the message is potentially misleading?

That's a valid point, although actually I don't think this case should be an error - it would be better to just return an empty string for dump even if we extracted no text, and then omindex's handling for empty documents will kick in:

```
  -e, --empty-docs=ARG      how to handle documents we extract no text from:
                            ARG can be index, warn (issue a diagnostic and
                            index), or skip.  (default: warn)
```

It can be useful to index documents that no text is extracted from since they can be searched for by their metadata still (timestamps, owner, filename, etc).

xapian-applications/omega/docs/overview.rst
```diff
@@ -315,7 +315,8 @@ other filters too - see below):
 * TCR (simple compressed text format) files (.tcr) if libe-book is available
 * eReader files (.pdb) if libe-book is available
 * Sony eBook files (.lrf) if libe-book is available
-* Image files that contain text (.png, .jpg) if libtesseract-dev is available.
+* Image JPEG files that contain text (.png, .jpg, .jpeg, .jfif, .jpe, .webq, .tif, .tiff,
+  .pbm, .gif, .ppm, .pgm) if libtesseract-dev is available.
```
ojwb: The change from "Image files" to "Image JPEG files" doesn't seem helpful.

Also, `.webq` rather than `.webp` here - is that a typo?

xapian-applications/omega/handler_tesseract.cc
```diff
@@ -63,6 +63,12 @@ extract(const string& filename,
     }
     // Open Image
     Pix* image = pixRead(filename.c_str());
+
+    if (!image) {
+	error = "Tesseract Error: Error while openning the image";
```
ojwb: Typo: "opening"

Sorry about the English language...

xapian-applications/omega/mimemap.tokens
barufa: Yes, I am going to change it

xapian-applications/omega/handler_tesseract.cc
barufa: Okay, I will remove this check.

xapian-applications/omega/mimemap.tokens
barufa: Yes, I will change that

xapian-applications/omega/handler_tesseract.cc
barufa: I am going to change it.

xapian-applications/omega/docs/overview.rst
barufa: Yes, it is a mistake.

xapian-applications/omega/mimemap.tokens
barufa: Actually, I run `file --mime-type` and it returns `application/octet-stream` for these files but I am going to change the mimetypes. It is a good idea(and actually necessary) to get some kind of compatibility with libmagic.

jaylett: What `file --mime-type` says will depend on which version of `file(1)` you're running. `application/octet-stream` just means "it's not text but we don't know what it is".

ojwb: For me, running Debian unstable (which should have a recent version of `file`/libmagic) and looking at all the files with one of these extensions on my machine:

```
$ locate -0 '*.p[bgnp]m' | xargs -0 file --mime-type | awk '{print $2}' | sort | uniq -c
      2 application/octet-stream
      1 application/x-pie-executable
     52 image/x-portable-bitmap
     91 image/x-portable-greymap
    108 image/x-portable-pixmap
      1 inode/x-empty
      1 text/plain
```

The `application/x-pie-executable` is `/usr/lib/grass76/bin/r.out.ppm`, which seems to be a helper program to export PPMs from grass (a GIS), not an image file at all.

The `inode/x-empty` is an empty temporary file something left behind.

The `text/plain` and one `application/octet-stream` are deliberately broken example image files in a test suite in some source code I have lying around.

So for me, all the files on my machine are reported as one of the types, except one - looking at the contents that seems to have the end of the file padded with a load of zero bytes.  It displays OK in an image viewer though.

ojwb:
Looks good - rebased, squashed and pushed to master as bd3ee9a7d3bb2631fd0aa170df28b0f4ef925fd4.

PR #278, opened 2019-08-11
Fix implementation of rankers and indexing errors

VaibhavKansagara:


xapian-letor/api/featurelist.cc
```diff
@@ -132,6 +132,8 @@ FeatureList::create_feature_vectors(const Xapian::MSet & mset,
 	double wt = i.get_weight();
 	// Weight is added as a feature by default.
 	fvals.push_back(wt);
+	// Add a bias term.
+	fvals.push_back(1.0);
```
jaylett: I'm not at all sure I understand why a constant element in the feature vector is going to help any ranker, but even if it helps _some_ for some reason (such as an initialiser) I don't see why it would apply to _all_ rankers. This probably needs a bit more documentation to at least explain what's going on, and why it's always appropriate to add this element in.

This could and probably should have been a separate commit, as well.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -193,7 +195,7 @@ class XAPIAN_VISIBILITY_DEFAULT Ranker : public Xapian::Internal::intrusive_base
   protected:
     /// Method to train the model. Overridden in ranker subclass.
     virtual void
-    train(const std::vector<Xapian::FeatureVector> & training_data) = 0;
+    train(const vector<vector<Xapian::FeatureVector>> & training_data) = 0;
```
jaylett: This (and all the related changes) is under-described in the commit message. Before, we were dealing with a vector of `FeatureVector` objects, ie one `FeatureVector` per document in the input MSet. I _think_ (by reading the changes) that we now have this once per query in the training set?

If this is what is going on, then we can choose better variable names in the training code, write some slightly better docstrings, and probably also more generously comment the implementations. For algorithm implementations, it can be helpful to explain quite a lot of the code, linking back to specific bits of the paper by section or formula if that helps make it clear what's going on. (The fact that you've had to fix the rankers in this way suggests we could really do with more clarity around what the code is doing and how that is a correct implementation!)

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -126,21 +127,27 @@ batch_learning(vector<FeatureVector> feature_vectors,
...
-    // Initialize the parameters for neural network
+    // initialize the parameters for neural network with Xavier initialization.
+    // construct a trivial random generator engine from a time-based seed:
+    std::default_random_engine generator(469382313);
+    std::normal_distribution<double> distribution(0.0, sqrt(2.0 /
+							   (1 + feature_cnt)));
     vector<double> new_parameters;
     for (int feature_num = 0; feature_num < feature_cnt; ++feature_num) {
-	new_parameters.push_back(0.0);
+	new_parameters.push_back(distribution(generator));
```
jaylett: The Xavier initialisation could have been a separate commit.

xapian-letor/tests/api_letor.cc
```diff
@@ -220,8 +220,8 @@ DEFINE_TESTCASE(createfeaturevector, generated)
     TEST_EQUAL(mset.size(), 2);
     auto fv = fl.create_feature_vectors(mset, Xapian::Query("lions"), db);
     TEST_EQUAL(fv.size(), 2);
-    TEST_EQUAL(fv[0].get_fcount(), 19);
-    TEST_EQUAL(fv[1].get_fcount(), 19);
+    TEST_EQUAL(fv[0].get_fcount(), 20);
+    TEST_EQUAL(fv[1].get_fcount(), 20);
```
jaylett: These test changes should really be in the commit that made them necessary to maintain passing tests.

xapian-letor/api/featurelist.cc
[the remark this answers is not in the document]
AyushP123: Logic for adding bias
1. [link1](https://www.geeksforgeeks.org/effect-of-bias-in-neural-network/)
2. [link2](https://stackoverflow.com/questions/2480650/role-of-bias-in-neural-networks)

Although I would suggest that we refer to a more appropriate document for explaining the role of bias in a neural network.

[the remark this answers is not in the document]
jaylett: Link 1 says that bias is applied to the output…not to the input. (However link 2 talks about it in the middle of a process, so maybe that's what is happening here.)

If bias is interesting (I'm not doubting that it is, I just don't understand how adding a constant term to our feature vector acts as bias — this is probably my ignorance of the underlying maths of ListNet & ListMLE), why don't we make a `ConstantFeature(value=1.0)` and have that (a) in the default feature list, (b) available as an option for anyone customising their feature list?

xapian-letor/include/xapian-letor/ranker.h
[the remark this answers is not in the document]
ojwb: Yes, that's right - we're now aiming to follow the modern C++ convention of putting `*` and `&` with the type (so `const vector<vector<Xapian::FeatureVector>>&` here) with a space between that and a variable/function name that follows.

A lot of the existing code doesn't follow this, as it predates us adopting this, but please follow it for new and changed code so at least the situation doesn't get worse.  At some point we'll do an automated fix of the remaining cases.

(We used to follow typical C-style which is to write `char *p` or `char * p`, and used the obvious extension to reference types, which C itself doesn't have).

The patch checking script will now spot many cases of this, but maybe not this one (it needs to avoid flagging when the `&` is a bit-wise and operation as that should have a space either side).

ojwb:
The clang+libc++ build on Linux and the macOS build (which also uses clang+libc++) are both failing letor tests in travis-ci.  I'm guessing this means there's some undefined behaviour being triggered somewhere, which happens to behave differently for GCC+libstdc++ vs clang+libc++.

I'd try a build with asan (which is supported by both GCC and clang) and see if that flushes out any issues.

xapian-letor/docs/letor.rst
```diff
@@ -68,7 +68,6 @@ As mentioned before, this process requires a training file in the above format.
 Provided such information, API is capable of creating the training file which is in the mentioned format and can be easily used for learning a model. In xapian-letor we support the following learning algorithms:
 
     1. `ListNET <https://dl.acm.org/citation.cfm?id=1273513>`_
-    2. `Ranking-SVM <https://dl.acm.org/citation.cfm?id=775067>`_
```
ojwb: There's also documentation about libsvm in `xapian-letor/INSTALL` which should go as part of this commit.

It's a good idea when stripping out something like this to do something like `git grep -i svm` from the top of the source tree to see if there are any other lingering references to it - I just tried that and as well as `xapian-letor/INSTALL` I see some false positives in the stemming data, a mention in xapian-letor/ChangeLog (which is a historical reference so should stay), but also this which ought to go:

```
xapian-letor/tests/generate-api_generated:    'SVMRanker' => ''
```

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -35,6 +35,8 @@
 #include <string>
 #include <vector>
 
+using namespace std;
+
```
ojwb: We shouldn't do this in public API headers - it dumps a load of symbols into the global namespace for user code.

Explicitly qualify with `std::` instead.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -126,12 +126,12 @@ batch_learning(vector<FeatureVector> feature_vectors,
...
 
 void
-ListMLERanker::train(const vector<FeatureVector>& training_data)
+ListMLERanker::train(const vector<vector<FeatureVector>>& training_data)
 {
     LOGCALL_VOID(API, "ListMLERanker::train", training_data);
     if (training_data.empty())
 	throw InvalidArgumentError("Cannot train: no training data");
-    int feature_cnt = training_data[0].get_fcount();
+    int feature_cnt = training_data[0][0].get_fcount();
```
ojwb: What about the `training_data[0].empty()` case?

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -193,7 +195,7 @@ class XAPIAN_VISIBILITY_DEFAULT Ranker : public Xapian::Internal::intrusive_base
   protected:
     /// Method to train the model. Overridden in ranker subclass.
     virtual void
-    train(const std::vector<Xapian::FeatureVector> & training_data) = 0;
+    train(const vector<vector<Xapian::FeatureVector>>& training_data) = 0;
```
ojwb: Probably not something to address right now, but I'm already luke warm on baking requirements to use a particular container into the API like we were here, and this change makes that situation worse.  Taking begin and end iterators for the data is a more flexible (and arguably a more natural C++ API since that's how it's done in the STL).

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -140,7 +140,9 @@ ListMLERanker::train(const vector<FeatureVector>& training_data)
     }
 
     for (int iter_num = 1; iter_num <= iterations; ++iter_num) {
-	batch_learning(training_data, new_parameters, learning_rate);
+	for (size_t it = 0; it < training_data.size(); ++it) {
+	    batch_learning(training_data[it], new_parameters, learning_rate);
+	}
```
ojwb: You can use C++11 range `for` here, which is more readable and eliminates the explicit index:

```
    for (auto& item : training_data) {
        batch_learning(item, new_parameters, learning_rate);
    }
```

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -120,12 +120,13 @@ update_parameters(vector<double> &new_parameters, const vector<double> &gradient
 }
 
 void
-ListNETRanker::train(const std::vector<Xapian::FeatureVector> & training_data) {
+ListNETRanker::train(const std::vector<vector<Xapian::FeatureVector>>&
+		     training_data) {
```
ojwb: The opening `{` for an out-of-line function definition should really be on a fresh line by Xapian coding conventions (this is currently often wrong in letor code).

`using namespace std;` is already active in this file, so you don't need the `std::` qualification - if you remove that and fix the `{` location then there's (just) no need to wrap this.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -120,12 +120,13 @@ update_parameters(vector<double> &new_parameters, const vector<double> &gradient
...
     LOGCALL_VOID(API, "ListNETRanker::train", training_data);
     size_t fvv_len = training_data.size();
     int feature_cnt = -1;
     if (fvv_len != 0) {
-	feature_cnt = training_data[0].get_fcount();
+	feature_cnt = training_data[0][0].get_fcount();
```
ojwb: Similar to above, what if `training_data[0].empty()`?

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -134,12 +135,16 @@ ListNETRanker::train(const std::vector<Xapian::FeatureVector> & training_data) {
 
     // iterations
     for (int iter_num = 1; iter_num <= iterations; ++iter_num) {
-	// initialize Probability distributions of y and z
-	prob_distrib_vector prob = init_probability(training_data, new_parameters);
-	// compute gradient
-	vector<double> gradient = calculate_gradient(training_data, prob);
-	// update parameters: w = w - gradient * learningRate
-	update_parameters(new_parameters, gradient, learning_rate);
+	for (size_t it = 0; it < training_data.size(); ++it) {
```
ojwb: Again, an opportunity to use C++11 range `for`.

xapian-letor/ranker/ranker.cc
```diff
@@ -106,9 +106,13 @@ load_list_fvecs(const string & filename)
 	train_file.ignore();
 
 	fv.set_score(0);
-	fvv.push_back(fv);
+	fvv[qid].push_back(fv);
     }
-    return fvv;
+    std::vector<vector<FeatureVector>> vt;
+    for (auto it = fvv.begin(); it != fvv.end(); ++it) {
```
ojwb: Another place that where C++11 range `for` would be more readable.

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -133,10 +134,14 @@ ListMLERanker::train(const vector<vector<FeatureVector>>& training_data)
 	throw InvalidArgumentError("Cannot train: no training data");
     int feature_cnt = training_data[0][0].get_fcount();
 
-    // Initialize the parameters for neural network
+    // initialize the parameters for neural network with Xavier initialization.
+    // construct a trivial random generator engine:
```
ojwb: If a comment is a sentence, it really should start with a capital letter (like it did before this change).

xapian-letor/ranker/listmle_ranker.cc
```diff
@@ -133,10 +134,14 @@ ListMLERanker::train(const vector<vector<FeatureVector>>& training_data)
...
+    std::default_random_engine generator(469382313);
+    std::normal_distribution<double> distribution(0.0, sqrt(2.0 /
+							   (1 + feature_cnt)));
```
ojwb: We have `using namespace std;` active in this file, so there's no need for the `std::` qualifiers (which probably means no need to wrap the line above).

`469382313` seems an unexplained magic number - that sort of thing deserves a comment.  I tried a web search for it but found nothing that seemed relevant.

Also we seem to be duplicating this Xavier initialisation code in each ranker - could we not have a single common version of it?

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -120,6 +120,13 @@ update_parameters(vector<double> &new_parameters, const vector<double> &gradient
     }
 }
 
+static void
+normalize(vector<double>& gradient, int size) {
```
ojwb: The passed value in the only use of this new function so far is of type `std::vector::size_type`, which on a 64 bit platform is typically a 64 bit unsigned type whereas `int` is typically a 32 bit type (and signed).  So the passed value might not fit in an `int` for really large inputs (more than ~2 billion training data items) at which point this function would quietly do the wrong thing.  That's quite an extreme input size, but probably not completely inconceivable (and it becomes more conceivable with time - my current computer has 32 million times more RAM than my first computer did!)

In general it's better to use the appropriate type instead.  Using `int` usually works fine for small scale tests but fails eventually, often in hard to debug ways.  C++ doesn't have a generic `size_type` as such but `size_t` is a reasonable choice.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -120,6 +120,13 @@ update_parameters(vector<double> &new_parameters, const vector<double> &gradient
...
+    for (size_t i = 0; i < gradient.size(); ++i) {
```
ojwb: Again, C++11 range for...

xapian-letor/ranker/ranker.cc
```diff
@@ -239,10 +239,11 @@ initialise_queryparser(const Xapian::Database & db)
     Xapian::QueryParser parser;
     parser.add_prefix("title", "S");
     parser.add_prefix("subject", "S");
+    parser.add_prefix("description", "XD");
     parser.set_database(db);
     parser.set_default_op(Xapian::Query::OP_OR);
     parser.set_stemmer(stemmer);
-    parser.set_stemming_strategy(Xapian::QueryParser::STEM_SOME);
+    parser.set_stemming_strategy(Xapian::QueryParser::STEM_ALL_Z);
```
ojwb: What's the reason for changing the stemming strategy here?

(Indeed there's no useful explanation for either of the two changes that are effectively being made in this commit in the commit message.)

xapian-letor/ranker/ranker.cc
```diff
@@ -293,8 +294,16 @@ Xapian::prepare_training_file(const string & db_path, const string & queryfile,
...
+	Xapian::Query query_description_prefix = parser.parse_query(querystr,
+					parser.FLAG_SPELLING_CORRECTION,
+					"XD");
 	// Combine queries
-	Xapian::Query query = Xapian::Query(Xapian::Query::OP_OR, query_no_prefix, query_default_prefix);
+	Xapian::Query query = Xapian::Query(Xapian::Query::OP_OR,
+				query_no_prefix,
+				Xapian::Query(Xapian::Query::OP_OR,
+					      query_default_prefix,
+					      query_description_prefix));
```
ojwb: Like elsewhere, we seem to be pointlessly specifying `FLAG_SPELLING_CORRECTION` here.

You're not specifying `FLAG_DEFAULT` when parsing with `"XD"` but we are for the other cases, which I guess is not deliberate?

Also it seems what we actually ought to be doing here is setting the default prefix for the query parser to map to "", "S" and "XD", then parsing the query string just once.  Something like:

```
parser.add_prefix("", "");
parser.add_prefix("", "S");
parser.add_prefix("", "XD");
Xapian::Query query = parser.parse_query(querystr);
```

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -190,6 +190,14 @@ class XAPIAN_VISIBILITY_DEFAULT CollTfCollLenFeature : public Feature {
     std::string name() const;
 };
 
+// This feature consists of a constant value
+class XAPIAN_VISIBILITY_DEFAULT ConstantFeature : public Feature {
+  public:
+    ConstantFeature() {}
```
ojwb: The commit message says `Add ConstantFeature(value=1.0)` but you're actually only implementing `ConstantFeature()` with the value hard-coded as `1.0`.

I don't know if it's useful to be able to specify the constant value for a bias feature (so I'm not sure if it's the commit message or the code that's wrong here) - I just know that the commit message really ought to match what the commit actually implements!

xapian-letor/include/xapian-letor/feature.h
```diff
@@ -190,6 +190,14 @@ class XAPIAN_VISIBILITY_DEFAULT CollTfCollLenFeature : public Feature {
     std::string name() const;
 };
 
+// This feature consists of a constant value
```
ojwb: This should be a documentation comment (so `///`).

xapian-letor/tests/api_letor.cc
```diff
@@ -336,12 +336,12 @@ DEFINE_TESTCASE(preparetrainingfileonedb, generated && path && writable)
 	string temp2;
 	int i = 0;
 	while ((iss1 >> temp1) && (iss2 >> temp2)) {
-	    // The 0th, 1st and 21st literals taken as input, are strings,
+	    // The 0th, 1st and 22st literals taken as input, are strings,
```
ojwb: "22nd" (also below)

xapian-letor/api/featurelist.cc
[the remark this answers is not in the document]
AyushP123: Actually bias term is not a feature, its a weight term used by neural network to make predictions. 
(https://medium.com/deeper-learning/glossary-of-deep-learning-bias-cf49d9c895e2)

VaibhavKansagara:
sorry for the late fixups, I was busy with academic work. I will look into rest of the comments tomorrow morning.

xapian-letor/docs/letor.rst
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/include/xapian-letor/ranker.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listmle_ranker.cc
VaibhavKansagara: > Or having different number of features than the rest of the vectors ( We should have a better naming convention here, each instance corresponds to a query ). Although in the current use that seems unlikely, I think that would be a good check in case anyone decides to use ListMLERanker::train outside Ranker::train_model.

>
I really think that is not necessary in this case but I might be wrong.

xapian-letor/ranker/listmle_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listnet_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listnet_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listnet_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/listmle_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listmle_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listnet_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listnet_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/include/xapian-letor/ranker.h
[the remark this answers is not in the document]
ojwb: I just checked and it is reported by the current version of the script:

```
xapian-letor/include/xapian-letor/ranker.h:196: error: Preferred style is 'int& x' (not 'int &x' or 'int & x'): +    train(const vector<vector<Xapian::FeatureVector>> & training_data) = 0;
```

xapian-letor/ranker/listmle_ranker.cc
ojwb: We ought to at least ensure something sensible happens if there are different numbers of entries in some `FeatureVector` objects compared to others.  That could be throwing an exception, or we could do something like assuming that any missing features compared to the first one are zero and any extra features get ignored.  It definitely shouldn't mean reading or writing off the end of some vector or using uninitialised values in the calculations.

It's unclear to me what actually happens with the current code.

jaylett: Pretty easy to write a test for vectors of differing lengths, as well.

xapian-letor/api/featurelist.cc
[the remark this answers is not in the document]
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/include/xapian-letor/ranker.h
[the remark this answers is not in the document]
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/listmle_ranker.cc
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/ranker/ranker.cc

xapian-letor/ranker/ranker.cc

xapian-letor/include/xapian-letor/feature.h
VaibhavKansagara: fixed in a fixup commit.

xapian-letor/tests/api_letor.cc

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -119,27 +119,48 @@ update_parameters(vector<double> &new_parameters, const vector<double> &gradient
     }
 }
 
+static void
+normalize(vector<double>& gradient, size_t size) {
+    for (auto item: gradient) {
```
ojwb: Missing space before `:`

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -155,6 +156,11 @@ class XAPIAN_VISIBILITY_DEFAULT Ranker : public Xapian::Internal::intrusive_base
      *			   same as what was used during preparation of the
      *			   training file.
      */
+
+    /// Initialize the parameters for neural network with Xavier initialization.
+    std::vector<double>
+    xavier_initialisation(int feature_cnt);
```
ojwb: This is only intended to be used by subclasses so should probably be `protected` rather than exposed as a `public` API.

Also, this line doesn't need wrapping.

xapian-letor/ranker/listnet_ranker.cc
```diff
@@ -119,27 +119,48 @@ update_parameters(vector<double> &new_parameters, const vector<double> &gradient
...
-	vector<double> gradient = calculate_gradient(training_data, prob);
-	// update parameters: w = w - gradient * learningRate
-	update_parameters(new_parameters, gradient, learning_rate);
+	for (auto& item : training_data) {
+	    // initialize Probability distributions of y and z
+	    prob_distrib_vector prob = init_probability(item,
+							new_parameters);
+	    // compute gradient
+	    vector<double> gradient = calculate_gradient(item,
+							 prob);
```
ojwb: Neither of the two wrapped lines above need wrapping.

There's a typo in the comment below: "gradinet"

Also capitalisation of these comments is odd - in general, start a comment that's a sentence with a capital letter, and end it with a full stop (or other suitable punctuation).  And don't randomly capitalise words in the comment (like "Probability" above).

xapian-letor/ranker/ranker.cc
```diff
@@ -106,9 +107,13 @@ load_list_fvecs(const string & filename)
 	train_file.ignore();
 
 	fv.set_score(0);
-	fvv.push_back(fv);
+	fvv[qid].push_back(fv);
     }
-    return fvv;
+    std::vector<vector<FeatureVector>> vt;
+    for (auto it: fvv) {
```
ojwb: Missing space before `:`.

xapian-letor/ranker/ranker.cc
```diff
@@ -281,22 +291,15 @@ Xapian::prepare_training_file(const string & db_path, const string & queryfile,
...
+	Xapian::Query query = parser.parse_query(querystr);
 
 	Xapian::Enquire enquire(letor_db);
 	enquire.set_query(query);
 	Xapian::MSet mset = enquire.get_mset(0, msetsize);
 
-	std::vector<FeatureVector> fvv_mset = flist.create_feature_vectors(mset, query, letor_db);
+	std::vector<FeatureVector> fvv_mset =
+					flist.create_feature_vectors(mset,
+					query, letor_db, flag, bias);
```
ojwb: This isn't how we'd typically wrap a line.  Especially wrapping the parameters at the same indent as the object the method is called on is likely to make the reader have to stop and and think in a distracting way.

(This isn't really specific to Xapian even - I can't think of a project I've seen where the coding conventions would be to wrap like this!)

Also I notice we're in a file where `using namespace std;` is in use, so the `std::` can be omitted, which helps.

I'd suggest either:

```
        vector<FeatureVector> fvv_mset = flist.create_feature_vectors(mset,
                                                                      query,
                                                                      letor_db,
                                                                      flag,
                                                                      bias);
```

Or:

```
        vector<FeatureVector> fvv_mset =
            flist.create_feature_vectors(mset, query, letor_db, flag, bias);

```

xapian-letor/ranker/ranker.cc
```diff
@@ -390,11 +393,27 @@ Ranker::rank(Xapian::MSet & mset, const string & model_key, const Xapian::Featur
...
 
+vector<double>
+Ranker::xavier_initialisation(int feature_cnt)
+{
+    // Construct a trivial random generator engine:
+    // 469382313 is a random number for which we are getting the best
+    // performance of letor against standard benchmark datasets.
+    default_random_engine generator(469382313);
+    normal_distribution<double> distribution(0.0, sqrt(2.0 /
+						       (1 + feature_cnt)));
```
ojwb: Wrapping between parameters seems more readable than in the middle of a parameter:

```
    normal_distribution<double> distribution(0.0,
                                             sqrt(2.0 / (1 + feature_cnt)));
```

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -193,7 +199,8 @@ class XAPIAN_VISIBILITY_DEFAULT Ranker : public Xapian::Internal::intrusive_base
   protected:
     /// Method to train the model. Overridden in ranker subclass.
     virtual void
-    train(const std::vector<Xapian::FeatureVector> & training_data) = 0;
+    train(const std::vector<std::vector<Xapian::FeatureVector>>&
+	  training_data) = 0;
```
ojwb: We're inside `namespace Xapian {` ... `}` here, so the explicit `Xapian::` qualification isn't necessary here.  And without there's no need to wrap the parameter list.

xapian-letor/include/xapian-letor/ranker.h
```diff
@@ -256,7 +263,8 @@ class XAPIAN_VISIBILITY_DEFAULT ListNETRanker: public Ranker {
      *
      * @exception LetorInternalError will be thrown if training data is null.
      */
-    void train(const std::vector<Xapian::FeatureVector> & training_data);
+    void train(const std::vector<std::vector<Xapian::FeatureVector>>&
+	       training_data);
```
ojwb: This also doesn't need wrapping if the `Xapian::` is removed.  And another case below.

dipanshu124:
The changes suggested are made #320.

ojwb:
I've cherry-picked the first commit (the removal of SVMRanker) onto current master and pushed it along with an update to the new CI configuration (we've since switched from travis-ci to GHA).

I'll resolve the rest in #324.

PR #279, opened 2019-08-12
Add Gmime to Omega

barufa:
At the moment, omindex uses a perl script to index eml files. However, 
we can use Gmime to handle this kind of files.

### DIFF
--- a/xapian-applications/omega/docs/overview.rst
+++ b/xapian-applications/omega/docs/overview.rst
@@ -307,8 +307,8 @@ other filters too - see below):
 * Atom feeds (.atom)
 * MAFF (.maff) if unzip is available
 * MHTML (.mhtml, .mht) if perl with MIME::Tools is available
-* MIME email messages (.eml) and USENET articles if perl with MIME::Tools and
-  HTML::Parser is available
+* MIME email messages (.eml) and USENET articles if gmime 2.6 or perl with MIME::Tools and
+  HTML::Parser are available
 * vCard files (.vcf, .vcard) if perl with Text::vCard is available
 * FictionBook v.2 files (.fb2) if libe-book is available
 * QiOO (mobile format, for java-enabled cellphones) files (.jar) if libe-book is available
--- a/xapian-applications/omega/docs/overview.rst
+++ b/xapian-applications/omega/docs/overview.rst
@@ -308,7 +308,7 @@ other filters too - see below):
 * MAFF (.maff) if unzip is available
 * MHTML (.mhtml, .mht) if perl with MIME::Tools is available
 * MIME email messages (.eml) and USENET articles if gmime 2.6 or perl with MIME::Tools and
-  HTML::Parser are available
+  HTML::Parser is available
 * vCard files (.vcf, .vcard) if perl with Text::vCard is available
 * FictionBook v.2 files (.fb2) if libe-book is available
 * QiOO (mobile format, for java-enabled cellphones) files (.jar) if libe-book is available
--- a/xapian-applications/omega/handler_gmime.cc
+++ b/xapian-applications/omega/handler_gmime.cc
@@ -166,11 +166,7 @@ extract(const string& filename,
     author = g_mime_message_get_sender(message);
     title = g_mime_message_get_subject(message);
 
-    try {
-	parser_content(body, dump);
-    } catch (ReadError) {
-	dump.clear();
-    }
+    parser_content(body, dump);
 
     (void)pages;
     return true;
--- a/xapian-applications/omega/handler_gmime.cc
+++ b/xapian-applications/omega/handler_gmime.cc
@@ -52,27 +52,26 @@ extract_html(const string& text, string& charset, string& dump)
     }
 }
 
-size_t decode(unsigned char* text, size_t len, GMimeContentEncoding encoding)
+size_t decode(unsigned char* text, size_t len, GMimeContentEncoding encoding,
+	      int* state, guint32* save)
 {
     unsigned char buffer[SIZE];
-    static int state = 0;
-    static guint32 save = 0;
     switch (encoding) {
 	case GMIME_CONTENT_ENCODING_BASE64: {
 	    len = g_mime_encoding_base64_decode_step(text, len, buffer,
-						     &state, &save);
+						     state, save);
 	    memcpy(text, buffer, len);
 	    break;
 	}
 	case GMIME_CONTENT_ENCODING_UUENCODE: {
 	    len = g_mime_encoding_uudecode_step(text, len, buffer,
-						&state, &save);
+						state, save);
 	    memcpy(text, buffer, len);
 	    break;
 	}
 	case GMIME_CONTENT_ENCODING_QUOTEDPRINTABLE: {
 	    len = g_mime_encoding_quoted_decode_step(text, len, buffer,
-						     &state, &save);
+						     state, save);
 	    memcpy(text, buffer, len);
 	    break;
 	}
@@ -117,12 +116,14 @@ parser_content(GMimeObject* me, string& dump)
 	    char buffer[SIZE];
 	    GMimeStream* sr = g_mime_data_wrapper_get_stream(content);
 	    auto encoding = g_mime_part_get_content_encoding(part);
-	    int len = 0;
+	    int state = 0, len = 0;
+	    guint32 save = 0;
 	    do {
-		len = g_mime_stream_read(sr, buffer, SIZE / 2);
-		if (len > 0) {
-		    len = decode((unsigned char*)buffer, len, encoding);
-		    if (len > 0)
+		len = g_mime_stream_read(sr, buffer, SIZE);
+		if (0 < len) {
+		    auto u_buff = reinterpret_cast<unsigned char*>(buffer);
+		    len = decode(u_buff, len, encoding, &state, &save);
+		    if (0 < len)
 			text.append(buffer, len);
 		}
 	    } while (0 < len);
--- a/xapian-applications/omega/handler_gmime.cc
+++ b/xapian-applications/omega/handler_gmime.cc
@@ -1,0 +1,173 @@
+/** @file handler_gmime.cc
+ * @brief Extract text and metadata using gmime.
+ */
+/* Copyright (C) 2019 Bruno Baruffaldi
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+#include <config.h>
+#include "handler.h"
+
+#include "myhtmlparse.h"
+
+#include <glib.h>
+#include <gmime/gmime.h>
+
+using namespace std;
+
+const unsigned SIZE = 8192;
+
+static void
+extract_html(const string& text, string& charset, string& dump)
+{
+    MyHtmlParser parser;
+    if (charset.empty())
+	charset = "UTF-8";
+    try {
+	parser.ignore_metarobots();
+	parser.parse_html(text, charset, false);
+	dump += parser.dump;
+    } catch (const string& newcharset) {
+	parser.reset();
+	parser.ignore_metarobots();
+	parser.parse_html(text, newcharset, true);
+	dump += parser.dump;
+    } catch (...) {
+	return;
+    }
+}
+
+size_t decode(unsigned char* text, size_t len, GMimeContentEncoding encoding)
+{
+    unsigned char buffer[SIZE];
+    int state;
+    guint32 save;
+    switch (encoding) {
+	case GMIME_CONTENT_ENCODING_BASE64: {
+	    len = g_mime_encoding_base64_decode_step(text, len, buffer,
+						     &state, &save);
+	    memcpy(text, buffer, len);
+	    break;
+	}
+	case GMIME_CONTENT_ENCODING_UUENCODE: {
+	    len = g_mime_encoding_uudecode_step(text, len, buffer,
+						&state, &save);
+	    memcpy(text, buffer, len);
+	    break;
+	}
+	case GMIME_CONTENT_ENCODING_QUOTEDPRINTABLE: {
+	    len = g_mime_encoding_quoted_decode_step(text, len, buffer,
+						     &state, &save);
+	    memcpy(text, buffer, len);
+	    break;
+	}
+	default:
+	    break;
+    }
+    return len;
+}
+
+bool
+parser_content(GMimeObject* me, string& dump)
+{
+    GMimeContentType* ct = g_mime_object_get_content_type(me);
+    if (GMIME_IS_MULTIPART(me)) {
+	GMimeMultipart* mpart = (GMimeMultipart*)me;
+	string subtype = g_mime_content_type_get_media_subtype(ct);
+	int count = g_mime_multipart_get_count(mpart);
+	if (subtype == "alternative") {
+	    for (int i = 0; i < count; ++i) {
+		GMimeObject* part = g_mime_multipart_get_part(mpart, i);
+		if (parser_content(part, dump))
+		    return true;
+	    }
+	} else {
+	    bool ret = false;
+	    for (int i = 0; i < count; ++i) {
+		GMimeObject* part = g_mime_multipart_get_part(mpart, i);
+		ret |= parser_content(part, dump);
+	    }
+	    return ret;
+	}
+    } else if (GMIME_IS_PART(me)) {
+	GMimePart* part = (GMimePart*)me;
+	GMimeDataWrapper* content = g_mime_part_get_content_object(part);
+	string type = g_mime_content_type_get_media_type(ct);
+	string subtype = g_mime_content_type_get_media_subtype(ct);
+	string charset(g_mime_object_get_headers(me));
+	size_t start = charset.find("charset=");
+	if (start != string::npos) {
+	    start += 9;
+	    charset = charset.substr(start, charset.length() - start - 3);
+	    charset = g_mime_charset_canon_name(charset.c_str());
+	} else {
+	    charset.clear();
+	}
+	if (type == "text") {
+	    char* buffer = new char[SIZE];
+	    GMimeStream* sr = g_mime_data_wrapper_get_stream(content);
+	    size_t presz = 0, len = g_mime_stream_read(sr, buffer, SIZE);
+	    auto encoding = g_mime_part_get_content_encoding(part);
+	    len = decode((unsigned char*)buffer, len, encoding);
+	    if (subtype == "plain") {
+		if (!charset.empty() && charset != "UTF-8") {
+		    GMimeFilter* ch = g_mime_filter_charset_new(charset.c_str(),
+								"UTF-8");
+		    if (GMIME_IS_FILTER(ch)) {
+			g_mime_filter_complete(ch, buffer, len, presz, &buffer,
+						&len, &presz);
+		    }
+		}
+		dump += string(buffer, buffer + len);
+	    } else if (subtype == "html") {
+		extract_html((const char*)buffer, charset, dump);
+	    }
+	    delete [] buffer;
+	    return true;
+	}
+    }
+    return false;
+}
+
+bool
+extract(const string& filename,
+	string& dump,
+	string& title,
+	string& keywords,
+	string& author,
+	string& pages,
+	string& error)
+{
+    g_mime_init(0);
+
+    FILE* fd = fopen(filename.c_str(), "r");
+
+    if (fd == NULL) {
+	error = "Gmime Error: fail open " + filename;
+	return false;
+    }
+
+    GMimeStream* stream = g_mime_stream_file_new(fd);
+    GMimeParser* parser = g_mime_parser_new_with_stream(stream);
+    GMimeMessage* message = g_mime_parser_construct_message(parser);
+    GMimeObject* body = g_mime_message_get_body(message);
+    author = g_mime_message_get_sender(message);
+    title = g_mime_message_get_subject(message);
+
+    parser_content(body, dump);
+    (void)pages;
+    return true;
+}
--- a/xapian-applications/omega/handler_gmime.cc
+++ b/xapian-applications/omega/handler_gmime.cc
@@ -23,9 +23,7 @@
 
 #include "myhtmlparse.h"
 #include "utf8convert.h"
-#include <string.h>
 
-#include <glib.h>
 #include <gmime/gmime.h>
 
 using namespace std;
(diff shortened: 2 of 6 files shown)


xapian-applications/omega/docs/overview.rst
```diff
@@ -307,8 +307,8 @@ other filters too - see below):
 * Atom feeds (.atom)
 * MAFF (.maff) if unzip is available
 * MHTML (.mhtml, .mht) if perl with MIME::Tools is available
-* MIME email messages (.eml) and USENET articles if perl with MIME::Tools and
-  HTML::Parser is available
+* MIME email messages (.eml) and USENET articles if gmime 2.6 or perl with MIME::Tools and
+  HTML::Parser are available
```
ojwb: I think this should still be "is" - it's "... if A or B is available" where A is "gmime 2.6" and B is "perl with <module list>".

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	size_t start = charset.find("charset=");
+	if (start != string::npos) {
+	    start += 9;
+	    charset = charset.substr(start, charset.length() - start - 3);
+	    charset = g_mime_charset_canon_name(charset.c_str());
+	} else {
+	    charset.clear();
+	}
+	if (type == "text") {
+	    char* buffer = new char[SIZE];
```
ojwb: There's no need for dynamic memory allocation here - you can just declare this as a local array, like you do in `decode()` above.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	    if (subtype == "plain") {
+		if (!charset.empty() && charset != "UTF-8") {
+		    GMimeFilter* ch = g_mime_filter_charset_new(charset.c_str(),
+								"UTF-8");
+		    if (GMIME_IS_FILTER(ch)) {
+			g_mime_filter_complete(ch, buffer, len, presz, &buffer,
+						&len, &presz);
+		    }
+		}
+		dump += string(buffer, buffer + len);
```
ojwb: This needlessly creates a temporary string object.  Better to use:

```
dump.append(buffer, len);
```

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	    start += 9;
+	    charset = charset.substr(start, charset.length() - start - 3);
+	    charset = g_mime_charset_canon_name(charset.c_str());
+	} else {
+	    charset.clear();
+	}
+	if (type == "text") {
+	    char* buffer = new char[SIZE];
+	    GMimeStream* sr = g_mime_data_wrapper_get_stream(content);
+	    size_t presz = 0, len = g_mime_stream_read(sr, buffer, SIZE);
```
ojwb: `presz` seems rather a cryptic name.  Even looking where it is used I'm not really any the wiser.

I'd also avoid declaring multiple variables on the same line if any have non-trivial initialisers - it isn't particularly easy to read.

You also really want to avoid doing this if any of the types are pointers or references - `int* x, y` actually declare `y` as of type `int` - the C++ style of putting the `*` with the type misleads in this case.

Something like `int x = 0, y = 0;` would be OK.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	    auto encoding = g_mime_part_get_content_encoding(part);
+	    len = decode((unsigned char*)buffer, len, encoding);
+	    if (subtype == "plain") {
+		if (!charset.empty() && charset != "UTF-8") {
+		    GMimeFilter* ch = g_mime_filter_charset_new(charset.c_str(),
+								"UTF-8");
+		    if (GMIME_IS_FILTER(ch)) {
+			g_mime_filter_complete(ch, buffer, len, presz, &buffer,
+						&len, &presz);
+		    }
```
ojwb: `presz` is only used in the block above, so better to declare it in this block rather than at a higher level.  That way the reader can more easily see what type `presz` is when looking at where it is used, and it's clear without searching that it isn't used again after this.

Also the wrapped line is over-indented by one column.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+		}
+		dump += string(buffer, buffer + len);
+	    } else if (subtype == "html") {
+		extract_html((const char*)buffer, charset, dump);
```
ojwb: This ignores `len` - I think you want `string(buffer, len)` for the first argument.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	}
+    } else if (GMIME_IS_PART(me)) {
+	GMimePart* part = (GMimePart*)me;
+	GMimeDataWrapper* content = g_mime_part_get_content_object(part);
+	string type = g_mime_content_type_get_media_type(ct);
+	string subtype = g_mime_content_type_get_media_subtype(ct);
+	string charset(g_mime_object_get_headers(me));
+	size_t start = charset.find("charset=");
+	if (start != string::npos) {
+	    start += 9;
```
ojwb: `charset=` is 8 characters.

If there's some reason why it's one extra, then you really should make that clear (ideally as a comment) or else someone is likely to come along later, think this is a typo and "fix" it.

I'm wondering if you're trying to ignore quotes around the charset value, but if so I think you ought to check they are there.  I looked at some MIME messages locally and they all just have a character set without quotes - e.g. `charset=us-ascii`.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	    charset.clear();
+	}
+	if (type == "text") {
+	    char* buffer = new char[SIZE];
+	    GMimeStream* sr = g_mime_data_wrapper_get_stream(content);
+	    size_t presz = 0, len = g_mime_stream_read(sr, buffer, SIZE);
+	    auto encoding = g_mime_part_get_content_encoding(part);
+	    len = decode((unsigned char*)buffer, len, encoding);
+	    if (subtype == "plain") {
+		if (!charset.empty() && charset != "UTF-8") {
```
ojwb: What does `g_mime_charset_canon_name()` return for `us-ascii`?  ASCII is a subset of UTF-8, so there's no need to convert it.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+bool
+extract(const string& filename,
+	string& dump,
+	string& title,
+	string& keywords,
+	string& author,
+	string& pages,
+	string& error)
+{
+    g_mime_init(0);
```
ojwb: Shouldn't this just be done once?

I think we probably want to add an `init()` function which gets called from `main()` in `assistant.cc` before it does anything else - in the tesseract module we also needed a one-off initialisation.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+
+    FILE* fd = fopen(filename.c_str(), "r");
```
ojwb: `fd` is an acronym for "file descriptor", but this is a "FILE pointer", so it's confusing to use `fd` as the variable name here.  Something like `fp` would be a better choice.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+    } catch (...) {
+	return;
+    }
+}
+
+size_t decode(unsigned char* text, size_t len, GMimeContentEncoding encoding)
+{
+    unsigned char buffer[SIZE];
+    int state;
+    guint32 save;
```
ojwb: I think you probably need to be preserving the values of `state` and `save` between calls to this function - otherwise it's not going to handle encoded data which is split between two chunks.

Consider quoted-printable sequence `=46` where `=4` is at the end of one chunk and `6` at the start of the next.  Something needs to remember that we're in the middle of a QP sequence and that we already saw `4` as the first digit.  I'd imagine currently this will wrongly decode as `6` instead of `F`.

xapian-applications/omega/handler_gmime.cc
barufa: Okay, I am going to move `presz` here and change it for `presize`.

xapian-applications/omega/handler_gmime.cc
barufa: Okay, I will try to make it more robust.

xapian-applications/omega/handler_gmime.cc
barufa: It return `us-ascii` and try to make some conversion, but it shouldn't as `us-ascii` is a subset of UTF-8.

xapian-applications/omega/handler_gmime.cc
barufa: Yes, I didn't notice. I am going to fix this.

xapian-applications/omega/handler_gmime.cc
barufa: Actually, I am not sure about how it works. I could not find any example and the manual doesn't say much about it.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,183 @@
...
+
+    GMimeStream* stream = g_mime_stream_file_new(fp);
+    GMimeParser* parser = g_mime_parser_new_with_stream(stream);
+    GMimeMessage* message = g_mime_parser_construct_message(parser);
+    GMimeObject* body = g_mime_message_get_body(message);
+    author = g_mime_message_get_sender(message);
+    title = g_mime_message_get_subject(message);
+
+    parser_content(body, dump);
+    (void)pages;
```
barufa: I forget `fclose(fp)`

ojwb: You shouldn't call that - see:

https://developer.gnome.org/gmime/stable/GMimeStreamFile.html#g-mime-stream-file-new

> Note: The created GMimeStreamFile object will own the FILE pointer passed in.

So that should close it for you - if you also close it, that's undefined behaviour - from `man fclose`:

> The  behaviour  of  fclose() is undefined if the stream parameter is an illegal pointer, or is a descriptor already passed to a previous invocation of fclose().

xapian-applications/omega/handler_gmime.cc
```diff
@@ -109,34 +109,39 @@ parser_content(GMimeObject* me, string& dump)
 	string type = g_mime_content_type_get_media_type(ct);
 	string subtype = g_mime_content_type_get_media_subtype(ct);
 	string charset(g_mime_object_get_headers(me));
-	size_t start = charset.find("charset=");
+	size_t end, start;
+	start = charset.find("charset=");
```
ojwb: Before `start` was initialised at the point of declaration, which nice and clean and generally what you want to do where it's easy to arrange.  If a variable is only declared when you can initialise it there's no point at which is has an undefined value, and that's one less thing to worry about.

But now you don't, and not for any good reason.

In fact `end` would be better declared in the block below where it is used rather than here.

xapian-applications/omega/handler_gmime.cc
ojwb: You don't actually seem to have addressed this though.

xapian-applications/omega/handler_gmime.cc
barufa: Okay, I am going to change it.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,183 @@
...
+	    return ret;
+	}
+    } else if (GMIME_IS_PART(me)) {
+	GMimePart* part = (GMimePart*)me;
+	GMimeDataWrapper* content = g_mime_part_get_content_object(part);
+	string type = g_mime_content_type_get_media_type(ct);
+	string subtype = g_mime_content_type_get_media_subtype(ct);
+	string charset(g_mime_object_get_headers(me));
+	size_t end, start;
+	start = charset.find("charset=");
```
ojwb: I was looking at this again as it seemed surprising that gmime didn't handle pulling the charset out from the header for us.  I think we ought to be using this function:

https://developer.gnome.org/gmime/stable/GMimeObject.html#g-mime-object-get-content-type-parameter

I tried this:

```
        string charset;  
        {                                                                            
            const char* p = g_mime_object_get_content_type_parameter(me,  
                                                                     "charset");  
            if (p) charset = g_mime_charset_canon_name(p);  
        }  
```

And it seems to give `UTF-8` in a simple test, same as your current code here does.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,183 @@
...
+	    auto encoding = g_mime_part_get_content_encoding(part);
+	    len = decode((unsigned char*)buffer, len, encoding);
+	    if (subtype == "plain") {
+		if (!charset.empty() && charset != "UTF-8") {
+		    GMimeFilter* ch = g_mime_filter_charset_new(charset.c_str(),
+								"UTF-8");
+		    if (GMIME_IS_FILTER(ch)) {
+			size_t presize = 0;
+			g_mime_filter_complete(ch, buffer, len, presize,
+					      (char**)&buffer, &len, &presize);
```
ojwb: Here you are specifying the input buffer as the output buffer.  That's unlikely to work well if translating to UTF-8 from a charset if the output is longer than the input, because the conversion will start to overwrite the input before it has read it.

For example, consider iso-8859-1 input "República" which consists of the 9 bytes:

52 65 70 fa 62 6c 69 63 61

I'll assume the input is read character by character (it's possible there's some funky optimisation which means the convertor reads more data at a time, but the general point still holds and it's simpler to consider the straightforward case to show how this might go wrong).

So as that gets converted to UTF-8 things are OK for the first 3 characters, as they each map to a single byte UTF-8 sequence (which consists of the same byte value as the input, since ASCII is a subset of UTF-8).  But when we convert the 4th character we get the two byte sequence c3 ba and the buffer now looks like this:

52 65 70 c3 ba 6c 69 63 61

Then we try to convert the 5th character, and we read byte `ba` which has overwritten `62`.  That's converted to c2 ba in UTF-8 so now we have:

52 65 70 c3 c2 ba 69 63 61

And again we have `ba`, so repeating these steps to the end the final result would be this, which isn't even valid UTF-8:

52 65 70 c3 c2 c2 c2 c2 c2 ba

That ought to be:

52 65 70 c3 ba 62 6c 69 63 61

Another problem is that the input might fit in `SIZE` bytes but the output might not.

To use this function safely I think you need to give it a separate output buffer, and to size that buffer such that it's large enough for any possibly expansion that character set conversion can give (or to retry if the conversion doesn't complete).

However Omega already has a function to convert text to UTF-8, so probably it makes more sense to just use that since it's much simpler to use from C++ - see `utf8convert.h`:

```
void convert_to_utf8(std::string & text, const std::string & charset);
```

This takes a string by reference and the charset it is in and modifies the string to be in UTF-8.  It even handles checking if it's already UTF-8 for you (it doesn't currently check for us-ascii but I'll push a fix for that shortly as that's a missed optimisation opportunity).

xapian-applications/omega/handler_gmime.cc
ojwb: You marked this as resolved, but haven't actually addressed it.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,173 @@
...
+	if (p) charset = g_mime_charset_canon_name(p);
+	if (type == "text") {
+	    string text;
+	    char buffer[SIZE];
+	    GMimeStream* sr = g_mime_data_wrapper_get_stream(content);
+	    auto encoding = g_mime_part_get_content_encoding(part);
+	    int len = 0;
+	    do {
+		len = g_mime_stream_read(sr, buffer, SIZE / 2);
+		if (len > 0) {
```
barufa: I am not sure if I should use `SIZE/2` here.
If I am reading a base64 encode string and I full all the buffer (which seems unlikely, but not impossible), I will run out of memory when I try to decode that and store the result in the buffer.
I don't feel that using `SIZE/2` is a good solution, but I cannot come up with a better one.

xapian-applications/omega/handler_gmime.cc
barufa: Okay, I has used that function before. I will change the filters for `convert_to_utf8`

xapian-applications/omega/handler_gmime.cc
```diff
@@ -54,8 +55,8 @@ extract_html(const string& text, string& charset, string& dump)
 size_t decode(unsigned char* text, size_t len, GMimeContentEncoding encoding)
 {
     unsigned char buffer[SIZE];
-    int state;
-    guint32 save;
+    static int state = 0;
+    static guint32 save = 0;
```
ojwb: Making these `static` isn't really a good solution.  Then they won't get reset between mime entities so if one leaves the decoder in a different state to (0,0) then the next will start from that state (and potentially in could be the state from a different decoder even).  It's also not thread safe, though we aren't currently using threads here.

Instead declare them in the caller and pass them in by reference - then they have the lifetime we actually want.

xapian-applications/omega/handler_gmime.cc
ojwb: Why will you run out of memory?

The decoded version should be **smaller** - e.g.:

```
$ BASE64=c2ltcGxlIHRlc3QK
$ echo "$BASE64"|wc -c
17
$ echo c2ltcGxlIHRlc3QK|base64 --decode
simple test
$ echo c2ltcGxlIHRlc3QK|base64 --decode|wc -c
12
```

(The `wc -c` counts both include a linefeed character so are actually one more than the actual length.)

xapian-applications/omega/handler_gmime.cc
```diff
@@ -108,39 +109,28 @@ parser_content(GMimeObject* me, string& dump)
...
+	    int len = 0;
+	    do {
+		len = g_mime_stream_read(sr, buffer, SIZE / 2);
+		if (len > 0) {
+		    len = decode((unsigned char*)buffer, len, encoding);
+		    if (len > 0)
+			text.append(buffer, len);
 		}
-		dump.append(buffer, len);
+	    } while (0 < len);
```
ojwb: It seems weirdly inconsistent to write `0 < len` here but `len > 0` above.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -108,39 +109,28 @@ parser_content(GMimeObject* me, string& dump)
...
-		    if (GMIME_IS_FILTER(ch)) {
-			size_t presize = 0;
-			g_mime_filter_complete(ch, buffer, len, presize,
-					      (char**)&buffer, &len, &presize);
-		    }
+	    int len = 0;
+	    do {
+		len = g_mime_stream_read(sr, buffer, SIZE / 2);
+		if (len > 0) {
+		    len = decode((unsigned char*)buffer, len, encoding);
```
ojwb: Quoting the dev guide:

> Use C++ style casts (``static_cast<>``, ``reinterpret_cast<>``, and
``const_cast<>``) or constructor-syntax (e.g. ``double(value)``) in
preference to C style casts.  The syntax of the C++ casts is ugly,
but they do make the intent much clearer which is definitely a good
thing, and they avoid issues such as casting away const when you
only meant to cast the type of a pointer.

xapian-applications/omega/handler_gmime.cc
barufa: Sorry, I am going to change it.

xapian-applications/omega/handler_gmime.cc
barufa: Sorry, I think that the decoded version was bigger

xapian-applications/omega/handler_gmime.cc
barufa: Okay, I am going to change it.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	parser.parse_html(text, charset, false);
+	dump += parser.dump;
+    } catch (const string& newcharset) {
+	parser.reset();
+	parser.ignore_metarobots();
+	parser.parse_html(text, newcharset, true);
+	dump += parser.dump;
+    } catch (...) {
+	return;
+    }
```
ojwb: Quietly ignoring all unexpected exceptions seems a bad idea.  If there particular exceptions that we should be ignoring, be more specific.  If not, I'd drop the default exception handling here.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+    return len;
+}
+
+bool
+parser_content(GMimeObject* me, string& dump)
+{
+    GMimeContentType* ct = g_mime_object_get_content_type(me);
+    if (GMIME_IS_MULTIPART(me)) {
+	GMimeMultipart* mpart = (GMimeMultipart*)me;
```
ojwb: Another C-style cast.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -0,0 +1,174 @@
...
+	} else {
+	    bool ret = false;
+	    for (int i = 0; i < count; ++i) {
+		GMimeObject* part = g_mime_multipart_get_part(mpart, i);
+		ret |= parser_content(part, dump);
+	    }
+	    return ret;
+	}
+    } else if (GMIME_IS_PART(me)) {
+	GMimePart* part = (GMimePart*)me;
```
ojwb: And another...

xapian-applications/omega/handler_gmime.cc
barufa: I am not sure if I should drop the default exception handling here. If there was any exception(different from newcharset) inside `extract_html`, it would return without appending anything to dump.
Maybe I can add some exception handling to `parser_content` at `extract`, and if there is any exception, the dump will be ignored (and return just the `author` and the `title`).

ojwb: Quietly ignoring an unexpected exception at any level is rarely helpful.  It might seem like you're writing more robust code by doing so, but what it actually means is that nobody's likely to report or debug a problem which results in an exception being thrown because they simply don't see any evidence of it (unless they look closely enough to notice some document content missing, but people rarely do unless they are looking for a particular affected document they know for sure is present but discover they can't find it in the search).

If you're going to handle unexpected exceptions at all just do it at the top level like the other helpers and return an error across the connection - e.g. for poppler you have:

```
    } catch (...) {
        error = "Poppler threw an exception";
        return false;
    } 
```

If there are particular exceptions that happen in the normal course of processing documents with a particular library, then they can be considered on a case-by-case basis - in some cases it might be reasonable to just treat them like an empty field.  But unless we know of such exceptions it's better to default to reporting exceptions as then we will at least find out about them.

xapian-applications/omega/handler_gmime.cc
```diff
@@ -166,7 +166,12 @@ extract(const string& filename,
     author = g_mime_message_get_sender(message);
     title = g_mime_message_get_subject(message);
 
-    parser_content(body, dump);
+    try {
+	parser_content(body, dump);
+    } catch (...) {
+	dump.clear();
+    }
```
ojwb: This isn't helpful handling of unexpected exceptions either.

ojwb:
Thanks for the updates - pushed to master as c34ca1efb4e495846317b3abacb40c4930926ca6.

BTW, I noticed gmime 2.6 is quite an old release - it was last updated 2.5 years ago and there are two newer release series - did you look at supporting gmime 3.0 and/or 3.2?

https://developer.gnome.org/gmime/

ojwb:
We can easily specify to use bionic for the CI job that's testing with gmime 3.0.

PR #280, opened 2019-08-15
Add omindextest to Omega 

barufa:
Currently, there is no program to test omindex and how the files are 
indexed. To solve this it is possible to index a small group of files and 
check the information extracted from them.
It is still work in progress.


xapian-applications/omega/Makefile.am
```diff
@@ -98,7 +98,8 @@ TESTS =	atomparsetest$(EXEEXT)\
 	sdbntest$(EXEEXT)\
 	sourcetest\
 	urlenctest$(EXEEXT)\
-	utf8converttest$(EXEEXT)
+	utf8converttest$(EXEEXT)\
+	omindextest
```
ojwb: As a general principal, when adding a new entry to a list check if the list is currently in sorted order - if it is, you almost always want to add it so you preserve that sorted order...

xapian-applications/omega/omindexcheck.cc
```diff
@@ -0,0 +1,106 @@
...
+{
+    tests.insert({"test.txt", {"Zjoey", "Zfood",
+		  "Zедой", "Z喬伊不分享食物"}});
+#if defined HAVE_POPPLER
+    tests.insert({"pdf/hello.pdf", {"Shello", "Sworld", "Ajeroen", "Aooms",
+		  "Zhello", "Zworld", "Zsubsect"}});
+    tests.insert({"pdf/poppler.pdf", {"Zabstract", "Zkeyword", "Ztopic",
+		  "Ztext", "Spdftitle"}});
+#endif
+}
```
barufa: I know that this test cases are poor. I just want to ask if this is way of testing is okay.

xapian-applications/omega/Makefile.am
```diff
@@ -95,11 +95,11 @@ TESTS =	atomparsetest$(EXEEXT)\
 	jsonesctest$(EXEEXT)\
 	md5test$(EXEEXT)\
 	omegatest\
+	omindextest\
```
ojwb: This is a compiled test program, so ought to have the `$(EXEEXT)` suffix here (for platforms which require `.exe` on compiled executables).

ojwb: Oh sorry, I was mixing up `omindexcheck` and `omindextest` - this is correct as-is.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -1,5 +1,5 @@
 /** @file omindexcheck.cc
- * @brief test omindex
+ * @brief Auxiliar program of omindextest
```
ojwb: "Auxiliar" isn't an English word - I think you want "Auxiliary".

xapian-applications/omega/omindexcheck.cc
```diff
@@ -40,7 +40,8 @@ unordered_map<string, testcase> tests;
 static void
 index_test(void)
```
ojwb: ISO C uses `(void)` for an empty parameter list.  That also works in C++, but it's more natural to just write `()` in C++ code when there are no parameters (in C `()` means an unspecified number of parameters, for compatibility with pre-standardised C).

.travis.yml
```diff
@@ -83,7 +83,6 @@ matrix:
             - python3-dev
             - tcl
             - libsvm-dev
-            - libpoppler-cpp-dev
```
ojwb: Presumably this change is accidental?

If it's actually wanted (because poppler isn't used in this case or something like that) then it doesn't seem to belong here and should really be submitted as a separate patch.

xapian-applications/omega/.gitignore
```diff
@@ -39,6 +39,7 @@
 /md5test
 /urlenctest
 /utf8converttest
+/omindexcheck
```
ojwb: The other test programs are listed in sorted order here...

xapian-applications/omega/omindextest
```diff
@@ -0,0 +1,34 @@
...
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+# USA
+
+DIR=$(pwd)
+TEST_FILES="$DIR/testfiles/"
+TEST_DB="$DIR/testdatabase"
+
+./omindex --db $TEST_DB --url $TEST_FILES $TEST_FILES
+./omindexcheck $TEST_FILES $TEST_DB > "omindextest.log"
+failed=$(echo "$?")
```
ojwb: Why not just use `failed=$?`?  Your version invokes an extra subshell, and is less clear to the reader.

xapian-applications/omega/omindextest
```diff
@@ -0,0 +1,34 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+# USA
+
+DIR=$(pwd)
+TEST_FILES="$DIR/testfiles/"
+TEST_DB="$DIR/testdatabase"
+
+./omindex --db $TEST_DB --url $TEST_FILES $TEST_FILES
```
ojwb: I'd suggest `--url=/` so that the terms in the database produced don't depend on the directory the source code is checked out in.

xapian-applications/omega/omindextest
```diff
@@ -0,0 +1,34 @@
...
+./omindexcheck $TEST_FILES $TEST_DB > "omindextest.log"
```
ojwb: You ought to quote these variables (`"$TEST_FILES"` etc) as otherwise things go wrong if the directory pathname contains whitespace or shell metacharacters.

xapian-applications/omega/omindextest
```diff
@@ -0,0 +1,34 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+# USA
+
+DIR=$(pwd)
+TEST_FILES="$DIR/testfiles/"
+TEST_DB="$DIR/testdatabase"
```
ojwb: This isn't right when building outside the source tree (which we aim to support).

Assuming `/testfiles/` is static data (i.e. it isn't generated as part of the build) than the correct place to look is `$srcdir` which will be set when `omindextest` is run via `make check` but not if `omindextest` is run by hand.

But the `omindextest` script is also not generated so should also be in the source tree, so we can use the directory part of `$0` to give us the path to the source directory, e.g.:

```
srcdir=`echo "$0"|sed 's!/*[^/]*$!!'`
```

So if you run `./omindextest` then that will set srcdir to `.` which is what we want.

If you're building in a `BUILD` subdirectory of the source tree, you would run `../omindextest` from there and then srcdir is set to `..`.  And a full path also works.

However `TEST_DB` should go in the build directory (because the source directory might be shared with other build directories, or might even be read-only).  The build directory is just the current directory, and there's no need for the database path to be an absolute path, so there's no need to run `pwd` - just `TESTDB=testdatabase` should work.

xapian-applications/omega/omindextest
ojwb: Also, the automake test harness will redirects output from `omindextest` to `omindextest.log` so you really don't want to also do that.  Just omit the redirection and things should work fine.

xapian-applications/omega/omindextest
```diff
@@ -0,0 +1,34 @@
...
+./omindex --db $TEST_DB --url $TEST_FILES $TEST_FILES
+./omindexcheck $TEST_FILES $TEST_DB > "omindextest.log"
+failed=$(echo "$?")
+
+rm -r $TEST_DB
+
+if [ $failed = 0 ] ; then
+    exit 0
+fi
+exit 1
```
ojwb: Simpler to `exit "$failed"` which also preserves the actual non-zero exit status of the command rather than squashing any non-zero exit status to `1`.

But there's a better way to achieve this using the `trap` shell built-in - you can specify a command to be run on shell exit with:

```
trap 'rm -rf "$TEST_DB"' 0
```

So do that just before running `omindex` and the database will always be removed for you.

If you want the database to be removed if the script is killed by a signal, you can list signal numbers too - the portable ones are apparently `1 2 13 15`.  E.g. `2` is `SIGINT` which means that the database is removed if you hit Ctrl+C mid test:

```
trap 'rm -rf "$TEST_DB"' 0 1 2 13 15
```

With the clean-up outsourced to `trap`, you can add `set -e` at the start of the script so that the script fails if a command exists with non-zero status, and then there's no need to check `$?` or explicitly invoke `exit` at all - the script will exit with the status of the failed command.  And as a bonus, it will now also exit right away if `omindex` fails.

Generally you probably want to use `set -e` for most shell scripts - if there are particular commands that are allowed to fail you can use `command || true` to allow that.

xapian-applications/omega/omindexcheck.cc
barufa: Sorry, I am going to correct it.

xapian-applications/omega/omindexcheck.cc
barufa: Okay, I am going to change it

xapian-applications/omega/.gitignore
barufa: Sorry, I didn't notice it. I will sort it.

xapian-applications/omega/omindextest
barufa: I am going to apply these changes.

xapian-applications/omega/omindextest
barufa: Sorry, you are right. That is much simpler.

xapian-applications/omega/omindextest
barufa: Okay, I am going to change it.

xapian-applications/omega/omindextest
barufa: Okay, I am going to add `trap` and `set`

barufa: Okay, I am going to add `trap` and `set`

xapian-applications/omega/omindextest
barufa: Okay, I am going to change it.

barufa:
Hello, it seems that the only thing left to do is solving Travis problem (I add a few eBook files and they are causing problems with the check-path).
Can @ojwb @jaylett tell me how to fix that?
Or if there is something else to do?

jaylett:
@barufa I don't think there's a way of telling the check script to ignore those files, so we may just have to merge this one ignoring the error.

.travis.yml
barufa: It was a mistake. It should not be there.

xapian-maintainer-tools/xapian-check-patch
```diff
@@ -152,7 +153,13 @@ while (<>) {
...
 	$check_trailing = 1;
 	$doxygen_first_line = 1;
-	if ($ext eq 'cc') {
+        $check_end_new_line = 1;
+        if  ($fnm =~ m!xapian-applications/omega/testfiles!) {
+	    $check_space_tab = 0;
+	    $check_trailing = 0;
+	    $doxygen_first_line = 0;
+	    $check_end_new_line = 0;
+        } elsif ($ext eq 'cc') {
```
ojwb, line 162: I've cherry-picked this commit to master as dca1a763b4ca5a0185cbd3a1d17adb44ceca440b (with some whitespace fixes squashed in - there's a double space after `if` above, and some of the indentation in this hunk doesn't use tabs.)

xapian-applications/omega/omindexcheck.cc
```diff
@@ -0,0 +1,142 @@
...
+#include "worker.h"
+
+#include <xapian.h>
+#include <iostream>
+#include <string>
+#include <unordered_map>
+#include <cstdlib>
+
+#include "gnu_getopt.h"
+#include "hashterm.h"
```
ojwb, line 31: This header doesn't seem to be used.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -0,0 +1,142 @@
...
+#include "common/stringutils.h"
+
+using namespace std;
+
+typedef vector<string> testcase;
+
+unordered_map<string, testcase> tests;
```
ojwb, line 38: `tests` is only used in this file, so it's better to declare it as `static` - that potentially allows the compiler to optimise better as it then knows that it can't be modified by external functions (unless a pointer or reference to tests is passed to such a function, etc).  It also means fewer external symbols which helps to speed up linking.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -0,0 +1,142 @@
...
+
+#include <xapian.h>
+#include <iostream>
+#include <string>
+#include <unordered_map>
+#include <cstdlib>
+
+#include "gnu_getopt.h"
+#include "hashterm.h"
+#include "common/stringutils.h"
```
ojwb, line 32: We don't usually include the `common/` when including headers from there (I'm guessing you copied this code from `omindex-list.cc`, which gets this wrong too, and that's why `hashterm.h` is needlessly included).

xapian-applications/omega/omindexcheck.cc
```diff
@@ -0,0 +1,142 @@
...
+	    start = CONST_STRLEN("url=");
+	} else {
+	    start = data.find("\nurl=");
+	    if (start == string::npos)
+		continue;
+	    start += CONST_STRLEN("\nurl=");
+	}
+	url.assign(data, start, data.find('\n', start) - start);
+	start = url.find(current_dir) + current_dir.length();
+	url = url.substr(start, url.length());
```
ojwb, line 136: I would suggest simplifying this code by more carefully specifying the options to `omindex` in `omindextest`.  The only reason why need to remove `current_dir` here is because we specified it in the `--url="/$TEST_FILES"` option.  If instead we specify `--url=/` then the URL in the database will just be e.g. `/pdf/poppler.pdf`.

And with that change, we can rely on the URL term being short enough that it'll contain the full URL (if the URL term would be >= `MAX_SAFE_TERM_LENGTH` (which is 240) bytes then the end of it is hashed.

Taken together that means we can then just do `string url(term, 2);` to extract the relative path from the `U` prefix term (2 because we want to skip both the `U` and the ledding `/`) instead of having to fetch the document data and scan through it for `url=`.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -0,0 +1,142 @@
...
+	auto it = tests.find(url);
+	if (it != tests.end())
+	    succeed &= compare_test(it->second, doc, url);
```
ojwb, line 139: If there's no entry in `tests` found corresponding to `url` then we quietly ignore that and keep going - shouldn't that situation cause the tests to fail?

xapian-applications/omega/omindextest
```diff
@@ -0,0 +1,30 @@
...
+
+set -e
+
+srcdir=`echo "$0"|sed 's!/*[^/]*$!!'`
+TEST_FILES="$srcdir/testfiles/"
+TEST_DB="testdatabase"
+
+trap 'rm -rf "$TEST_DB"' 0 1 2 13 15
+
+./omindex --db "$TEST_DB" --url="/$TEST_FILES" "$TEST_FILES"
```
ojwb, line 29: I think this should use `--overwrite` too for robustness - we really want to always start from a new empty database or else the user may get unexpected problems if there's an old `testdatabase` left behind from a previous run.

barufa: I agree, I am going to add it.

xapian-applications/omega/omindexcheck.cc
barufa: You are right, that would make the code much more readable and clear. Moreover, we are avoiding using more than `MAX_SAFE_TERM_LENGTH` characters in the `url`.

xapian-applications/omega/omindexcheck.cc
barufa: I am going to fix the header part. It seems that I have some mistakes here.

xapian-applications/omega/omindexcheck.cc
barufa: Okay, I am going to add `static` to `test`

xapian-applications/omega/omindexcheck.cc
barufa: It seems that I made some mistakes in the includes, I am going to change them.

xapian-applications/omega/omindexcheck.cc
barufa: Not sure if we should fails here.
I think that the main(or first) idea of this pr was to test the code of some handlers and run them during the CI build. That is why I used conditional compilation flags to add(or not) the cases for each file. This way, I just compared the test cases if I was sure that the workers were available in the system. 

As there may be more than one way to index the same file in some systems, it is possible that some workers were not available and omindex indexed some files anyway. That means that some term could be missing and `omindexcheck` would fail.

Maybe we can think in something more "robust" and remove the conditional flags. That way we check that if a file was indexed by omindex, it will have at least the terms of `omindexcheck`. However, we will need to be really careful when choosing the terms as they should be always present in the documents.

ojwb: Oh I see - your thinking that (for example) libarchive isn't available and then we fall back to indexing the file a different way, which might not result in exactly the same terms.

We could probably just disable indexing of the extensions in question in `#else` of the conditions where we add the testcases.

My concerns here are really that it's currently possible to half-add a testcase and not realise that your new test isn't actually being checked, and that a bug which results in a document being misindexed might not be flagged.  I think ideally we would also check that all the testcases which got enabled were checked exactly once.

Or perhaps we should just let the documents be indexed anyway and this serves as a check of whatever method is used.  As you say, we'd might need to select the terms to check with more care.

I'm OK to leave this as-is for now in the interests of getting this merged and usable for building on.

barufa: Yes, I agree. Maybe some improvements could be done, but I think it is okay as a first check.

PR #281, opened 2019-08-18
Fix testcases to not use BackendManager_local

tanmaysachan:


ojwb:
As a general point, don't end the commit summary with a full stop - think of it as being like the title of a book or an email subject line.  Don't worry about fixing this here - I can tweak before I merge.

xapian-core/tests/api_percentages.cc
```diff
@@ -197,11 +196,9 @@ DEFINE_TESTCASE(pctcutoff5, backend) {
...
 
 // Regression test for bug fixed in 1.0.14.
-DEFINE_TESTCASE(topercent3, remote) {
-    BackendManagerLocal local_manager(test_driver::get_srcdir() + "/testdata/");
-    Xapian::Database db;
-    db.add_database(get_database("apitest_simpledata"));
-    db.add_database(local_manager.get_database("apitest_simpledata"));
+DEFINE_TESTCASE(topercent3, multi && remote) {
+    Xapian::Database db = get_database("apitest_simpledata",
+				       "apitest_simpledata");
```
ojwb: This works as a regression test for the bug (I double-checked by reintroducing the equivalent of the old bug fixed in 1.0.14), but actually there's no reason we need to limit this to `multi && remote` - it only needs `backend` in terms of features.  We generally enable testcases for all the backends they work with even if (as here) they're a regression test for a particular backend (or backend combination) - experience shows that tends to help shake out more bugs.

Also, there isn't actually a need to double up the source data now - before we were really just doing that as a simple way to have two shards from different backends, but now your enhancements to the "multi" backend give us that from a single source data file.

With the old bug reintroduced, this testcase also fails with just `Xapian::Database db = get_database("apitest_simpledata");` here.

I'll just add these tweaks as an extra commit after merging.

ojwb:
Rebased onto master and pushed.

PR #282, opened 2019-08-27
Remove use of backendmanager_local from netstats1

tanmaysachan:


ojwb:
The appveyor failure is due to a problem on git master, so please ignore that for now.

xapian-core/tests/api_db.cc
```diff
@@ -1687,14 +1687,54 @@ DEFINE_TESTCASE(sortrel1, backend) {
...
 }
 
-// Test network stats and local stats give the same results.
-DEFINE_TESTCASE(netstats1, remote) {
-    XFAIL_FOR_BACKEND("multi_glass_remoteprog_glass",
-		      "Multi remote databases are currently buggy");
+static void
+make_netstats1_db(Xapian::WritableDatabase& db, const string&)
+{
+    struct { Xapian::docid did; const string text; } content[] = {
```
ojwb: We don't need to be able to modify this, so if you make this `static const` and use `const char* text` then the compiler should be able to put this in the read-only data section at compile time.

If you use `string` here, that prevents that happening (at least with current compilers - I guess theoretically they could spot that compile-time construction is possible).

xapian-core/tests/api_db.cc
```diff
@@ -1687,14 +1687,54 @@ DEFINE_TESTCASE(sortrel1, backend) {
...
+	{1, "This is a test document used with the API test. "
+	    "This paragraph must be at least three lines (including the blank "
+	    "line) to be counted as a \"paragraph\"."},
+	{2, "This is a second simple data test, used to test multiple "
+	    "(inmemory anyway) databases.  The text in this file is"
+	    " unimportant, although I suppose it ought to include "
```
ojwb: It would be better to be consistent about whether the space goes before or after the line break when splitting a string.  Xapian convention is to put it before, and you've mostly done that, but not here and also not in a few other cases below.

Being inconsistent about this makes it much harder for the reader to easily see that there are no missing spaces where
the string has been broken up.

xapian-core/tests/api_db.cc
```diff
@@ -1687,14 +1687,54 @@ DEFINE_TESTCASE(sortrel1, backend) {
...
+	    "the standard word \"word\" in a few places."},
+	{3, "This file will be indexed by paragraph, and the simple query "
+	    "will search for the word \"word\".  Well expect the mset to"
+	    " contain two documents, including this paragraph and the fourth"
+	    ", below.  Since this paragraph uses the word \"word\" so much,"
```
ojwb: And please don't break between a word and following punctuation either.

xapian-core/tests/api_db.cc
```diff
@@ -1687,14 +1687,54 @@ DEFINE_TESTCASE(sortrel1, backend) {
...
+	    " useful information in it either."},
+	{5, "This paragraph only has a load of absolute rubbish, "
+	    "and nothing of any use whatsoever."},
+	{7, "This is the other paragraph with the word in the simple"
+	    " query in it.  For simplicity, all paragraphs are at least "
+	    "two lines, due to how the hacked up indexer works."},
+	{9, "This is another paragraph which wont be returned.  Well, "
+	    "not with the simple query, anyway."},
+	{11, "And yet another.  This one does mention banana splits, "
+	    "though, so cant be that bad."}
```
ojwb: This line is under-indented by one column.

xapian-core/tests/api_db.cc
```diff
@@ -1687,14 +1687,54 @@ DEFINE_TESTCASE(sortrel1, backend) {
...
+	indexer.set_document(doc);
+	indexer.index_text(i.text);
+	db.replace_document(i.did, doc);
+    }
+
+    db.commit();
+}
+
+// Test network stats and local stats give the same results.
+DEFINE_TESTCASE(netstats1, remote) {
```
ojwb: We should run this testcase for `generated` now - otherwise we aren't testing the "all local" case.

xapian-core/tests/api_db.cc
```diff
@@ -1705,63 +1745,28 @@ DEFINE_TESTCASE(netstats1, remote) {
...
-	TEST_EQUAL(mset.get_max_attained(), mset_alllocal.get_max_attained());
-	TEST_EQUAL(mset.size(), mset_alllocal.size());
-	TEST(mset_range_is_same(mset, 0, mset_alllocal, 0, mset.size()));
+	TEST_EQUAL(mset.get_matches_lower_bound(), 7);
+	TEST_EQUAL(mset.get_matches_upper_bound(), 7);
+	TEST_EQUAL(mset.get_matches_estimated(), 7);
+	TEST_EQUAL(mset.get_max_attained(), 1.445962071042388164);
+	TEST_EQUAL(mset.size(), 7);
+
+	vector<pair<Xapian::docid, double>> to_compare = {
```
ojwb: We don't need to be able to modify this, so better to make it an array and `static const` - i.e. `static const pair<Xapian::docid, double> to_compare[]` so the compiler can put it in the read-only data section at compile time.

xapian-core/tests/api_db.cc
```diff
@@ -1705,63 +1745,28 @@ DEFINE_TESTCASE(netstats1, remote) {
...
+	    make_pair(7, 1.445962071042388164),
+	    make_pair(3, 1.4140112748017070743),
+	    make_pair(1, 1.3747698831232337824),
+	    make_pair(5, 1.1654938419498412916),
+	    make_pair(9, 1.1654938419498412916),
+	    make_pair(4, 1.1543806706320836053),
+	    make_pair(2, 0.12268031290495594321)
```
ojwb: You can just write `{7, 1.445962071042388164},` etc here.  It should be equivalent, but a bit easier to read.

xapian-core/tests/harness/testutils.cc
```diff
@@ -109,6 +109,31 @@ mset_range_is_same_weights(const Xapian::MSet &mset1, unsigned int first1,
...
+	// FIXME: don't use internal macro here...
+	if (!TEST_EQUAL_DOUBLE_(i.get_weight(), to_compare[l].second)) {
+	    tout << "weights differ at item " << (l + 1) << " in range: "
+		    << i.get_weight() << " != " << to_compare[l].second << "\n";
+	    return false;
+	}
+	++i;
+    }
+    return true;
+}
```
ojwb: This only checks the weights - `to_compare[l].first` is ignored completely.

I think you wanted to create an overloaded version of `mset_range_is_same()` not `mset_range_is_same_weights()`.

xapian-core/tests/api_db.cc
```diff
@@ -1687,14 +1687,54 @@ DEFINE_TESTCASE(sortrel1, backend) {
...
+	indexer.index_text(i.text);
+	db.replace_document(i.did, doc);
+    }
+
+    db.commit();
+}
+
+// Test network stats and local stats give the same results.
+DEFINE_TESTCASE(netstats1, remote) {
     BackendManagerLocal local_manager(test_driver::get_srcdir() + "/testdata/");
```
ojwb: This variable is now unused and can be removed, and then `tests/harness/backendmanager_local.h` can go too, as this is the last use of it.

ojwb:
Thanks - rebased and pushed as 8d3aaa213657cb4120b2f7109152047b7acaa8c9.

PR #283, opened 2019-08-29
Make generated testcases run on remote multi db backends

tanmaysachan:


ojwb:
Thanks - I've fixed the bugs this reveals in e3863b019a10635a546ba95c1b3726147c686611 and then applied this change in aa75cf22ffc406b9e8bd1a12e971fd6e596c1353.

PR #284, opened 2020-01-03
Add synonym functionalities for remote

gauravaror:
This pull request implements synonyms for Remote databases.
Instead of creating new class remote_synonymtermlist. I preferred to use VectorTermList which is also used for synonyms in normal backends like glass.

One caveat is now in VectorTermList, the string might be taking less space then what we reserve as I changed the logic of how we pack the vector.

`qp_scale1` was failing with the timing issue when I was using `RemoteMetaTermList`. Now it seems to be fine when I ran tests again. Probably best to see in CI if it's an issue.


xapian-core/api/vectortermlist.h
```diff
@@ -66,14 +70,33 @@ class VectorTermList : public TermList {
...
-	    pack_string(data, *i);
+	    if (rare(prev.size() > 255))
+		prev.resize(255);
+	    const std::string& term = *i;
+	    size_t reuse = common_prefix_length(prev, term);
+	    data.append(1, char(reuse));
+	    pack_uint(data, term.size() - reuse);
+	    data.append(term, reuse, std::string::npos);
+	    prev = term;
 	}
```
ojwb: This makes the encoding for existing uses of `VectorTermList` more verbose, and also increases the cost of encoding - I'm not sold that's a good trade-off.  There are inherently differences between the sorted and unsorted cases which are highlighted by the changes you've had to make here.

Can't we just use `RemoteMetaDataTermList` for this?  That seems to handle essentially the same situation of a sorted `TermList` from a remote database.

xapian-core/backends/remote/remote-database.h
```diff
@@ -1,4 +1,6 @@
 /** @file remote-database.h
+    TermList * open_synonym_termlist(const string & term) const;
+    TermList * open_synonym_keylist(const string & prefix) const;
  *  @brief RemoteDatabase is the baseclass for remote database implementations.
```
ojwb: junk...

xapian-core/net/remote_protocol.rst
```diff
@@ -318,3 +318,33 @@ Reconstruct text
...
+Synonym Key List
+-----------------
+
+-  ``MSG_SYNONYMKEYLIST <word>``
+-  ``REPLY_SYNONYMKEYLIST [C<chars of previous term to reuse> S<string to append>]...``
+
+Add synonym
+-----------
+
+- ``MSG_ADDSYNONYM <word> <synonym>
```
ojwb: This is wrong - `pack_string()` is used on `word` so it's `S<word> <synonym>`.  Same for `MSG_REMOVESYNONYM`.

xapian-core/net/remote_protocol.rst
```diff
@@ -318,3 +318,33 @@ Reconstruct text
 
 -  ``MSG_RECONSTRUCTTEXT I<did> I<length> I<start_pos> I<end_pos> <prefix>``
 -  ``REPLY_RECONSTRUCTTEXT <text>``
+
+Synonym Term List
+-----------------
+
+-  ``MSG_SYNONYMTERMLIST <word>``
+-  ``REPLY_SYNONYMTERMLIST [C<chars of previous term to reuse> S<string to append>]...``
```
ojwb: This doesn't match what's actually implemented (the count is missing).  Same below.

xapian-core/net/remoteserver.cc
```diff
@@ -968,3 +983,91 @@ RemoteServer::msg_removespelling(const string & message)
     pack_uint_last(reply, wdb->remove_spelling(string(p, p_end - p), freqdec));
     send_message(REPLY_REMOVESPELLING, reply);
 }
+
+
```
ojwb: Extra blank line here.

xapian-core/tests/api_wrdb.cc
```diff
@@ -1589,6 +1589,50 @@ DEFINE_TESTCASE(nomoredocids1, writable) {
     return true;
 }
 
+DEFINE_TESTCASE(synonym_merge1, writable && synonyms) {
+    Xapian::WritableDatabase db = get_writable_database();
+    Xapian::WritableDatabase db1 = get_named_writable_database("db1");
+    Xapian::WritableDatabase db2 = get_named_writable_database("db3");
+    Xapian::WritableDatabase db3 = get_named_writable_database("db4");
```
ojwb: The strings passed for the last two don't match the variable names.

xapian-core/tests/api_wrdb.cc
```diff
@@ -1589,6 +1589,50 @@ DEFINE_TESTCASE(nomoredocids1, writable) {
...
+    db1.add_synonym("abc", "asd3");
+
+    db2.add_synonym("abc", "asd4");
+    db2.add_synonym("abc", "asd5");
+    db2.add_synonym("abc", "asd6");
+
+    db3.add_synonym("abc", "asd7");
+    db3.add_synonym("abc", "asd8");
+    db3.add_synonym("abc", "asd9");
+    db3.add_synonym("abc", "asd10");
```
ojwb: What exactly are you trying to test here?

In general a comment before each testcase explaining what the testcase is actually for is useful (a lot of older testcases lack this, but most newer ones include it).  I can see it's testing merging of synonyms across subdbs, but I'm not really sure about why in detail.

The synonyms from each subdb mostly form disjoint ranges (except that `asd10` string sorts between `asd1` and `asd2`) which seems like a lost opportunity to shake out problems in the merging code.  At least for the testcase as it is, it seems it would provide equally good coverage with just 2 sub databases.

xapian-core/tests/api_wrdb.cc
```diff
@@ -1589,6 +1589,50 @@ DEFINE_TESTCASE(nomoredocids1, writable) {
...
+	s = "|";
+	t = db_multi.synonyms_begin("abc");
+	while (t != db_multi.synonyms_end("abc")) {
+	    s += *t++;
+	    s += '|';
+	}
+	e = "|asd|asd1|asd10|asd2|asd3|asd4|asd5|asd6|asd7|asd8|asd9|";
+	TEST_STRINGS_EQUAL(s, e);
+	db.commit();
+    }
```
ojwb: Is repeating these tests before and after committing useful?  It feels like this is probably a cargo-cult copy from synonymitor1.

ojwb:
`qp_scale1` can be flaky as it's timing based, but it's disabled for builds from CI.

ojwb:
The build is failing in CI (looks like a missing `#include <xapian/error.h>`).

xapian-core/backends/remote/remote-database.cc
```diff
@@ -906,7 +907,7 @@ RemoteDatabase::open_synonym_termlist(const string& word) const
     string message;
     send_message(MSG_SYNONYMTERMLIST, word);
     get_message(message, REPLY_SYNONYMTERMLIST);
-    return new RemoteMetadataTermList(word, std::move(message));
+    return new VectorTermList(std::move(message));
```
ojwb: Oh, looking at the individual commits I see you tried `RemoteMetadataTermList` first.

Testing that version, I see it fails because `RemoteMetadataTermList::get_approx_size()` isn't implemented.  You can just return `1` there - that's what `GlassMetadataTermList::get_approx_size()` does:

```
Xapian::termcount
GlassMetadataTermList::get_approx_size() const
{
    // Very approximate!  This is only used to build a balanced or tree, so
    // at least we'll get an even tree by returning a constant answer.
    return 1;
}
```

Also I think you want `string()` instead of `word` when creating the `RemoteMetadataTermList` here - the first argument is any common prefix, but in general there isn't a common prefix here.

xapian-core/tests/api_wrdb.cc
gauravaror: I added it because I thought the merging code might behave a bit differently. But Since commit is tested at other places. it might be worth it to change this test case to just test merging of posting list and

xapian-core/tests/api_wrdb.cc
gauravaror: Yep! I will change to be just two dbs.

xapian-core/backends/remote/remote-database.h
gauravaror: ah! Strange It skipped my eyes.

xapian-core/api/vectortermlist.h
gauravaror: Yes, Since now RemoteMetaDataTermList isn't just handling only metadata but also synonyms and probably spellings in future when it's implemented.

Would it make sense to change the name from RemoteMetaDataTermList to something which depicts it's just a list of sorted terms (without term frequency and num_entries )?
maybe RemoteSortedTermList

gauravaror:
> qp_scale1 can be flaky as it's timing-based, but it's disabled for builds from CI.
Since it is always failing when I run it locally. 
I think it's probably worth increasing the factor to 4 or Disabling this test case for remote databases. 
Probably disabling is better as we don't want to dilute the test case for dbs it was intended testcase.

Noting that it was not enabled before for remote databases as this test case was marked synonym.

I am disabling these test cases for remote db. I can switch them back if you feel otherwise.

xapian-core/api/vectortermlist.h
ojwb: > Would it make sense to change the name from RemoteMetaDataTermList to something which depicts it's just a list of sorted terms (without term frequency and num_entries )?
maybe RemoteSortedTermList

I think we should rename, but let's do it as a separate step so the functional changes here can be seen without the noise of the renaming.

I'm also not sure what the best name is - `RemoteSortedTermList` is confusing as we already have `RemoteTermList` which is also a sorted list of terms.  Indeed almost all `TermList` subclasses are sorted (I think `VectorTermList` is the only one which isn't) and none of the others have `Sorted` in the name.

Probably the key difference compared with `RemoteTermList` is that `RemoteMetaDataTermList` is just terms without any frequency information.

xapian-core/backends/remote/remote-database.cc
```diff
@@ -900,6 +900,52 @@ RemoteDatabase::remove_spelling(const string & word,
...
+    return new RemoteMetadataTermList(string(), std::move(message));
+}
+
+TermList*
+RemoteDatabase::open_synonym_keylist(const string& prefix) const
+{
+    string message;
+    send_message(MSG_SYNONYMKEYLIST, prefix);
+    get_message(message, REPLY_SYNONYMKEYLIST);
+    return new RemoteMetadataTermList(string(), std::move(message));
```
ojwb: I think here you did want `prefix` not `string()` - the first argument is the common prefix on the entries, which means we can avoid the redundancy of sending that in the reply for every entry.

The difference compared to `RemoteDatabase::open_synonym_termlist()` is that there we're returning the synonyms for `word` (so there's no common prefix, at least not that the client knows in advance) whereas here we're returning all words with synonyms and which start with `prefix`.

xapian-core/net/remoteserver.cc
```diff
@@ -968,3 +983,83 @@ RemoteServer::msg_removespelling(const string & message)
...
+	++t;
+    }
+    send_message(REPLY_SYNONYMTERMLIST, reply);
+}
+
+void
+RemoteServer::msg_getsynonymkeylist(const string& message)
+{
+    Xapian::TermIterator t = db->synonym_keys_begin(message);
+    string reply, prev;
```
ojwb: And here `prev` should be initialised to `message`, like in `RemoteServer::msg_metadatakeylist()`.

xapian-core/tests/api_wrdb.cc
ojwb: `synonymitor1` essentially tests the inputs the merging code would get, so I think it's better to just test the merging here.

One thing this probably ought to test is that the merging handles a duplicate synonym suitably.

ojwb:
> Since it is always failing when I run it locally. I think it's probably worth increasing the factor to 4

The testcase is essentially comparing the time to parse a query string 5 times with the time to parse 5 copies of the query string concatenated - e.g. 5 times `foo` vs `foo foo foo foo foo ` (except actually it's a much longer query string).

We want this to scale linearly, but there's some noise in the timing so we need to allow for that.  But the factor we allow needs to be small enough that O(n.log(n)) scaling is caught so just increasing it is problematic.

> or Disabling this test case for remote databases.
> Probably disabling is better as we don't want to dilute the test case for dbs it was intended testcase.
>
> Noting that it was not enabled before for remote databases as this test case was marked synonym.
>
> I am disabling these test cases for remote db. I can switch them back if you feel otherwise.

I think this testcase needs rethinking - it's the flakiest we have and for some people seems to fail a lot (https://trac.xapian.org/ticket/764).  But that isn't something to do as part of this change.

In principle I'm OK to disable it for remote if it's just flaky, but I think it may actually be highlighting a performance issue - the check for multi-word synonyms in `queryparser/queryparser.lemony` does:

```
        string key;
        vector<Term*>::const_iterator begin = terms.begin();
        vector<Term*>::const_iterator i = begin;
        while (i != terms.end()) {
            TermIterator synkey(db.synonym_keys_begin((*i)->name));
            TermIterator synend(db.synonym_keys_end((*i)->name));
            if (synkey == synend) {
                // No multi-synonym matches.
                if (stopper && (*stopper)((*i)->name)) {
                    state->add_to_stoplist(*i);
                } else {
                    if (default_op_is_positional)
                        (*i)->need_positions();
                    subqs.push_back((*i)->get_query_with_auto_synonyms());
                }
                begin = ++i;
                continue;
            }
```

(and then more stuff is done with `synkey` if there is a match)

This potentially does a synonym keys iteration for every term in the query, which is OK locally but for remote use is pretty horrible.  I don't immediately see why this affects the O() behaviour though - I'll investigate a bit.

xapian-core/tests/api_queryparser.cc
```diff
@@ -2745,6 +2745,11 @@ qp_scale1_helper(const Xapian::Database &db, const string & q, unsigned n,
 // Regression test: check that query parser doesn't scale very badly with the
 // size of the query.
 DEFINE_TESTCASE(qp_scale1, synonyms) {
+    const char* xfail_msg = "Querying synonyms for multiple query words is "
+	               "expected to incur a higher time in remote DB";
+    XFAIL_FOR_BACKEND("remote", xfail_msg);
+    XFAIL_FOR_BACKEND("multi_remoteprog_glass", xfail_msg);
+    XFAIL_FOR_BACKEND("multi_glass_remoteprog_glass", xfail_msg);
```
ojwb: I don't think we want to `XFAIL` here - that means an `XPASS` result (which means "unexpected pass" and the testcase is treated as failing) if the time is OK and that's what I seem to reliably get:

```
$ ./runtest ./apitest qp_scale1 -v
Running test './apitest qp_scale1 -v' under eatmydata and valgrind
Running tests with backend "honey"...
Running tests with backend "inmemory"...
Running test: qp_scale1... POSSIBLE UNRELEASED MEMORY - RETRYING TEST ok
Running tests with backend "singlefile_glass"...
Running tests with backend "multi_glass"...
Running test: qp_scale1... ok
./apitest backend multi_glass: All 1 tests passed.
Running tests with backend "multi_glass_remoteprog_glass"...
Running test: qp_scale1... POSSIBLE UNRELEASED MEMORY - RETRYING TEST XPASS (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
Running tests with backend "multi_remoteprog_glass"...
Running test: qp_scale1... POSSIBLE UNRELEASED MEMORY - RETRYING TEST XPASS (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
Running tests with backend "remoteprog_glass"...
Running test: qp_scale1... XPASS (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
Running tests with backend "remotetcp_glass"...
Running test: qp_scale1... XPASS (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
./apitest total: 2 tests passed, 4 expected failures passed.
```

That's with your branch rebased onto current git master, but I don't think there are any changes on git master which would have affected this.

Probably just making the condition `synonyms && !remote` will do for now (unless we manage to solve this before merging).

If you run `./runtest ./apitest -v qp_scale1` without these, what's the output?  I'm wondering which part of the testcase is actually failing.  I tried printing out the times and most of the ratios of large/small are < 1 for me.

ojwb: If I suppress use of valgrind then I mostly seem to get XFAIL, with the occasional XPASS:

```
$ VALGRIND= ./runtest ./apitest qp_scale1 -v
Running test './apitest qp_scale1 -v' under eatmydata
Running tests with backend "honey"...
Running tests with backend "inmemory"...
Running test: qp_scale1... ok
Running tests with backend "singlefile_glass"...
Running tests with backend "multi_glass"...
Running test: qp_scale1... ok
./apitest backend multi_glass: All 1 tests passed.
Running tests with backend "multi_glass_remoteprog_glass"...
Running test: qp_scale1... XFAIL (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
small=0.010077s, large=0.004406s
small=0.048833s, large=0.095696s
small=0.056213s, large=0.150561s
api_queryparser.cc:2742: (time2) < (time1 * 2.15)
Evaluates to: 0.150561 < 0.120858


Running tests with backend "multi_remoteprog_glass"...
Running test: qp_scale1... XFAIL (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
small=0.00565s, large=0.004249s
small=0.101363s, large=0.193138s
small=0.105056s, large=0.311184s
api_queryparser.cc:2742: (time2) < (time1 * 2.15)
Evaluates to: 0.311184 < 0.22587


Running tests with backend "remoteprog_glass"...
Running test: qp_scale1... XPASS (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
small=0.009589s, large=0.003246s
small=0.052061s, large=0.082922s
small=0.065391s, large=0.117193s
small=0.058717s, large=0.03465s

Running tests with backend "remotetcp_glass"...
Running test: qp_scale1... XFAIL (Querying synonyms for multiple query words is expected to incur a higher time in remote DB)
small=0.007938s, large=0.004389s
small=0.085934s, large=0.151948s
small=0.113013s, large=0.290744s
api_queryparser.cc:2742: (time2) < (time1 * 2.15)
Evaluates to: 0.290744 < 0.242978


./apitest total: 2 tests passed, 1 expected failures passed, 3 expected failures.
```

So for me it seems to always be the third part that fails - that's parsing repetitions of `foo` with a synonym `foo` -> `bar` and no multi-word synonyms (but with `FLAG_AUTO_MULTIWORD_SYNONYMS` on).

If I switch that part to using `FLAG_AUTO_SYNONYMS` then I seem to reliably get `XPASS`, so it seems that checking for but not finding multi-word synonyms isn't scaling well - that's something I should be able to investigate.

The second sub-testcase seems to reliably take longer in the larger version, but that changes to usually being quicker with just `FLAG_AUTO_SYNONYMS`.  The fourth sub-testcase is reliably quicker in the larger version though, and that actually uses multi-word synonyms.  That reinforces that it's probably not finding multi-word synonyms that's the issue here.

gauravaror: Third one consistently fails for me aswell:
```
Running tests with backend "multi_glass_remoteprog_glass"...
Running test: qp_scale1... XFAIL (Querying synonyms for multiple query words is expected to incur a higher cost in remote DB)
small=0.010694s, large=0.009417s
small=0.040276s, large=0.056852s
small=0.030887s, large=0.083621s
api_queryparser.cc:2742: (time2) < (time1 * 2.15)
Evaluates to: 0.083621 < 0.066407


Running tests with backend "multi_remoteprog_glass"...
Running test: qp_scale1... XFAIL (Querying synonyms for multiple query words is expected to incur a higher cost in remote DB)
small=0.011954s, large=0.010813s
small=0.085493s, large=0.102148s
small=0.054816s, large=0.155688s
api_queryparser.cc:2742: (time2) < (time1 * 2.15)
Evaluates to: 0.155688 < 0.117854


Running tests with backend "remoteprog_glass"...
Running test: qp_scale1... XFAIL (Querying synonyms for multiple query words is expected to incur a higher cost in remote DB)
small=0.003145s, large=0.002939s
small=0.036848s, large=0.058566s
small=0.039391s, large=0.091513s
api_queryparser.cc:2742: (time2) < (time1 * 2.15)
Evaluates to: 0.091513 < 0.0846907


Running tests with backend "remotetcp_glass"...
Running test: qp_scale1... XFAIL (Querying synonyms for multiple query words is expected to incur a higher cost in remote DB)
small=0.010367s, large=0.009933s
small=0.073369s, large=0.088867s
small=0.052924s, large=0.132596s
api_queryparser.cc:2742: (time2) < (time1 * 2.15)
Evaluates to: 0.132596 < 0.113787
```

gauravaror: Is there a documentation for XFAIL?
I was trying to look for it but couldn't find. I think I tried developer guide and few other places. I was bit confused on usage of XFAIL myself.

gauravaror: Yes, removing `FLAG_AUTO_SYNONYMS` seems to fix the test case and return `XPASS`. I will remove XFAIL part for now and probably it's worth wait if it's possible for you to investigate this.

ojwb: > Is there a documentation for XFAIL?
> I was trying to look for it but couldn't find. I think I tried developer guide and few other places. I was bit confused on usage of XFAIL myself.

There's a note which covers how it works and a common use case here: https://xapian-developer-guide.readthedocs.io/en/latest/contributing/contributing-changes.html#automated-tests

There's also a documentation comment where it's defined in the code: xapian-core/tests/harness/testsuite.h

ojwb:
> This potentially does a synonym keys iteration for every term in the query, which is OK locally but for remote use is pretty horrible. I don't immediately see why this affects the O() behaviour though - I'll investigate a bit.

I think it's not the O() behaviour, but that all those remote operations increases the risk that the time taken varies a lot.

I've optimised this code and pushed to master, but CI is showing a problem so I really need to sort that out before merging this PR.  I have this branch with a few clean-ups ready to go though.

xapian-core/net/remote_protocol.rst
```diff
@@ -318,3 +318,33 @@ Reconstruct text
...
+
+Synonym Key List
+-----------------
```
ojwb: RST allows the heading underline to be longer like this, but it's nicer to align it.  I'll tweak that as I merge.

xapian-core/net/remoteserver.cc
```diff
@@ -255,6 +255,21 @@ RemoteServer::run()
 		case MSG_RECONSTRUCTTEXT:
 		    msg_reconstructtext(message);
 		    continue;
+		case MSG_SYNONYMTERMLIST:
+		    msg_getsynonymtermlist(message);
+		    continue;
+		case MSG_SYNONYMKEYLIST:
+		    msg_getsynonymkeylist(message);
```
ojwb: The other method names match the message names, but these two add `get`.  I'll adjust that as I merge.

xapian-core/tests/api_wrdb.cc
```diff
@@ -1589,6 +1589,41 @@ DEFINE_TESTCASE(nomoredocids1, writable) {
...
+    Xapian::WritableDatabase db = get_writable_database();
+    Xapian::WritableDatabase db1 = get_named_writable_database("db1");
+
+    db.add_synonym("abc", "asd");
+    db.add_synonym("abc", "asd2");
+    db.add_synonym("abc", "asd4");
+    db.add_synonym("abc", "asd6");
+    db.commit();
+
+    db1.add_synonym("abc", "asd1");
```
ojwb: I'll add `asd2` to this shard too so we test that merging correctly handles the same synonym coming from more than one shard.

PR #285, opened 2020-02-18
Add automatic boost option cgi ops and maybe

gauravaror:
Initial implementation of automatic boost for terms provided using `AUTOBOOST`  CGI parameters.

specify one or more terms to be considered in weight contribution for a document without affecting selection using Query operation `OP_AND_MAYBE`.

Most of the weighting schemes like bm25, tf-idf don't use wdf as 1 in weighted contexts for boolean terms. This considers wdf = 1 for boolean terms in weighted contexts.

This is implemented using `OP_AND_MAYBE` query.


xapian-applications/omega/omega.cc
```diff
@@ -364,6 +364,45 @@ try {
...
+		    filters.append(boost_term, b, e + 1 - b);
+		    b = e;
+		    e = boost_term.find(filter_sep, b + 1);
+		}
+		filters.append(boost_term, b, string::npos);
+	    }
+	    filters += filter_sep;
+	    // old_filters predates 'AUTOBOOST' terms, so if there are
+	    // 'AUTOBOOST' terms this is definitely a different query.
+	    old_filters.clear();
```
gauravaror: Most of this is a common code for B, N, AUTOBOOST. I wonder if we should consider replacing it with a lambda function.
Only difference is `!` is added for `N` cgi param.

xapian-applications/omega/omega.cc
```diff
@@ -364,6 +364,45 @@ try {
...
+	    const string& v = i->second;
+	    if (!v.empty() && C_isalnum(v[0])) {
+		add_boost_term(v);
+		auto_boost.push_back(v);
+	    }
+	}
+	sort(auto_boost.begin(), auto_boost.end());
+	vector<string>::const_iterator j;
+	for (j = auto_boost.begin(); j != auto_boost.end(); ++j) {
+	    const string& boost_term = *j;
```
gauravaror: Moreover, since filters are mostly used for reloading if the query has changed. Is it necessary to add a symbol to distinguish terms added with `AUTOBOOST` and with `B`? I don't think it's required but would be great to know your views.

xapian-applications/omega/omegatest
```diff
@@ -1451,6 +1451,36 @@ printf '$hitlist{$list{$terms{T},|}/$field{tag1}|$field{tag2}|$field{tag3}}' > "
...
+boost=five
+boost=six
+pfx_boost=ten
+
+END
+
+# Test CGI param AUTOBOOST boost boolean terms.
+# Currently this tests don't work for following weighting schemes
+# dicecoeff dph ineb2 lm inl2 pl2plus pl2 tradweight
+for i in coord tfidf bm25 bm25plus
```
gauravaror: I got the implementation for few initial weighting schemes. I can edit try doing similar for other weightings schemes if this is a correct approach.

xapian-applications/omega/omegatest
```diff
@@ -1451,6 +1451,36 @@ printf '$hitlist{$list{$terms{T},|}/$field{tag1}|$field{tag2}|$field{tag3}}' > "
...
+boost=four
+boost=five
+boost=six
+pfx_boost=ten
+
+END
+
+# Test CGI param AUTOBOOST boost boolean terms.
+# Currently this tests don't work for following weighting schemes
+# dicecoeff dph ineb2 lm inl2 pl2plus pl2 tradweight
```
gauravaror: `bm25plus` weighting scheme currently works without changes because it adds param_delta.
```
133     RETURN(termweight * ((param_k1 + 1) * wdf_double / denom + param_delta));
```

But probably we should be consistent and assign wdf_double to 1.0 when its 0.0

xapian-core/weight/tfidfweight.cc
```diff
@@ -184,21 +184,22 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
 {
     switch (c) {
 	case 'b':
-	    if (wdf == 0) return 0;
+	    if (wdf == 0) return 1.0;
```
gauravaror: Use less anyways `1.0` is returned always.

xapian-core/weight/tfidfweight.cc
```diff
@@ -184,21 +184,22 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
...
 	    return 1.0;
 	case 's':
+	    if (wdf == 0) return 1.0;
 	    return (wdf * wdf);
 	case 'l':
-	    if (wdf == 0) return 0;
+	    if (wdf == 0) return 1.0;
 	    return (1 + log(double(wdf)));
```
gauravaror: Without the current check not sure of `wdf==0`. Implementation before looks wrong as log(0) would be -inf.

```
-inf
```

xapian-core/api/queryinternal.h
```diff
@@ -72,6 +76,10 @@ class QueryTerm : public Query::Internal {
 
     void serialise(std::string & result) const;
 
+    void set_weight(const Xapian::Weight& wt_) noexcept {
+	weight.reset(wt_.clone());
+    }
```
ojwb, line 81: `Query` objects are meant to act as if they are immutable, so a setter method like this doesn't really work properly.

I say "as if" - mostly they actually are immutable, but methods like `&=` are optimised for a `Query` with a ref count of 1 on its internals to modify in place so `&=` can be used to assemble a query in a loop and the loop can be `O(n)` overall.

I think this would need to be done via something like an `OP_WEIGHTING_SCHEME` operator which takes a subquery and a `Xapian::Weight` object to use for that subquery.

gauravaror: Sure, that sounds better. So, we want to have a new Query operator
Query(const std::string & term,
           Xapian::Weight* weight,
           Xapian::termcount wqf = 1,
           Xapian::termpos pos = 0)

which internally instantiates a QueryTerm object with Weight object set via the constructor.

xapian-applications/omega/omegatest
```diff
@@ -663,6 +663,7 @@ testcase '.~~X' DOCIDORDER=A # Buggy, but kept for compatibility.
 testcase '.~~D' DOCIDORDER=D
 testcase '.~~' DOCIDORDER=X # Buggy, but kept for compatibility.
 testcase '.~~' DOCIDORDER=x # Buggy, but kept for compatibility.
+testcase 'and,40.000000~five,1.000000~work,34.344200~.~~' BOOST=1,five BOOST=40,and BOOST=34.3442,work
```
gauravaror, line 666: Would it be better to use `:` here :thinking: ?

xapian-core/include/xapian/query.h
```diff
@@ -375,6 +375,22 @@ class XAPIAN_VISIBILITY_DEFAULT Query {
...
+     *  	    to calculate weight for term.
+     *  @param wqf  The within-query frequency. (default: 1)
+     *  @param pos  The query position.  Currently this is mainly used to
+     *		    determine the order of terms obtained via
+     *		    get_terms_begin(). (default: 0)
+     */
+    Query(const std::string& term,
+	  const Xapian::Weight*  wt,
+	  Xapian::termcount wqf = 1,
+	  Xapian::termpos pos = 0);
```
gauravaror, line 392: We probably need to update the Query Description of QueryTerm::get_description to show we added an external weight

I am probably thinking of something like this:
```
+    if (weight != 0) {
+       desc  += "^coord";
+    }

```

PR #286, opened 2020-02-18
Add Thai, Lana Tai and Pali codepoints to codepoint_is_cjk() test

rsto:
The Thai, Lana Tai and Pali codepoints currently are not detected by the CJK tokenizer, and text in these scripts are only word-segmented around spaces and punctuation.

This pull request fixes that.

### DIFF
--- a/xapian-core/tests/api_termgen.cc
+++ b/xapian-core/tests/api_termgen.cc
@@ -184,6 +184,9 @@ static const test test_simple[] = {
     // would split this differently as '申込み ！月額 円'
     { "", "申込み！月額円", "み[2] 円[4] 月額[3] 申込[1]" },
 
+    // Thai word segmentation
+    { "", "โดยตั้งใจ", "ตั้งใจ[2] โดย[1]" },
+
     // Test set_stemming_strategy():
     { "stem=en,none,!cjkwords",
 	  "Unstemmed words!", "unstemmed[1] words[2]" },
--- a/xapian-core/queryparser/cjk-tokenizer.cc
+++ b/xapian-core/queryparser/cjk-tokenizer.cc
@@ -53,6 +53,8 @@ CJK::codepoint_is_cjk(unsigned p)
     // Array containing the last value in each range of codepoints which
     // are either all CJK or all non-CJK.
     static const unsigned splits[] = {
+	// 0E00..0E7F; Thai, Lanna Tai, Pali
+	0x0E00 - 1, 0x0E7F,
 	// 1100..11FF; Hangul Jamo
 	0x1100 - 1, 0x11FF,
 	// 2E80..2EFF; CJK Radicals Supplement


xapian-core/tests/api_termgen.cc
```diff
@@ -184,6 +184,9 @@ static const test test_simple[] = {
     // would split this differently as '申込み ！月額 円'
     { "", "申込み！月額円", "み[2] 円[4] 月額[3] 申込[1]" },
 
+    // Thai word segmentation
+    { "", "โดยตั้งใจ", "ตั้งใจ[2] โดย[1]"},
```
ojwb: Tiny nit - this is missing a space before the closing `}` compared to other entries (and to the opening `{`).

ojwb:
I don't know much about Thai, but this looks reasonable to me.

It does mean that `FLAG_CJK_WORDS` is now rather misnamed.  It hasn't been in a stable release series so now is a good time to address that.  Something like `FLAG_SEGMENTATION` might be a better name.

This change also affects `FLAG_CJK_NGRAMS`, which probably makes sense but also means that is similarly misnamed.  That one is present in 1.4.x but we could rename and keep a compatibility alias.  Perhaps `FLAG_NGRAMS` or `FLAG_SEGMENTATION_NGRAMS`.

I'm happy to merge without renaming the flags, but we should make sure we decide what to do before the next stable release series.

rsto:
I'd vouch for FLAG_NGRAMS and FLAG_WORD_SEGMENTATION.

I'll be happy to come up with a patch for that, but I'll most likely not get to it this week. In addition to renaming the enums, we should also take care to rename related variable and argument names. For example, I think there's an argument `cjk_flags` somewhere in TermGenerator.

xapian-core/tests/api_termgen.cc
rsto: Fixed and updated the commit. Sorry for that.

ojwb:
Thanks for the update - now merged.

> I'll be happy to come up with a patch for that, but I'll most likely not get to it this week.

That sounds fine.  It just needs to happen before the next stable release series (and not get forgotten!)

> In addition to renaming the enums, we should also take care to rename related variable and argument names. For example, I think there's an argument `cjk_flags` somewhere in TermGenerator.

And update the documentation to make it clear these work for languages where some sort of wor

The only other thing in the API is `SNIPPET_CJK_NGRAM` and `SNIPPET_CJK_WORDS`, but quest, omega and the bindings expose these flags with "cjk" names.  There are also various other internal things that it would be good to update for clarity.  I think a `git grep -i cjk` and check through the results should catch anything that might need an update (`xapian-core/unicode/UnicodeData.txt` is the Unicode data and `ChangeLog` and `NEWS` files are historical information which leaves 1520-1231-6-8-1-1 = 273 references, but I think it's a lot less work than that might suggest).

rsto:
> I'll be happy to come up with a patch for that, but I'll most likely not get to it this week.

@ojwb I had started on renaming the flags a few weeks ago, but it turned out to be more work than I originally anticipated. When is the next stable release planned, and on which date should we have a patch ready to review?

ojwb:
We don't have a date estimate yet.  I'd really like to get the new backend (honey) functional for indexing first as with that done a generally useful non-GPL build becomes possible.

ojwb:
@rsto New release plan is to not stall the next stable release on getting honey update working: https://lists.xapian.org/pipermail/xapian-discuss/2023-March/009961.html

I don't have a firm timeline currently, but I'd guess it'll probably be a month or two elapsed time.

Where did you get to with the flag renaming to drop the now-misleading CJK references?  If you're up for completing it that's great; otherwise I can work on it, but if you've already done significant work maybe I should start from where you got to.

I also noticed that the script used by the Lao language should probably be added to the list (codepoints U+0E80 to U+0EFF) - https://en.wikipedia.org/wiki/Lao_script says:

> Spaces for separating words and punctuation were traditionally not used, but space is used and functions in place of a comma or period.

and:

> Lao is traditionally not written with spaces between words. Spaces are reserved for ends of clauses or sentences.

Similarly Tibetan (U+0F00 to U+0FFF) - https://en.wikipedia.org/wiki/Tibetan_script says:

> In the Tibetan script, the [syllables](https://en.wikipedia.org/wiki/Syllable) are written from left to right. Syllables are separated by a tsek (་); since many Tibetan words are monosyllabic, this mark often functions almost as a space. Spaces are not used to divide words.

I don't really know anything specific about these languages/scripts apart from what I've read on wikipedia though, so I was wondering if you have any useful insights that would make me more confident about adding them.  And indeed if you know of any other scripts which we should consider including.

ojwb:
More support for adding Lao from https://en.wikipedia.org/wiki/Text_segmentation:

> Languages which do not have a trivial word segmentation process include Chinese, Japanese, where [sentences](https://en.wikipedia.org/wiki/Sentences) but not words are delimited, [Thai](https://en.wikipedia.org/wiki/Thai_language) and [Lao](https://en.wikipedia.org/wiki/Lao_language), where phrases and sentences but not words are delimited, and [Vietnamese](https://en.wikipedia.org/wiki/Vietnamese_language), where syllables but not words are delimited.

Also from ICU docs https://unicode-org.github.io/icu/userguide/boundaryanalysis/#dictionary-based-breakiterator says:

> ICU provides dictionary support for word boundaries in Chinese, Japanese, Thai, Lao, Khmer and Burmese.

This supports adding Lao, but also suggests U+1780 to U+17FF and U+19E0 to U+19FF for Khmer, and U+1000 to U+109F, U+AA60 to U+AA7F, U+A9E0 to U+A9FF for Burmese.

https://github.com/unicode-org/icu/tree/main/icu4c/source/data/brkitr/dictionaries is the actual dictionary files, so it seems the list in the documentation is up to date.

This suggests there's no ICU dictionary for Tibetan, so (currently at least) maybe that's not useful to add.

rsto:
@ojwb I did not get to renaming the flag and then lost track of this task. I don't have more insight to add to your findings on Lao, Khmer, Burmese and potentially other languages. The Unicode BreakIterator dictionary looks to me to be a good basis to determine which code points hint at using the Unicode iterator for text segmentation.

I do not have any work planned in that area, but if you need support or would like to split the effort I might be able to allocate some time for some hacking.

ojwb:
Mostly I was wanting to avoid wasting time and effort on redoing something if you'd already mostly done it.

I've eliminated the references to CJK where they're now wrong in 13295e9142f56911d4876fb92271df348759c34b.

`FLAG_CJK_NGRAM` and `SNIPPET_CJK_NGRAM` are still supported, but the preferred names are now `FLAG_NGRAMS` and `SNIPPET_NGRAMS` (which I'll add in for 1.4.23 too to aid people migrating to them).

I've simply renamed `FLAG_CJK_WORDS` to `FLAG_WORD_BREAKS` and `SNIPPET_CJK_WORDS` to `SNIPPET_WORD_BREAKS` since these haven't been in a stable release yet.  Hopefully that's not going to create too much hassle for you.

> The Unicode BreakIterator dictionary looks to me to be a good basis to determine which code points hint at using the Unicode iterator for text segmentation.

Yes - I'm more confident about which languages to add since I found the dictionaries.  I'll add those.

I am wondering whether we should perhaps have separate lists of languages/scripts for the n-grams and word break finding though.  The word break finding is only really useful to try when ICU has support, but n-grams are perhaps more widely useful (e.g. we could enable n-grams for Tibetan).  Or perhaps they're really only useful in languages where words are typically only one or two characters - we only generate unigrams and bigrams which may result in too many false matches when words are longer (consider n-gram search applied to English text...)

And perhaps we should consider the n-gram support as a legacy feature which would probably argue for applying it as we do in 1.4.x, i.e. only to CJK (or for keeping things simple internally and just having one list!)

I don't have any direct experience of using either CJK n-grams or the ICU word breaking though.  If you have any useful insights that'd be helpful.  I should probably ask on the mailing list too.

> I do not have any work planned in that area, but if you need support or would like to split the effort I might be able to allocate some time for some hacking.

ojwb:
> I do not have any work planned in that area, but if you need support or would like to split the effort I might be able to allocate some time for some hacking.

It'd be useful if you could check the updated code works with your code before we actually make a stable release with it, as problems which require ABI changes to solve will be much harder to address then.  Probably worth waiting until the dust settles though.

rsto:
I'll take note to test the latest main branch in a couple of weeks. Please ping me if you would like me to test earlier.

PR #287, opened 2020-02-22
#801  fix ,mono toolchain bug where it fails to parse a newer terminfo file format.

dipanshu124:
While Building c# on Ubuntu 18.10 ,an error occurrs which is resolved by setting terminal to an empty value.

### DIFF
--- a/xapian-bindings/csharp/Makefile.am
+++ b/xapian-bindings/csharp/Makefile.am
@@ -92,10 +92,10 @@ $(ASSEMBLY).dll: $(XAPIAN_SWIG_CS_SRCS) AssemblyInfo.cs XapianSharp.snk
 ## this is the first time we're signing, the key in the file is used and
 ## then stored in the key container.  From then on, the key in the container
 ## is used.
-	$(CSC) -unsafe -target:library -out:$(ASSEMBLY).dll \
-	    -keyfile:XapianSharp.snk -keycontainer:"Xapian Signing Key" \
-	    `for f in $(XAPIAN_SWIG_CS_SRCS); do if test -f "$$f"; then echo $$f; else echo $(srcdir)/$$f ; fi ; done` \
-	    AssemblyInfo.cs
+	TERM= $(CSC) -unsafe -target:library -out:$(ASSEMBLY).dll \
+	          -keyfile:XapianSharp.snk -keycontainer:"Xapian Signing Key" \
+	          `for f in $(XAPIAN_SWIG_CS_SRCS); do if test -f "$$f"; then echo $$f; else echo $(srcdir)/$$f ; fi ; done` \
+	          AssemblyInfo.cs
 
 SWIG_GENERATED = xapian_wrap.cc xapian_wrap.h $(XAPIAN_SWIG_CS_SRCS)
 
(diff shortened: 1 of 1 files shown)


xapian-bindings/csharp/Makefile.am
```diff
@@ -92,10 +92,10 @@ $(ASSEMBLY).dll: $(XAPIAN_SWIG_CS_SRCS) AssemblyInfo.cs XapianSharp.snk
...
 ## then stored in the key container.  From then on, the key in the container
 ## is used.
-	$(CSC) -unsafe -target:library -out:$(ASSEMBLY).dll \
-	    -keyfile:XapianSharp.snk -keycontainer:"Xapian Signing Key" \
-	    `for f in $(XAPIAN_SWIG_CS_SRCS); do if test -f "$$f"; then echo $$f; else echo $(srcdir)/$$f ; fi ; done` \
-	    AssemblyInfo.cs
+	TERM= $(CSC) -unsafe -target:library -out:$(ASSEMBLY).dll \
+	          -keyfile:XapianSharp.snk -keycontainer:"Xapian Signing Key" \
+	          `for f in $(XAPIAN_SWIG_CS_SRCS); do if test -f "$$f"; then echo $$f; else echo $(srcdir)/$$f ; fi ; done` \
+	          AssemblyInfo.cs
```
ojwb: The extra indent added to these lines doesn't seem to be related to the change you're actually trying to make here (working around a mono bug).

It also seems wrong - we generally indent wrapped lines in makefiles by 4 columns, not 10.

ojwb:
The CI failure on appveyor is an issue already present on master - I have a fix mostly ready for it, but just ignore that for now.

Aside from the whitespace issue noted above, the code change looks good as far as it goes.

However, I wonder if there are more cases which also need this workaround - have you tried running the tests (`make check`) for example? That also runs compiled C# code.

And installing/uninstalling uses `gacutil` which appears to be implemented in C#.  You can install/uninstall to a temporary staging directory to test this without actually installing files to system locations or needing root access:

```
STAGING=`mktemp -d`
make install DESTDIR="$STAGING"
make uninstall DESTDIR="$STAGING"
rm -rf "$STAGING"
```

Your commit message could be better too: "_#801 bug fix_" doesn't actually tell the reader what was fixed unless they go and look up the ticket.  The reader ought to know what the commit is doing just from the commit message.  External references such as the ticket should be included, but they should supplement that rather than replacing it.

See: https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages

dipanshu124:
When I ran _**make**_ _**check**_ It passed all tests.But while _**gacutil**_ , it showed  error -
[ERROR] FATAL UNHANDLED EXCEPTION: System.TypeInitializationException: The type initializer for 'System.Console' threw an exception. ---> System.TypeInitializationException: The type initializer for 'System.ConsoleDriver' threw an exception. ---> System.Exception: Magic number is wrong: 542

For complete run  [https://pastebin.com/g2dBRJ9C](url)

dipanshu124:
Sorry for wrong indentation.Is there a way  to change commit message now?

ojwb:
> Is there a way to change commit message now?

The commit message (and other metadata) is included in the commit hash, so changing the commit message changes the commit hash (this is a deliberate feature as it makes it hard to tamper with the commit message) - that means to actually change the commit message you'd have to "force push", which you don't want to be doing with branches you're collaborating on with others (like a PR branch) unless everyone involved agrees that's the right thing to do.

For a PR branch people _might_ agree, but once merged they're very unlikely to, so it's a good idea to take a bit of time and care writing commit messages in the first place.  It might be tempting to think of them as something to just bang out quickly after you've done the hard work, but really the commit message is an important communication to future developers looking at the code, perhaps including your future self after enough time has passed that you can't entirely remember the details of this change.

An alternative to rewriting history is just to put the replacement commit message in fixup commit or even a comment on the PR, and make it clear what you intend - then the person merging can replace it as they merge.  It's extra work for them though, so better to write a good commit message to start with.

Are you OK to agree to the patch licensing requirements?

https://xapian-developer-guide.readthedocs.io/en/latest/contributing/contributing-changes.html#license-grant

dipanshu124:
> Are you OK to agree to the patch licensing requirements?
>>
YES.

dipanshu124:

ojwb:
Thanks, squashed and merged.

PR #288, opened 2020-02-26
Allow  command to fail with an option to run failover script

gauravaror:
Implement option for allow failure in $include command in Omega

This adds a change to allow include to run an optional failover script if the provided template is non-existant.

### DIFF
--- a/xapian-applications/omega/docs/omegascript.rst
+++ b/xapian-applications/omega/docs/omegascript.rst
@@ -1018,8 +1018,9 @@ $if{COND[,THEN[,ELSE]]}
 
         The ability to omit ``THEN`` was added in Omega 1.4.15.
 
-$include{FILE}
-	include another OmegaScript file
+$include{FILE,[FAILOVER_SCRIPT]}
+        include another OmegaScript file ``FILE``, runs optional
+        OmegaScript ``FAILOVER_SCRIPT`` on failure.
 
 $switch{EXPR,CASE1,VALUE1,[CASE2,VALUE2]...[,DEFAULT]}
         first evaluates ``EXPR``, and then evaluates ``CASE1``, ``CASE2``, ...
--- a/xapian-applications/omega/omegatest
+++ b/xapian-applications/omega/omegatest
@@ -1451,6 +1451,17 @@ printf '$hitlist{$list{$terms{T},|}/$field{tag1}|$field{tag2}|$field{tag3}}' > "
 testcase 'Tone/one|two|three' B=Tone
 testcase 'Tthree|Ttwo/one|two|three' B=Ttwo B=Tthree
 
+
+# Test failover script is run when the failover script is provided
+# and non-existent template file is passed in $include
+printf '$include{non_existant.template,$set{data,replacement_for_template}}$opt{data}' > "$TEST_TEMPLATE"
+testcase 'replacement_for_template' B=Tone
+
+# Test exception is thrown when non-existent template file is passed in
+# $include and optional failover script is not provided
+printf '$include{non_existant.template}$opt{data}' > "$TEST_TEMPLATE"
+testcase "Exception: Couldn't read format template 'non_existant.template' (No such file or directory)" B=Tone
+
 rm "$OMEGA_CONFIG_FILE" "$TEST_INDEXSCRIPT" "$TEST_TEMPLATE"
 rm -rf "$TEST_DB"
 if [ "$failed" = 0 ] ; then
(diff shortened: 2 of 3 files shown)


xapian-applications/omega/docs/omegascript.rst
```diff
@@ -1018,8 +1018,9 @@ $if{COND[,THEN[,ELSE]]}
 
         The ability to omit ``THEN`` was added in Omega 1.4.15.
 
-$include{FILE}
-	include another OmegaScript file
+$include{FILE,[FAILOVER_SCRIPT]}
+        include another OmegaScript file ``FILE``, runs optional
+        OmegaScript ``FAILOVER_SCRIPT`` on failure.
```
ojwb, line 1023: The `,` should be omitted when the second argument is, so that should be inside the `[`...`]` too - `$include{non_existent.template,}` will quietly expand to nothing!

xapian-applications/omega/query.cc
```diff
@@ -1789,7 +1789,15 @@ eval(const string &fmt, const vector<string> &param)
...
-		value = eval_file(args[0]);
+		try {
+		    value = eval_file(args[0]);
+		} catch (...) {
+		    if (args.size() == 2) {
+			value = eval(args[1], param);
+		    } else {
+			throw;
+		    }
+		}
```
ojwb, line 1800: I don't think this should work via exception handling.  One issue is that this treats any exception the same, whereas I think really the case we're interested in here is when the file isn't present (or maybe is there but not readable).  Also it would be reasonable to use this to include an optional template (kind of like a hook point), and so that optional template might never exist for a particular site, so the "file not found" case really isn't exceptional.

C++ exceptions are implemented on the assumption that they rarely happen and the design puts the cost when an exception is thrown/handled, and not for the no-exception case.

xapian-applications/omega/omegatest
```diff
@@ -1451,6 +1451,17 @@ printf '$hitlist{$list{$terms{T},|}/$field{tag1}|$field{tag2}|$field{tag3}}' > "
 testcase 'Tone/one|two|three' B=Tone
 testcase 'Tthree|Ttwo/one|two|three' B=Ttwo B=Tthree
 
+
+# Test failover script is run when the failover script is provided
+# and non-existent template file is passed in $include
+printf '$include{non_existant.template,$set{data,replacement_for_template}}$opt{data}' > "$TEST_TEMPLATE"
```
ojwb, line 1457: Typo - it's "existent" not "existant".

This tests that the fallback code is evaluated, but doesn't test that the result of the evaluation is used.  It looks from the code like it will be, but if you'd written or a later programmer maintaining the code refactors it to

```
    eval(args[1], param);
```

(i.e. failed to assign to `value`) that would evaluate it but not use the value, but your tests here wouldn't catch that.

It would be better to test something like this:

```
$include{non_existent.template,foo$set{data,bar}}$opt{data}
```

That should give `foobar`.

xapian-applications/omega/omegatest
```diff
@@ -1451,6 +1451,17 @@ printf '$hitlist{$list{$terms{T},|}/$field{tag1}|$field{tag2}|$field{tag3}}' > "
...
+testcase 'replacement_for_template' B=Tone
+
+# Test exception is thrown when non-existent template file is passed in
+# $include and optional failover script is not provided
+printf '$include{non_existant.template}$opt{data}' > "$TEST_TEMPLATE"
+testcase "Exception: Couldn't read format template 'non_existant.template' (No such file or directory)" B=Tone
```
ojwb, line 1463: It's good to check this case works as before, but having the `$opt{data}` in there isn't checking anything useful as nothing in this testcase sets it.

It would however be useful to have a test that including a file which **does** exist doesn't evaluate the second argument since that's quite easier to get wrong in the implementation.  Something like checking this gives what's in `template`:

```
$include{template,foo$set{data,bar}}$opt{data}
```

xapian-applications/omega/docs/omegascript.rst
ojwb: I think it would be clearer to word this like the docs for `$if` above - "runs" is a bit vague, whereas "evaluates and returns" makes it clear that the second argument is evaluated and the result of that returned.

I worry `FAILOVER_SCRIPT` might be misunderstood to mean a script file (i.e. another file to include instead of `FILE`.  Perhaps `$include{FILE[,FALLBACK]}`?

We should almost make it clear what "failure" means.  I think (see below) that it probably should only be a failure to open the file to be included.

ojwb (changes requested):

ojwb:
Sorry for taking so long to review.

ojwb:
I've applied the changes I suggested, squashed and merged.  Thanks for the patch.

PR #289, opened 2020-03-08
Add OP_VALUE_GT and OP_VALUE_LT

bodasadallah:
As per #224 
Here's a small plan for this 
https://docs.google.com/document/d/1NXYDDMhkcSfULaJSVkp2e1u1MfCpd0GdqAV39_ylyMs/edit?usp=sharing

Still needs work on the tests


xapian-core/api/queryinternal.cc
```diff
@@ -1314,7 +1379,7 @@ QueryValueGE::postlist(QueryOptimiser *qopt, double factor) const
     if (limit > db.get_value_upper_bound(slot)) {
 	RETURN(NULL);
     }
-    if (limit < lb) {
+    if (limit <= lb) {    // should't this be <= ??
```
jaylett: This change is unrelated to your other changes. If this change should be made, it should be done at least as a separate commit, with a testcase that fails before the change happens. (It could also be done as entirely separate PR.)

xapian-core/include/xapian/query.h
```diff
@@ -211,6 +211,14 @@ class XAPIAN_VISIBILITY_DEFAULT Query {
 	 */
 	OP_VALUE_GE = 11,
 
+    /** Match only documents where a value slot is > a given value.
+	 *
+	 *  Similar to @a OP_VALUE_RANGE, but open-ended.
+	 *
+	 *  This operator never contributes weight.
+	 */
+	OP_VALUE_GT = 17,
```
jaylett: It would be better to keep them in numerical order, I think. That could then be the same order you put them in switch or chained else/if statements.

xapian-core/include/xapian/query.h
```diff
@@ -219,6 +227,22 @@ class XAPIAN_VISIBILITY_DEFAULT Query {
...
+	 */
+	OP_VALUE_LT = 18,
+
+    /** Match only documents where a value slot is = a given value.
+	 *
+	 *  Similar to @a OP_VALUE_RANGE, but open-ended.
+	 *
+	 *  This operator never contributes weight.
+	 */
+	OP_VALUE_EQ = 19,
```
jaylett: As noted, you probably want to remove this. Note also that the doccomment is incorrect: this is not an open-ended operator.

jaylett (changes requested):
As well as figuring out why your new tests are failing, there are a number of other problems with your PR, mostly to do with formatting and committing the wrong files.

If you haven't already, please read the sections of the developer guide for new contributors and on contributing changes, as they contain some tips which can help you with git and pull requests.

xapian-core/api/queryinternal.cc
jaylett: (If the change is about efficiency rather than correctness, then a test case is going to be hard to impossible to write. This change should still be in its own commit in that case.)

jaylett:
> Here's a small plan for this
> https://docs.google.com/document/d/1NXYDDMhkcSfULaJSVkp2e1u1MfCpd0GdqAV39_ylyMs/edit?usp=sharing

I love that you've written out a plan — that's a really helpful habit to get into :)

.gitignore
```diff
@@ -1,7 +1,10 @@
...
-
+*.valgrind
+*.vscode
 /configure
 /Makefile
 /config.cache
 /BUILD
 /INST
+vscode
+
```
ojwb: This change really belong in a separate PR - it's nothing to do with adding these new operators at all.

Please don't reformat whitespace in existing files when modifying them (here you removed an existing blank line separately glob patterns and verbatim names, and added an extra blank line at the end.

`*.valgrind` looks wrong to me - did you actually mean to add `.valgrind.log.*`?  The test harness creates those as temporary files, but it should also remove them again (unless interrupted).  We didn't add them to `.gitignore` previously because it seems better to know they've been left behind - otherwise you can end up with a lot of diskspace used by these files as the name includes the process id.  I found I had thousands at one point, though we now remove them on "make clean" which helps avoid them building up indefinitely.

ojwb:
guthub doesn't seem to let me comment on an empty file, but `xapian-bindings/python/xapian_wrap.stamp-t` and `xapian-bindings/python3/xapian_wrap.stamp-t` are temporary files which shouldn't be committed.

xapian-core/api/queryinternal.cc
```diff
@@ -1295,6 +1297,70 @@ QueryValueLE::get_description() const
...
+	// postlist, which is especially efficient if there are no gaps in
+	// the docids.
+	if (db.get_value_freq(slot) == db.get_doccount()) {
+	    RETURN(db.open_post_list(string()));
+	}
+    }
+    RETURN(new ValueLtPostList(&db, slot, limit));
+}
+// This is the same as QueryValueLE::serialise,
+// not sure if it should be the same
```
ojwb: No, it needs to be different - otherwise if you serialise OP_VALUE_LT then unserialise the result youl get OP_VALUE_LE out:

```
olly@gemse:~/git/xapian/xapian-bindings/python$ ./run-python-test 
Python 2.7.17 (default, Jan 19 2020, 19:54:54) 
[GCC 9.2.1 20200110] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import xapian
>>> q = xapian.Query(xapian.Query.OP_VALUE_LT, 0, "abc")
>>> str(q)
'Query(VALUE_LT 0 abc)'
>>> str(xapian.Query.unserialise(q.serialise()))
'Query(VALUE_LE 0 abc)'
>>> 
```

xapian-core/api/queryinternal.cc
```diff
@@ -1315,6 +1381,7 @@ QueryValueGE::postlist(QueryOptimiser *qopt, double factor) const
 	RETURN(NULL);
     }
     if (limit < lb) {
+	// should't this be <= ??
```
ojwb: I think you're right.  It means the optimisation won't be applied in the edge case so it's an efficiency problem not a correctness one.  But as James said, that would be better handled as a separate change.

xapian-core/api/queryinternal.cc
```diff
@@ -1355,6 +1422,69 @@ QueryValueGE::get_description() const
...
+void
+QueryValueGT::serialise(string& result) const
+{
+    if (slot < 15) {
+	result += static_cast<char>(0x20 | 0x10 | slot);
+    } else {
+	result += static_cast<char>(0x20 | 0x10 | 15);
+	pack_uint(result, slot - 15);
+    }
+    pack_string(result, limit);
```
ojwb: That's the same serialisation as OP_VALUE_GE...

ojwb:
This provides open-ended non-inclusive ranges.  We already have open-ended inclusive ranges, and non-open-ended inclusive ranges.

I think we should consider what the plan is for the other combinations.  Do we expect people to use "`OP_VALUE_LT` `OP_AND`  `OP_VALUE_GT`" if they want a non-open-ended non-inclusive range?  What about if they want one end inclusive and the other not?

Having to compose with `OP_AND` is perhaps a little cumbersome to use as an API, but adding an operator for every combination doesn't seem great to use either.  Passing boolean flags is an option but would be a bit cryptic - consider `Query(Query::OP_VALUE_RANGE, 0, "a", "b", true, false)` - it's hard to remember whether `true` means inclusive or exclusive because there isn't a natural meaning.  Named values avoid that problem, but tend to be rather verbose - e.g. `Query(Query::OP_VALUE_RANGE, 0, "a", "b", Query::VALUE_RANGE_INCLUSIVE, Query::VALUE_RANGE_EXCLUSIVE)`.

Internally, if we do expect people to compose with `OP_AND` then do we try to optimise so we don't end up fetching the value slot data twice (once for each end of the range) and if so what's the plan for how to achieve that?  And if we add every combination as an operator, can we tame all the very similar classes which the current approach would lead to?

Also, can we eliminate the source code duplication in `ValueXXPostList` by having a common base class, or perhaps by templating?

ojwb:
> > Here's a small plan for this
> > https://docs.google.com/document/d/1NXYDDMhkcSfULaJSVkp2e1u1MfCpd0GdqAV39_ylyMs/edit?usp=sharing
> 
> I love that you've written out a plan — that's a really helpful habit to get into :)

Yes, though perhaps better to use a plain text file and commit it to your branch (we used to have a convention of `README.branch` at the top level of long lived development branches which is probably not a bad name to use).  That's easier to look at than an online doc since it doesn't require being online to view and update it.  It also means updates to the plan are version controlled along with the corresponding changes to the code.  This file can be removed before merging while still being present in the branch history.

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +869,158 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
-		    op combiner = static_cast<op>(*(*p)++);
-		    string pattern;
-		    if (!unpack_string(p, end, pattern)) {
-			throw SerialisationError("not enough data");
-		    }
-		    return new Xapian::Internal::QueryWildcard(pattern,
-							       max_expansion,
-							       flags,
-							       combiner);
+		switch (ch >> 4)
```
ojwb: The formatting of this hunk of the patch is wrong in various ways - e.g. the `{` should be on the same line as the `switch`, some lines are overindented, some blocks aren't indented compare to their parent.

As James said on IRC, you really want to get the formatting correct when you write the code.  It's no fun for reviewers to have to point out glaring problems, and no fun for you to have to go through fixing them all.

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +869,158 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
-			throw SerialisationError("not enough data");
-		    }
-		    return new Xapian::Internal::QueryTerm(string(), wqf, pos);
-		}
-		case 0x0f:
-		    return new Xapian::Internal::QueryTerm();
-		default: // Others currently unused.
-		    break;
-	    }
+		case 0: {
```
ojwb: Formatting is off in this hunk too...

xapian-core/include/xapian/query.h
```diff
@@ -268,6 +268,22 @@ class XAPIAN_VISIBILITY_DEFAULT Query {
...
+	 *  Similar to @a OP_VALUE_RANGE, but open-ended.
+	 *
+	 *  This operator never contributes weight.
+	 */
+	OP_VALUE_GT = 17,
+
+    /** Match only documents where a value slot is < a given value.
+	 *
+	 *  Similar to @a OP_VALUE_RANGE, but open-ended.
+	 *
```
ojwb: The new `/**` lines are underindented compare to the comment bodies and code.

jaylett:
License declaration received in https://github.com/xapian/xapian/pull/291

xapian-core/api/queryinternal.cc
```diff
@@ -800,7 +803,8 @@ Query::Internal::unserialise(const char ** p, const char * end,
 	    //     0: wqf = 0; pos = 0
 	    //     1: wqf = 1; pos = 0
 	    //     2: wqf = 1; pos -> encoded value follows
-	    //     3: wqf -> encoded value follows; pos -> encoded value follows
+	    //     3: wqf -> encoded value follows;
+	    //	   pos -> encoded value follows
```
ojwb, line 807: Wrapping this line isn't necessary and makes the comment a bit less readable, but really the key thing is that it isn't part of the change you're trying to make (i.e. "Add OP_VALUE_GT and OP_VALUE_LT").  So even if this were a desirable change, this PR isn't the place to be making it.

xapian-core/api/queryinternal.cc
ojwb: You marked the conversation as resolved, which is really something you should leave to the reviewer.

But while you have updated the serialisation code you haven't removed the comment with your question from the patch, and that's not something useful to include - it's just confusing now.

xapian-core/api/queryinternal.cc
```diff
@@ -783,7 +785,8 @@ Query::Internal::unserialise(const char ** p, const char * end,
 		    break;
 		default:
 		    // 8 to 12 are currently unused.
-		    throw SerialisationError("Unknown multi-way branch Query operator");
+		    throw SerialisationError(
+			    "Unknown multi-way branch Query operator");
```
ojwb, line 789: This line is over-length in the existing code, and fixing formatting problems is useful, but you aren't changing this part of the code so this isn't the PR to be addressing it in.  Mixing reformatting and functional changes just makes for a needlessly larger patch which is harder to review.

BodaSadalla98:
Updated, please review

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +867,156 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
+	    case 1: {
+		// 0001oxxx where:
+		// o is the operation: LT or GT
+		// xxx is the slot number
+		Xapian::valueno slot = ch & 7;
+		if (slot == 7) {
+		    if (!unpack_uint(p, end, &slot)) {
+			unpack_throw_serialisation_error(*p);
+		}
+		slot += 7;
```
jaylett, line 880: The two lines here are underindented.

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +867,156 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
-		    using Xapian::Internal::QueryEditDistance;
-		    return new QueryEditDistance(pattern,
-						 max_expansion,
-						 flags,
-						 combiner,
-						 edit_distance,
-						 fixed_prefix_len);
+		return new Xapian::Internal::QueryValueLT(slot, end_);
+		}
 		}
```
jaylett, line 901: No way should two close braces be at the same level of indentation.

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +867,156 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
+			unpack_throw_serialisation_error(*p);
+		}
+		slot += 7;
+		}
+		string begin;
+		if (!unpack_string(p, end, begin)) {
+		    unpack_throw_serialisation_error(*p);
+		}
+		switch (ch >> 3)
+		{
```
jaylett, line 887: We 'cuddle' opening braces for switch statements (as indeed you've done up at line 870).

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +867,156 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
-		    int flags = static_cast<unsigned char>(*(*p)++);
-		    op combiner = static_cast<op>(*(*p)++);
-		    string pattern;
-		    if (!unpack_string(p, end, pattern)) {
+		    case 0x0b: { // Wildcard
+			Xapian::termcount max_expansion;
+			if (!unpack_uint(p, end,
+					 &max_expansion) ||
+					 end - *p < 2) {
+				throw SerialisationError(
```
jaylett, line 947: I think this is overindented (it should be four spaces to the right of the containing `if()`).

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +867,156 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
-		    return new Xapian::Internal::QueryTerm();
-		default: // Others currently unused.
-		    break;
+		    case 0x0f:
+			return new Xapian::Internal::QueryTerm();
+		    default: // Others currently unused.
+			break;
+			}
+	    }
 	    }
```
jaylett, line 1015: Another double close brace at the same level of indentation.

xapian-core/api/queryinternal.cc
```diff
@@ -865,98 +867,156 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
 	    break;
 	}
     }
+
```
jaylett, line 1019: Please avoid introducing whitespace changes unless there's a good reason (I can't see one here). Even then, they would ideally be in a separate commit, which I don't think is true here.

jaylett (changes requested):
As well as removing the return values from your tests and rebasing onto master (so you can run tests locally), you need to do a pass over the formatting. It may be helpful to skim read each file you've changed, looking for things that don't sit right. I've noted some from the diffs, but looking at [queryinternal.cc](https://github.com/xapian/xapian/blob/2c8f2a94d8ed14b5470e0d63bfebf06efa446b55/xapian-core/api/queryinternal.cc) for instance, some of them are even clearer (such as [the case block at lines 893–901](https://github.com/xapian/xapian/blob/2c8f2a94d8ed14b5470e0d63bfebf06efa446b55/xapian-core/api/queryinternal.cc#L893)) and you should spot some others (such as [an underindented if leg](https://github.com/xapian/xapian/blob/2c8f2a94d8ed14b5470e0d63bfebf06efa446b55/xapian-core/api/queryinternal.cc#L917)).

xapian-core/api/queryinternal.cc
BodaSadalla98: one brace if for the "case" statement and the other is for the "switch"
and the convention used for this entire file that the switch statement and its cases are in the same indentation level

jaylett: I've checked this file on master, and that is not true of any of the 10 switch statements. You may be getting confused by your editor displaying tabs as 4 spaces rather than 8; I can't think of another reason why you're seeing something different to me.

BodaSadalla98: yes, my bad. Sorry for this inconvenience

xapian-core/api/queryinternal.cc
```diff
@@ -865,94 +867,138 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
+				throw SerialisationError("not enough data");
+			    }
+			    return new Xapian::Internal::QueryTerm(
+						    string(), wqf, pos);
+			}
+			case 0x0f:
+			    return new Xapian::Internal::QueryTerm();
+			default: // Others currently unused.
+			    break;
+			    }
```
jaylett, line 1000: This still looks wrong. The close brace here is presumably closing the switch, so should be eight spaces left?

xapian-core/api/queryinternal.cc
jaylett: @BodaSadalla98 as mentioned, this comment shouldn't be here.

xapian-core/include/xapian/query.h
```diff
@@ -257,6 +257,22 @@ class XAPIAN_VISIBILITY_DEFAULT Query {
 	 */
 	OP_EDIT_DISTANCE = 16,
 
+    /** Match only documents where a value slot is > a given value.
```
jaylett, line 260: This is still underindented — it's four spaces but the next one (added in this patch) is a tab.

BodaSadalla98:
Please review @jaylett

xapian-core/api/query.cc
```diff
@@ -92,10 +94,16 @@ Query::Query(op op_, Xapian::valueno slot, const std::string & limit)
 	    internal = new Xapian::Internal::QueryTerm();
 	else
 	    internal = new Xapian::Internal::QueryValueGE(slot, limit);
+    } else if (usual(op_ == OP_VALUE_GT)) {
+	internal = new Xapian::Internal::QueryValueGT(slot, limit);
     } else if (usual(op_ == OP_VALUE_LE)) {
 	internal = new Xapian::Internal::QueryValueLE(slot, limit);
+    } else if (usual(op_ == OP_VALUE_LT)) {
+	internal = new Xapian::Internal::QueryValueLT(slot, limit);
```
ojwb, line 102: The `usual(...)` macro here hints to the compiler that each of these branches is usually taken, which isn't correct.

Before the change it was, since supplying an invalid `op` is not the usual case.

So we shouldn't annotate these conditions now (except perhaps for the last one, but I did some experiments which suggest the compiler assumes that a branch which raises an exception shouldn't be predicted as taken, so it's probably superfluous to hint that it isn't explicitly).

xapian-core/api/queryinternal.cc
```diff
@@ -865,94 +867,138 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
-							       max_expansion,
-							       flags,
-							       combiner);
-		}
-		case 0x0c: { // PostingSource
-		    string name;
-		    if (!unpack_string(p, end, name)) {
-			throw SerialisationError("not enough data");
+	    switch (ch >> 4) {
+		case 1: {
```
ojwb, line 871: This seems to be just handling a single bit, so I'd suggest an `if` would be clearer than using `switch` - with the `switch` it's not clear to the reader that either `case 0` or `case 1` should be taken, and that other values should be impossible.

That would also reduce the indentation of the code inside by a level.

xapian-core/api/queryinternal.cc
```diff
@@ -865,94 +867,138 @@ Query::Internal::unserialise(const char ** p, const char * end,
...
+		    //     tttt -> encodes which OP_XXX
+		    switch (ch & 0x0f) {
+			case 0x00: // OP_INVALID
+			    return new Xapian::Internal::QueryInvalid();
+			case 0x0a: { // Edit distance
+			    Xapian::termcount max_expansion;
+			    if (!unpack_uint(p, end, &max_expansion) ||
+						     end - *p < 2) {
+			    throw SerialisationError("not enough data");
+			    }
```
ojwb, line 917: There are still indentation problems such as just above - the wrapped parameter list is over-indented, while the code inside the `if` body is under-indented.

PR #290, opened 2020-03-09
Datetime

Exter-dg:


xapian-applications/omega/templates/query
```diff
@@ -78,8 +78,8 @@ $set{clickurl,$if{$opt{clicklog},$or{$env{SCRIPT_NAME},omega.cgi}?FMT=clicklog&D
 $if{$and{$field{modtime},$ne{$field{modtime},-1}},
 <script type="text/javascript">write_date($add{$field{modtime}});</script>
 <noscript>
-<span title="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
-Modified:<br><b>$html{$date{$field{modtime},%Y-%m-%d}}</b></span>
+<time title="$html{$date{$field{modtime},%Y-%m-%d %H:%M:%S}}">
```
jaylett: You don't have a `datetime` attribute here, although you do in the Javascript-generated one. Given you then want to replace the Javascript with something that reads that static datetime and replaces the element contents with a localised date string, this is important to get right. While you could read from `title`, that's intended for people to read (so we will probably end up with both — `title` and element contents that our Javascript will localised, and `datetime` which will have the full timezone, probably in UTC for simplicity).

ojwb:
Your expanded commit message is inaccurate:

> Introduced time tag as span is no longer supported in HTML5.

`<span>` **is** still supported in HTML5:

https://www.w3.org/TR/html52/textlevel-semantics.html#the-span-element

The motivation for this change is that `<span>` carries no specific semantic meaning while HTML5 added `<time>` as a way to mark up dates and times, which is useful as it means they are easily and directly machine readable. A commonly given example (though probably not so useful here) is that the browser could allow you to click on the `<time>` and add it to your calendar. The benefits are perhaps less immediately obvious to us right now (maybe such elements are more accessible to blind people using screen reader software because that software can read out the date in a clearer way?) but it still makes sense to present the information with a semantic structure where we can.

xapian-applications/omega/templates/query
ojwb: > You also don't need to merge master in. It just makes your branch more complex. (If you actually need changes more recent then you can rebase your changes onto latest master, which is generally how we merge PRs into master once ready.)

Though don't rebase and force push a branch already under review without discussion - it tends to make review more complicated.

Travis internally actually does a temporary merge of the PR branch with master and tests the result of that merge (which is helpful since that tests what the current result of merging the branch would actually be) so typically you only need to update to the current master if the branch stops merging cleanly.

Not all CI systems test a merged version though - e.g. gitlab's definitely just tests the branch, and I think appveyor is the same.

Exter-dg:
> You also don't need to merge master in. It just makes your branch more complex. (If you actually need changes more recent then you can rebase your changes onto latest master, which is generally how we merge PRs into master once ready.)
> 
> A tiny tiny note about your commit message (which you don't have to correct because we can do it when we merge): for the first line of a commit you shouldn't put a full stop at the end. The first line is special (it should be shorter, and followed by a blank line if the commit message is longer) — there are some notes on this in the developer guide, including a link to a fairly detailed discussion of commit messages.
> 
> Can you also please confirm that you are OK to agree to the patch licensing requirements?
> 
> https://xapian-developer-guide.readthedocs.io/en/latest/contributing/contributing-changes.html#license-grant
> 
> This is a requirement for changes to Xapian.

 I agree to dual-license all my contributions to Xapian under the GNU GPL version 2 and all later versions, and the MIT/X licence.





> Your expanded commit message is inaccurate:
> 
> > Introduced time tag as span is no longer supported in HTML5.
> 
> `<span>` **is** still supported in HTML5:
> 
> https://www.w3.org/TR/html52/textlevel-semantics.html#the-span-element
> 
> The motivation for this change is that `<span>` carries no specific semantic meaning while HTML5 added `<time>` as a way to mark up dates and times, which is useful as it means they are easily and directly machine readable. A commonly given example (though probably not so useful here) is that the browser could allow you to click on the `<time>` and add it to your calendar. The benefits are perhaps less immediately obvious to us right now (maybe such elements are more accessible to blind people using screen reader software because that software can read out the date in a clearer way?) but it still makes sense to present the information with a semantic structure where we can.

Thank you @ojwb for pointing out my mistake, I have corrected it and have updated my PR.


As discussed with @jaylett and @ojwb, I have collapsed multiple commits into a single commit to make the history clean.
If any changes, please let me know.

xapian-applications/omega/templates/query
```diff
@@ -25,7 +25,7 @@ function write_date(time_t) {
 var d=new Date();d.setTime(time_t*1000);
 var D=d.getFullYear()+"-"+f2(d.getMonth()+1)+"-"+f2(d.getDate());
 var T=f2(d.getHours())+":"+f2(d.getMinutes())+":"+f2(d.getSeconds());
-document.write("<span title=\""+D+" "+T+"\">Modified:<br><b>"+D+"</b></span>");
+document.write("<time datetime=\""+D+" "+T+"\" title=\""+D+" "+T+"\">Modified:<br><b>"+D+"</b></time>");
```
jaylett: From my previous comment:

> so we will probably end up with both — title and element contents that our Javascript will localised, and datetime which will have the full timezone, probably in UTC for simplicity

But that isn't the case here; `datetime` and `title` have exactly the same content. Although I think this is technically a valid use of datetime, it's  ambiguous without the timezone. Likewise, it'd be good to put the timezone in the static version below (which will be in UTC, because that's how `$date{}` works.)

Exter-dg: Thank you for pointing this out, I have reworked the code and have added the Zulu time zone (Z) to the 'datetime' attribute.
 Please let me know if this is correct or should I make some changes.

jaylett: This looks good to me at a quick glance. Note that in future it's a good idea to use fixup commits rather than creating completely new ones. The total change here is going to end up as a single commit (because the noscript version is technically wrong without the second commit), and fixup commits make it easier to squash things together later. Not a problem with only a couple of commits, but when you're dealing with larger PRs with half a dozen commits, using fixups to indicate which commit you're updating can save a lot of time.

Exter-dg: Thank you for letting me know, I have squashed the commits and have updated the PR with a single commit. From now on, I will be using fixup commits for the same.

ojwb:
This looks good to me now, though I'd like to get James' OK on it too.

Now we're using `<time>` we could get rid of all the calls to `write_date()` and `<noscript>` tags and instead use `onLoad` to run some javascript which iterates over all the `<time>` elements on the page (perhaps restricted to those with a particular class?) and rewrites their `innerHTML`.  

But let's get this merged as it's a definite improvement as it is now, and think about that as a possible follow-on change.

jaylett:
I'm happy with this — thanks @Exter-dg. In future, please don't force push to PRs that are in the middle of review, because it can cause problems with larger changes. (For instance, github tries to keep track of the last changes you've looked at, so you can see what additional commits including fixups have been added, and check they've dealt with earlier concerns. When you force push, that history is lost.)

ojwb:
Thanks.  Rebased onto master as 1a35991fb0fc664397beac4f310a7724c67ffdb9, so closing.

PR #291, opened 2020-03-10
Adding .vscode folder to .gitignore

BodaSadalla98:
.vscode folder contains a user-specific settings


BodaSadalla98:
> Is there an existing testcase which gives test coverage for this? We can't really provide a regression test (since the previous behaviour was correct and the efficiency gain is hard to test reliably) but it would be good to make sure that there's a test with `OP_VALUE_GE` with the limit equal to the lower bound.

In "valuege1" test, one of the limits is empty string ( "" ), doesn't that insure that the limit is equal to the lower bound ?

BodaSadalla98:
> Also, are you OK to agree to the patch licensing requirements?
> 
> https://xapian-developer-guide.readthedocs.io/en/latest/contributing/contributing-changes.html#license-grant

I do

ojwb:
> > Is there an existing testcase which gives test coverage for this? We can't really provide a regression test (since the previous behaviour was correct and the efficiency gain is hard to test reliably) but it would be good to make sure that there's a test with `OP_VALUE_GE` with the limit equal to the lower bound.
> 
> In "valuege1" test, one of the limits is empty string ( "" ), doesn't that insure that the limit is equal to the lower bound ?

No - the lower bound is the lowest **non-empty** value in the slot.  Empty values aren't explicitly stored (empty is the same as not-present).  That's why we have to also check `(db.get_value_freq(slot) == db.get_doccount())` (since if that's not true then we know there's at least one empty value.

This also means a useful testcase would have a value in the slot set for every document.

However, you are looking in the right place - checking the values set in slot 1 of the DB used by that testcase:

```
olly@pippikin:~/git/xapian/xapian-core/tests$ ../bin/xapian-delve .glass/db__apitest_phrase -V1
Value 1 for each document: 1:h 2:h 3:o 4:a 5:r 6:n 7:h 8:o 9:  10:e 11:e 12:n 13:e 14:o 15:i 16:a 17:a
```

So the lowest is `a` and it's set for all docs - there are 17 docs in this DB:

```
olly@pippikin:~/git/xapian/xapian-core/tests$ ../bin/xapian-delve .glass/db__apitest_phrase/
UUID = 302b4b1f-88bf-494b-ad7f-ba55781c9405
number of documents = 17
average document length = 11.1176
document length lower bound = 5
document length upper bound = 26
highest document id ever used = 17
has positional information = true
revision = 1
currently open for writing = false
```

And `a` is one of the tested bounds, so we do have test coverage.

ojwb:
Oh, actually ` ` is the lower bound (doc 9) but that's also tested.

.gitignore
```diff
@@ -5,3 +5,4 @@
 /config.cache
 /BUILD
 /INST
+.vscode
```
ojwb: Can `.vscode` really be in any directory of the source tree?  If it's only at the top-level then it's better to say `/.vscode` as then git only checks for it in the same directory as the `.gitignore` file it's listed in.  Checking one extra ignore entry per directory doesn't make much difference, but if many needlessly broad entries can.

Also, you've misspelling "ignoring" in the commit message.  And it's conventional to start the commit summary with a capital letter (see the advice in the dev guide on writing commit messages).

BodaSadalla98: .vscode folder, would always be in the root of the project (top-level), although there's something called  [Multi-root Workspaces](https://code.visualstudio.com/docs/editor/multi-root workspaces), where we would have multiple roots and maybe multiple setting files. 
I think we shouldn't worry about that and we can only ignore the folder in the root

jaylett: I agree that we shouldn't worry about more complex setups. People can figure those out using global gitignore as needed.

For commit messages the "subject" (the first line) should generally be of the form "Fix something", "Do something", so in this case it'd be better to use something like "Ignore .vscode directory" or "Ignore VSCode settings" (both with additional explanation in the "body" of the commit message if necessary to make it clear what's going on, although in this case the change is probably clear enough with a subject only).

jaylett: (The reason for doing it like that is that git generates commit messages automatically sometimes — for merges, for instance — and phrases them in that way.)

jaylett:
@BodaSadalla98 did you want to finish up this PR? It looks like there's a stray file that should be removed in a fixup (you should be able to revert the first commit actually, and it'll disappear in a rebase before we merge)), and you'll want to rewrite the commit message for the optimisation — for now can you just write what the commit message should be in a comment here?

BodaSadalla98:
Olly wrote a good one [here]https://github.com/xapian/xapian/commit/3c56e5db5b8f3696fd8f311793c62921eb413ef8l).

I think we can close this now, I reverted the 2 unnecessary commits

ojwb:
> Revert adding vscode setting in .gitignore
>
> We setteled that this should be added in the git  global ignore

I'm happy to add sensible ignore entries for commonly use editors/IDEs.  I'd agree people probably ought to set up a global `.gitignore` entry for the editor/IDE they use, but they aren't set up by default and not everyone will have added them, so having an entry in our `.gitignore` helps avoid people accidentally including such unwanted files in PRs (which takes reviewer time to sort out that would be better spent elsewhere).

My question above was just about whether it should be `/.vscode` rather than `.vscode` (which is seems it probably should).

BodaSadalla98:
Closed as it adds no new changes.

PR #292, opened 2020-03-15
go bindings

srinivasyadav18:
configure your go path correctly before  installlation of bindings


.travis.yml
```diff
@@ -169,11 +169,14 @@ matrix:
...
   # Bootstrap everything (including letor, which isn't done
   # by default), then configure using our chosen compiler.
+  - eval "$(gimme 1.14)"
+  - eval env
   - travis_retry ./bootstrap xapian-core xapian-applications/omega swig xapian-bindings xapian-letor
   - ./configure $confargs CC="$USE_CC" CXX="$USE_CXX"
 script:
   - make -j2
   - $installcore
+  #  - sudo env "PATH=$PATH" make install
```
ojwb: This commented out code looks like a left-over from debugging/development.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+python3-sphinx mono-devel default-jdk lua5.3 liblua5.3-dev \
+php-dev php-cli python-dev python3-dev ruby-dev tcl-dev texinfo -y
+
+RUN wget https://dl.google.com/go/go1.14.linux-amd64.tar.gz ;
+RUN tar -C /usr/local -xzf go1.14.linux-amd64.tar.gz
+RUN echo "export PATH=$PATH:/usr/local/go/bin " >> ~/.bashrc
+
+#run these commands once you are in
+#source ~/.bashrc
+#cd xapian && git pull && ./bootstrap && ./configure && make && make install
```
ojwb: A Dockerfile is perhaps useful to provide, but really should be a separate PR.

xapian-bindings/configure.ac
```diff
@@ -287,11 +291,31 @@ case $with_python$with_python3$with_php7$with_ruby$with_tcl$with_csharp$with_jav
...
+    fi
+    if test -z "${GOROOT}" ; then
+      AC_MSG_ERROR([goroot not set correctly])
+    fi
+  fi
+  BINDINGS="$BINDINGS go"
+  if test -z "$GO" ; then
+    AC_MSG_ERROR("go not installed or go path has not been set")
+  fi
+fi
```
ojwb: You've added entries for Go last above and below (which seems reasonable to me as that's how previous new languages have been added) but here you've added it as the **first** language probed for.  I think a consistent order (i.e. after Lua) would make this easier to maintain.

xapian-bindings/go/Makefile.am
```diff
@@ -0,0 +1,38 @@
...
+	awk 'NR==18{print "#cgo CXXFLAGS: ${XAPIAN_CXXFLAGS}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==19{print "#cgo CXXFLAGS: -std=gnu++11"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo LDFLAGS: ${USE_LIB}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo CPPFLAGS: ${CPPFLAGS}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo LDFLAGS: -L${abs_top_srcdir}/../xapian-core/.libs -Wl,-rpath,${abs_top_srcdir}/../xapian-core/.libs -lxapian-1.5"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
```
ojwb: It would seem better to add these lines all in one go rather than one by one.

xapian-bindings/go/Makefile.am
```diff
@@ -0,0 +1,38 @@
...
+	${SWIG} ${SWIG_FLAGS} -c++ -go -cgo -intgosize 64 go.i
+	awk 'NR==18{print "#cgo CXXFLAGS: ${XAPIAN_CXXFLAGS}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==19{print "#cgo CXXFLAGS: -std=gnu++11"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo LDFLAGS: ${USE_LIB}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo CPPFLAGS: ${CPPFLAGS}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo LDFLAGS: -L${abs_top_srcdir}/../xapian-core/.libs -Wl,-rpath,${abs_top_srcdir}/../xapian-core/.libs -lxapian-1.5"}1' xapian.go > xapian2.go
```
ojwb: This is hard-coding information which should come from `XAPIAN_CXXFLAGS`.  Your hard-coded version will work currently when building from git, but will fail when building against an already installed xapian-core, or indeed once the current version is no longer 1.5.x.

xapian-bindings/go/Makefile.am
```diff
@@ -0,0 +1,38 @@
...
+
+modify : build
+	mkdir -p ${GOSRCDIR}
+	cp -r * ${GOSRCDIR}
+
+build :
+	${SWIG} ${SWIG_FLAGS} -c++ -go -cgo -intgosize 64 go.i
+	awk 'NR==18{print "#cgo CXXFLAGS: ${XAPIAN_CXXFLAGS}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==19{print "#cgo CXXFLAGS: -std=gnu++11"}1' xapian.go > xapian2.go
```
ojwb: This option isn't supported by all C++ compilers.  You should use the value of `CXXFLAGS` which is set in the Makefile (to which configure will have added any option needed to enable C++11 support).

xapian-bindings/go/Makefile.am
```diff
@@ -0,0 +1,38 @@
...
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo LDFLAGS: ${USE_LIB}"}1' xapian.go > xapian2.go
```
ojwb: But `USE_LIB` is something you invented and set in the CI script - it won't be set for user builds.

Why is this needed exactly?

xapian-bindings/go/docs/examples/index/index1.go
```diff
@@ -0,0 +1,59 @@
...
+import ("fmt"
+	"io"
+	"encoding/csv"
+	"os"
+	"xapian"
+)
+
+func main(){
+	//a csv reader to get the values from csv file
+	csvfile,err := os.Open("../data/100-objects-v2.csv")
```
ojwb: We don't want to add this data here as it's unfortunately under a Creative Commons "Attribution-NonCommercial-ShareAlike" licence which doesn't allow commercial use.  For example, I'd have to repack the source archive to remove it for the Debian packages (I maintain the Debian packages of Xapian).

I'd suggest it would make more sense to just translate the "simple" examples here (like is done for other languages) and add the more sophisticated examples which use this data to the "Getting Started" guide.

xapian-bindings/go/docs/index.rst
```diff
@@ -0,0 +1,7 @@
+GO BINDINGS FOR XAPIAN
+
+To Install go bindings run these commands
+./bootstrap && ./configure && ./make && make install
```
ojwb: That's only true when installing from git...

xapian-bindings/go/docs/index.rst
```diff
@@ -0,0 +1,7 @@
...
+
+if u have already installed xapian-core :
+        just do make install
```
ojwb: Don't use "SMS-speak" in documentation.  Write out "you" instead of "u".

xapian-bindings/go/test/smoke_test.go
```diff
@@ -0,0 +1,24 @@
...
+	doc := xapian.NewDocument()
+        input_str := "hello go xapian"
+        doc.Set_data(input_str)
+        output_str := doc.Get_data()
+
+        if input_str != output_str{
+		t.Error("Build Failed")
+	}
+}
+
```
ojwb: Blank line at end of file.

xapian-bindings/xapian-headers.i
```diff
@@ -477,3 +477,5 @@ class Remote {
 %include <xapian/dbfactory.h>
 
 #endif
+
+
```
ojwb: You've added blank lines at the end of this file.

.travis.yml
```diff
@@ -169,11 +169,14 @@ matrix:
 before_script:
   # Bootstrap everything (including letor, which isn't done
   # by default), then configure using our chosen compiler.
+  - eval "$(gimme 1.14)"
+  - eval env
```
ojwb: What's the purpose of `eval env` here?

xapian-bindings/configure.ac
```diff
@@ -287,11 +291,31 @@ case $with_python$with_python3$with_php7$with_ruby$with_tcl$with_csharp$with_jav
...
 
 BINDINGS=
-
+if test no != "$with_go" ; then
+  AC_PATH_PROG([GO],["go"],[])
+  AC_MSG_CHECKING(go)
+  if test -n "$GO" ; then
+    AC_MSG_RESULT(yes)
+    AC_MSG_CHECKING(goroot)
+    GOROOT=`go env GOROOT`
```
ojwb: You carefully probe for `go` above (which is good) but then hard-code `go` here instead of using the probed value in `$GO`.

xapian-bindings/configure.ac
```diff
@@ -287,11 +291,31 @@ case $with_python$with_python3$with_php7$with_ruby$with_tcl$with_csharp$with_jav
...
   ;;
 esac
 
 BINDINGS=
-
+if test no != "$with_go" ; then
+  AC_PATH_PROG([GO],["go"],[])
+  AC_MSG_CHECKING(go)
+  if test -n "$GO" ; then
+    AC_MSG_RESULT(yes)
```
ojwb: You don't need `AC_MSG_CHECKING` + `AC_MSG_RESULT` here - you just used `AC_PATH_PROG` which will have said `checking for go...` followed by the path to the found `go` or else `no` if none was found, so your extra messages are look odd and confusing.

xapian-bindings/configure.ac
```diff
@@ -287,11 +291,31 @@ case $with_python$with_python3$with_php7$with_ruby$with_tcl$with_csharp$with_jav
...
+    AC_MSG_CHECKING(goroot)
```
ojwb: You should say *for* goroot here for consistency with other configure probe messages.

xapian-bindings/configure.ac
```diff
@@ -287,11 +291,31 @@ case $with_python$with_python3$with_php7$with_ruby$with_tcl$with_csharp$with_jav
...
+    GOROOT=`go env GOROOT`
+    AC_SUBST(GOROOT)
+    if test -n "${GOROOT}" ; then
+      AC_MSG_RESULT([${GOROOT} yes])
```
ojwb: Drop the "yes", again for consistency.

But if nothing is found, you want to report a result of `no` (again for consistency).

xapian-bindings/go/Makefile.am
```diff
@@ -0,0 +1,38 @@
...
+AM_TESTS_ENVIRONMENT = \
+	abs_builddir='$(abs_builddir)' ;\
+	srcdir='$(srcdir)'
+	export abs_builddir ;\
+	export srcdir ;
+LOG_COMPILER = ${GO} test ./test
+
+install-data-hook:
+	cd ${GOROOT}/src/xapian
+	${GO} install -i -x -a
```
ojwb: Each line in a make rule is run is a separate shell, so your `cd` has no effect on your `go install` command.

xapian-bindings/go/Makefile.am
```diff
@@ -0,0 +1,38 @@
...
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo LDFLAGS: ${USE_LIB}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo CPPFLAGS: ${CPPFLAGS}"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	awk 'NR==17{print "#cgo LDFLAGS: -L${abs_top_srcdir}/../xapian-core/.libs -Wl,-rpath,${abs_top_srcdir}/../xapian-core/.libs -lxapian-1.5"}1' xapian.go > xapian2.go
+	mv xapian2.go xapian.go
+	${GO} env -w CXX="${CXX}"
+check :
+	cd ${GOSRCDIR}/test && ${GO} test
```
ojwb: You carefully set up the automake test harness above, but then override it with your own here.

It'd be much better to use the automake harness for consistency with everything else.

.travis.yml
srinivasyadav18: yea ill remove it

xapian-bindings/configure.ac
srinivasyadav18: yea sure ill replace it

xapian-bindings/go/Makefile.am
srinivasyadav18: oh right....I didn't notice it..

.travis.yml
ojwb: I guess you're just wanting to show the currently set environment variables here?  If so, just `env` will do that (I don't think you're never going to want to use eval to run a fixed command). 

But this is really debugging code so shouldn't be added - there's a length limit on travis logs, so printing out everything you might ever want to know isn't a workable approach.

xapian-bindings/go/Makefile.am
srinivasyadav18: sir I have made changes for almost all that you have mentioned in the comments, for this one I need the path of shared library(libxapian-[verison.no].so) in the cloned git repo (source code) or if installed ..but the XAPIAN_CXXFLAGS is holding the include path for xapian lib and XAPIAN_LIBS is holding path for xapian lib tool archive (libxapian-1.5.la) ....so  since after running make, a (.libs) folder is created in xapian-core right..so im using path for it..and for -lxapian or -lxapian-1.5, there's a variable in xapian-core configure.ac called LIBRARY_VERSION_SUFFIX so can I use that ..? so that is could write it down as      -lxapian"$LIBRARY_VERSION_SUFFIX"            thank you :-)

.travis.yml
srinivasyadav18: yea..I have used it for debugging ..ill be more careful to remove all debugging statements from now on before merging to my master branch.. :-)

xapian-bindings/configure.ac
srinivasyadav18: sir could you please explain little bit in detail..I didn't understand..you mean to use ${GO} in line 305?

ojwb: No need for "sir" here.

Say for example I have the version of go I want to use installed as `/home/olly/private/go/bin/go` - I specify `./configure GO=home/olly/private/go/bin/go` and the probe sets `GO` inside configure to that, but then here you just run `go` which searches `PATH` instead.  If there isn't a `go` on `PATH` this command will fail; if there is, it'll set `GOROOT` to the value for the wrong install of go.

(The user could instead set `PATH` to point first to the `go` you want to use to achieve this effect, but given we're supporting `GO=...` we ought to actually implement it correctly).

xapian-bindings/configure.ac
```diff
@@ -287,11 +291,11 @@ case $with_python$with_python3$with_php7$with_ruby$with_tcl$with_csharp$with_jav
   test -z "$with_ruby" && with_ruby=no
   test -z "$with_perl" && with_perl=no
   test -z "$with_lua" && with_lua=no
+  test -z "$with_go" && with_go=no
   ;;
 esac
 
 BINDINGS=
-
```
ojwb: You've deleted a blank line here - don't reformat existing code as part of a functional change.

.appveyor.yml
```diff
@@ -58,16 +58,18 @@ matrix:
...
       $env:CONFIGOPTS="--disable-documentation"
+  - ps: >-
+      $env:BINDINGS=""
   - ps: >-
       if ($env:TOOLCHAIN -eq "mingw") {
         $env:PATH="C:\Mingw\bin;C:\Mingw\MSYS\1.0\bin;$env:PATH"
       } elseif ($env:TOOLCHAIN -eq "mingw64") {
         if ($env:Platform -eq "x86") {
           $env:PATH="C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\bin;C:\msys64\usr\bin;$env:PATH"
-          $env:CONFIGOPTS+=" --host=i686-pc-mingw32"
```
ojwb, line 67: Why are you removing this?

.appveyor.yml
```diff
@@ -88,13 +90,13 @@ install:
       }
   - if "%TOOLCHAIN%"=="mingw" bash -c "echo 'c:/mingw /mingw' >> /etc/fstab"
   - if "%TOOLCHAIN%"=="mingw" bash -c 'mingw-get install libz-dev'
-  - if "%TOOLCHAIN%"=="mingw64" bash -c 'pacman -Sy --noconfirm zlib-devel'
+  - if "%TOOLCHAIN%"=="mingw64" bash -c 'pacman -Sy --noconfirm --needed base-devel libuuid-devel mingw-w64-x86_64-graphviz doxygen mingw-w64-x86_64-doxygen doxygen swig zlib-devel'
```
ojwb, line 93: > ... `doxygen mingw-w64-x86_64-doxygen doxygen` ...

Should doxygen really be listed twice like that?

Alnd what's the difference between doxygen and mingw-w64-x86_64-doxygen?  Do we really need to install both?

Installing the `swig` package here seems wrong - we deliberately have bootstrap install a particular version of SWIG (specified by a git commit hash) so that we're using a known-good version of SWIG.  We've found that we sometimes end up needing a more recent SWIG (for a particular bug fix or new feature) than is readily available as a package on common platforms.  E.g. currently git master is using an unreleased snapshot of SWIG because the most recent release doesn't work with Ruby 2.7.  Using the version of SWIG mingw64 have packaged may work OK for Go right now, but this really should work like a developer's tree would or it'll cause problems going forwards (e.g. when someone hooks up Ruby bindings for the appveyor builds).

xapian-bindings/configure.ac
```diff
@@ -1234,6 +1238,39 @@ if test no != "$with_lua" ; then
...
+    if test -z "${GOROOT}" ; then
+      AC_MSG_ERROR([goroot not set correctly])
+    fi
+  fi
+  USE_CLANG_LIB=`echo ${CXX} | grep stdlib`
+  AC_MSG_RESULT([${USE_CLANG_LIB}])
+  if test -n "${USE_CLANG_LIB}" ; then
+      AC_MSG_RESULT([${USE_CLANG_LIB}])
+      AC_SUBST(USE_CLANG_GO_LIB,'-lc++')
+      AC_MSG_RESULT([${USE_CLANG_GO_LIB}])
```
ojwb, line 1267: The body of this `if` is indented by 4 columns, but all the others only by 2.

xapian-bindings/configure.ac
```diff
@@ -1234,6 +1238,39 @@ if test no != "$with_lua" ; then
...
+    GOPATH=`"${GO}" env GOPATH`
+    AC_SUBST(GOROOT)
+    AC_SUBST(GOPATH)
+    if test -n "${GOROOT}" ; then
+      AC_MSG_RESULT([${GOROOT}])
+      AC_MSG_CHECKING(for gopath)
+      if test -n "${GOPATH}" ; then
+        AC_MSG_RESULT([${GOPATH}])
+	else
+	AC_MSG_RESULT(no)
```
ojwb, line 1255: `else` body not indented.

.appveyor.yml
ojwb: Why are you installing `libuuid-devel`?  We should generate UUIDs via a win32 API call under mingw.

.appveyor.yml
```diff
@@ -58,16 +58,18 @@ matrix:
...
         $env:PATH="C:\Mingw\bin;C:\Mingw\MSYS\1.0\bin;$env:PATH"
       } elseif ($env:TOOLCHAIN -eq "mingw64") {
         if ($env:Platform -eq "x86") {
           $env:PATH="C:\mingw-w64\i686-6.3.0-posix-dwarf-rt_v5-rev1\mingw32\bin;C:\msys64\usr\bin;$env:PATH"
-          $env:CONFIGOPTS+=" --host=i686-pc-mingw32"
         } else {
           $env:PATH="C:\mingw-w64\x86_64-7.2.0-posix-seh-rt_v5-rev1\mingw64\bin;C:\msys64\usr\bin;$env:PATH"
-          $env:CONFIGOPTS+=" --host=x86_64-w64-mingw32"
+          $env:CONFIGOPTS+=" --host=x86_64-w64-mingw32 --with-go"
+          $env:BINDINGS+=" xapian-bindings "
```
ojwb, line 72: I'd just set (rather than append) since `BINDINGS` is either empty or `xapian-bindings`.  And I'd drop the extra spaces around the value - they aren't needed and make the substituted command-line look a bit messy.

.appveyor.yml
srinivasyadav18: I would change the swig part soon,..there are many parts I need to clean up...I will complete them asap.. and for lu

> > ... `doxygen mingw-w64-x86_64-doxygen doxygen` ...
> 
> Should doxygen really be listed twice like that?
> 
> Alnd what's the difference between doxygen and mingw-w64-x86_64-doxygen? Do we really need to install both?
> 
> Installing the `swig` package here seems wrong - we deliberately have bootstrap install a particular version of SWIG (specified by a git commit hash) so that we're using a known-good version of SWIG. We've found that we sometimes end up needing a more recent SWIG (for a particular bug fix or new feature) than is readily available as a package on common platforms. E.g. currently git master is using an unreleased snapshot of SWIG because the most recent release doesn't work with Ruby 2.7. Using the version of SWIG mingw64 have packaged may work OK for Go right now, but this really should work like a developer's tree would or it'll cause problems going forwards (e.g. when someone hooks up Ruby bindings for the appveyor builds).

I would change this soon..there are still many parts for clean up ..I would finish them up soon ..

srinivasyadav18: > Why are you installing `libuuid-devel`? We should generate UUIDs via a win32 API call under mingw.

it failed saying this --> https://ci.appveyor.com/project/srinivasyadav18/xapian/builds/32381253
so after installing libuuid-devel it worked

Dockerfile
ojwb: You've not addressed this.

xapian-bindings/configure.ac
ojwb: Not addressed.

xapian-bindings/configure.ac
ojwb: Not addressed.

xapian-bindings/configure.ac
```diff
@@ -1234,6 +1238,48 @@ if test no != "$with_lua" ; then
...
+      AC_MSG_ERROR([goroot not set correctly])
+    fi
+  fi
+  USE_CLANG_LIB=`echo ${CXX} | grep stdlib`
+  AC_MSG_RESULT([${USE_CLANG_LIB}])
+  if test -n "${USE_CLANG_LIB}" ; then
+    AC_MSG_RESULT([${USE_CLANG_LIB}])
+    AC_SUBST(USE_CLANG_GO_LIB,'-lc++')
+    AC_MSG_RESULT([${USE_CLANG_GO_LIB}])
+  fi
```
ojwb, line 1277: There should be an `AC_MSG_CHECKING` for each `AC_MSG_RESULT` (for each one called that is) or else the output will look weird, reporting results but not clear for what.

This also looks very dubious to me, and could misdetect if the `stdlib` is present elsewhere - e.g. `GXX=/home/stdlib/bin/gcc`, or misdetect the other way if `CXXFLAGS=-stdlib=libc++` is used instead of `CXX='clang -stdlib=libc++'`.

What is this actually trying to achieve?

xapian-bindings/go/docs/index.rst
ojwb: Not dealt with.

.appveyor.yml
```diff
@@ -95,14 +95,14 @@ install:
...
   # cygwin1.dll is first on PATH.
-  - if "%TOOLCHAIN%-%Platform%"=="cygwin-x64" C:\cygwin64\setup-x86_64.exe -qnNdO -R C:/cygwin64 -s http://cygwin.mirror.constant.com -l C:/cygwin64/var/cache/setup -P zlib-devel -P git -P patch
-  - if "%TOOLCHAIN%-%Platform%"=="cygwin-x86" C:\cygwin64\setup-x86_64.exe -qnNdO -R C:/cygwin64 -s http://cygwin.mirror.constant.com -l C:/cygwin64/var/cache/setup -P cygwin32-zlib -P cygwin32-gcc-core -P cygwin32-gcc-g++ -P git -P patch
-  - bash -c 'time ./bootstrap --fetch-url-command="curl -L" xapian-core $BINDINGS'
+  - if "%TOOLCHAIN%-%Platform%"=="cygwin-x64" appveyor-retry C:\cygwin64\setup-x86_64.exe -qnNdO -R C:/cygwin64 -s http://cygwin.mirror.constant.com -l C:/cygwin64/var/cache/setup -P zlib-devel -P git -P patch
+  - if "%TOOLCHAIN%-%Platform%"=="cygwin-x86" appveyor-retry C:\cygwin64\setup-x86_64.exe -qnNdO -R C:/cygwin64 -s http://cygwin.mirror.constant.com -l C:/cygwin64/var/cache/setup -P cygwin32-zlib -P cygwin32-gcc-core -P cygwin32-gcc-g++ -P git -P patch
+  - bash -c 'time ./bootstrap --fetch-url-command="curl --retry 5 --retry-connrefused -L" xapian-core'
```
ojwb: While resolving conflicts during the merge from master you've removed the only place that actually uses `$BINDINGS`, so all your changes towards CI of the bindings on appveyor no longer do anything useful.

rafaelsierra:
This PR also doesn't cover exceptions, for example, calling `xapian.NewWritableDatabase("/foo", xapian.DB_OPEN)` will result in un`recover()`able `SIGABRT` if there is no database under `/foo`:

```
terminate called after throwing an instance of 'Xapian::DatabaseNotFoundError'
SIGABRT: abort

[dump and trace]
```

I tried adding `%include generic/except.i` but none of the `SWIG_*Error` are defined

ojwb:
@rafaelsierra It looks like the SWIG_*Error codes are irrelevant for SWIG/Go as its `SWIG_exception` macro just ignores `code`:

```
#define SWIG_exception(code, msg) _swig_gopanic(msg)
```

So for Go we can skip the "Re-rethrow" part and just return `SWIG_UnknownError` (and define that to some dummy value) for Go.

A bigger problem though is that `_swig_gopanic` just calls `_cgo_panic` which doesn't play nicely with C++ object lifetimes:

https://github.com/swig/swig/issues/1981

Are you interested in taking this patch and cleaning it up to get it merged?

It seems to have stalled currently with no response to many of the review comments.  It's promising, but rather messy with a lot of changes that seems extraneous.  I don't really know anything substantial about Go so it's unlikely to progress unless someone with Go knowledge steps in.

rafaelsierra:
Hey @ojwb, I will give it a shot, though my skills with some of the tools used are quite limited.

I will see if I can get this building and working and later we can round the corners

.appveyor.yml
ojwb: It seems we aren't detecting that the platform is "win32" there.  Debugging that...

ojwb:
Closing in favour on #325 which has picked up work on this branch.

.appveyor.yml
ojwb: I can reproduce.

The problem is that the `--host` you pass doesn't match the prefix on an install compiler so configure picks unprefixed `gcc` which is for msys2, essentially it's a modified cygwin compiler, so doesn't define `_WIN32`, etc:

```
[00:00:06] $env:CONFIGOPTS+=" --host=x86_64-w64-mingw32"
```
...
```
[00:05:31] checking for x86_64-w64-mingw32-gcc... no
[00:05:31] checking for gcc... gcc
```

And `gcc -v` reports:

```
[00:05:28] COLLECT_GCC=gcc
[00:05:28] COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-msys/11.3.0/lto-wrapper.exe
[00:05:28] Target: x86_64-pc-msys
[00:05:28] Configured with: /c/S/gcc/src/gcc-11.3.0/configure --build=x86_64-pc-msys --prefix=/usr --libexecdir=/usr/lib --enable-bootstrap --enable-shared --enable-shared-libgcc --enable-static --enable-version-specific-runtime-libs --with-arch=x86-64 --with-tune=generic --disable-multilib --enable-__cxa_atexit --with-dwarf2 --enable-languages=c,c++,fortran,lto --enable-graphite --enable-threads=posix --enable-libatomic --enable-libgomp --disable-libitm --enable-libquadmath --enable-libquadmath-support --disable-libssp --disable-win32-registry --disable-symvers --with-gnu-ld --with-gnu-as --disable-isl-version-check --enable-checking=release --without-libiconv-prefix --without-libintl-prefix --with-system-zlib --enable-linker-build-id --with-default-libstdcxx-abi=gcc4-compatible --enable-libstdcxx-filesystem-ts
[00:05:28] Thread model: posix
[00:05:28] Supported LTO compression algorithms: zlib
[00:05:28] gcc version 11.3.0 (GCC) 
```

Checking the preprocessor symbols:

```
[00:06:16] bash -c 'echo __WIN32__ _WIN32 WIN32 __WIN64__ _WIN64 WIN64 __CYGWIN__|gcc -E -'
[00:06:16] # 0 "<stdin>"
[00:06:16] # 0 "<built-in>"
[00:06:16] # 0 "<command-line>"
[00:06:16] # 1 "<stdin>"
[00:06:16] __WIN32__ _WIN32 WIN32 __WIN64__ _WIN64 WIN64 1
```

So it defines `__CYGWIN__` but not the others.

Essentially this CI job is misconfigured because it's not testing what we want to test here (which is mingw64), but it should ideally be clearer what's going on - it took me a while to work it out and I'm more familiar with Xapian's build system than most.  I'm not entirely sure how to improve that as such though.

I also think we want to use `UuidCreate()` in this configuration (we do in real cygwin) - I think that just needs `*msys*` adding as an alternative here:

```
    *-yes | *cygwin*) dnl Use built-in win32 API.
```

I'll do that and also check other places that special case `cygwin` like this.

PR #294, opened 2020-03-22
Remote backend errors testing 

slesarev-hub:
tests are not passing, because of backendmanager::kill_server calls instead of backendmanager_remotetcp::kill_server. there several problems: 

I have tried to make particular remote server killable, but noticed that it is not only backendmanager_remote class used in "remote" test cases. there are backendmanager_mutli used during test invoking with several db classes i guess.
I want to know is my logic correct of making changes in xapian-core/tests/harness/backendmanager_remotetcp.cc correct? and can you explain how backendmanager_multi used in "remote" tests?

### DIFF
--- a/xapian-core/backends/dbfactory_remote.cc
+++ b/xapian-core/backends/dbfactory_remote.cc
@@ -28,9 +28,15 @@
 
 #include <string>
 
+#ifdef HAVE_FORK
+# include <signal.h>
+#endif
+
 using namespace std;
 
 namespace Xapian {
+
+map<string, pid_t> uuid_to_pid;
 
 Database
 Remote::open(const string &host, unsigned int port, unsigned timeout_,
@@ -57,7 +63,9 @@ Remote::open(const string &program, const string &args,
 	     unsigned timeout_)
 {
     LOGCALL_STATIC(API, Database, "Remote::open", program | args | timeout_);
-    RETURN(Database(new ProgClient(program, args, timeout_ * 1e-3, false, 0)));
+	auto rc = new ProgClient(program, args, timeout_ * 1e-3, false, 0);
+	uuid_to_pid[rc->get_uuid()] = rc->get_child();
+	RETURN(Database(rc));
 }
 
 WritableDatabase
@@ -65,8 +73,29 @@ Remote::open_writable(const string &program, const string &args,
 		      unsigned timeout_, int flags)
 {
     LOGCALL_STATIC(API, WritableDatabase, "Remote::open_writable", program | args | timeout_ | flags);
-    RETURN(WritableDatabase(new ProgClient(program, args,
-					   timeout_ * 1e-3, true, flags)));
+	auto rc = new ProgClient(program, args,
+					   timeout_ * 1e-3, true, flags);
+	auto db = WritableDatabase(rc);
+	uuid_to_pid[db.get_uuid()] = rc->get_child();
+	RETURN(db);
+}
+
+bool
+Remote::kill_server(const std::string& uuid)
+{
+#ifdef HAVE_FORK
+    auto pidit = uuid_to_pid.find(uuid);
+    if (pidit != uuid_to_pid.end()) {
+	if (kill((*pidit).second, SIGKILL) == -1) {
+	    string msg("Couldn't kill the remote server");
+	    msg += strerror(errno);
+	    throw msg;
+	}
+	uuid_to_pid.erase(pidit);
+	return true;
+    }
+#endif
+    return false;
 }
 
 }
--- a/xapian-core/tests/api_closedb.cc
+++ b/xapian-core/tests/api_closedb.cc
@@ -490,8 +490,11 @@ DEFINE_TESTCASE(closedb10, writable && metadata) {
 DEFINE_TESTCASE(closedb11, remote) {
 	Xapian::WritableDatabase db(get_writable_database());
 	kill_server(db.get_uuid());
-	TEST_EXCEPTION(Xapian::NetworkError, db.add_document(Xapian::Document()));
-    TEST_EXCEPTION(Xapian::NetworkError, db.get_doccount());
-    TEST_EXCEPTION(Xapian::NetworkError, db.commit());
+	TEST_EXCEPTION(Xapian::NetworkError,
+		db.add_document(Xapian::Document()));
+    TEST_EXCEPTION(Xapian::NetworkError,
+		db.get_doccount());
+    TEST_EXCEPTION(Xapian::NetworkError,
+		db.commit());
 	return true;
 }
--- a/xapian-core/tests/api_closedb.cc
+++ b/xapian-core/tests/api_closedb.cc
@@ -484,6 +484,14 @@ DEFINE_TESTCASE(closedb10, writable && metadata) {
 		   db.get_metadata("bar"));
     TEST_EXCEPTION(Xapian::DatabaseClosedError,
 		   db.metadata_keys_begin());
+	return true;
+}
 
-    return true;
+DEFINE_TESTCASE(closedb11, remote) {
+	Xapian::WritableDatabase db(get_writable_database());
+	kill_server(db.get_uuid());
+	TEST_EXCEPTION(Xapian::NetworkError, db.add_document(Xapian::Document()));
+    TEST_EXCEPTION(Xapian::NetworkError, db.get_doccount());
+    TEST_EXCEPTION(Xapian::NetworkError, db.commit());
+	return true;
 }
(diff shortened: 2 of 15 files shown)


jaylett:
@slesarev-hub you don't have to make every line change as a separate commit :) please read the notes in our developer guide for suggestions on how to structure commits.

Given you've only just opened this and no one has reviewed it yet, I've changed the name to WIP (so people won't review) to give you time to sort your commits out; there are suggestions in the developer guide, and articles we link to that explain what we're looking for. When it's ready to review, please drop the WIP flag and let us know on IRC.

As I noted on IRC, I think you need to discuss how things work inside the test system, to help you figure out how to proceed to get the tests to pass. At that point it'll be possible to review this PR.

jaylett:
@slesarev-hub in future, try to use fixup commits rather than giving commits subjects like "style fix"; it's easier to squash the "fix" bit to the original change when we merge, which makes for a more useful git history.

xapian-core/tests/api_closedb.cc
```diff
@@ -488,7 +488,7 @@ DEFINE_TESTCASE(closedb10, writable && metadata) {
 }
 
 DEFINE_TESTCASE(closedb11, remote) {
-	Xapian::WritableDatabase db(get_writable_database());
+	Xapian::WritableDatabase db = get_writable_database();
```
jaylett: You didn't need to change this, so it's causing the commit to be about more than it needs to be. You can create a fixup that undoes just this change.

xapian-core/tests/harness/backendmanager_multi.cc
```diff
@@ -248,3 +248,18 @@ BackendManagerMulti::get_writable_database_path_again()
 {
     return last_wdb_path;
 }
+
+bool
+BackendManagerMulti::kill_server(const std::string& uuid) {
```
jaylett: This is also not covered by the commit message.

xapian-core/backends/dbfactory_remote.cc
```diff
@@ -28,10 +28,16 @@
...
+#ifdef HAVE_FORK
+# include <signal.h>
+#endif
+
 using namespace std;
 
 namespace Xapian {
 
+map<string, pid_t> uuid_to_pid;
+
```
ojwb: This is the wrong place to be tracking this sort of thing.  This should be handled entirely in the test harness rather than requiring modifications to the library.  Doing this here adds overhead for library users for no good reason.

xapian-core/tests/api_closedb.cc
```diff
@@ -484,6 +484,17 @@ DEFINE_TESTCASE(closedb10, writable && metadata) {
 		   db.get_metadata("bar"));
     TEST_EXCEPTION(Xapian::DatabaseClosedError,
 		   db.metadata_keys_begin());
+	return true;
+}
 
-    return true;
+DEFINE_TESTCASE(closedb11, remote) {
+	Xapian::WritableDatabase db(get_writable_database());
+	kill_server(db.get_uuid());
```
ojwb: I think the UUID of the database is not a good way to specify which server to kill.  The same database can be open multiple times for reading, as well as potentially once for writing.  While the UU stands for "Universally Unique", that just means each database on disk should have a different UUID - if you open the same database twice the UUID won't be unique, and if the specified UUID maps to more than one open database then we won't know which server process to kill.

Passing `kill_server()` a pointer to the `Database`/`WritableDatabase` object might work.  At least for a single-shard database it can just read the internal pointer and use that to find the process id.

xapian-core/tests/harness/backendmanager.cc
```diff
@@ -302,6 +302,14 @@ BackendManager::clean_up()
 {
 }
 
+bool
+BackendManager::kill_server(const std::string& uuid)
+{
+    throw Xapian::InvalidOperationError("kill_server called for " +
+    uuid + ", but available only for remote servers");
+    return false;
```
ojwb: There's no point calling `return` after an unconditional `throw` as it can never be reached.

xapian-core/tests/harness/backendmanager_remotetcp.cc
```diff
@@ -211,19 +216,21 @@ launch_xapian_tcpsrv(const string & args)
...
+	unsigned pididx = 0;
+	for (unsigned i = 0; i < sizeof(pid_to_fd) / sizeof(pid_fd); ++i) {
 	if (pid_to_fd[i].pid == 0) {
-	    pid_to_fd[i].fd = tracked_fd;
-	    pid_to_fd[i].pid = child;
-	    break;
+		pid_to_fd[i].fd = tracked_fd;
+		pid_to_fd[i].pid = child;
+		pididx = i;
+		break;
```
ojwb: You've reindented the above and now it no longer conforms with our coding conventions.

xapian-core/tests/harness/backendmanager_remotetcp.cc
```diff
@@ -341,6 +348,9 @@ launch_xapian_tcpsrv(const string & args)
 # error Neither HAVE_FORK nor __WIN32__ is defined
 #endif
 
+// map exact database object to it's pid_fd position in pid_to_fd
+static std::map<std::string, unsigned> uuid_to_pididx;
```
ojwb: We don't want to use a `std::map` here - it causes problems because it dynamically allocates memory, and that confuses the test harness leak checking machinery as the existing comment above (~line 73) notes.

Just use a static fixed size array and linearly scan like we already do.  It doesn't need a lot of entries, so it's probably as cheap to linearly scan anyway (it being O(n) is irrelevant when n is small and fixed).

xapian-core/tests/harness/backendmanager_remotetcp.cc
```diff
@@ -364,41 +374,65 @@ BackendManagerRemoteTcp::get_writable_database(const string & name,
 					       const string & file)
 {
     string args = get_writable_database_args(name, file);
-    int port = launch_xapian_tcpsrv(args);
-    return Xapian::Remote::open_writable(LOCALHOST, port);
+	auto pp = launch_xapian_tcpsrv(args);
+	auto db = Xapian::Remote::open_writable(LOCALHOST, pp.port);
+	uuid_to_pididx[db.get_uuid()] = pp.pididx;
+	return db;
```
ojwb: Another incorrect indentation change.

xapian-core/tests/harness/backendmanager_remotetcp.cc
```diff
@@ -364,41 +374,65 @@ BackendManagerRemoteTcp::get_writable_database(const string & name,
...
 
 Xapian::Database
 BackendManagerRemoteTcp::get_remote_database(const vector<string> & files,
 					     unsigned int timeout)
 {
     string args = get_remote_database_args(files, timeout);
-    int port = launch_xapian_tcpsrv(args);
-    return Xapian::Remote::open(LOCALHOST, port);
+    auto pp = launch_xapian_tcpsrv(args);
+    return Xapian::Remote::open(LOCALHOST, pp.port);
```
ojwb: Here you don't record the mapping needed to kill the server, nor below.

Rather than add the same handling in 3 places, it would be better for `launch_xapian_tcpsrv()` to take care of updating that mapping, and then it can just return the port number as it currently does.

xapian-core/backends/dbfactory_remote.cc
```diff
@@ -57,16 +63,39 @@ Remote::open(const string &program, const string &args,
 	     unsigned timeout_)
 {
     LOGCALL_STATIC(API, Database, "Remote::open", program | args | timeout_);
-    RETURN(Database(new ProgClient(program, args, timeout_ * 1e-3, false, 0)));
+	auto rc = new ProgClient(program, args, timeout_ * 1e-3, false, 0);
+	uuid_to_pid[rc->get_uuid()] = rc->get_child();
+	RETURN(Database(rc));
```
ojwb: More incorrect reindentation.  And yet more below.  Please check your editor settings.

jaylett:
License declaration received on mailing list (22nd March 2020).

jaylett:
@slesarev-hub I've been trying to contact you about your GSoC application by email and on IRC. Can you please try to get in touch in the next day or two (IRC will be easiest)?

ojwb:
I've pushed an implementation of this using the approach I suggested in the comments above:

> Passing kill_server() a pointer to the Database/WritableDatabase object might work. At least for a single-shard database it can just read the internal pointer and use that to find the process id.

The initial implementation only handles a single-shard which is remotetcp, but that's give pretty good coverage.  I've noted approaches for extending this support to remoteprog and sharded databases in the commit message.

PR #296, opened 2020-04-23
ruby: Don't print iterator sizes to stdout

danc86:
This was introduced in commit 4b6a37cd, presumably just some debugging
left in by mistake. It causes unwanted mess in the output of the
application using xapian bindings.


ojwb:
Thanks, rebased and pushed as 3e848a9a3d20a10bc7431feda1a2e51b43233045.

PR #297, opened 2020-05-04
remove emscripten-specific workaround

petersalomonsen:
fixed in https://github.com/emscripten-core/emscripten/pull/10631

refered by issues:
https://github.com/emscripten-core/emscripten/issues/8187
https://github.com/emscripten-core/emscripten/issues/10628

### DIFF
--- a/xapian-core/backends/glass/glass_version.cc
+++ b/xapian-core/backends/glass/glass_version.cc
@@ -295,25 +295,9 @@ GlassVersion::write(glass_revision_number_t new_rev, int flags)
 	else
 	    tmpfile += "/v.tmp";
 
-#ifdef __EMSCRIPTEN__
-	// Emscripten fails to create a file if O_TRUNC is specified and the
-	// filename is the previous name of a renamed file (which it will be
-	// the second time we write out the version file for a DB):
-	//
-	// https://github.com/emscripten-core/emscripten/issues/8187
-	//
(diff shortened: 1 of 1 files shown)


ojwb:
It's really great that you got to the bottom of this and got a fix merged.  Thanks for pushing through on that.

I wonder if we want to hold of merging this revert of the workaround for the time being though - while this is now fixed in the emscripten git repo, existing releases will still have the bug.  The workaround itself is safe and only a tiny amount of extra work which is not on a hot code path.  We could also make the check for whether to use it look at the emscripten version (https://emscripten.org/docs/compiling/Building-Projects.html#detecting-emscripten-in-preprocessor).

petersalomonsen:
Yeah good point. The latest emscripten release has the fix though (the travis test uses the latest release), but for earlier releases the workaround will still have to be used.

ojwb:
OK, https://github.com/emscripten-core/emscripten/commit/9fdc8af8e9f5a419b1751cadf1304c88210a6b4c shows the fix was in 1.39.10.  I think for now I'll push an update to the comment about the workaround to note it's fixed in that version.

ojwb:
8987d572faba01a3e130a93dea69875ecab690b5 updates the comment and adds a note to `xapian-core/emscripten/README.md` recommending using at least 1.39.10.

PR #298, opened 2020-06-07
Add New normalizations for Tf-Idf weighting schemes

dipanshu124:
Add **Global frequency IDF , Augmented log , Log Global frequency IDF and Square root** Normalizations along with tests .


xapian-core/weight/tfidfweight.cc
```diff
@@ -236,6 +249,10 @@ TfIdfWeight::get_idfn(char c) const
 	    return pow(log(N / termfreq), 2.0);
 	case 'P':
 	    return log((N + 1) / termfreq);
+	case 'G':
+	    return (double(collfreq) / termfreq);
+	case 'l':
+	    return log(double(collfreq) / termfreq + 1);
```
ojwb, line 255: There's no `case 'E':` here.

xapian-core/weight/tfidfweight.cc
```diff
@@ -89,6 +89,9 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
 	need_stat(DOC_LENGTH_MAX);
 	need_stat(UNIQUE_TERMS);
     }
+    if (!strchr("ntpfsP", normalizations[1])) {
+	need_stat(COLLECTION_FREQ);
+    }
```
ojwb, line 94: Perhaps better to check against a list of those normalisations which do need it, rather than those than don't?  Especially as that's a shorter list.

xapian-core/weight/tfidfweight.cc
```diff
@@ -223,6 +234,8 @@ TfIdfWeight::get_idfn(char c) const
     if (c != 'n') termfreq = get_termfreq();
     double N = 1.0;
     if (c != 'n' && c != 'f') N = get_collection_size();
+    Xapian::termcount collfreq = 1;
+    if (c == 'E' || c == 'G' || c == 'l') collfreq = get_collection_freq();
```
ojwb, line 238: We're about to switch on `c`, so it seems cleaner to call `get_collection_freq()` directly in the cases that need it, rather than effectively have a pre-dispatch on `c`.

(I realise this is an existing pattern in this code, but I don't think it's a good pattern to extend further)

xapian-core/weight/tfidfweight.cc
```diff
@@ -223,6 +234,8 @@ TfIdfWeight::get_idfn(char c) const
     if (c != 'n') termfreq = get_termfreq();
     double N = 1.0;
     if (c != 'n' && c != 'f') N = get_collection_size();
```
ojwb, line 236: Your new normalisations don't seen to use `N`.

xapian-core/weight/tfidfweight.cc
```diff
@@ -210,6 +213,14 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
 	    double den = 1 + log(wdf_avg);
 	    return num / den;
 	}
+	case 'a': {
+	    if (wdf == 0) return 0;
+	    return (0.2 + 0.8 * log(1.0 + wdf));
```
ojwb, line 218: This doesn't seem to match the definition of "a" for the first normalisation at http://people.csail.mit.edu/jrennie/ecoc-svm/smart.html : 

> "**a**ug-norm" : new-tf = 0.5 + 0.5 * (tf / max-tf)
>              augmented normalized tf.  0.5 < new-tf <= 1.0

xapian-core/weight/tfidfweight.cc
```diff
@@ -210,6 +213,14 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
...
+	}
+	case 'S': {
+	    if (wdf == 0) return 0;
+	    return (sqrt(wdf - 0.5) + 1);
+	}
```
ojwb, line 223: Where's "S" defined?  I don't see it listed at http://people.csail.mit.edu/jrennie/ecoc-svm/smart.html or https://nlp.stanford.edu/IR-book/html/htmledition/document-and-query-weighting-schemes-1.html

xapian-core/weight/tfidfweight.cc
```diff
@@ -63,8 +63,8 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
     : normalizations(normals), param_slope(slope), param_delta(delta)
 {
     if (normalizations.length() != 3 ||
-	!strchr("nbslPL", normalizations[0]) ||
-	!strchr("ntpfsP", normalizations[1]) ||
+	!strchr("nbslPLaS", normalizations[0]) ||
+	!strchr("ntpfsPEGl", normalizations[1]) ||
```
ojwb, line 67: You've not documented any of your new normalisations in the API docs (see `include/xapian/weight.h` for where all the currently supported ones are defined).

xapian-core/tests/api_weight.cc
```diff
@@ -882,6 +882,39 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
...
+    TEST_EQUAL_DOUBLE(mset[0].get_weight(), sqrt(8 - 0.5) + 1);
+    TEST_EQUAL_DOUBLE(mset[1].get_weight(), sqrt(1 - 0.5) + 1);
+
+    // Check for "nln" .
+    enquire.set_weighting_scheme(Xapian::TfIdfWeight("nln"));
+    mset = enquire.get_mset(0, 10);
+    TEST_EQUAL(mset.size(), 2);
+    mset_expect_order(mset, 2, 4);
+    TEST_EQUAL_DOUBLE(mset[0].get_weight(), 8 * log(9.0 / 2 + 1));
+    TEST_EQUAL_DOUBLE(mset[1].get_weight(), 1 * log(9.0 / 2 + 1));
```
ojwb, line 916: There's no test of 'E' for the second character here (which would probably have revealed that you didn't actually provide an implementation for it).

xapian-core/tests/api_weight.cc
```diff
@@ -882,6 +882,39 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
...
+
 }
```
ojwb, line 918: Xapian coding style is to avoid blank lines at the end of a block.

xapian-core/weight/tfidfweight.cc
ojwb: Also, where are "E", "G" and "l" defined?

ojwb: OK, it seems these are from the paper you link to from https://trac.xapian.org/wiki/GSoC2020/WeightingSchemes/ProjectPlan

However that doesn't seem to define the single letter codes you use here.  I think we need to take a lot of care if we're inventing new single letter codes as this is a pre-existing naming scheme rather than something we came up with, there are only 52 different letters (allowing for case-sensitivity) and it would be awkward to deal with if we picked one which was already used for a different normalisation elsewhere (I'm guessing that's why your implementation of "a" below doesn't match the definition of "a" I was looking at).  Or indeed if we picked an unused one and somebody else picked the same one later for a different purpose.

xapian-core/tests/api_weight.cc
dipanshu124: Updated.

xapian-core/weight/tfidfweight.cc
dipanshu124: Right, I will take of it .

xapian-core/tests/api_weight.cc
dipanshu124: 'E' is for Entropy normalization . I haven't implemented it yet .
It requires visibility of wdf of the term in every document in the collection . The idea was to use the postlist to get that.
But the concern is that won't be very efficient.
So it is still to be implemented

xapian-core/weight/tfidfweight.cc
dipanshu124: 'A' is for Augmented log (I have changed from 'a' to 'A') .

xapian-core/weight/tfidfweight.cc
dipanshu124: All these characters I used, represent the weighting schemes mentioned in the paper I link to from https://trac.xapian.org/wiki/GSoC2020/WeightingSchemes/ProjectPlan .
Since the paper used 4 letters to define a particular normalization (for which we use only 1 letter) , I took the liberty of changing the 4 letter name to a relevant 1 letter parameter (Sorry, I should have consulted before doing that.)
I have ensured that the parameters I used haven't been used elsewhere till now .(I have changed 'a' to 'A' for Augmented log.)
I have now defined these parameters in API docs .
If we can't use an unused character , what else can be done?

jaylett: Can we come up with a way of maintaining compatibility with single letter approach we used that matched the original paper (for people who are close to the literature and might expect that), but also start providing a more user-friendly API using a bitfield or similar?

We could provide both, but the bitfield would enable us to support normalisations that don't have single letter variants (giving us options beyond 26 letters), which could include this situation rather than trying to come up with our own single letter abbreviations.

(The constants for a bitfield approach could if necessary include both "verbose and clear" and "mnemonic from papers", which could be say `IDFB` or `CK_IDFB` or similar, with API doc comments explaining that CK means that particular paper.)

Thinking aloud here a little, so I suspect there's a discussion and maybe worth sketching out different ways of doing the API to determine a good route forward.

ojwb: Yes, I think enums might be a better approach (though perhaps with a separate parameter for each of the 3 normalisation types rather than a bitmask, since that provides more scope for compile-time type-checking).

It's inherently problematic to have such a small space to allocate from without a central allocation authority (though it is 52 not 26 as it's case sensitive) .  For academic papers it presumably works out OK because academics who write papers which extend the SMART codes will have read the existing literature in the area they're working in (and if they miss one, the paper gets peer reviewed and there's a second opportunity for someone to catch a collision).  But I suspect academics are less likely to spot an "allocation" which is only recorded in Xapian.

The 3 letter codes are also rather cryptic.  They come from SMART which originated more than 50 years ago - we can afford to be a bit more verbose now than they could then.

The one advantage of a string code which remains is it's easy to support as a config option in an application built on Xapian, but we now support creating a `Weight` object from a string description so that's no longer really relevant.

dipanshu124:
This PR has a lot of commits. (They aren't fixup commits either.)
I don't know what to do with them at this point.

ojwb:
> This PR has a lot of commits. (They aren't fixup commits either.)
> I don't know what to do with them at this point.

There isn't a lot you can usefully do at this point (you could rebase and merge the commits to group changes sensibly, or otherwise rebuild the branch, but that makes review harder so isn't a good option at this point given we're already well into the review process).  That's why it's important to think about how to structure the changes before you open a PR, and when using fixup commits to make sure they're correctly specified and targetted at the appropriate existing commit.

I think we'll just have to review the changes as a whole and squash down to a single commit when we merge.

xapian-core/tests/api_weight.cc
dipanshu124: I will make a separate PR for Entropy normalization.

ojwb: Good plan.

PR #299, opened 2020-06-09
Update in README file

divyanshugit:
The change in **README** file will better the _viewing experience_ and _readability_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Welcome to the Xapian repository!
+# Welcome to the Xapian repository!
 =================================
 
 Xapian is split into several modules, which are held in sub-directories of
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for the patch.

However, this file is already formatted using [restructured text](https://docutils.sourceforge.io/docs/ref/rst/introduction.html) - this is broadly similar to markdown in concept, but better thought out.  It's what we use for all our other documentation so I think it makes sense to stick to it.  Otherwise developers would have to deal with two different text markup languages.

It's also supported by github, but only if the file has a `.rst` extension (at least I don't know of another way).  So if you want nicer formatting in our README on github, adding a `.rst` extension here is the way to do that, but that needs a careful check for places that refer to this file by name (`xapian-maintainer-tools/make-xapian-git-snapshot-tarballs` does, but there are probably other references).

divyanshugit:
@ojwb  Can you suggest me ways to get start to contributing to this repository ?

divyanshugit:
@ojwb  If I change extension of the **README** file to .rst it will be get accepted ?

ojwb:
@divyanshugit I think it'd be OK to rename it to `README.rst`, but as I noted above we should make sure we update references to the current name.  The snapshot tarball script is probably the only one actually - any links would probably just be to https://github.com/xapian/xapian (which shows the content of the README file) rather than directly to the README.  I did a quick grep in the likely places and didn't find any other references.

We're happy to welcome new contributors.  What were you thinking of working on?

If you just find it an interesting project but don't really have an idea of something specific to do then I'd suggest looking at:

https://trac.xapian.org/wiki/GSoC%20Guide

Most of that is useful even if you're not thinking about taking part in GSoC, but you might want to look at the ideas for shorter projects at https://trac.xapian.org/wiki/ProjectIdeas (GSoC involves 3 months of full-time coding work.)

PR #300, opened 2020-06-10
Omindex libarchive zip

Exter-dg:
Handler for extracting data from OpenDocument Formats using Libarchive


xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+	string& keywords,
+	string& author,
+	string& pages,
+	string& error)
+{
+    try {
+	struct archive* a;
+	struct archive_entry* entry;
+	int status_code;
+	const char* file = &filename[0];
```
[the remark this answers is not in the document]
Exter-dg: I have changed this and committed again,

`const char* file = filename.c_str();`

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+	status_code = archive_read_open_filename(a, file, 10240);
+
+	if (status_code != ARCHIVE_OK) {
+	    error = "Libarchive failed to open the file " + filename;
+		return false;
+	}
+
+	size_t total;
+	ssize_t size;
+	string s = "";
```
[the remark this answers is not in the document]
Exter-dg: Thanks for pointing this out, I have edited this.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+extract(const string& filename,
+	string& dump,
+	string& title,
+	string& keywords,
+	string& author,
+	string& pages,
+	string& error)
+{
+    try {
+	struct archive* a;
```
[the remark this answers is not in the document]
Exter-dg: a, here, is an archive object (pointer to struct archive). I have renamed `a` to `archive_obj`.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+ */
+#include <config.h>
+#include "handler.h"
+
+#include "metaxmlparse.h"
+#include "opendocparse.h"
+#include "str.h"
+
+#include<archive.h>
+#include<archive_entry.h>
```
barufa, line 29: Missing space

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+		char buf1[total];
+		size = archive_read_data(archive_obj, buf1, total);
+
+		if (size <= 0) {
+		    error = "Libarchive was not able to extract data from "
+		    "content.xml";
+		    return false;
+		}
+
+		s += str(buf1);
```
barufa, line 76: It would be better using different strings to store the information extracted as we cannot assume any particular order in which libarchive browses on the file.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+
+	size_t total;
+	ssize_t size;
+	string s;
+
+	// extracting data from content.xml and styles.xml
+	while (archive_read_next_header(archive_obj, &entry) == ARCHIVE_OK) {
+	    if (strcmp(archive_entry_pathname(entry), "content.xml") == 0) {
+		total = archive_entry_size(entry);
+		char buf1[total];
```
barufa, line 67: This is not a proper way to declare variable length arrays in C++ as some version of the standard doesn't support it. 
You can use a constant-expression or the operators `new` and `delete`.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+		if (size <= 0) {
+		    error = "Libarchive was not able to extract data from "
+		    "styles.xml";
+		    return false;
+		}
+
+		s += str(buf2);
+		OpenDocParser parser;
+		parser.parse(s);
+		dump = parser.dump;
```
barufa, line 92: As you cannot assume any order in which files are shown, you cannot parse the content of the file here. It is highly possible that `s` doesn't store the information extracted from `content.xml` at this point so you should parse the information at the end of the program.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+		MetaXmlParser metaxmlparser;
+		metaxmlparser.parse(s);
+		title = metaxmlparser.title;
+		keywords = metaxmlparser.keywords;
+		author = metaxmlparser.author;
+		}
+	}
+	status_code = archive_read_free(archive_obj);
+	if (status_code != ARCHIVE_OK) {
+	    return false;
```
barufa, line 113: Not sure if we should return `false` here and drop all the extracted information.

.travis.yml
```diff
@@ -141,6 +144,7 @@ matrix:
             - python
             - python2
             - tesseract
+            - libarchive
```
ojwb, line 147: When you're adding a new item to an existing list, look to see if the list is in sorted order (as this one was before your addition) - if so, you should probably add the item to preserve that sorted order.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+{
+    try {
+	struct archive* archive_obj;
+	struct archive_entry* entry;
+	int status_code;
+	const char* file = filename.c_str();
+	archive_obj = archive_read_new();
+	archive_read_support_filter_all(archive_obj);
+	archive_read_support_format_all(archive_obj);
+	status_code = archive_read_open_filename(archive_obj, file, 10240);
```
ojwb, line 52: In C++ code, it's usually better to declare variables at the point where you can assign to them (at least if that's easily possible as it is here for `status_code` and `archive_obj`).  That way there's no point where the variable is uninitialised, which makes it very clear that the uninitialised value is never used (and harder for someone changing the code in future to accidentally introduce a use of an uninitialised value).

The style of declaring everything at the top of the function (or block) is a hang-over from older C code - before C99 you couldn't portably declare a variable except at the top of a block.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+
+	if (status_code != ARCHIVE_OK) {
+	    error = "Libarchive failed to open the file " + filename;
+		return false;
```
ojwb, line 56: over-indented

xapian-applications/omega/index_file.cc
```diff
@@ -189,6 +190,63 @@ index_add_default_libraries()
     index_library("message/rfc822", omindex_gmime);
     index_library("message/news", omindex_gmime);
 #endif
+#if defined HAVE_LIBARCHIVE
+    Worker* omindex_libarchive = new Worker("omindex_libarchive");
+    index_library("application/vnd.oasis.opendocument.text",
+    omindex_libarchive);
```
ojwb, line 196: When wrapping a parameter list, Xapian-style is to line up the wrapped parameters under the first parameter (so the `o` should be directly below the opening `"`.

(I don't think I've seen the style you've used here before - it seems confusing as it makes it look like `omindex_libarchive);` is a separate statement.)

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+	// extracting data from content.xml and styles.xml
+	while (archive_read_next_header(archive_obj, &entry) == ARCHIVE_OK) {
+	    if (strcmp(archive_entry_pathname(entry), "content.xml") == 0) {
+		total = archive_entry_size(entry);
+		char buf1[total];
+		size = archive_read_data(archive_obj, buf1, total);
+
+		if (size <= 0) {
+		    error = "Libarchive was not able to extract data from "
+		    "content.xml";
```
ojwb, line 72: The opening `"` of the wrapped part of the string should be below the opening quote on the line above.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+		    "styles.xml";
+		    return false;
+		}
+
+		s += str(buf2);
+		OpenDocParser parser;
+		parser.parse(s);
+		dump = parser.dump;
+		} else if (strcmp(
+			archive_entry_pathname(entry), "meta.xml") == 0) {
```
ojwb, line 94: Wrap after the `,`, not before the first parameter.

In general when writing code in an existing project you should look at how existing code is formatted for a situation similar to the one you're in.

xapian-applications/omega/handler_libarchive.cc
ojwb: Indeed.  I've just added a check to xapian-check-patch for this.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+		return false;
+	}
+
+	size_t total;
+	ssize_t size;
+	string s;
+
+	// extracting data from content.xml and styles.xml
+	while (archive_read_next_header(archive_obj, &entry) == ARCHIVE_OK) {
+	    if (strcmp(archive_entry_pathname(entry), "content.xml") == 0) {
```
ojwb, line 65: Currently the structure of this is:

```
    if (strcmp(archive_entry_pathname(entry), "...") == 0) {
        // ...
    } else if (strcmp(archive_entry_pathname(entry), "...") == 0) {
        // ...
    } else if (strcmp(archive_entry_pathname(entry), "...") == 0) {
        // ...
    }
```

If an entry doesn't match any of these checks (which most probably don't - a typical ODF zip container has a lot more than just these 3 files) or matches the last one then we call `archive_entry_pathname(entry)` 3 times.  It'll give the same answer each time, but the compiler has no way to tell that.

I'd suggest just calling it once and putting the result in a variable.  That'll also make the code much clearer - the conditions will be short enough to fit on one line, and if you use a string variable you can just write them as equality checks:

```
    string pathname = archive_entry_pathname(entry);
    if (pathname == "content.xml") {
        // And so on...
```

xapian-applications/omega/handler_libarchive.cc
ojwb: Yes - ISO C99 added VLAs, and some C++ compilers support them in C++ code as a non-standard extension, but we shouldn't rely on that as it limits portability.

The size needs to be a compile-time constant - you could allocate a few KB and then call `archive_read_data()` repeatedly to read the file in chunks.

If you want a variable size block, the portable C++ version is something like:

```
unique_ptr<char> buf1 = new char[total];
```

Though here you can avoid allocating a temporary buffer entirely by resizing the string to have enough space and then reading directly into that space, which also avoids an extra copy operation:

```
auto i = s.size();
s.resize(i + total);
size = archive_read_data(archive_obj, &s[i], total);
if (size <= 0) {
    // ...
}
s.resize(i + size);

Comment (ojwb) [reply xapian-applications/omega/handler_libarchive.cc:67]: Also, `buf1`, `buf2`, `buf3` aren't great variable names unless the numbers have some actual meaning (which they don't here).  You could actually just call them all `buf` in this case I think, since they each exist in disjoint blocks, but if they were in the same block then `buf_content`, `buf_styles`, `buf_meta` would be easier for the reader to follow than arbitrary numbers.

File: xapian-applications/omega/handler_libarchive.cc (hunk 5 above)
Comment (ojwb) [reply xapian-applications/omega/handler_libarchive.cc:76]: Also you really need to use the length read (which is in `size`), not rely on there being a terminating zero byte (which the libarchive documentation doesn't mention being added).  You'll often happen to have a zero byte after the read data as memory is typically zero-initialised on first allocation in modern machines, but in a longer running process as heap memory gets reused this will likely end up with extra junk getting appended to the read archive member data.

File: xapian-applications/omega/handler_libarchive.cc [hunk 16]
```diff
@@ -0,0 +1,122 @@
...
+		    // error handling
+		}
+
+		s = str(buf3);
+		MetaXmlParser metaxmlparser;
+		metaxmlparser.parse(s);
+		title = metaxmlparser.title;
+		keywords = metaxmlparser.keywords;
+		author = metaxmlparser.author;
+		}
```
ojwb, line 109: `}` overindented

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,122 @@
...
+		parser.parse(s);
+		dump = parser.dump;
+		} else if (strcmp(
+			archive_entry_pathname(entry), "meta.xml") == 0) {
+		total = archive_entry_size(entry);
+		char buf3[total];
+		size = archive_read_data(archive_obj, buf3, total);
+
+		if (size <= 0) {
+		    // error handling
```
ojwb, line 100: Needs doing!

Exter-dg: I noticed, in `index_file.cc`, that the file is indexed even if extracting from `meta.xml` fails.
So when size is <=0, I thought I shouldn't be returning false.
Also, should I be appending to `error`? `error` will be used only when bool extract returns false, am I right?

xapian-applications/omega/handler_libarchive.cc
barufa: It seems that if `archive_read_free` fails, some file descriptors can remain opened and that can cause problems(run out of file descriptors) in some scenarios. 
Have @Exter-dg  seen [`archive_error_string`](https://nxmnpg.lemoda.net/3/archive_error_string)? Maybe we can get more information to append to the error string.

xapian-applications/omega/handler_libarchive.cc
barufa: The error string is going to be used only when `extract` returns false.
I think we should ignore this error, as indexing partial information from the document can be useful. (This apply for others cases in the handler).

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have changed it.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have used string resizing as Olly suggested so char array is not being used anymore.

Also, I will use two different strings for `content.xml` and 'styles.xml`, I haven't done this yet, will make the change and push it again.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have used string resizing now. Thanks for pointing it out, I have renamed the buffers to `buf`. As they are in different blocks, it shouln't be a problem.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have now used OpenDocParser at the end of the file.
@barufa, should I use different strings to store data extracted from content.xml and styles.xml and later append them? 
Or is it alright if they are in a different order?

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Yes, I read about 'archive_error_string`, it can provide us with more information about the error produced incase archive_read_free() fails. I have changed this.

.travis.yml
Exter-dg: Thanks for pointing out, I have changed this.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I will take care of using this style, I have changed this.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have corrected this.

xapian-applications/omega/index_file.cc
Exter-dg: Thanks for pointing this out. I will take care of this in future and have changed it.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have changed this, will take care of this in future.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Thanks for pointing this out. I have used `string pathname` here, hence it doesn't require wrapping now. I will take care of this in the future.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Thanks for pointing this out. Yes, multiple calls to archive_entry_pathname() were redundant. I have changed this, used `string pathname` instead.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have corrected this.

xapian-applications/omega/handler_libarchive.cc
barufa: Not sure if it is necessary, but I would try to preserve the order. It would be great to process the file in the same way that we were doing it on `index_file.cc`.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,125 @@
...
+
+		s.resize(i + size);
+	    } else if (pathname == "meta.xml") {
+		total = archive_entry_size(entry);
+		string s_meta;
+		auto i = s_meta.size();
+		s_meta.resize(i + total);
+		size = archive_read_data(archive_obj, &s_meta[i], total);
+
+		if (size <= 0) {
```
barufa, line 98: In this part I would write some like 
```
if (size > 0) {
    s_meta.resize(i + size);
    .....
}
```
That way we execute this code only if there is some work to do and we emphasize that `s_meta` is not empty.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,125 @@
...
+			    "styles.xml";
+		    return false;
+		}
+
+		s.resize(i + size);
+	    } else if (pathname == "meta.xml") {
+		total = archive_entry_size(entry);
+		string s_meta;
+		auto i = s_meta.size();
+		s_meta.resize(i + total);
```
barufa, line 95: As we declare s_meta here, we know that s_meta is empty so it is no necessary to call `size`.

By the way, I think you want to use [`reserve`](http://www.cplusplus.com/reference/string/string/reserve/) rather than [`resize`](http://www.cplusplus.com/reference/string/string/resize/). `reserve` has a constant complexity(generally) while the complexity of `risize` is linear.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,125 @@
...
+	status_code = archive_read_free(archive_obj);
+	if (status_code != ARCHIVE_OK) {
+	    error += archive_error_string(archive_obj);
+	    return false;
+	}
+
+	OpenDocParser parser;
+	parser.parse(s);
+	dump = parser.dump;
+	} catch (...) {
```
barufa, line 119: Over indented

xapian-applications/omega/handler_libarchive.cc
ojwb: It is necessary - that's why `index_file.cc` calls `unzip` separately for `context.xml` and then for `styles.xml`, rather than calling `unzip` once and listing both files - see 5811711bb0f3b9dca7f8be6279290df31fa80397 (which fixes a bug where we were processing them in the order they happened to be inside the zip container).

xapian-applications/omega/handler_libarchive.cc
ojwb: But we need to resize `s_meta` before we call `archive_read_data()` so there's space there to read the data into, so we can't conditionalise the resize on the value returned by `archive_read_data()`!

barufa: Yes, but we are doing it already in line 95. This second call is to shorten the string if necessary.
Here I meant that we could move all the code inside the if, that way we use `MetaXmlParser` only when we get some information to process.
```
if (size > 0) {
    s_meta.resize(size);
    MetaXmlParser metaxmlparser;
    metaxmlparser.parse(s_meta);
    title = metaxmlparser.title;
    keywords = metaxmlparser.keywords;
    author = metaxmlparser.author;
}
```

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have changed this, have used different strings for content.xml and styles.xml and appended them to s.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I changed this in the last commit only, I am not sure why it is not updated in the remote repo, I have added this now.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Shouldn't this be true for s_content and s_styles too? I have removed `auto i = s_file.size()` for all three strings.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Thanks for pointing this out, Corrected.

xapian-applications/omega/handler_libarchive.cc
barufa: Yes, that can also  be applied to s_content and s_styles.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -80,6 +80,46 @@ index_test()
     tests.insert({"img/scan-page.png", {"Zannual", "Zfed", "Zreturn",
 		  "Zwhile"}});
 #endif
+#if defined HAVE_LIBARCHIVE
+    // blank file
+    tests.insert({"odf/blank.odt", {}});
+    // corrupted file (ODP)
+    // "no text extracted from document body, but indexing metadata anyway"
+    // error on searching for any term - term not found
+    // tests.insert({"corrupt_file.odp", {"ZSnatur"}});
```
barufa, line 89: Not sure if these comments should be here. I know they are going to be relevant in the near future if some modification are applied, but at the moment the doesn't add any relevant information.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,121 @@
...
+	    error = "Libarchive failed to open the file " + filename;
+	    return false;
+	}
+
+	size_t total;
+	ssize_t size;
+	string s, s_content, s_styles;
+	struct archive_entry* entry;
+
+	// extracting data from content.xml and styles.xml
```
barufa, line 62: I think you forget meta.xml in this comment, right?

xapian-applications/omega/omindexcheck.cc
Exter-dg: I have removed the comments (except those describing the file - blank, corrupt, ODF, ApacheOffice)

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Thank you for pointing this out. I have added meta.xml (No, there was no reason for not adding it, I missed it)

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,121 @@
...
+#include "handler.h"
+
+#include "metaxmlparse.h"
+#include "opendocparse.h"
+#include "str.h"
+
+#include <archive.h>
+#include <archive_entry.h>
+
+#include <cstring>
```
ojwb, line 31: Are you actually using anything from `<cstring>`?

I'm wondering if this was added for `strcmp()` in a previous iteration of the patch, but you aren't using that now.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,121 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+#include <config.h>
+#include "handler.h"
+
+#include "metaxmlparse.h"
+#include "opendocparse.h"
+#include "str.h"
```
ojwb, line 26: `"str.h"` only defines an overloaded `str()` function, but that function doesn't seem to be used in this file.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,121 @@
...
+	string& keywords,
+	string& author,
+	string& pages,
+	string& error)
+{
+    try {
+	const char* file = filename.c_str();
+	struct archive* archive_obj = archive_read_new();
+	archive_read_support_filter_all(archive_obj);
+	archive_read_support_format_all(archive_obj);
```
ojwb, line 48: This filetype is meant to be a zip container so I think we don't want to allow any filters (which seems to be the default if you don't do anything), and only want the zip formats (i.e. `archive_read_support_format_zip(archive_obj)`, which enables both the streamable and seekable zip format handlers).

For example, if somebody renames a tar.bz2 archive to `.odt` then it's not helpful for us to open it and read its contents - nothing else which supports `.odt` is likely to.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,121 @@
...
+	int status_code = archive_read_open_filename(archive_obj,
+						    file, 10240);
```
ojwb, line 50: Why 10KB blocks?

ojwb, line 50: Also, the wrapped line is under-indented by one column (the `f` of `file` should be under the `a` of `archive_obj`).

This line doesn't actually need wrapping - if I unwrap it, it only reaches column 79...

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -0,0 +1,121 @@
...
+		    metaxmlparser.parse(s_meta);
+		    title = metaxmlparser.title;
+		    keywords = metaxmlparser.keywords;
+		    author = metaxmlparser.author;
+		}
+	    }
+	}
+	status_code = archive_read_free(archive_obj);
+	if (status_code != ARCHIVE_OK) {
+	    error += archive_error_string(archive_obj);
```
ojwb, line 108: Did you mean to assign a string to `error` before appending the error message from libarchive to it?

xapian-applications/omega/omindexcheck.cc
```diff
@@ -80,6 +80,44 @@ index_test()
...
+    // corrupted file (ODP)
+    // tests.insert({"corrupt_file.odp", {"ZSnatur"}});
+
+    // ODF
+    tests.insert({"odf/test.odt", {"Zедой", "Z喬伊不分享食物"}});
+    tests.insert({"odf/libarchive_text.odt", {"Stesttitle", "Atestauthor",
+		  "Zsampl", "Zhead", "Ztext", "Zhello",
+		  "Zworld"}});
+    tests.insert({"odf/libarchive_text_template.ott", {"Zjane", "Zdoe",
+		  "Zstructur"}});
```
ojwb, line 95: This free-form wrapping makes things harder to read.  I'd wrap between the items instead (and then wrap the list as well but only if it's still too long), e.g.:

```
    tests.insert({"odf/libarchive_text_template.ott",
                  {"Zjane", "Zdoe", "Zstructur"}});
    tests.insert({"odf/libarchive_presentation.odp",
                  {"Zfascin", "Zfact", "Zpustak", "Zmahal", "Zmillion", "Zpeopl",
                   "Zbirthday", "501"}});
```

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Thanks for pointing this. These are redundant now and I have removed them.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: It was used in earlier versions but it is redundant now. I have removed this.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have changed this.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have added a comment for this. As per discussions(on IRC), Libarchive automatically detects the required block size for files and for regular file size it usually takes a default one - https://sources.debian.org/src/libarchive/3.4.3-1/libarchive/archive_read_open_filename.c/#L353.

The specified block size will only be used in case of tape drives and 10240 is chosen as it is default size used in Libarchive's documentation (size of 20 records of 512 bytes each)

Exter-dg: Thanks for pointing this out, I have changed this.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I believe I was appending to a "not-empty" error string. But this has probably changed over the commits. Thanks for pointing this out, I have changed this. 

`error = archive_error_string(archive_obj);`

xapian-applications/omega/omindexcheck.cc
Exter-dg: I have changed this.

ojwb:
Thanks for the updates.  Squashed and merged via git.xapian.org.

I adjusted your commit message before pushing - the summary line was a bit over-long, and the longer description wasn't wrapped.

PR #302, opened 2020-06-27
Enum parameters for Tf-idf weight

dipanshu124:
Now tfidf weight will also support 3 separate normalisation parameters besides the normalisation string .
This will help to implement normalizations that require more than 1 character to specify normalization.

### DIFF
--- a/xapian-bindings/ruby/ruby.i
+++ b/xapian-bindings/ruby/ruby.i
@@ -112,9 +112,9 @@
 %rename("weight") get_weight;
 
 // Name of constants should start wth upper_case
-%rename("Wdfn_type") wdfn_type;
-%rename("Idfn_type") idfn_type;
-%rename("Wtn_type") wtn_type;
+%rename("Wdf_norm") wdf_norm;
+%rename("Idf_norm") idf_norm;
+%rename("Wt_norm") wt_norm;
 
 // These are 'dangerous' methods; i.e. they can cause a segfault if used
 // improperly.  We prefix with _dangerous_ so that Ruby users will not use them
--- a/xapian-bindings/ruby/ruby.i
+++ b/xapian-bindings/ruby/ruby.i
@@ -111,6 +111,11 @@
 %rename("wdf") get_wdf;
 %rename("weight") get_weight;
 
+// Name of constants should start wth upper_case
+%rename("Wdfn_type") wdfn_type;
+%rename("Idfn_type") idfn_type;
+%rename("Wtn_type") wtn_type;
+
 // These are 'dangerous' methods; i.e. they can cause a segfault if used
 // improperly.  We prefix with _dangerous_ so that Ruby users will not use them
 // inadvertently.
--- a/xapian-bindings/ruby/ruby.i
+++ b/xapian-bindings/ruby/ruby.i
@@ -111,7 +111,7 @@
 %rename("wdf") get_wdf;
 %rename("weight") get_weight;
 
-// Name of constants should start wth upper_case
+// Name of constants should start with upper case
 %rename("Wdf_norm") wdf_norm;
 %rename("Idf_norm") idf_norm;
 %rename("Wt_norm") wt_norm;
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -613,8 +613,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     /** Construct a TfIdfWeight using the default normalizations ("ntn"). */
     TfIdfWeight()
-	: normalizations("ntn"), wdf_norm(WDF_NORM::NONE), idf_norm(IDF_NORM::TFIDF),
-	  wt_norm(WT_NORM::NONE), param_slope(0.2), param_delta(1.0)
+	: normalizations("ntn"), wdf_norm(WDF_NORM::NONE),
+	  idf_norm(IDF_NORM::TFIDF), wt_norm(WT_NORM::NONE),
+	  param_slope(0.2), param_delta(1.0)
     {
 	need_stat(TERMFREQ);
 	need_stat(WDF);
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -467,51 +467,44 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
        during serialise(), the value may get truncated by static_cast<> and
        the desired result will not be achieved. */
   public:
-    /* The WDF_NORM works like so:
-     *
-     *    NONE: None.		wdfn=wdf
-     *    BOOLEAN: Boolean	wdfn=1 if term in document else wdfn=0
-     *    SQUARE: Square	wdfn=wdf*wdf
-     *    LOG: Logarithmic	wdfn=1+log<sub>e</sub>(wdf)
-     *    PIVOTED: Pivoted	wdfn=(1+log(1+log(wdf)))*
-     *				(1/(1-slope+(slope*doclen/avg_len)))+delta
-     *    LOG_AVERAGE: Log average wdfn=(1+log(wdf))/
-     *				   (1+log(doclen/unique_terms))
-     */
+    /** Wdf normalizations. */
     enum class WDF_NORM: int {
+    /*  NONE: None.  wdfn=wdf */
 	NONE = 1,
+    /*  BOOLEAN: Boolean  wdfn=1 if term in document else wdfn=0 */
 	BOOLEAN = 2,
+    /*  SQUARE: Square  wdfn=wdf*wdf */
 	SQUARE = 3,
+    /*  LOG: Logarithmic  wdfn=1+log<sub>e</sub>(wdf) */
 	LOG = 4,
+    /*  PIVOTED: Pivoted  wdfn=(1+log(1+log(wdf)))*
+     *			       (1/(1-slope+(slope*doclen/avg_len)))+delta */
 	PIVOTED = 5,
+    /*  LOG_AVERAGE: Log average  wdfn=(1+log(wdf))/(1+log(doclen/unique_terms)) */
 	LOG_AVERAGE = 6
     };
 
-    /* The IDF_NORM works like so:
-     *
-     *    NONE: None		idfn=1
-     *    TFIDF: TfIdf		idfn=log(N/Termfreq) where N is the number of
-     *    documents in collection and Termfreq is the number of documents
-     *    which are indexed by the term t.
-     *    PROB: Prob		idfn=log((N-Termfreq)/Termfreq)
-     *    FREQ: Freq		idfn=1/Termfreq
-     *    SQUARE: Squared	idfn=log(N/Termfreq)^2
-     *    PIVOTED: Pivoted	idfn=log((N+1)/Termfreq)
-     */
+    /** Idf normalizations. */
     enum class IDF_NORM: int {
+    /*  NONE: None  idfn=1 */
 	NONE = 1,
+    /*  TFIDF: TfIdf  idfn=log(N/Termfreq) where N is the number of documents
+     *  in collection and Termfreq is the number of documents which are
+     *  indexed by the term t. */
 	TFIDF = 2,
+    /*  SQUARE: Squared	 idfn=log(N/Termfreq)^2 */
 	SQUARE = 3,
+    /*  FREQ: Freq  idfn=1/Termfreq */
 	FREQ = 4,
+    /*  PROB: Prob  idfn=log((N-Termfreq)/Termfreq) */
 	PROB = 5,
+    /*  PIVOTED: Pivoted  idfn=log((N+1)/Termfreq) */
 	PIVOTED = 6
     };
 
-    /* The WT_NORM works like so:
-     *
-     *    NONE: None	wtn=tfn*idfn
-     */
+    /** Weight normalizations. */
     enum class WT_NORM: int {
+    /*  NONE: None  wtn=tfn*idfn */
 	NONE = 1
     };
   private:
@@ -546,6 +539,10 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
   public:
     /** Construct a TfIdfWeight
+     *
+     *  @param normalizations	A three character string indicating the
+     *				normalizations to be used for the tf(wdf), idf
+     *				and document weight.  (default: "ntn")
      *
      * The @a normalizations string works like so:
      *
@@ -587,6 +584,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     /** Construct a TfIdfWeight
      *
+     *  @param normalizations	A three character string indicating the
+     *				normalizations to be used for the tf(wdf), idf
+     *				and document weight.  (default: "ntn")
      *	@param slope		Extra parameter for "Pivoted" tf normalization.  (default: 0.2)
      *	@param delta		Extra parameter for "Pivoted" tf normalization.  (default: 1.0)
      *
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -577,6 +577,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     /* When additional normalizations are implemented in the future, the additional statistics for them
        should be accessed by these functions. */
+    wdf_norm decode_wdf_norm(char c);
+    idf_norm decode_idf_norm(char c);
+    wt_norm decode_wt_norm(char c);
     double get_wdfn(Xapian::termcount wdf,
 		    Xapian::termcount len,
 		    Xapian::termcount uniqterms,
@@ -627,7 +630,8 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
      */
-    explicit TfIdfWeight(const std::string &normalizations);
+    explicit TfIdfWeight(const std::string &normalizations)
+	: TfIdfWeight(normalizations, 0.2, 1.0) {}
 
     /** Construct a TfIdfWeight
      *
@@ -683,7 +687,11 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
      * Implementing support for more normalizations of each type would require
      * extending the backend to track more statistics.
      */
-    TfIdfWeight(wdf_norm wdf_norm_, idf_norm idf_norm_, wt_norm wt_norm_);
+    TfIdfWeight(wdf_norm wdf_normalization,
+		idf_norm idf_normalization,
+		wt_norm wt_normalization)
+	: TfIdfWeight(wdf_normalization, idf_normalization,
+		      wt_normalization, 0.2, 1.0) {}
 
     /** Construct a TfIdfWeight
      *
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -480,7 +480,8 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     /*  PIVOTED: Pivoted  wdfn=(1+log(1+log(wdf)))*
      *			       (1/(1-slope+(slope*doclen/avg_len)))+delta */
 	PIVOTED = 5,
-    /*  LOG_AVERAGE: Log average  wdfn=(1+log(wdf))/(1+log(doclen/unique_terms)) */
+    /*  LOG_AVERAGE: Log average  wdfn=(1+log(wdf))/
+     *				       (1+log(doclen/unique_terms)) */
 	LOG_AVERAGE = 6
     };
 
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -463,12 +463,8 @@ class XAPIAN_VISIBILITY_DEFAULT BoolWeight : public Weight {
 
 /// Xapian::Weight subclass implementing the tf-idf weighting scheme.
 class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
-    /* Three character string indicating the normalizations for tf(wdf), idf and
-       tfidf weight. */
-    std::string normalizations;
-
-    /* These three enum class should use integers only upto 256 . Otherwise
-       during serialise(), the value may get truncated by static_cast() and
+    /* These three enum classes should use integers only up to 255. Otherwise
+       during serialise(), the value may get truncated by static_cast<> and
        the desired result will not be achieved. */
   public:
     enum class WDF_NORM: int {
@@ -616,9 +612,8 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
 
     /** Construct a TfIdfWeight using the default normalizations ("ntn"). */
     TfIdfWeight()
-	: normalizations("ntn"), wdf_norm(WDF_NORM::NONE),
-	  idf_norm(IDF_NORM::TFIDF), wt_norm(WT_NORM::NONE),
-	  param_slope(0.2), param_delta(1.0)
+	: wdf_norm(WDF_NORM::NONE), idf_norm(IDF_NORM::TFIDF),
+	  wt_norm(WT_NORM::NONE), param_slope(0.2), param_delta(1.0)
     {
 	need_stat(TERMFREQ);
 	need_stat(WDF);
--- a/xapian-core/include/xapian/weight.h
+++ b/xapian-core/include/xapian/weight.h
@@ -468,7 +468,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
        the desired result will not be achieved. */
   public:
     /** Wdf normalizations. */
-    enum class WDF_NORM : int {
+    enum class wdfn_type : int {
 	/** None
 	 *
 	 *  wdfn=wdf
@@ -509,7 +509,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     };
 
     /** Idf normalizations. */
-    enum class IDF_NORM : int {
+    enum class idfn_type : int {
 	/** None
 	 *
 	 *  idfn=1
@@ -550,7 +550,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     };
 
     /** Weight normalizations. */
-    enum class WT_NORM : int {
+    enum class wtn_type : int {
 	/** None
 	 *
 	 *  wtn=tfn*idfn
@@ -559,11 +559,11 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
     };
   private:
     /// The parameter for normalization for the wdf.
-    WDF_NORM wdf_norm;
+    wdfn_type wdf_norm;
     /// The parameter for normalization for the idf.
-    IDF_NORM idf_norm;
+    idfn_type idf_norm;
     /// The parameter for normalization for the document weight.
-    WT_NORM wt_norm;
+    wtn_type wt_norm;
 
     /// The factor to multiply with the weight.
     double wqf_factor;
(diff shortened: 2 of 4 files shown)


xapian-core/weight/tfidfweight.cc
```diff
@@ -89,12 +128,131 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
 	need_stat(DOC_LENGTH_MAX);
 	need_stat(UNIQUE_TERMS);
     }
+    switch (normalizations[0]) {
```
jaylett, line 131: You can pull this into a private initialiser method to avoid the duplication.

xapian-core/weight/tfidfweight.cc
```diff
@@ -211,42 +372,44 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
...
 }
 
 double
-TfIdfWeight::get_wtn(double wt, char c) const
+TfIdfWeight::get_wtn(double wt, WT_NORM wt_norm_) const
 {
-    (void)c;
-    AssertEq(c, 'n');
+    (void)wt_norm_;
+    AssertEq(wt_norm_, WT_NORM::NONE);
```
jaylett, line 412: Since this is now an enum, you don't need to assert anything — the compiler should pick up any attempt to violate this.

xapian-core/weight/tfidfweight.cc
```diff
@@ -211,42 +372,44 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
...
 	    return (1.0 / termfreq);
-	case 's':
+	case IDF_NORM::SQUARE:
 	    return pow(log(N / termfreq), 2.0);
-	case 'P':
+	case IDF_NORM::PIVOTED:
 	    return log((N + 1) / termfreq);
 	default:
-	    AssertEq(c, 't');
+	    AssertEq(idf_norm_, IDF_NORM::TFIDF);
```
jaylett, line 403: Since this is an enum, you can change the `default` into a specific case (nothing else will appear). Some compilers will warn if you don't provide cases for every possibility of an enum. Then you won't have to assert, either.

xapian-core/weight/tfidfweight.cc
```diff
@@ -211,42 +372,44 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
 	    return num / den;
 	}
 	default:
-	    AssertEq(c, 'n');
+	    AssertEq(wdf_norm_, WDF_NORM::NONE);
```
jaylett, line 375: Same comment about `default`.

xapian-core/weight/tfidfweight.cc
```diff
@@ -138,19 +298,20 @@ TfIdfWeight::unserialise(const string & s) const
     const char *end = ptr + s.size();
     double slope = unserialise_double(&ptr, end);
     double delta = unserialise_double(&ptr, end);
-    string normals(ptr, end);
-    ptr += 3;
+    WDF_NORM wdf_norm_ = static_cast<WDF_NORM>(*(ptr)++);
+    IDF_NORM idf_norm_ = static_cast<IDF_NORM>(*(ptr)++);
+    WT_NORM wt_norm_ = static_cast<WT_NORM>(*(ptr)++);
```
jaylett, line 303: I think this at least warrants a comment somewhere that it's serialising as single a byte, so that in the (unlikely) event we end up with > 256 there's documentation somewhere of what will go wrong. (What would happen? Will static_cast<> fail to compile, or will it truncate or something?)

xapian-core/weight/tfidfweight.cc
```diff
@@ -89,12 +128,131 @@ TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
...
+TfIdfWeight::TfIdfWeight(WDF_NORM wdf_norm_,
+			 IDF_NORM idf_norm_,
+			 WT_NORM wt_norm_)
+    : wdf_norm(wdf_norm_), idf_norm(idf_norm_), wt_norm(wt_norm_),
+      param_slope(0.2), param_delta(1.0)
+{
+    if (!((wdf_norm == WDF_NORM::NONE) || (wdf_norm == WDF_NORM::BOOLEAN) ||
+	(wdf_norm == WDF_NORM::LOG_AVERAGE) || (wdf_norm == WDF_NORM::LOG) ||
+	(wdf_norm == WDF_NORM::PIVOTED) || (wdf_norm == WDF_NORM::SQUARE)))
+	throw Xapian::InvalidArgumentError("WDF_NORM is invalid");
```
jaylett, line 181: Can the compiler allow through something that would trigger this? It feels like this is just going to increase maintenance effort without adding anything useful.

xapian-core/include/xapian/weight.h
```diff
@@ -467,6 +467,33 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
        tfidf weight. */
     std::string normalizations;
 
+  public:
+    enum class WDF_NORM: int {
+	NONE = 1,
```
jaylett, line 472: Is there a reason not to start at `0`?

xapian-core/include/xapian/weight.h
```diff
@@ -467,6 +467,33 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
        tfidf weight. */
     std::string normalizations;
```
jaylett, line 468: Is this still used anywhere?

dipanshu124: Yes ,In switch I added.

xapian-core/include/xapian/weight.h
dipanshu124: I wasn't sure that 0 will be fine during serialise(). ( \0 may cause problem.)
Also I looked at other enums defined in weight class .They were also beginning form 1 .
So I chose 1 instead of 0.

xapian-core/weight/tfidfweight.cc
dipanshu124: Yes , This is avoidable. Thanks for pointing it out .

xapian-core/weight/tfidfweight.cc
dipanshu124: I have now added comments where I declared these enum classes.

xapian-core/weight/tfidfweight.cc
dipanshu124: When I removed default and added a case, I got this error -
error: control reaches end of non-void function [-Werror=return-type]

xapian-core/weight/tfidfweight.cc
dipanshu124: Updated

xapian-core/weight/tfidfweight.cc
dipanshu124: I have used delegating constructors . This would help avoid other duplication (apart from switch) as well.

xapian-core/include/xapian/weight.h
```diff
@@ -467,6 +467,9 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
...
 
+    /* These three enum class should use integers only upto 256 . Otherwise
+       during serialise(), the value may get truncated by static_cast() and
+       the desired result will not be achieved. */
```
jaylett, line 472: Surely 255? Sorry, I think I said 256 but that was when I was thinking we'd renumber to start at 0. At the number 256 it will overflow 1 byte.

More minor: "three enum classes" rather than "class", "up to" rather than "upto", and don't leave a space before the full stop. Also `static_cast<>` instead of `static_cast()`.

xapian-core/include/xapian/weight.h
jaylett: I don't think you need to though — since that switch is in a constructor, you could just use the formal parameter (`normals`) directly.

xapian-core/include/xapian/weight.h
dipanshu124: Updated

xapian-core/include/xapian/weight.h
dipanshu124: Updated to use normals

jaylett:
@dipanshu124 your API docstrings still aren't right in all cases. I strongly suggest you build the docs and look at them, and use that as a guide for where you're off.

xapian-core/include/xapian/weight.h
```diff
@@ -463,9 +463,58 @@ class XAPIAN_VISIBILITY_DEFAULT BoolWeight : public Weight {
...
 class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
-    /* Three character string indicating the normalizations for tf(wdf), idf and
-       tfidf weight. */
-    std::string normalizations;
+    /* These three enum classes should use integers only up to 255. Otherwise
+       during serialise(), the value may get truncated by static_cast<> and
+       the desired result will not be achieved. */
+  public:
+    /** Wdf normalizations. */
+    enum class WDF_NORM: int {
```
ojwb, line 471: While constants such as the values of an `enum` are conventionally all-caps, the `enum` type itself usually isn't.  We don't have any other `enum class` in the API so far but we do have `typedef`-ed `enum` - e.g.

```
    typedef enum {
	STEM_NONE, STEM_SOME, STEM_ALL, STEM_ALL_Z, STEM_SOME_FULL_POS
    } stem_strategy;
```

So I think this type should be `wdf_norm` or similar for consistency.

(We only don't use `enum class` yet because it was new in C++11, and for 1.4.x we require a C++11 compiler to build Xapian, but aim to support use of a non-C++11 compiler for building applications using Xapian so we avoid C++11 features in the API headers - for git master we now require a C++11 compiler for building applications too, so it's OK to start using it.)

There should probably be a space before the `:` too (there isn't a precedent since we don't have an `enum class` yet, but it's `class Foo : public Base` and this seems equivalent to that situation.)

xapian-core/weight/tfidfweight.cc
ojwb: You could make the final `case` in the `switch` just do `break;` and then have its `return` at the end of the function - then you won't get the "control reaches end of non-void function".

As James says, the major advantage of having all the enum values listed as explicit cases with no `default:` is that compilers will warn if a new enum value is added without new cases being added for it (which is an easy mistake for a future developer working on the code to make).  If there's a `default:` then there's no such warning (since the new value is handled by the `default:`).

xapian-core/include/xapian/weight.h
dipanshu124: Updated

xapian-core/include/xapian/weight.h
```diff
@@ -468,7 +468,7 @@ class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
        the desired result will not be achieved. */
   public:
     /** Wdf normalizations. */
-    enum class WDF_NORM : int {
+    enum class wdfn_type : int {
```
ojwb, line 471: Not sure I like `wdfn_type`, etc - we don't have any other type names that end `_type` in the API currently (except for sub-types like `Xapian::TermIterator::value_type` which are required for compatibility with the C++ STL).

ojwb: Also, `wdfn_type::BOOLEAN`, etc seem less clear than `wdf_norm::BOOLEAN`.

xapian-core/include/xapian/weight.h
```diff
@@ -463,9 +463,107 @@ class XAPIAN_VISIBILITY_DEFAULT BoolWeight : public Weight {
 
 /// Xapian::Weight subclass implementing the tf-idf weighting scheme.
 class XAPIAN_VISIBILITY_DEFAULT TfIdfWeight : public Weight {
-    /* Three character string indicating the normalizations for tf(wdf), idf and
-       tfidf weight. */
-    std::string normalizations;
+    /* These three enum classes should use integers only up to 255. Otherwise
+       during serialise(), the value may get truncated by static_cast<> and
+       the desired result will not be achieved. */
```
ojwb, line 468: Rather than just documenting this in the comment (which could be missed), I think it makes more sense to instead specify the underlying type of the `enum` as `unsigned char`.  Then the compiler knows to complain if an out of range value is specified - e.g. GCC gives the error:

```
error: enumerator value ‘256’ is outside the range of underlying type ‘unsigned char’
```

Better to let the compiler check such requirements for us when we can get it to.

xapian-core/weight/tfidfweight.cc
```diff
@@ -211,42 +268,41 @@ TfIdfWeight::get_wdfn(Xapian::termcount wdf, Xapian::termcount doclen,
...
 {
     Xapian::doccount termfreq = 1;
-    if (c != 'n') termfreq = get_termfreq();
+    if (idf_norm_ != idfn_type::NONE) termfreq = get_termfreq();
     double N = 1.0;
-    if (c != 'n' && c != 'f') N = get_collection_size();
-    switch (c) {
-	case 'n':
+    if ((idf_norm_ == idfn_type::PROB) || (idf_norm_ == idfn_type::SQUARE) ||
+	(idf_norm_ == idfn_type::TFIDF) || (idf_norm_ == idfn_type::PIVOTED))
```
ojwb, line 282: There's no need to put extra parentheses around each equality test, and the code seems harder to read that way.

Why did you change the sense of the tests here?  Previously we checked the norm wasn't one of two, but now we check it is one of the other four - twice as many comparisons!

xapian-core/include/xapian/weight.h
dipanshu124: I have now changed the class name to wdf_norm, etc. The parameter names are now wdf_norm_, etc.

xapian-core/include/xapian/weight.h
dipanshu124: Yes, This is better. Updated.

xapian-core/weight/tfidfweight.cc
dipanshu124: This change was supposed to be after I add more normalization. For now I have restored the previous version.
Thanks for pointing it out.

xapian-bindings/ruby/ruby.i
```diff
@@ -111,6 +111,11 @@
 %rename("wdf") get_wdf;
 %rename("weight") get_weight;
 
+// Name of constants should start wth upper_case
```
ojwb, line 114: Comment typo: "wth"

Also, "upper_case" is a bit weird here - "upper case" with a space would be better.

xapian-core/weight/tfidfweight.cc
```diff
@@ -36,54 +36,110 @@ using namespace std;
 
 namespace Xapian {
 
-TfIdfWeight::TfIdfWeight(const std::string &normals)
-    : normalizations(normals), param_slope(0.2), param_delta(1.0)
+TfIdfWeight::TfIdfWeight(const std::string& normals)
+    : TfIdfWeight::TfIdfWeight(normals, 0.2, 1.0) {}
```
ojwb, line 40: It would be better to just inline this forwarding constructor from the header.

The extra code that adds to the application is very small (I'd expect it's just loading two floating point constants) and it saves having an extra symbol exported by the shared library just for the forwarding version (such symbols need to be resolved at library load time, and that slows start up so it's good to avoid adding symbols for no good reason).

xapian-core/weight/tfidfweight.cc
```diff
@@ -36,54 +36,110 @@ using namespace std;
...
+    wt_norm_ = wt_norm::NONE;
 }
 
-TfIdfWeight::TfIdfWeight(const std::string &normals, double slope, double delta)
-    : normalizations(normals), param_slope(slope), param_delta(delta)
+TfIdfWeight::TfIdfWeight(wdf_norm wdf_normalization,
+			 idf_norm idf_normalization,
+			 wt_norm wt_normalization)
+    : TfIdfWeight::TfIdfWeight(wdf_normalization, idf_normalization,
+			       wt_normalization, 0.2, 1.0) {}
```
ojwb, line 117: Same arguments for inlining apply here too.

xapian-core/weight/tfidfweight.cc
```diff
@@ -36,54 +36,110 @@ using namespace std;
...
+
+TfIdfWeight::TfIdfWeight(wdf_norm wdf_normalization,
+			 idf_norm idf_normalization,
+			 wt_norm wt_normalization,
+			 double slope, double delta)
+    : wdf_norm_(wdf_normalization), idf_norm_(idf_normalization),
+      wt_norm_(wt_normalization), param_slope(slope), param_delta(delta)
 {
```
ojwb, line 125: This and the constructor which takes a 3 character code for compatibility share all the logic to select which stats are needed, and to vet the other parameters.

It'd be much cleaner for the compatibility one to forward to this, then we only need to have one copy of that code.

You just need to arrange that the decoding code can be done in the forwarding constructor call, e.g. with a function for each one, something like:

```
TfIdfWeight::TfIdfWeight(const string& normals, double slope, double delta)
    : TfIdfWeight(decode_wdf_norm(normals).
                  decode_idf_norm(normals),
                  decode_wt_norm(normals),
                  slope,
                  delta) {}
```

xapian-core/weight/tfidfweight.cc
```diff
@@ -138,19 +197,21 @@ TfIdfWeight::unserialise(const string & s) const
     const char *end = ptr + s.size();
     double slope = unserialise_double(&ptr, end);
     double delta = unserialise_double(&ptr, end);
-    string normals(ptr, end);
-    ptr += 3;
+    wdf_norm wdf_normalization = static_cast<wdf_norm>(*(ptr)++);
+    idf_norm idf_normalization = static_cast<idf_norm>(*(ptr)++);
+    wt_norm wt_normalization = static_cast<wt_norm>(*(ptr)++);
```
ojwb, line 202: This doesn't check that there's actually enough data in the string supplied (the existing code isn't quite right either).  (The first byte read is OK as there's guaranteed to be a zero byte on the end (since C++11), but the others are off the end of the string, so undefined behaviour).

These serialisations are passed across the network, so we need to take extra care not to assume they are valid.

Instead of throwing an exception if `ptr != end` after we process these bytes, we can instead throw if `end - ptr != 3` before we process them - that way if the string is too short we throw an exception instead of invoking undefined behaviour.

dipanshu124: Updated

xapian-core/weight/tfidfweight.cc
dipanshu124: Updated

xapian-core/weight/tfidfweight.cc
dipanshu124: Updated

xapian-bindings/ruby/ruby.i
dipanshu124: Fixed

xapian-core/weight/tfidfweight.cc
```diff
@@ -36,85 +36,68 @@ using namespace std;
...
+	    throw Xapian::InvalidArgumentError
+		  ("Normalization string is invalid");
+    }
+}
+
+TfIdfWeight::TfIdfWeight(const std::string& normals,
+			 double slope, double delta)
+    : TfIdfWeight::TfIdfWeight(decode_wdf_norm(normals[0]),
+			       decode_idf_norm(normals[1]),
+			       decode_wt_norm(normals[2]),
```
ojwb, line 99: There's no longer a check on the length of `normals` before we access its characters - pass in an empty string and this will read off the end of it.  That's why in my sketch of the code I passed `normals` to these helpers - then they can check `normals.size() == 3`.  We probably need to repeat that check in each helper since the order in which parameters are evaluated isn't specified in C++, but these functions are only called once so the compiler is likely to decide to inline them all, and then it can see that the repeat length checks are redundant and remove them.  And if it fails to, then the length check is cheap (and this is compatibility code anyway).

xapian-core/weight/tfidfweight.cc
```diff
@@ -197,11 +180,12 @@ TfIdfWeight::unserialise(const string & s) const
     const char *end = ptr + s.size();
     double slope = unserialise_double(&ptr, end);
     double delta = unserialise_double(&ptr, end);
+    if (rare(end - ptr != 3))
+	throw Xapian::SerialisationError
+	      ("Extra data in TfIdfWeight::unserialise()");
```
ojwb, line 185: OK, except that it might be more or less data now - "Incorrect data ..." maybe?

xapian-core/weight/tfidfweight.cc
```diff
@@ -36,85 +36,68 @@ using namespace std;
...
+	    return idf_norm::PROB;
+	case 't':
+	    return idf_norm::TFIDF;
 	default:
-	    idf_norm_ = idf_norm::TFIDF;
+	    throw Xapian::InvalidArgumentError
+		  ("Normalization string is invalid");
     }
-    wt_norm_ = wt_norm::NONE;
 }
```
ojwb, line 81: Since all the other cases `return`, you could handle the default case outside the `switch` which would avoid the wrapping.

ojwb:
This looks good to me aside from the few comments I've just made.

I didn't look at the formatting of the doxygen output which James flagged though - are you happy with that now, @jaylett?

xapian-core/weight/tfidfweight.cc
dipanshu124: Done.

xapian-core/weight/tfidfweight.cc
dipanshu124: Updated

xapian-core/weight/tfidfweight.cc
dipanshu124: ohh, I missed that. I have now updated it.

jaylett:
@ojwb yes, everything that should be a docstring now seems to be!

ojwb:
"Fixup!" isn't recognised by `git rebase`'s `--autosquash` option - you almost always want to use `git commit --fixup abcdef` where abcdef is the SHA of the commit you're wanting to fixup (or another way to reference it such as `HEAD`).

It'll take me longer to merge this because of that, so it'll have to wait until tomorrow.

ojwb:
Having look through the commits in more detail I think all I can sensibly do is to squash it down to a single commit.  That's not an unreasonable thing to do here, but for a more complex set of changes it wouldn't be great - you really need to be specify the fixup commits properly.

Your commit message is not properly formatted - see https://xapian-developer-guide.readthedocs.io/en/latest/contributing/workflow.html#good-commit-messages

The summary (the first line) is too long, and also ends in a full stop which it conventionally doesn't (think of it like the title of a book or document):

```
Add support for 3 normalization parameters of Tf-Idf weighting scheme.
```

The longer description should be line-wrapped (at 72 characters is common, but it definitely should be at < 76 characters since git shows commits indented by 4 columns in the log.  So your commit looks like this:

```
commit d23e53d72b5ddbe3771b557e90f65cdb2ffaeccf (HEAD)
Author: dipanshu124 <dipanshu@iitk.ac.in>
Date:   Sat Jun 27 13:55:10 2020 +0530

    Add support for 3 normalization parameters of Tf-Idf weighting scheme.
    
    Now tfidf weight will also support 3 separate normalisation parameters besid
es the normalisation string .
    This will help to implement normalizations that require more than 1 characte
r to specify normalization.
```

There are some other formatting/spelling issues:

* There's an odd space before the full stop at the end of the first sentence.

* You use two different spellings of "normalisation".  We've tended to use British English spellings in Xapian, where it's "normalisation".  I don't feel strongly enough to insist we avoid the US English "normalization", but I think we should avoid mixing different spellings of a word within a paragraph as that's liable to distract the reader.

The content could be better too:

* "besides" suggests that the new parameters are in addition to the string, whereas they're really an alternative - e.g. "instead of" would be clearer.

* The point of this change isn't that there are normalisations where we need to use more than one character - there aren't anything near 52 proposed normalisations yet.

  I'd say the points to make are that not all normalisations have an existing single letter code and allocating them ourselves is likely to result in a clash sooner or later, and that using enums allows for clear self-describing names, whereas single letter codes are inherently rather cryptic.

I think developers sometimes view the commit message as just an obstacle before they can push their code, and bash out something just to fill the space.  But the commit message is important - it's a way to communicate to future developers why you changed something (quite likely including your future self - it's pretty common to pull up and old change you made which you can remember nothing useful about and either curse your past self for not providing a useful commit message, or be glad you took the time to write something helpful.

ojwb:
The other thing I just noticed is that there aren't any testcases for this new API.  These could just be the existing testcases which specify a 3-character string but modified to pass the corresponding 3 `enum` values instead, but we really should have testcases exercising the new API as well as the old one.  Sorry for not spotting this earlier.

dipanshu124:
Apologies, I will be more careful while writing commit messages and while making fix up commits. I have now modified some of the existing tests to use enum parameters.

xapian-core/tests/api_weight.cc
```diff
@@ -882,6 +882,152 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
...
+	    Xapian::TfIdfWeight::wdf_norm::LOG,
+	    Xapian::TfIdfWeight::idf_norm::NONE,
+	    Xapian::TfIdfWeight::wt_norm::NONE));
+    mset = enquire.get_mset(0, 10);
+    TEST_EQUAL(mset.size(), 2);
+    mset_expect_order(mset, 2, 4);
+    TEST_EQUAL_DOUBLE(mset[0].get_weight(), 1 + log(8.0));
+    TEST_EQUAL_DOUBLE(mset[1].get_weight(), 1.0);
+
+    // Check for SQUARE, NONE , NONE.
```
ojwb, line 966: There's an unwanted space in front of the second `,` here.

xapian-core/tests/api_weight.cc
```diff
@@ -882,6 +882,152 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
...
+	    Xapian::TfIdfWeight::wt_norm::NONE)); // idf=1 and tfn=tf*tf
+    mset = enquire.get_mset(0, 10);
+    TEST_EQUAL(mset.size(), 5);
+    mset_expect_order(mset, 2, 1, 4, 3, 5);
+    TEST_EQUAL_DOUBLE(mset[0].get_weight(), 9.0);
+    TEST_EQUAL_DOUBLE(mset[4].get_weight(), 1.0);
+
+    // Check for NONE, TFIDF, NONE when termfreq=N
+    enquire.set_query(Xapian::Query("this"));
+    // N=termfreq and so idfn=0 for "t"
```
ojwb, line 981: `"t"` doesn't make sense here...

xapian-core/tests/api_weight.cc
```diff
@@ -882,6 +882,152 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
...
+    mset = enquire.get_mset(0, 10);
+    TEST_EQUAL(mset.size(), 6);
+    mset_expect_order(mset, 1, 2, 3, 4, 5, 6);
+    for (int i = 0; i < 6; ++i) {
+	TEST_EQUAL_DOUBLE(mset[i].get_weight(), 0.0);
+    }
+
+    // Check for NONE, PROB, NONE and for both branches of PROB
+    enquire.set_query(Xapian::Query("this"));
+    // N=termfreq and so idfn=0 for "p"
```
ojwb, line 996: Nor `"p"` here - when updating code you really need to also fully update the comments to match the updated code.

xapian-core/tests/api_weight.cc
```diff
@@ -882,6 +882,152 @@ DEFINE_TESTCASE(tfidfweight3, backend) {
...
+	    Xapian::TfIdfWeight::wt_norm::NONE));
+    mset = enquire.get_mset(0, 10);
+    TEST_EQUAL(mset.size(), 6);
+    mset_expect_order(mset, 1, 2, 3, 4, 5, 6);
+    for (int i = 0; i < 6; ++i) {
+	TEST_EQUAL_DOUBLE(mset[i].get_weight(), 0.0);
+    }
+
+    enquire.set_query(Xapian::Query("word"));
+    enquire.set_weighting_scheme(Xapian::TfIdfWeight("npn"));
```
ojwb, line 1010: You appear to be adding another copy of an existing test of `"npn"` here - presumably you failed to convert this one to the enums?

xapian-core/tests/api_weight.cc
dipanshu124: Updated.

xapian-core/tests/api_weight.cc
dipanshu124: Updated.

xapian-core/tests/api_weight.cc
dipanshu124: Yes, I missed this one. I have now fixed it.

PR #303, opened 2020-07-12
Omindex mimetype modifications

Exter-dg:
* Libarchive now supports OOXML formats (.docx, .xlsx, .pptx, etc)


xapian-applications/omega/handler_libarchive.cc
```diff
@@ -31,92 +34,262 @@
...
+	    size = archive_read_data(archive_obj, &s_styles[0], total);
+
+	    if (size <= 0) {
+		error = "Libarchive was not able to extract data from "
+			"styles.xml";
+		return false;
+	    }
+	    s_styles.resize(size);
+	} else if (pathname == "meta.xml") {
+		total = archive_entry_size(entry);
```
barufa, line 89: overindented

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -31,92 +34,262 @@
...
+		s_meta.resize(total);
+		size = archive_read_data(archive_obj, &s_meta[0], total);
+
+		if (size > 0) {
+		    // indexing file even if this fails
+		    parse_metadata(s_meta, title, keywords, author);
+		}
+	    s_meta.resize(size);
```
barufa, line 97: Here I think there is a problem.
The type of `size` is `ssize_t`. In short, `ssize_t` is similar to `size_t`(unsigned) but it can store -1 to notify an error.
It seems that if `archive_read_data` fails, size will hold -1 and an exception is going to be raised if you call resize with negative values.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -31,92 +34,262 @@
...
+		 string& dump,
+		 string& title,
+		 string& keywords,
+		 string& author,
+		 string& error)
+{
+    size_t total;
+    ssize_t size;
+    struct archive_entry* entry;
+    string s, s_meta;
```
barufa, line 119: Maybe `s_content` is a more meaningful name for `s`

xapian-applications/omega/handler_libarchive.cc
ojwb: Indeed - `resize(-1)` on a string will cast the -1 to the largest value of the unsigned `std::string::size_type`, which will (hopefully) be larger than you can actually resize a string to.

Also, to be useful this resize needs to happen *before* we pass `s_meta` to `parse_metadata()` (though you'd usually expect the number of bytes read to match the size reported for the entry in `total`, so we probably get away with this most of the time).

xapian-applications/omega/handler_libarchive.cc
ojwb: There's not really a good reason for all these `s_` prefixes.  Some coding styles (like hungarian notation which Microsoft uses a bastardised version of) try to encode the types of variables into their names, but we don't use that style in Xapian.

ojwb:
It would be simpler to review this if the mime type passing was one PR and the support for the MS formats a separate PR.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have resolved this.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Thanks for pointing this out. I have changed this. Now resize() is called before it is sent to parse_metadata().

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I will keep this in mind. For now, I have changed few variable names 
* s_content -> content
* s_meta -> metadata
* s -> content
* s_styles -> styles
Are these a better choice?

ojwb: Yes (at least assuming `s_content` and `s` aren't used in the same scope!)

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -31,8 +34,221 @@
...
+	dump = xmlparser.dump;
+    } else {
+	XlsxParser parser;
+	parser.parse(content);
+	dump = parser.dump;
+    }
+
+    // Parse if docProps/core.xml is found
+    if (metaxml)
+	parse_metadata(metadata, title, keywords, author);
```
ojwb, line 245: Calling `parse_metadata()` here sets fields `title`, `keywords`, and `author` which doesn't overlap with what's set by parsing the other file above, so there's no need to delay this parsing until here - we can just parse the data where we extract it from the zip archive above (there are several places, but you've put the code to do the parsing into a function so you can just call it in each case).

That reduces how long we need to hold that data in memory, reduces the lifetime of the `metadata` variable, and eliminates the need for the `metaxml` flag.  It'll also mean the code is easier to follow.

Exter-dg: Thanks for pointing this out. I have made the required changes and have updated the PR.

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -55,68 +268,30 @@ extract(const string& filename,
...
-	    }
+	bool succeed;
+	if (startswith(mimetype, "application/vnd.sun.xml.") ||
+	    startswith(mimetype, "application/vnd.oasis.opendocument.")) {
+	    succeed = extract_odf(archive_obj, dump, title, keywords, author,
+				  error);
+	    if (!succeed)
+		return false;
+	} else if (startswith(mimetype, "application/"
+					"vnd.openxmlformats-officedocument.")) {
```
ojwb, line 279: It would be better to wrap the line by aligning the function parameters vertically rather than splitting the string - that's more typical of how this sort of case is wrapped in existing code in the codebase, it seems easier to read that way, and it means things like `grep application/vnd` match here.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -130,6 +132,14 @@ index_test()
 		  {"Zhead", "Zhello", "Zopenoffic", "Zimpress"}});
     tests.insert({"sof/libarchive_openoffice_presentation_template.sti",
 		  {"ZSproject", "ZSresearch", "Zhead", "Ztext"}});
+
+    // OpenXML formats
```
ojwb, line 136: I think we should above calling these formats "OpenXML", and just say "OOXML" instead.  Microsoft subverted the ISO standardisation process (for some history see https://en.wikipedia.org/wiki/Office_Open_XML#Standardization_process).

It's not really an "Open" format - sure the specification is public, but the format was just submitted fully formed by Microsoft rather than developed in an open way.  It's really no more "Open" than the older OLE-stream Microsoft formats they replaced with it (which are also now publicly documented, but only after Microsoft lost an anti-trust case).

The format is sadly widely used so it's unhelpful to refuse to support it, but we should avoid being sucked into their PR games.

Exter-dg: Thank you for the insight. I have made the necessary changes.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have changed this. But in order to maintain 80 chars per line, I had to move the opening bracket of the else if statement to the next line (against the usual practice). Will that be ok?

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -56,68 +268,31 @@ extract(const string& filename,
...
-		    author = metaxmlparser.author;
-		}
-	    }
+	bool succeed;
+	if (startswith(mimetype, "application/vnd.sun.xml.") ||
+	    startswith(mimetype, "application/vnd.oasis.opendocument.")) {
+	    succeed = extract_odf(archive_obj, dump, title, keywords, author,
+				  error);
+	    if (!succeed)
+		return false;
```
ojwb, line 277: If you're going to put the returned value in a variable, better to declare the variable when you can assign to it - as this currently is the scope of the variable is greater than needed.

But there isn't really much point putting the result in a variable here - you can just test the result directly - e.g. for the above:

```
if (!extract_odf(...)) {
    return false;
}
```

A boolean return for "success" is pretty common and reads nicely inline - "if we didn't extract the odf then return false" - but if the returned thing is more obscure a variable can provide a way to document what it is.  Perhaps `extract_foo()` returns the number of errors encountered, in which case:

```
int error_count = extract_foo(...);
if (error_count > 0)
    return false;
```

is perhaps more helpful than:

```
if (extract_foo(...) > 0)
    return false;
```

And probably clearer than adding a comment such as:

```
// Check if extract_foo() encountered any errors.
if (extract_foo(...) > 0)
    return false;
```

xapian-applications/omega/handler_libarchive.cc
```diff
@@ -31,6 +34,215 @@
...
+
+		if (size <= 0) {
+		    error = "Libarchive was not able to extract data from " +
+			    pathname;
+		    return false;
+		}
+		sheets.resize(i + size);
+	    } else if (pathname == "docProps/core.xml") {
+		total = archive_entry_size(entry);
+		string metadata(total, ' ');
```
ojwb, line 193: The content we fill the string to have here shouldn't matter as we resize to remove any excess we didn't read into, but it'd be more consistent to use the same fill byte in cases like this as we get for `sheets.resize(i + total);` which would be a zero byte (i.e. `'\0'`).  Zero bytes should also make it more obvious to the person debugging if we did end up leaving some of this fill in the read data, whereas spaces will probably look like valid data.

xapian-applications/omega/handler_libarchive.cc
ojwb: Yes, I think that's OK.

When it comes down to it, the point of having such formatting rules as part of coding standards is mainly to aid readability and make it easier to work with the source code, and to help spot problems (if things are consistent then inconsistencies stand out, and bugs are often due to inconsistencies).

ojwb:
Thanks for the updates (and the rebase).  This look good to me now, aside from some minor nits noted above.

Are you happy @barufa?

xapian-applications/omega/handler_libarchive.cc
Exter-dg: Thank you for pointing this out. I have made the necessary changes.

xapian-applications/omega/handler_libarchive.cc
Exter-dg: I have changed this.

barufa:
> Thanks for the updates (and the rebase). This look good to me now, aside from some minor nits noted above.
> 
> Are you happy @barufa?

It seems that everything is okay..

PR #304, opened 2020-07-14
Omindex mimetype

Exter-dg:
Add mimetype as a parameter to bool extract().


xapian-applications/omega/omindexcheck.cc
```diff
@@ -131,6 +131,44 @@ index_test()
...
+    tests.insert({"sof/libarchive_openoffice_text_template.stw", {"Zhello",
+		  "Zworld", "Zsampl", "Zhead", "ZStemplat", "ZStext"}});
+    tests.insert({"sof/libarchive_openoffice_presentation.sxi", {"Zhead",
+		  "Zhello", "Zopenoffic", "Zimpress"}});
+    tests.insert({"sof/libarchive_openoffice_presentation_template.sti",
+		 {"ZSproject", "ZSresearch", "Zhead", "Ztext"}});
+#endif
 }
 
 static bool
```
Exter-dg, line 174: When I did git commit -v, this wasn't there. I even compared before creating a PR. I don't know how this end up here.

What I did was I performed a git reset --mixed on my omindex_mimetype-modifications branch to reset it to a commit that has been merged. And then did git add -p to only add changes which were required.

Exter-dg: > When I did git commit -v, this wasn't there. I even compared before creating a PR. I don't know how this end up here.
> 
> What I did was I performed a git reset --mixed on my omindex_mimetype-modifications branch to reset it to a commit that has been merged. And then did git add -p to only add changes which were required.

I have fixed this, It was added twice, I will find out how that happened. I have fixed it for this PR.

ojwb:
The code changes look good, but again your commit message could be better:

```
Add parameter mimetype to bool extract() in worker.h and handler.h

Mimetype is now passed to the handlers along with filename.
This can be used to decide the parser to be used inside the
handlers.
```

Mostly this is reasonable, but the summary line is longer than the recommended 50 characters.

This "50 characters" isn't a hard and fast limit and it's OK to push it a little if you need to in order to write a useful summary, but overlong summaries make the log output harder to skim (for example `git log --oneline --graph`) so it's better to keep to it when you can.

Here the source files changed probably aren't worth mentioning in the commit message at all since git can tell you what files a commit changed if you want to know (some projects have conventions that you should itemise changes by source file in the commit message, but they're really hangovers from pre-version-control days where those projects put that data in a ChangeLog text file by hand - we stopped doing that for Xapian when we switched to using git).  If you want to find commits which changed `worker.h` then `git log worker.h` will show you them with total accuracy.

The return type of `extract()` isn't important here either.  It's not like there's an `int extract()` to distinguish it from (you can't overload on the return type in C++, only on parameter types).

So "Add parameter mimetype to extract()" would be a better summary, though I'd probably focus on the high level overview for the summary rather than the mechanics of what you changed and essentially promote the first sentence of the longer description to be the overview - i.e. something like:

```
Pass mimetype to handlers along with the filename

This can be used to decide the parser to be used inside the
handlers.
```

There's definitely a skill to writing good commit messages, and it gets easier when you've had to spend time digging back through the history of changes as that gives insights in to what's helpful and not in commit messages.

PR #306, opened 2020-07-21
Omindex libextractor

Exter-dg:
Adds support for Audio and Video formats in Omega (Extracting meta-data).


ojwb:
> I noticed that if the global variables are not cleared/resized to 0, the values from previous file stays and is added to the new file too.
>
> Why does this happen?

Well, that's how global variables work - they have the lifetime of the program (in this case of the handler program).

> Is there a better alternative to global variables here?

Yes - the `cls` parameter to `EXTRACTOR_extract()` is provided to address this need.  You can make a class or structure that holds pointers or references to the `title`, etc parameters and pass a pointer to that structure in as `cls`.   Then cast that back to the right type in your callback function.  It's a pretty common idiom that a callback function takes a `void*` pointer so you can pass in a pointer to something in this way.

Note that you really want to pointers or references in this class/structure to avoid copying these strings needlessly (or else I guess you could use `std::move()` - the point is that these are potentially long strings so we want to avoid the overhead of unnecessary copying of them).

xapian-applications/omega/configure.ac
```diff
@@ -286,6 +286,11 @@ PKG_CHECK_MODULES([LIBARCHIVE], [libarchive], [
     AC_DEFINE(HAVE_LIBARCHIVE, 1, [Define HAVE_LIBARCHIVE if the library is available])
     OMINDEX_MODULES="$OMINDEX_MODULES omindex_libarchive"],[ ])
 
+dnl check if libextractor is available.
+PKG_CHECK_MODULES([LIBEXTRACTOR], [libextractor], [
+AC_DEFINE(HAVE_LIBEXTRACTOR, 1, [Define HAVE_LIBEXTRACTOR if the library is available])
+OMINDEX_MODULES="$OMINDEX_MODULES omindex_libextractor"],[ ])
```
ojwb, line 292: This isn't indented like the similar constructs just above.

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,120 @@
...
+	// Returns null and not error.
+	title = ::title;
+	author = ::author;
+	keywords = ::keywords;
+	pages = ::pages;
+
+	::keywords.resize(0);
+	::title.resize(0);
+	::author.resize(0);
+	::pages.resize(0);
```
barufa, line 113: A better option than resizing the string to zero would be calling the string method [clear](http://www.cplusplus.com/reference/string/string/clear/).

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,120 @@
...
+#include <config.h>
+#include "handler.h"
+
+#include <extractor.h>
+
+#include <cstring>
+
+using namespace std;
+
+string title, keywords, author, pages;
```
barufa, line 30: Not sure if using the same names for this variables is a good option as this practice can induce bugs. I think we should try to avoid it.

We can use an structure(as Olly've suggested) to store the necessary information and pass a pointer to it in `cls` instead of global variables.

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,120 @@
...
+	    // EXTRACTOR_METAFORMAT_BINARY
+	    // EXTRACTOR_METAFORMAT_C_STRING
+	    return 0;
+    }
+
+    switch (type) {
+	case EXTRACTOR_METATYPE_TITLE:
+	case EXTRACTOR_METATYPE_BOOK_TITLE:
+	case EXTRACTOR_METATYPE_SUBJECT:
+	    title = data;
```
barufa, line 65: I couldn't find if `data` is null terminated on libextractor docs. Maybe using `data_len` would be a good idea.
There is actually a [string constructor](http://www.cplusplus.com/reference/string/string/string/) that uses it(see constructor number 5).

xapian-applications/omega/configure.ac
Exter-dg: I have changed this.

xapian-applications/omega/handler_libextractor.cc
Exter-dg: Thanks for pointing this out. I was using clear(), I chose to use resize(0) because I read that clear (in case of vector) will make the vector empty but space will still be reserved, So I used resize instead of clear. I am not sure how does it work in case of strings. I will research more about this.
For the handler - I have used a struct as suggested by Olly.

xapian-applications/omega/handler_libextractor.cc
Exter-dg: I have changed this.

xapian-applications/omega/handler_libextractor.cc
Exter-dg: Thanks for pointing this out. I have changed to code and now it uses the string constructor 

`string (const char* s, size_t n);`

xapian-applications/omega/handler_libextractor.cc
Exter-dg: Should I avoid using the same variables inside the structure too? As I have done right now.

barufa: No, that's not necessary. Only when two variables have the same name under a scope. This is not an error, but it can introduce bugs in the code (for example, if you forget :: when you want to refer to a global variable) and makes the code less readable.

xapian-applications/omega/handler_libextractor.cc
ojwb: https://en.cppreference.com/w/cpp/string/basic_string/clear says:

> Unlike for std::vector::clear, the C++ standard does not explicitly require that capacity is unchanged by this function, but existing implementations do not change capacity. This means that they do not release the allocated memory (see also shrink_to_fit). 

(I'd definitely recommends cppreference.com as a resource - I think there's an offline version too.)

But if you're reusing the strings over and over you probably want the reserved space to be kept.  It's not so helpful if you process one enormous file but mostly small ones, but it's also not helpful to be constantly releasing and reallocating space.

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    switch (type) {
+	case EXTRACTOR_METATYPE_BOOK_TITLE:
+	case EXTRACTOR_METATYPE_JOURNAL_NAME:
+	case EXTRACTOR_METATYPE_ORIGINAL_TITLE:
+	case EXTRACTOR_METATYPE_SUBJECT:
+	case EXTRACTOR_METATYPE_TITLE:
+	    md->title += string(data, data_len);
+	    md->title += " ";
```
ojwb, line 78: This creates a temporary `std::string` object to append the data, which will result in needlessly copying the data, and also in a heap allocation if the string is at all long (short strings are typically actually just stored within the `std::string` object itself).  This temporary object can be avoided by using the `std::string::append()` method - that has an overload form which takes a pointer and length:

```
md->title.append(data, data_len);
```

Also, this always results in a trailing space on the result - it'd be neater to only add it if there's already something in the variable, which is easy to do by first checking and adding the space if non-empty, then appending the data.

Exter-dg: I have made the necessary changes here.

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -74,12 +74,13 @@ process_metadata(void* cls,
...
-	    md->title += string(data, data_len);
-	    md->title += " ";
+	    if (!md->title.empty())
+		md->title.append(" ");
+	    md->title.append(data, data_len);
 	    break;
 
 	case EXTRACTOR_METATYPE_PAGE_COUNT:
-	    md->pages = string(data, data_len);
+	    md->pages.append(data, data_len);
```
ojwb, line 83: If we get multiple `EXTRACTOR_METATYPE_PAGE_COUNT` items (which seems like it would happen if more than one plugin reports a number of pages for a format) then this will now concatenate them as strings - e.g. if we're told 12 pages twice, we'll turn that into `1212`!

The behaviour before of taking the last answer in case of duplicates seems better, so I think you want the "assign" method instead of "append" here.

xapian-applications/omega/mimemap.tokens
```diff
@@ -214,6 +214,23 @@ tif		image/tiff
 tiff		image/tiff
 webp		image/webp
 
+# Audio files:
+flac		audio/flac
+mp3		audio/mpeg
+ogg		audio/ogg
```
ojwb, line 220: There's also `oga` and `spx` for `audio/ogg`, and `ogv` for `video/ogg` - see https://tools.ietf.org/html/rfc5334#section-10.2

I think we probably don't want an extension to mime-type mapping for the `ogg` extension itself though - `ogg` can be `audio/ogg` or `video/ogg` (as your example files demonstrate!)  If we don't provide a mapping then omindex will use libmagic which can tell the difference from the file content:

```
$ file --mime-type testfiles/*/*.ogg
testfiles/audio/file_example_OOG_1MG.ogg:       audio/ogg
testfiles/video/file_example_OGG_480_1_7mg.ogg: video/ogg
```

(`file` uses libmagic internally.)

Both get mapped to libextractor by default, but the recorded mime-type will be wrong (which could matter if that's provided as a "file type" filter to users) and it'll also be confusing to the user if they overrides the default mappings.

xapian-applications/omega/handler_libextractor.cc
Exter-dg: Thanks for pointing this out. I have changed this.

xapian-applications/omega/mimemap.tokens
Exter-dg: Thank you for the insight, I have changed this.

ojwb: You've addressed the second part, but not the missing extensions I noted in the first paragraph.

We should pick up the type of such files via libmagic, but it's good to provide extension to mime type mappings where they're reliable since libmagic needs to look at the start of the file to determine the type, which takes more time than just looking at the end of the filename - not much per file, but it can add up.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -170,8 +181,14 @@ main(int argc, char** argv)
 	Xapian::Document doc = db.get_document(did);
 	auto iter = tests.find(url);
 	if (iter != tests.end()) {
-	    succeed &= compare_test(iter->second, doc, url);
+	    if (startswith(iter->first, "audio") ||
+		startswith(iter->first, "video"))
+		skip &= compare_test(iter->second, doc, url);
+	    else
+		succeed &= compare_test(iter->second, doc, url);
```
ojwb, line 188: I think we need to find a better approach than this.

As written, the test can never fail for an audio or video file, but there are test results which are clearly a failure because we should either get all the terms that come from a plugin or none of them.  So for example, for `audio/file_example_OOG_1MG.ogg` if we get `Akevin` we should also get `Simpact`.

You need to think carefully about what cases might happen due to a missing plugin and what can't, so that the test can actually catch problems.

Also, the variable `skip` seems to be confusingly named - it looks to me like `skip` being `false` means we're skipping the tests if they'd otherwise pass.  The sense seems reversed from what I'd expect.

(And a coding style point - it's clearer to put `{`...`}` on the blocks for an `if` where the condition has been split over multiple lines like this.)

xapian-applications/omega/mimemap.tokens
Exter-dg: Oh, I misunderstood the comment. I will fix this and update the PR.

Exter-dg: I have fixed this

xapian-applications/omega/Makefile.am
```diff
@@ -44,7 +44,7 @@ BUILT_SOURCES = extra/omegascript.vim my-html-tok.h mimemap.h namedents.h
 
 MAINTAINERCLEANFILES = my-html-tok.h mimemap.h mimemap.stamp namedents.h
 
-EXTRA_PROGRAMS = omindex_poppler omindex_libebook omindex_libetonyek omindex_tesseract omindex_gmime omindex_libarchive omindex_libabw omindex_libcdr
+EXTRA_PROGRAMS = omindex_poppler omindex_libebook omindex_libetonyek omindex_tesseract omindex_gmime omindex_libarchive omindex_libabw omindex_libcdr omindex_libextractor
```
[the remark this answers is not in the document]
Exter-dg: I have changed this.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -34,6 +34,8 @@ typedef vector<string> testcase;
 
 static unordered_map<string, testcase> tests;
 
+enum err_code { PASS, FAIL, SKIP };
```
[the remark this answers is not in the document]
Exter-dg: Thank you for pointing this out. I have changed the name to `test_result`.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -92,8 +94,8 @@ index_test()
 		  {"Zannual", "Zfed", "Zreturn", "Zwhile"}});
 #endif
 #if defined HAVE_LIBARCHIVE
-    // blank file
-    tests.insert({"odf/blank.odt", {}});
+    // blank file - will skip the test as no terms will be found.
+    // tests.insert({"odf/blank.odt", {}});
```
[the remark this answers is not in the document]
Exter-dg: I have made the necessary changes in a new commit.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -157,29 +159,54 @@ index_test()
...
+    tests.insert({"audio/file_example_OOG_1MG.ogg",
+		  {"Akevin", "Amacleod", "Simpact", "ZSmoderato", "Zlibrari",
+		   "Zcinemat"}});
+    tests.insert({"audio/file_example_WAV_1MG.wav",
+		  {"Zstereo", "wav", "Zms"}});
+#endif
 }
 
-static bool
+static err_code
```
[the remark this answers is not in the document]
Exter-dg: Thank you for your insight. I will try and make separate commits for distinct changes in the future.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -157,29 +159,54 @@ index_test()
...
+    bool term_found = false, exists = true;
     for (auto& t : test) {
 	term_iterator.skip_to(t);
 	if (term_iterator == doc.termlist_end() || *term_iterator != t) {
 	    cerr << "Error in " << file << ": Term " << t <<
 		 " does not belong to this file" << endl;
-	    return false;
+	    exists = false;
+	} else if (!term_found) {
+	    term_found = true;
```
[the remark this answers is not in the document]
Exter-dg: Thank you for pointing this out. I have changed the `else if` to `else`.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -157,29 +159,54 @@ index_test()
...
 	}
     }
-    return true;
+    // Skips if no term is found
+    // Passes if all terms are found
+    // Fails if only some terms are found
```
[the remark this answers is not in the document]
Exter-dg: I have changed this. The comments are now at the top of the function.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -157,29 +159,54 @@ index_test()
...
+	return SKIP;
+    }
 }
 
 int
 main(int argc, char** argv)
 {
     Xapian::Database db;
-    bool succeed = true;
+    err_code succeed = PASS;
```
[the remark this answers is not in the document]
Exter-dg: Thank you for pointing this out. Right now, I have changed 2 variable names - `result` will indicate the result of the entire test and `outcome` will indicate the result of individual tests on files. Are these better names?

* result -> outcome
* succeed -> result

xapian-applications/omega/omindexcheck.cc
```diff
@@ -196,8 +223,17 @@ main(int argc, char** argv)
...
+		succeed = SKIP;
 	}
     }
-    return succeed ? 0 : 1;
+    if (succeed == PASS)
+	return 0;
+    else if (succeed == FAIL)
+	return 1;
+    else
+	return 77;
```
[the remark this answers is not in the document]
Exter-dg: Thank you for pointing this out. I have added a comment for the same.

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,136 @@
...
+
+/*
+* cls: passed as last parameter from EXTRACTOR_extract
+* plugin_name: name of the plugin
+* type: mime-type of file according to libextractor
+* format: format information about data
+* data_mime_type: mimetype of data according to libextractor
+* data: actual meta-data found
+* data_len: number of bytes in data
+*/
```
ojwb, line 47: For Xapian code, we'd usually indent the `*` and `*/` lines so the `*`s all line up vertically with the one in `/*`.

It's also better to format function parameter docs as a doxygen documentation comment - that way someone can run doxygen over the code and get collated docs for internal APIs.  Going through and fixing up existing docs that lack such formatting is tedious work which is hard to automate, so it's the sort of thing which doesn't ever get done, so doing this when the comments are added is better.

What that actually means is starting the comment block with `/**` followed by a brief description of the purpose of the function, and then formatting the parameter list entries like: `@param data_len  number of bytes in data`

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,136 @@
...
+static int
+process_metadata(void* cls,
+		 const char* plugin_name,
+		 enum EXTRACTOR_MetaType type,
+		 enum EXTRACTOR_MetaFormat format,
+		 const char* data_mime_type,
+		 const char* data,
+		 size_t data_len)
+{
+    struct metadata* md = (struct metadata*)cls;
```
ojwb, line 57: Better to avoid C-style casts in C++ code.  A C-style cast can be used for one of several purposes and it may not be clear to the reader which are wanted.  Also the compiler doesn't know which, and that allows errors to slip through.  C++ casts make the purpose clear to the reader, and the compiler can complain if they're misused.  `static_cast<>` is the appropriate one here: https://en.cppreference.com/w/cpp/language/static_cast

As an example of the sort of problem this avoids, if we had `const void* cls2` and tried:

```
    struct metadata* md = (struct metadata*)cls2;
```

Then this would compile without even a warning (because one use of a C-style cast is to remove `const` qualification), and we'd be able to modify the object despite getting a `const` pointer to it - that's potentially very bad.  If the object is actually in the `RODATA` section, it could give an access violation.  If it's constant data but not protected in that way, it's undefined behaviour to modify it.  Even if it's not constant, someone reading code that passes in an object as a `const` pointer won't expect it to be modified, and reasoning about calling code may be invalid.

If you try to compile this you'll get an error:

```
    struct metadata* md = static_cast<struct metadata*>(cls2);
```

If you actually want to remove `const` with a C++ cast you need to use `const_cast<>` (and that only removes `const` - it won't accidentally change the type at the same time).

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,136 @@
...
+	string& author,
+	string& pages,
+	string& error)
+{
+    try {
+	struct metadata md(title, author, keywords, pages);
+
+	// Add all default plugins
+	struct EXTRACTOR_PluginList* plugins
+	 = EXTRACTOR_plugin_add_defaults(EXTRACTOR_OPTION_DEFAULT_POLICY);
```
ojwb, line 121: Xapian coding style is to wrap after `=` not before.

But you can probably avoid needing to wrap by using `auto` for the type.  The exact type isn't interesting here - this is really just an opaque thing representing plugins which we need to pass to other libextractor API functions.

xapian-applications/omega/handler_libextractor.cc
```diff
@@ -0,0 +1,136 @@
...
+
+	// If plugin not found/ File format not recognised/ corrupt file
+	// returns null and not an error.
+	EXTRACTOR_extract(plugins, filename.c_str(),
+			  NULL, 0,
+			  &process_metadata, &md);
+	EXTRACTOR_plugin_remove_all(plugins);
+
+    } catch (...) {
```
ojwb, line 130: Blank line at end of block.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -92,8 +94,6 @@ index_test()
 		  {"Zannual", "Zfed", "Zreturn", "Zwhile"}});
 #endif
 #if defined HAVE_LIBARCHIVE
-    // blank file
-    tests.insert({"odf/blank.odt", {}});
```
ojwb, line 96: This removes a testcase, which seems undesirable.

Adding per-testcase flags would address this in a much more satisfactory way as it would allow marking the testcases where "skip for no terms" is appropriate.  Then we could keep this testcase, and we'd also catch "missing all terms" bugs for formats where this missing plugin situation doesn't apply (which is probably at least everything that doesn't use libextractor, and perhaps even everything except libextractor with particular plugins).

barufa: Having a flag seems something really useful, but perhaps it would be better making that modification on a new PR (to avoid this become even larger).

xapian-applications/omega/handler_libextractor.cc
Exter-dg: I have made these changes and updated the comment.

xapian-applications/omega/handler_libextractor.cc
Exter-dg: Thank you for your insight. I have changed this.

xapian-applications/omega/handler_libextractor.cc
Exter-dg: I have changed the type to `auto`. But I had to still wrap the lines as it was still exceeding 80 chars.

xapian-applications/omega/handler_libextractor.cc
Exter-dg: I have corrected this.

xapian-applications/omega/omindexcheck.cc
Exter-dg: I have made the necessary changes here.
* Change implementation of testcase to a structure.
* Added an enum for flags with 3 values as of now - `FAIL_IF_NO_TERMS`, `SKIP_IF_NO_TERMS`, `PASS_IF_NO_TERMS`

barufa:
It seems that the result `omindextest` is `SKIP` in all the environment of Travis, Wouldn't it be a good idea to add the necessary plugins to get at least one `PASS`?

xapian-applications/omega/omindexcheck.cc
```diff
@@ -30,156 +30,217 @@
...
 {
-    sort(test.begin(), test.end());
+    // when all terms are found - PASS
+    // when no terms are found - SKIP/PASS/FAIL depending on the value of
+    // testcase flags.
+    // when no terms are found - FAIL
+    sort(test.terms.begin(), test.terms.end());
     Xapian::TermIterator term_iterator = doc.termlist_begin();
-    for (auto& t : test) {
+    bool term_found = false, exists = true;
```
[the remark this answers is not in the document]
jaylett: `all_terms_exist` would be slightly better.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -30,156 +30,217 @@
 
 using namespace std;
 
-typedef vector<string> testcase;
+enum test_result { PASS, FAIL, SKIP };
+enum flag { FAIL_IF_NO_TERMS = 1, SKIP_IF_NO_TERMS = 2, PASS_IF_NO_TERMS = 4 };
```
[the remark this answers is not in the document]
jaylett: `test_flag` would be one option, or you could nest it in the `struct testcase` which might be more appropriate.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -196,8 +257,18 @@ main(int argc, char** argv)
 	Xapian::Document doc = db.get_document(did);
 	auto iter = tests.find(url);
 	if (iter != tests.end()) {
-	    succeed &= compare_test(iter->second, doc, url);
+	    test_result outcome = compare_test(iter->second, doc, url);
```
jaylett, line 260: I'd probably call this `individual_result` or something. It's not going to be clear to someone just reading the variable names what the difference between "outcome" and "result" is.

xapian-applications/omega/omindexcheck.cc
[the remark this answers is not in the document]
Exter-dg: Thank you for pointing this out. I have changed the variable name to `test_flag`.
I haven't nested the declaration inside the struct as I was not sure if I understood it correctly.

@jaylett  Should I declare the enum `test_flag` inside the structure definition and then the insertion in tests would be of the form-
`tests.insert({"test.txt", {testcase::FAIL_IF_NO_TERMS, {"Zjoey"}}})`

xapian-applications/omega/omindexcheck.cc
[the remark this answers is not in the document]
Exter-dg: I have changed the variable name to `all_terms_exist`

xapian-applications/omega/omindexcheck.cc
Exter-dg: Thank you for pointing this out. I have changed the variable name to `individual_result` instead of `outcome`.

Exter-dg:
> It seems that the result `omindextest` is `SKIP` in all the environment of Travis, Wouldn't it be a good idea to add the necessary plugins to get at least one `PASS`?

We test 3 formats for libextractor - `avi`, `ogg`, `wav`. For `ogg` libvorbis is a dependency and I have added it to one of the distro in Travis CI. `avi` and `wav` plugins seem to use their own parsers and don't require any additional libraries. The test should pass now. I have updated the PR.

xapian-applications/omega/omindexcheck.cc
ojwb: I don't think we need 3 flags for this.  If there are 3 flags, there are 8 possible combinations of them, but most of the combinations don't make any sense.  As a general design point, it's cleaner to have a design where there aren't invalid combinations like this to worry about.

I think there are really just 2 cases here:

* We're using an extractor plugin which may not extract anything due to a missing dependency - this is `SKIP_IF_NO_TERMS`
* We aren't - not specifying `SKIP_IF_NO_TERMS` covers this case (and if you add a constructor to `struct testcase` with the parameter for `flags` last and with a default of 0 then this you should be able to only specify the flag on the testcases that need it and avoid the clutter of flags on every testcase in the list).

Note that you can tell the cases that you currently have marked with `FAIL_IF_NO_TERMS` and `PASS_IF_NO_TERMS` apart because the latter has `terms.empty()` and the former don't.  It wouldn't really make sense to list terms but have `PASS_IF_NO_TERMS`, and a testcase listing no terms with `FAIL_IF_NO_TERMS` would have to always fail.

Exter-dg: I have changed this and have updated the PR.

PR #307, opened 2020-07-30
Omindex libabw

Exter-dg:
* Adds support for Abiword documents using Libabw
* Also indexes meta-data

### DIFF
--- a/xapian-applications/omega/handler_libabw.cc
+++ b/xapian-applications/omega/handler_libabw.cc
@@ -1,0 +1,63 @@
+/** @file handler_libarchive.cc
+ * @brief Extract text and metadata using libarchive.
+ */
+/* Copyright (C) 2020 Parth Kapadia
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation; either version 2 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+#include <config.h>
+#include "handler.h"
+
+#include <cstring>
+
+#include <librevenge-generators/librevenge-generators.h>
+#include <librevenge-stream/librevenge-stream.h>
+#include <libabw/libabw.h>
+
+using namespace std;
+using namespace librevenge;
+
+static void
+clear_text(string& str, const char* text)
+{
+    if (text) {
+	for (int i = 0; text[i] != '\0'; ++i)
+	    if (!isspace(text[i]) || (i && !isspace(text[i - 1])))
+		str.push_back(text[i]);
+    }
+}
+
+bool
+extract(const string& filename,
+	const string& mimetype,
+	string& dump,
+	string& title,
+	string& keywords,
+	string& author,
+	string& pages,
+	string& error)
+{
+    RVNGFileStream input(filename.c_str());
+    RVNGString metadata_dump, content_dump;
+
+    RVNGTextTextGenerator metadata(metadata_dump, true);
+    if (libabw::AbiDocument::parse(&input, &metadata))
+	cout<<metadata_dump.cstr();
+
+    RVNGTextTextGenerator content(content_dump, false);
+    if (libabw::AbiDocument::parse(&input, &content))
+	cout<<content_dump.cstr();
+}
--- a/xapian-applications/omega/handler_libabw.cc
+++ b/xapian-applications/omega/handler_libabw.cc
@@ -1,7 +1,8 @@
 /** @file handler_libabw.cc
- * @brief Extract text and metadata using libarchive.
+ * @brief Extract text and metadata using libabw.
  */
-/* Copyright (C) 2020 Parth Kapadia
+/* Copyright (C) 2019 Bruno Baruffaldi
+ * Copyright (C) 2020 Parth Kapadia
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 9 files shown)


xapian-applications/omega/handler_libabw.cc
```diff
@@ -0,0 +1,168 @@
...
+	RVNGTextTextGenerator content(content_dump, false);
+	if (libabw::AbiDocument::parse(&input, &content)) {
+	    clear_text(dump, content_dump.cstr());
+	    succeed = true;
+	} else {
+	    if (!error.empty())
+		error.push_back('\n');
+	    error += "Libabw Error: Fail to extract text";
+	}
+	return succeed;
```
ojwb, line 163: `succeed` is initialised to `true`, and then set to `true` conditionally in two places - it never gets set anywhere else so what's the point of it?

Exter-dg: Thanks for pointing this out. It was a mistake. The plan is to initialise `succeed` to `false` and set it to `true` if libabw extracts metadata/content from the file.
I have corrected this.

xapian-applications/omega/omindexcheck.cc
```diff
@@ -133,7 +133,7 @@ index_test()
 #endif
 #if defined HAVE_LIBABW
     tests.insert({"abw/test.abw",
-		  {"ZAparth", "ZSabiword", "Zsampl", "Zdocument"}});
+		  {"ZAparth", "Zabiword", "Zsampl", "Zdocument"}});
```
ojwb, line 136: I said on IRC adding a comment here would be good:

> [07:58:00] <olly> so probably use something other than the document title in this testcase
> [07:58:26] <olly> and note in a comment that we deliberately aren't testing the title, because of this bug fix not being present in some versions still in use

It's useful to note in the code when there's a reason why we don't test something that it might seem to someone reading the code that it would be good to test (i.e. a term from the subject here), since otherwise they might try to be helpful and improve the test coverage by adding `ZSabiword` or similar to the terms checked, and because developer machines tend to have fairly recent versions of everything it'll probably work fine for them and then we'll have to do another loop around the cycle of discovering it doesn't work with older libabw versions and taking that back out again.

If you also note the version of libabw this was fixed in (which you can see from the libabw commit I linked to) then the comment also serves to tell us what version of libabw we need to assume if we want to restore this check in the future.

xapian-applications/omega/handler_libabw.cc
```diff
@@ -0,0 +1,168 @@
+/** @file handler_libabw.cc
+ * @brief Extract text and metadata using libarchive.
+ */
+/* Copyright (C) 2020 Parth Kapadia
```
ojwb, line 4: This is clearly significantly based on an existing file (e.g. all the `PARSE_FIELD` code is taken from one of the existing handler_*.cc files), so you should preserve copyright information from that file and add yourself rather than replacing it with yourself.

xapian-applications/omega/handler_libabw.cc
```diff
@@ -0,0 +1,168 @@
...
+static void
+clear_text(string& str, const char* text)
+{
+    if (text) {
+	for (int i = 0; text[i] != '\0'; ++i)
+	    if (!isspace(text[i]) || (i && !isspace(text[i - 1])))
+		str.push_back(text[i]);
+	str.push_back('\n');
+    }
+}
```
ojwb, line 121: There's code in omindex to clean up whitespace in the stored sample, so it really shouldn't be necessary to clean up whitespace here too.

The only reason to do so I can see would be if libabw produces output with so much superfluous whitespace that the cost of sending it over the pipe to the parent process is greater than the extra cost of this otherwise redundant cleaning up of whitespace.

xapian-applications/omega/handler_libabw.cc
ojwb: The revised code doesn't entirely make sense either.

The value in `error` is only used if this function indicates there was an error by returning `false`.  If it returns `true` that means "it worked", so it doesn't make sense to also try to report an error.

I think unless we have evidence that failure here is actually OK it makes more sense to be strict and just set `error` and `return false` right away if any call to `libabw::AbiDocument::parse()` fails.

If that turns out to reject certain abiword documents which can be usefully opened with abiword or libreoffice then we can relax the checks appropriately in response to bug reports.

If we start off being lax here it will be very hard to tighten up later - users will be more forgiving of a new feature being overly strict than an existing feature changing in a way which stops things working for them which were fine before.

(This is different to the situation for e.g. OpenDocument where we were reusing the existing parsers and matching how they were being applied currently made sense.)

xapian-applications/omega/handler_libabw.cc
```diff
@@ -0,0 +1,168 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301
+ * USA
+ */
+#include <config.h>
+#include "handler.h"
+#include "stringutils.h"
+
+#include <memory>
```
ojwb, line 25: Is anything from `<memory>` actually used in this file?

xapian-applications/omega/handler_libabw.cc
Exter-dg: Thank you for the insight. I have changed this.

* Removed the bool variable (succeed) altogether
* Returns false if it fails to extract either metadata or content

xapian-applications/omega/omindexcheck.cc
Exter-dg: Thanks for pointing this out. I have fixed this.

xapian-applications/omega/handler_libabw.cc
Exter-dg: Thank you for pointing this out. I have changed the copyright portion.

xapian-applications/omega/handler_libabw.cc
Exter-dg: I analysed the output generated by libabw's parse function for some files and it doesn't seem to generate extra whitespaces.
I removed the function from the file.

xapian-applications/omega/handler_libabw.cc
Exter-dg: Thanks for pointing this out. I noticed that some header files were used in the previous versions of the handler but after the changes, these were not required. I have removed them and I will try and make sure to keep a habit of checking if the header files are still required after making some changes.

PR #308, opened 2020-07-31
Support more normalisations in TfIdfWeight

dipanshu124:
Add support for wdf normalisation "augmented average term frequency",
and IDF normalisations "incremented global frequency" and
"square root global frequency".


ojwb:
The CI failures are due to a current appveyor problem, not to do with your changes.

ojwb:
Thanks, looks good - merged via git.xapian.org.

PR #309, opened 2020-08-06
Extract stat "wdfdocmax" for TfIdf Weight

dipanshu124:
Some TfIdf normalisations require the max wdf in the document.
This change adds remote and honey backend support for wdfdocmax,
hence fixes one issue of  #744.

### DIFF
--- a/xapian-core/backends/honey/honey_database.cc
+++ b/xapian-core/backends/honey/honey_database.cc
@@ -164,6 +164,22 @@ HoneyDatabase::get_unique_terms(Xapian::docid did) const
     return HoneyTermList(this, did).get_unique_terms();
 }
 
+Xapian::termcount
+HoneyDatabase::get_wdfdocmax(Xapian::docid did) const
+{
+    Assert(did != 0);
+    HoneyTermList termlist(this, did);
+    Xapian::termcount max_wdf = 0;
+    termlist.next();
+    while (!termlist.at_end()) {
+	Xapian::termcount current_wdf = termlist.get_wdf();
+	if (current_wdf > max_wdf) max_wdf = current_wdf;
+	termlist.next();
+    }
+    return max_wdf;
+
+}
+
 void
 HoneyDatabase::get_freqs(const string& term,
 			 Xapian::doccount* termfreq_ptr,
--- a/xapian-core/backends/honey/honey_database.cc
+++ b/xapian-core/backends/honey/honey_database.cc
@@ -177,7 +177,6 @@ HoneyDatabase::get_wdfdocmax(Xapian::docid did) const
 	termlist.next();
     }
     return max_wdf;
-
 }
 
 void
--- a/xapian-core/matcher/synonympostlist.cc
+++ b/xapian-core/matcher/synonympostlist.cc
@@ -41,6 +41,7 @@ SynonymPostList::set_weight(const Xapian::Weight * wt_)
     delete wt;
     wt = wt_;
     want_wdf = wt->get_sumpart_needs_wdf_();
+    want_wdfdocmax = wt->get_sumpart_needs_wdfdocmax_();
 }
 
 PostList *
@@ -85,12 +86,15 @@ SynonymPostList::get_weight(Xapian::termcount doclen,
 	    // that this is true.  Note that this requires doclen to be fetched
 	    // even if the weight object doesn't want it.
 	    if (doclen == 0) {
-		Xapian::termcount dummy;
-		pltree->get_doc_stats(pl->get_docid(), doclen, dummy, dummy);
+		doclen = pltree->get_doclength(pl->get_docid());
 	    }
 	    if (wdf > doclen) wdf = doclen;
 	}
-	// Set wdfdocmax to doclen in case of a synonym.
+    }
+    if (want_wdfdocmax) {
+	if (doclen == 0) {
+	    doclen = pltree->get_doclength(pl->get_docid());
+	}
 	wdfdocmax = doclen;
     }
     RETURN(wt->get_sumpart(wdf, doclen, unique_terms, wdfdocmax));
--- a/xapian-core/matcher/synonympostlist.cc
+++ b/xapian-core/matcher/synonympostlist.cc
@@ -61,7 +61,8 @@ SynonymPostList::skip_to(Xapian::docid did, double w_min)
 
 double
 SynonymPostList::get_weight(Xapian::termcount doclen,
-			    Xapian::termcount unique_terms) const
+			    Xapian::termcount unique_terms,
+			    Xapian::termcount wdfdocmax) const
 {
     LOGCALL(MATCH, double, "SynonymPostList::get_weight", doclen | unique_terms);
     Xapian::termcount wdf = 0;
@@ -85,12 +86,12 @@ SynonymPostList::get_weight(Xapian::termcount doclen,
 	    // even if the weight object doesn't want it.
 	    if (doclen == 0) {
 		Xapian::termcount dummy;
-		pltree->get_doc_stats(pl->get_docid(), doclen, dummy);
+		pltree->get_doc_stats(pl->get_docid(), doclen, dummy, dummy);
 	    }
 	    if (wdf > doclen) wdf = doclen;
 	}
     }
-    RETURN(wt->get_sumpart(wdf, doclen, unique_terms));
+    RETURN(wt->get_sumpart(wdf, doclen, unique_terms, wdfdocmax));
 }
 
 double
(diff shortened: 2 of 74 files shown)


xapian-core/backends/honey/honey_database.cc
```diff
@@ -164,6 +164,22 @@ HoneyDatabase::get_unique_terms(Xapian::docid did) const
...
+    Xapian::termcount max_wdf = 0;
+    termlist.next();
+    while (!termlist.at_end()) {
+	Xapian::termcount current_wdf = termlist.get_wdf();
+	if (current_wdf > max_wdf) max_wdf = current_wdf;
+	termlist.next();
+    }
+    return max_wdf;
+
+}
```
ojwb, line 181: Unwanted blank line at the end of a block here.

xapian-core/net/remoteprotocol.h
```diff
@@ -132,6 +133,7 @@ enum reply_type {
     REPLY_METADATAKEYLIST,	// Iterator for metadata keys
     REPLY_FREQS,		// Get termfreq and collfreq
     REPLY_UNIQUETERMS,		// Get number of unique terms in doc
+    REPLY_WDFDOCMAX,		// Get the max_wdf in doc
```
ojwb, line 136: `remote_protocol.rst` needs updating with the details of these new messages.

xapian-core/tests/api_weight.cc
```diff
@@ -1392,6 +1395,7 @@ class CheckStatsWeight : public Xapian::Weight {
 	TEST_REL(uniqueterms,>=,1);
 	TEST_REL(uniqueterms,<=,doclen);
 	TEST_REL(wdf,<=,wdf_upper);
+	TEST_REL(wdfdocmax,>=,1);
```
ojwb, line 1398: We can definitely check more than this (and I'm not sure this condition is actually correct - if all the wdf values are 0 in a document then wdfdocmax would be 0).

For example, wdfdocmax should be the largest wdf in the document, so wdfdocmax => wdf.

But it should be equal to some wdf value, so wdfdocmax <= wdf_upper.

There are probably some others.

ojwb:
Aside from the fairly minor notes above, this seems good.

I'm still not entirely happy that every time we add a new stat of this type we have to change the API of `Xapian::Weight` incompatibly, which means any user subclasses need updating (and conditionalised code to work with old and new versions of Xapian).  I suspect that's fairly rare, but it'll be annoying for those affected.

We've not yet come up with a plan for how to supply these stats efficiently which would avoid this though, and I don't think it's worth holding up progress on this.  It could be reworked before the next release series if we came up with a plan.

jaylett:
@dipanshu124 it'd be worth capturing olly's point about new stats and API compatibility somewhere in the wiki pages associated with your project. If you look on the wiki at what our other student has done, they've started to collect ideas like this in a separate place from the journal.

xapian-core/backends/honey/honey_database.cc
dipanshu124: Fixed.

xapian-core/net/remoteprotocol.h
dipanshu124: Updated.

xapian-core/tests/api_weight.cc
dipanshu124: I add `TEST_REL(wdfdocmax,>=,1)` assuming `TEST_REL(uniqueterms,>=,1)` is correct.
I think it would make sense to remove both of these.

wdfdocmax <= wdf_upper is not true in general. (Suppose the current term has Coll freq 1, then wdf_upper takes value 2. But wdfdocmax can take a value greater than 2 (due to some other term in the same doc.))

For wdfdocmax => wdf, wdfdocmax has to be made synonym aware.(I am figuring how to do that.)

I have added `TEST_REL(wdfdocmax,<=doclen)` for now.

ojwb: `TEST_REL(uniqueterms,>=,1);` is indeed interesting.  The case of a document not indexed by any terms isn't relevant here, as no terms means `get_sumpart()` won't get called, so the question comes down to whether uniqueterms is just the number of different terms in the document (in which case it counts terms which have zero wdf too, and the check is OK) or whether it only counts terms with non-zero wdf.  I thought we'd actually settled on the latter interpretation (since it seems undesirable that adding more filter terms changes the relevance weights, and it's also sometimes handy that the document length provides an upper bound on the number of unique terms), but the API documentation currently says:

> This is the number of different terms which index the given document.

If this count includes zero wdf terms too, then `TEST_REL(uniqueterms,<=,doclen);` is actually wrong instead.

I'll dig into this a bit more, but I think we probably just failed to clarify the API docs when we decided on this interpretation so let's assume that here for now.

Rather than removing these two `>= 1` checks, I think we should conditionalise them on `if (doclen > 0)` since `doclen` is the sum of all the wdfs in the document, so if that's non-zero there must be at least one non-zero wdf value in the document.

ojwb:
This is partly based on changes on the tfidf-maxwdf-norm branch isn't it?

If so, it's good to credit the original author in a situation like this where their original commits aren't merged into the tree - just saying something in the commit message like "Partly based on changes by Nishad Dawkhar on the tfidf-maxwdf-norm branch."

Acknowledging contributions helps make contributors feel appreciated and doesn't really cost us anything to do.  And hopefully it also makes it more likely someone will contribute again.

xapian-core/tests/api_weight.cc
dipanshu124: I have now conditionalise it.
Also added `wdfdocmax>=wdf`  by setting `wdfdocmax` to `doclen` in case of synonym.

dipanshu124:
> This is partly based on changes on the tfidf-maxwdf-norm branch isn't it?
> 
> If so, it's good to credit the original author in a situation like this where their original commits aren't merged into the tree - just saying something in the commit message like "Partly based on changes by Nishad Dawkhar on the tfidf-maxwdf-norm branch."
> 
> Acknowledging contributions helps make contributors feel appreciated and doesn't really cost us anything to do. And hopefully it also makes it more likely someone will contribute again.

Yes, we should do that.
Since I can't edit the commit message now, probably you would add that at the time of merge. ( I will take care from now on.)

jaylett:
@dipanshu124 changing the description of the PR to include that information is one way to at least flag this. At the least, it provides the information up front in a way that whoever merges it can include in the merge commit (or in the rebased commits if a project works like that, as Xapian does). In many projects, you would at the point of merge be asked to "rebase onto latest master", which involves a force push to the branch behind the PR — and then it would be up to you to change the commit message.

However there's a better way, at least for Xapian. You can create an empty (ie with zero changes) "squash" commit with the extra information. Then, when rebasing with autosquash (which is the typical way of collapsing commits), the commit message for the single collapsed commit will contain the base commit message plus the commit message from the squash commit.

You can create an empty squash commit mostly like you would for a fixup, using something like:

```bash
$ git commit --allow-empty -v --squash <commit>
```

Make sure you have nothing in the stage/index first! (Use `git status` to check.) That should open an editor asking for your commit message, with the "squash! …" subject line in place. Don't touch the subject line, but add the extra information in the body. Then you just push as you would for a fixup commit.

xapian-core/matcher/synonympostlist.cc
```diff
@@ -90,6 +90,8 @@ SynonymPostList::get_weight(Xapian::termcount doclen,
 	    }
 	    if (wdf > doclen) wdf = doclen;
 	}
+	// Set wdfdocmax to doclen in case of a synonym.
+	wdfdocmax = doclen;
```
ojwb, line 94: This dosn't seem quite right.

One problem is it only currently gets executed if `want_wdf` is set, but there's no inherent reason why a weighting scheme which uses wdfdocmax would also use wdf so this should be outside that `if`.

Another problem is that `doclen` may not have been read here (a weighting scheme could use wdfdocmax, but not doclen, and in that case the doclen won't be set unless the `(!wdf_disjoint && wdf > doclen_lower_bound)` condition above was true.  The approach we currently take to this for clamping wdf to doclen can be seen just above, though unlike that situation, for wdfdocmax we'll always want doclen available here so it's probably better to have `WDFDOCMAX` automatically enable `DOCLEN` when weighting a synonym (which is handled by a separate overloaded form of `Weight::init_()`).

I've pushed a commit to your branch which adds a testcase to demonstrate these problems - it fails currently, but I think it should pass for all 3 flag combinations tested.

ojwb:
Thanks for the updates.

The travis CI is failing on macOS - something's up with the doxygen output it seems.  But the diff from the last good version on the PR branch doesn't touch anything related to doxygen so I think this must be unrelated to the changes here so I've squashed and merged.

PR #310, opened 2020-08-11
Pass wdfdocmax to get_sumextra()

dipanshu124:
Since we pass doclen and uniqterms to get_sumextra(), it makes
sense that we also pass wdfdocmax to it. This fixes the second
issue of #744. (the first was fixed by PR 309)

### DIFF
--- a/xapian-core/matcher/extraweightpostlist.cc
+++ b/xapian-core/matcher/extraweightpostlist.cc
@@ -33,11 +33,12 @@ ExtraWeightPostList::get_weight(Xapian::termcount doclen,
 				Xapian::termcount unique_terms,
 				Xapian ::termcount wdfdocmax) const
 {
-    /* Weight::get_sumextra() takes two parameters (document length and number
-     * of unique terms) but none of the currently implemented weighting schemes
-     * actually use the latter - it was added because it's likely to be wanted
-     * at some point, and so that we got all the incompatible changes needed to
-     * add support for the number of unique terms over with in one go.
+    /* Weight::get_sumextra() takes three parameters (document length, number of
+     * unique terms and the max wdf in the document) but none of the currently
+     * implemented weighting schemes actually use the max wdf - it was added
+     * because it's likely to be wanted at some point, and so that we got all
+     * the incompatible changes needed to add support for the max wdf over with
+     * in one go.
      */
     double sum_extra = weight->get_sumextra(doclen, unique_terms, wdfdocmax);
     AssertRel(sum_extra,<=,max_extra);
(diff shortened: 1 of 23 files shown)


xapian-core/matcher/extraweightpostlist.cc
```diff
@@ -39,7 +39,7 @@ ExtraWeightPostList::get_weight(Xapian::termcount doclen,
      * at some point, and so that we got all the incompatible changes needed to
      * add support for the number of unique terms over with in one go.
```
ojwb, line 40: This comment needs updating too...

xapian-core/weight/tfidfweight.cc
```diff
@@ -214,13 +214,16 @@ TfIdfWeight::get_maxpart() const
 {
     Xapian::termcount wdf_max = get_wdf_upper_bound();
     Xapian::termcount len_min = get_doclength_lower_bound();
-    double wdfn = get_wdfn(wdf_max, len_min, len_min, wdf_max, wdf_norm_);
+    double wdfn = get_wdfn(wdf_max, len_min, len_min,
+			   wdf_max, wdf_norm_);
```
ojwb, line 218: This seems to be an unrelated formatting change (and not really a useful one either).

ojwb:
This looks good to me aside from a couple of minor things I've noted.

It'd be good to have test coverage for the new parameter to check the value passed is correct, but we don't actually have any of the unique_terms value passed to `get_sumextra()` either.  And really the main reason to have added that and to add this now is so that we get the public API changes out of the way at the same time as we're changing `get_sumpart()`.

xapian-core/matcher/extraweightpostlist.cc
dipanshu124: I have updated it. (Not sure if it's completely correct.)

xapian-core/weight/tfidfweight.cc
dipanshu124: Corrected.

xapian-core/matcher/extraweightpostlist.cc
```diff
@@ -33,11 +33,12 @@ ExtraWeightPostList::get_weight(Xapian::termcount doclen,
...
-     * of unique terms) but none of the currently implemented weighting schemes
-     * actually use the latter - it was added because it's likely to be wanted
-     * at some point, and so that we got all the incompatible changes needed to
-     * add support for the number of unique terms over with in one go.
+    /* Weight::get_sumextra() takes three parameters (document length, number of
+     * unique terms and the max wdf in the document) but none of the currently
+     * implemented weighting schemes actually use the max wdf - it was added
+     * because it's likely to be wanted at some point, and so that we got all
+     * the incompatible changes needed to add support for the max wdf over with
+     * in one go.
```
ojwb, line 41: What it said before about unique terms being unused is still correct - of the parameters, only `doclen` is currently used (which is perhaps the simpler way to put it now).

dipanshu124: I have corrected it. (I was confused as we have TfIdf normalisations that use unique terms.)

PR #311, opened 2020-08-12
Add support for CorelDRAW files

Exter-dg:


xapian-applications/omega/handler_libcdr.cc
```diff
@@ -0,0 +1,71 @@
...
+{
+    try {
+	RVNGFileStream input(filename.c_str());
+	RVNGStringVector cdr_pages;
+	RVNGTextDrawingGenerator content(cdr_pages);
+
+	// check if cdr file supported
+	if (libcdr::CDRDocument::isSupported(&input)) {
+	    if (libcdr::CDRDocument::parse(&input, &content)) {
+
```
barufa, line 49: Maybe we can remove this empty line

xapian-applications/omega/index_file.cc
```diff
@@ -265,6 +265,11 @@ index_add_default_libraries()
     index_library("application/x-abiword", omindex_libabw);
     index_library("application/x-abiword-compressed", omindex_libabw);
 #endif
+#if defined HAVE_LIBCDR
+    Worker* omindex_libcdr = new Worker("omindex_libcdr");
+    index_library("application/zip", omindex_libcdr);
+    index_library("application/octet-stream", omindex_libcdr);
```
barufa, line 271: These mimetypes are too broad. `application/octet-stream` is defined as "arbitrary binary data" and maybe it would be better to handle application/zip with omindex_libarchive.

Perhaps `image/x-coreldraw cdr` would be more appropriated.

barufa:
It seems that Travis is failing due to an external issue and omindex_libcdr haven't been checked in MacOs jet.

xapian-applications/omega/index_file.cc
ojwb: Indeed - the code above will try to process zip files and arbitrary binary data with libcdr.

For a case like this where libmagic gives us a mime-type which isn't specific to the type of file we should add extension mappings (in this case for `cdr` and `cmx`) to an appropriate mime-type or types, and then use that type or types here.

Looking in the EDRM data set (which has sample files of quite a lot of different types), the libmagic from debian unstable reports `image/x-coreldraw` when it reports a specific type, so that's a good choice:

```
$ file --mime-type /mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/*/*.{cdr,CDR}
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/bmp/4seafoodc.cdr: application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/bmp/PO.cdr:        image/x-coreldraw
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cdr/5radialg.cdr:  application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cdr/award.cdr:     image/x-coreldraw
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cdr/itmenu.cdr:    application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/bmp/911.CDR:       application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/bmp/C5MENU.CDR:    application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/bmp/EYE.CDR:       application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cdr/BOXES.CDR:     application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cdr/FROMWPG.CDR:   application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cdr/GOODPEAR.CDR:  application/octet-stream
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cdr/UNIFILL.CDR:   application/octet-stream
```

There's only one CMX file in there which gives a non-specific type:

```
$ file --mime-type /mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cmx/NEWSFLSH.CMX
/mnt/data/scratch/EDRM_Data-Set_File-Formats_1-0/data-set/cmx/NEWSFLSH.CMX: application/octet-stream
```

While looking to see if I had any other .cmx files, I noticed it's used by ocaml:

```
$ file /usr/lib/ocaml/ocamldoc/odoc_texi.cmx
/usr/lib/ocaml/ocamldoc/odoc_texi.cmx: OCaml native object file (.cmx) (Version 025)
$ file --mime-type /usr/lib/ocaml/ocamldoc/odoc_texi.cmx
/usr/lib/ocaml/ocamldoc/odoc_texi.cmx: application/octet-stream
```

But since libmagic doesn't currently seem to identify coreldraw CMX files I think we probably need to assume that `cmx` is coreldraw CMX at least for now (if libmagic learns about them then we could change that).  That will mean we pointlessly try to process any ocaml cmx files we find, but that seems less bad than ignoring the ones we can handle.

xapian-applications/omega/handler_libcdr.cc
Exter-dg: I have removed the extra line.

xapian-applications/omega/index_file.cc
Exter-dg: Thanks for pointing this out.
Actually, for all versions of CMX files that I tested, libcdr doesn't seem to extract any data from them. I tested CMX files on both libcdr-0.1.5 and libcdr-0.1.6 and hence have not included them for now. 

I used two mime-types earlier as older versions of `.cdr` files returned `application/octet-stream` while the newer zip containers returned `application/zip`. I have replaced these with `image/x-coreldraw`.

PR #312, opened 2020-08-14
Support more normalisations for TfIdfWeight

dipanshu124:
Add support for wdf normalisations "max-norm" and "aug-norm"
described by SMART.


ojwb:
Thanks - rebased and merged via git.xapian.org.

PR #313, opened 2020-08-17
Fix glass function readahead_key to not abort for overlong keys

rsto:
Currently, a query may abort with an InvalidArgument error if one of
the terms exceed the maximum allowed length for glass btree keys.
However, this error is thrown only if the query triggers a readahead_key
call on the glass table.

This patch updates readahead_key to report overlong keys as not found,
rather than aborting the whole query.


ojwb:
Thanks.

This highlights that we don't actually currently seem to have any tests for searching for a term which isn't present, so I've adjusted the testcase to be more general - it now runs for all backends and tests a wide range of term lengths.  This didn't turn up any other bugs, but at least means it'll be harder to introduce similar problems in the future.

PR #314, opened 2020-08-19
Update TfIdfWeight::create_from_parameters() 

dipanshu124:
We now use enums to specify normalisations. The parameters will be
specified as constant names.

### DIFF
--- a/xapian-core/weight/Makefile.mk
+++ b/xapian-core/weight/Makefile.mk
@@ -2,10 +2,10 @@ noinst_HEADERS +=\
 	weight/weightinternal.h
 
 EXTRA_DIST +=\
-	weight/collate-wdf-norm\
 	weight/collate-idf-norm\
-	weight/wdf-norm-dispatch.h\
+	weight/collate-wdf-norm\
 	weight/idf-norm-dispatch.h\
+	weight/wdf-norm-dispatch.h\
 	weight/Makefile
 
 lib_src +=\
@@ -28,12 +28,12 @@ lib_src +=\
 	weight/weight.cc\
 	weight/weightinternal.cc
 
-weight/wdf-norm-dispatch.h: weight/collate-wdf-norm weight/Makefile.mk common/Tokeniseise.pm
-	$(PERL) -I'$(srcdir)/common' '$(srcdir)/weight/collate-wdf-norm' '$(srcdir)'
-
 weight/idf-norm-dispatch.h: weight/collate-idf-norm weight/Makefile.mk common/Tokeniseise.pm
 	$(PERL) -I'$(srcdir)/common' '$(srcdir)/weight/collate-idf-norm' '$(srcdir)'
 
+weight/wdf-norm-dispatch.h: weight/collate-wdf-norm weight/Makefile.mk common/Tokeniseise.pm
+	$(PERL) -I'$(srcdir)/common' '$(srcdir)/weight/collate-wdf-norm' '$(srcdir)'
+
 BUILT_SOURCES +=\
-	weight/wdf-norm-dispatch.h\
-	weight/idf-norm-dispatch.h
+	weight/idf-norm-dispatch.h\
+	weight/wdf-norm-dispatch.h
--- a/xapian-core/weight/Makefile.mk
+++ b/xapian-core/weight/Makefile.mk
@@ -2,6 +2,10 @@ noinst_HEADERS +=\
 	weight/weightinternal.h
 
 EXTRA_DIST +=\
+	weight/collate-wdf-norm\
+	weight/collate-idf-norm\
+	weight/wdf-norm-dispatch.h\
+	weight/idf-norm-dispatch.h\
 	weight/Makefile
 
 lib_src +=\
@@ -23,3 +27,13 @@ lib_src +=\
 	weight/tradweight.cc\
 	weight/weight.cc\
 	weight/weightinternal.cc
+
+weight/wdf-norm-dispatch.h: weight/collate-wdf-norm weight/Makefile.mk common/Tokeniseise.pm
+	$(PERL) -I'$(srcdir)/common' '$(srcdir)/weight/collate-wdf-norm' '$(srcdir)'
+
+weight/idf-norm-dispatch.h: weight/collate-idf-norm weight/Makefile.mk common/Tokeniseise.pm
+	$(PERL) -I'$(srcdir)/common' '$(srcdir)/weight/collate-idf-norm' '$(srcdir)'
+
+BUILT_SOURCES +=\
+	weight/wdf-norm-dispatch.h\
+	weight/idf-norm-dispatch.h
--- a/xapian-core/weight/collate-idf-norm
+++ b/xapian-core/weight/collate-idf-norm
@@ -23,15 +23,19 @@ EOF
 use Tokeniseise;
 
 my $hdr = Tokeniseise->new('weight/idf-norm-dispatch.h', 'Map string to idf normalisation code', $copyright, 'XAPIAN_INCLUDED_IDF_NORM_DISPATCH_H', 'idf_norm', 2, 'false');
-$hdr->add('NONE', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::NONE)');
-$hdr->add('TFIDF', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::TFIDF)');
-$hdr->add('SQUARE', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::SQUARE)');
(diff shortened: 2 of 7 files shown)


xapian-core/weight/tfidfweight.cc
```diff
@@ -353,12 +355,77 @@ TfIdfWeight::get_wtn(double wt, wt_norm wt_normalization) const
...
+	{ "PROB", idf_norm::PROB },
+	{ "PIVOTED", idf_norm::PIVOTED },
+	{ "GLOBAL_FREQ", idf_norm::GLOBAL_FREQ },
+	{ "LOG_GLOBAL_FREQ", idf_norm::LOG_GLOBAL_FREQ },
+	{ "INCREMENTED_GLOBAL_FREQ", idf_norm::INCREMENTED_GLOBAL_FREQ },
+	{ "SQRT_GLOBAL_FREQ", idf_norm::SQRT_GLOBAL_FREQ }
+    };
+    map<string, wt_norm> wtnorm {
+	{ "NONE", wt_norm::NONE }
+    };
```
ojwb, line 408: This will have to construct these maps each time this method is called and then does one lookup in each - in general it very rarely makes sense to construct a complex data structure at run-time for a single lookup.  Building the data structure inevitably has to make a pass over all the data being put in it making in at least O(n), which is what a simple linear search over the data would be.

Constructing a map is O(n*log(n)) if the input data isn't sorted, as here (pre-sorting the input would help as then it's O(n)) and then looking up in the map is O(log(n)) - if you just did a simple `if`-chain that'd be O(n).  If the input data is sorted, binary chop is O(log(n)) (and `std::lower_bound()` provides a binary chop implementation) though that does rely on people adding new entries actually maintaining the sorted order.

Admittedly big-O scaling isn't necessarily the key concern when you have a small number of entries because the scale factor can matter more - e.g. a fast O(n²) can beat a slow O(log(n)) for bounded n - but I'd expect the scale factor to be worse for building a std::map keyed on strings.

Another issue this hits is that the initialiser is probably essentially an array with string constants in it, and in a shared library each string constant in an array needs relocating by the dynamic linker at library load time, which makes loading the library slower so we try to avoid these.  If the string constants are sensibly bounded in length (as here) there's a trick to avoid this problem - however we have a mechanism to build lookup tables for string to enum (or integer) mapping at compile time which I think would make sense to also use here.

It's a perl module - you write a script which instantiates an object `$hdr` and then for each (string, enum) pair calls:

```
$hdr->add($string, $enum);
```

This generates a header file which you then include from the code and to look up string `s`:

```
int r = keyword(whatever_the_generated_table_is_called, s.data(), s.size())
if (r < 0) {
    // Handle an unknown token.
    throw some_exception();
}
// Known token, so safe to cast to enum_type.
auto code = static_cast<enum_type>(r);
```

(There's also `keyword2()` which handles lookup tables which are large enough to need 2-byte offsets.)

This is used for keyword lookups in a number of places already - e.g. for stemmer names see `xapian-core/languages/collate-sbl` which generates `xapian-core/languages/sbl-dispatch.h` and is used in `xapian-core/languages/stem.cc`.  This script also has to allocate enum values and create an enum definition, which you don't need here as they are already defined so you don't need that part.

You'll want a separate lookup table for each type (or maybe just hard-code a simple `== "NONE"` check for the last one for now.

The way the lookup works is essentially to dispatch first on the length of the input via a lookup table (which is O(1)) and then perform a binary chop just within a list of known tokens with that length, so it should be more efficient that a binary chop on the whole list and the lookup tables are compact.

xapian-core/weight/Makefile.mk
```diff
@@ -2,6 +2,10 @@ noinst_HEADERS +=\
 	weight/weightinternal.h
 
 EXTRA_DIST +=\
+	weight/collate-wdf-norm\
+	weight/collate-idf-norm\
+	weight/wdf-norm-dispatch.h\
+	weight/idf-norm-dispatch.h\
```
ojwb, line 8: It'd be better to add these in sorted order - that way it's easier to quickly locate an entry, and that also makes it harder to accidentally add a duplicate of an entry which is already there (because you'd be trying to add it in the place it already is).

xapian-core/weight/collate-idf-norm
```diff
@@ -0,0 +1,37 @@
...
+$hdr->add('NONE', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::NONE)');
+$hdr->add('TFIDF', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::TFIDF)');
+$hdr->add('SQUARE', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::SQUARE)');
+$hdr->add('FREQ', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::FREQ)');
+$hdr->add('PROB', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::PROB)');
+$hdr->add('PIVOTED', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::PIVOTED)');
+$hdr->add('GLOBAL_FREQ', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::GLOBAL_FREQ)');
+$hdr->add('LOG_GLOBAL_FREQ', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::LOG_GLOBAL_FREQ)');
+$hdr->add('INCREMENTED_GLOBAL_FREQ', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::INCREMENTED_GLOBAL_FREQ)');
+$hdr->add('SQRT_GLOBAL_FREQ', 'static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::SQRT_GLOBAL_FREQ)');
```
ojwb, line 35: This seems like an unhelpful amount of duplication (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  Someone adding a new entry would be likely to copy and existing one and could easily fail to update it fully.

Better to loop over the names and have a single call to `$hdr->add(`...`)` - then the boilerplate code appears just once and is separate from the data - something like (untested):

```
for my $enum (qw/
    NONE
    TFIDF
    etc
/) {
    $hdr->add($enum, "static_cast<unsigned char>(Xapian::TfIdfWeight::idf_norm::$enum)");
}
```

Better still would be to extract the list of enum values automatically so a new one only needs adding to the header, but that seems more challenging.

xapian-core/weight/tfidfweight.cc
```diff
@@ -23,6 +23,10 @@
 #include <config.h>
 
 #include "xapian/weight.h"
+#include "weightinternal.h"
+#include "wdf-norm-dispatch.h"
+#include "idf-norm-dispatch.h"
+#include "keyword.h"
```
ojwb, line 29: Again, this new block ought to be in sorted order (but keep `xapian/weight.h` first as that's the header which has the definitions for this source file - we include that first as a simple way to try to ensure that our headers include any other headers they need (it doesn't ensure this for headers without a corresponding `.cc` file though).

xapian-core/weight/tfidfweight.cc
```diff
@@ -353,12 +357,50 @@ TfIdfWeight::get_wtn(double wt, wt_norm wt_normalization) const
...
 TfIdfWeight *
 TfIdfWeight::create_from_parameters(const char * p) const
 {
     if (*p == '\0')
 	return new Xapian::TfIdfWeight();
-    return new Xapian::TfIdfWeight(p);
+    string wdf_normalisation, idf_normalisation, wt_normalisation;
+    wdf_norm wdf_normalisation_ = wdf_norm::NONE;
+    idf_norm idf_normalisation_ = idf_norm::TFIDF;
+    wt_norm wt_normalisation_ = wt_norm::NONE;
```
ojwb, line 374: Better to declare these 3 below at the point you actually have the correct value to assign to them.  As this code is currently written the reader has to read it carefully to see that the values set here are never actually used.

xapian-core/weight/Makefile.mk
dipanshu124: Updated.

xapian-core/weight/collate-idf-norm
dipanshu124: Right, This was a lot of duplication which I could have avoided.
I have corrected that.

xapian-core/weight/tfidfweight.cc
dipanshu124: Updated

xapian-core/weight/tfidfweight.cc
dipanshu124: Updated

PR #315, opened 2020-08-23
Support Legacy Mac Documents using libmwaw

Exter-dg:
* Adds support for Apple Works documents


xapian-applications/omega/handler_libmwaw.cc
```diff
@@ -0,0 +1,264 @@
...
+    RVNGTextSpreadsheetGenerator content(pages_content, false);
+    result = MWAWDocument::parse(input, &content);
+    if (result != MWAWDocument::MWAW_R_OK) {
+	error = "Libmwaw Error: Failed to extract text";
+	return false;
+    } else {
+	for (unsigned i = 0; i < pages_content.size(); ++i) {
+	    dump.append(pages_content[i].cstr());
+	}
+	pages = pages_content.size();
```
ojwb, line 175: What type does `pages_content.size()` return?

Here you assign it to `pages`, which is a `std::string&` but above you compare `i < pages_content.size()` where `i` is unsigned.  I'm a bit surprised one doesn't fail to compile, and am concerned that at least one isn't doing what we want it to...

xapian-applications/omega/handler_libmwaw.cc
```diff
@@ -0,0 +1,264 @@
...
+	parse_metadata(data, len, author, title, keywords);
+    }
+
+    RVNGStringVector pages_content;
+    RVNGTextSpreadsheetGenerator content(pages_content, false);
+    result = MWAWDocument::parse(input, &content);
+    if (result != MWAWDocument::MWAW_R_OK) {
+	error = "Libmwaw Error: Failed to extract text";
+	return false;
+    } else {
```
ojwb, line 171: Control flow can't leave the block above, because there's a `return` at the end.  In a situation like this, the code which follows doesn't need to be in an `else` block, and putting it in one tends to make the code less clear - the code which follows is conceptually part of the actual purpose of the function (vs the code in the `if` block above which is handling an error case.)

xapian-applications/omega/handler_libmwaw.cc
Exter-dg: As per my understanding, `pages_content.size()` returns an integer (numerical value). 
`pages` in the handlers is supposed to store the number of the pages in the file, right?

So what I did was - iterate over the pages (pages_content.size() times) of the file and append the data to `dump`.
And later on line 175, I store the number of pages (`pages_content.size()`) in the document in the variable `pages`.

Am I interpreting things wrongly?

xapian-applications/omega/handler_libmwaw.cc
Exter-dg: Thanks for pointing this out. I have changed this.

xapian-applications/omega/handler_libmwaw.cc
ojwb: > `pages` in the handlers is supposed to store the number of the pages in the file, right?

Yes, but it's a `std::string&` and C++ doesn't implicitly convert an integer to a string for you.

It looks (from `/usr/include/librevenge-0.0/librevenge/RVNGStringVector.h`) like `size()` returns `unsigned` here.

Testing that:

```
$ cat string-from-unsigned.cc 
#include <iostream>
#include <string>

void f(std::string& pages, unsigned size) {
    pages = size;
}

int main() {
    std::string s;
    f(s, 42);
    std::cout << "'" << s << "'\n";
}
$ g++ -Wall -W -O2 -s string-from-unsigned.cc
$ ./a.out
'*'
```

So the code does compile, but my concern that it isn't doing what we want is well founded - it actually sets `pages` to a single character based on the value assigned!

You want `pages = str(pages_content.size());` instead.

xapian-applications/omega/handler_libmwaw.cc
```diff
@@ -0,0 +1,264 @@
...
+    MWAWDocument::Result result = MWAWDocument::parse(input, &content);
+    if (result != MWAWDocument::MWAW_R_OK) {
+	error = "Libmwaw Error: Failed to extract text";
+	return false;
+    } else {
+	for (unsigned i = 0; i < pages_content.size(); ++i) {
+	    dump.append(pages_content[i].cstr());
+	}
+	pages = pages_content.size();
+    }
```
ojwb, line 190: For the same reasons I think I pointed out in another PR, I'd get rid of the `else` here and put its contents at the top level of the function.  Same below.

xapian-applications/omega/handler_libmwaw.cc
```diff
@@ -0,0 +1,264 @@
...
+    return true;
```
ojwb, line 215: We seem to have 3 copies of the same code to iterate pages in an `MWAWDocument` - while it's not hugely complex code, it's probably more maintainable to factor that out into its own function so we just have it once.

xapian-applications/omega/handler_libmwaw.cc
Exter-dg: Thank you for pointing this out. I have made the necessary changes.

xapian-applications/omega/handler_libmwaw.cc
Exter-dg: I have made these changes.

xapian-applications/omega/handler_libmwaw.cc
Exter-dg: Thank you for pointing this out. I have changed this and created a separate function for iterating over pages.

ojwb:
Sorry for this having stalled for so long.  I've made the updates needed to apply it to current git master and pushed it.  Thanks for all your work on this.

PR #316, opened 2020-09-10
Add tests for TfIdfWeight::create_from_parameters()

dipanshu124:
Add tests to ensure that create_from_ parameters() works correctly.
Also, the wdf_norm and idf_norm tables aren't very large, so 1 byte
offset are sufficient.


ojwb:
Sorry we failed to deal with this for so long.  It looks good, so now merged.

PR #318, opened 2020-09-29
Headers missing to compile on msys with mingw.

alexsilva:
In msys with mingw-32bits (windows 10) I always have this problem and it is only solved by including this header in these files.
```
net/progclient.cc: In static member function 'static int ProgClient::run_program(const string&, const string&, void*&)':
net/progclient.cc:207:61: error: 'ERROR_PIPE_CONNECTED' was not declared in this scope
     if (!ConnectNamedPipe(hPipe, NULL) && GetLastError() != ERROR_PIPE_CONNECTED) {
                                                             ^~~~~~~~~~~~~~~~~~~~
make[2]: *** [net/progclient.lo] Error 1

net/remoteconnection.cc: In member function 'void RemoteConnection::shutdown()':
net/remoteconnection.cc:737:31: error: 'ERROR_IO_PENDING' was not declared in this scope
  if (!ok && GetLastError() == ERROR_IO_PENDING) {
```

### DIFF
--- a/xapian-core/net/progclient.h
+++ b/xapian-core/net/progclient.h
@@ -23,6 +23,10 @@
 
 #include <sys/types.h>
 
+#ifdef __WIN32__
+#include <winerror.h>
+#endif
+
 #include "backends/remote/remote-database.h"
 
 /** Implementation of RemoteDatabase using a spawned server.
--- a/xapian-core/net/remoteconnection.h
+++ b/xapian-core/net/remoteconnection.h
@@ -30,7 +30,7 @@
 
 #ifdef __WIN32__
 # include "safewinsock2.h"
-
+# include <winerror.h>
 # include <xapian/error.h>
 
 /** Class to initialise winsock and keep it initialised while we use it.


ojwb:
Thanks for raising this.  I'm a bit puzzled by why the build is failing for you, as we have CI builds which are all currently passing:

https://ci.appveyor.com/project/ojwb/xapian/builds/35393805

That tests mingw+msys (from http://mingw.org/), and both the 32-bit and 64-bit versions of mingw-w64+msys2 (http://mingw-w64.org/doku.php) - the latter seems to be rather more actively maintained.

In the headers I have to hand, it looks like <windows.h> includes <winbase.h> which includes <winerror.h>, but presumably that isn't the case for you?

Are you using the mingw.org or mingw-w64.org variant?  And what version of it?

alexsilva:
I downloaded and installed it from the website `mingw.org`

The version I have here is `gcc (MinGW.org GCC-6.3.0-1) 6.3.0`

The build is 32-bit on a Windows 10 x64 OS.

ojwb:
Thanks for the additional information.  These headers seem to be in the w32api package, which is currently version 5.4.1 on mingw.org.  I downloaded that and poked inside it on Linux, and it does indeed **not** seem to reach winerror.h from windows.h.

I'm not sure why the mingw CI passes.  The VM the CI uses has MSVC installed and my guess is that it's actually using the MSVC headers.  That would be good to address so we actually catch problems like yours which are specific to the mingw headers, but meanwhile we should address your problem.

Looking at [this example on MSDN](https://docs.microsoft.com/en-us/windows/win32/ipc/multithreaded-pipe-server) it uses `ERROR_PIPE_CONNECTED` and doesn't include `<winerror.h>`, so I'm going to address it via our `safewindows.h` wrapper header so we only need to workaround this in one place.

ojwb:
Backported ready for 1.4.18 in 1c88f6482cb7f802dddc7ec6133e6a0835b6fb15.

PR #319, opened 2020-11-19
Allow build on darwin20 (macOS Big Sur)

fxcoudert:
The configure script needs to be adapted to match on darwin20 (macOS Big Sur, 11.0)


ojwb:
The CI failure on macos seems unrelated (looks like asking homebrew to install ccache causes the already installed gdal and postgis to be upgraded, and the new gdal pulls in a huge pile of dependencies - then the build eventually times out after 10 minutes of no output while building cmake...)

Thanks for the patch.

This special handling for `MACOSX_DEPLOYMENT_TARGET` is only meant to kick in for 10.4 and earlier (10.3 and earlier for x86) and 10.4 went out of support in 2009, so after a brief discussion on IRC we're going to drop this handling from git master, but apply the patch to the stable 1.4.x series as it's less invasive.

PR #320, opened 2021-02-02
Fix rankers' implementation and indexing errors

dipanshu124:
This PR is a followup of #278 and attempts to resolve the remaining issues.

PR #324, opened 2022-06-29
Letor updates

ojwb:
This is a replacement for #320, updated for the change from travis to GHA.


ojwb:
I've rebased onto current master, which was clean apart from `.travis.yml` having been removed.

ojwb:
We are still seeing very similar looking test failures in the linux clang build and on macos (macos uses clang as the default compiler so these may well be the same issue).

I've tested with clang (with ubsan and asan enabled) locally and have a clean test run (after a few fixes for merge problems and one bug).

I think I'm going to try to peel out logically separate changes and test and merge them separately to try to get to the bottom of this.   I've already cherry-picked and merged the changes to drop SVMRanker.

I'll likely rebase and force-push this branch to drop out merged changes.  If anyone wants to help with this, please speak up and we can come up with a plan.

ojwb:
I've merged several commits to master, and have reached the point where I can't easily peel off more because there are 3 commits which change the testsuite results, and one commit which updates the testsuite to pass again.

I noticed the Xavier initialisation was using `default_random_engine` which may vary between implementations so I tried changing it to `minstd_rand0` as that's what `default_random_engine` actually is with current GCC.

That's helped, but there's still one test failing with clang and on macos, but passing elsewhere (and for me locally with clang).

There are also two XFAIL-ed testcases passing (which I don't see locally either):

```
./apitest total: 81 tests passed, 1 failed, 2 expected failures passed, 6 expected failures.
```

PR #331, opened 2024-01-25
QueryParser: do not try to parse unbroken words as group terms

rsto:
Fixes a bug where QueryParser fails with an error when parsing the sequence of: term, whitespace, unbroken words, term. The underlying issue is that unbroken words can not be part of a group according to the lemon grammar.

As the combination of unbroken words and terms is highly unlikely to ever form a multi-term synonym, this patch changes the query parser to leave group term mode after having parsed unbroken words.

### DIFF
--- a/xapian-core/queryparser/queryparser.lemony
+++ b/xapian-core/queryparser/queryparser.lemony
@@ -1506,6 +1506,9 @@ phrased_term:
 
 	    if (needs_word_break) {
 		Parse(&parser, UNBROKEN_WORDS, term_obj, &state);
+		// Drop out of IN_GROUP mode.
+		if (mode == IN_GROUP || mode == IN_GROUP2)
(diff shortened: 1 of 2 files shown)


ojwb:
Thanks.  Seems reasonable, but I'd like to have a look at the wider context before merging.

ojwb:
Sorry about taking so long to getting to looking into this.

I wondered if we should also be doing the `EMPTY_GROUP_OK` thing, but adding that in I can get parse errors for cases which should parse so it seems not.

I think you're right that we don't need to worry about trying multi-word synonyms in this situation - they'd need to include words from different scripts.  If someone really wants that to work we can revisit.

I think it'd be good to exercise the `IN_GROUP2` case as well as the `IN_GROUP` one so I'll add variants of your testcases which do that and merge via git.xapian.org in a few minutes.

PR #332, opened 2024-02-28
resolver.h: use AI_NUMERICSERV only when defined

barracuda156:
Older macOS does not have `AI_NUMERICSERV`, so use it as long as it is defined.

Otherwise the build is broken:
```
libtool: compile:  /opt/local/bin/g++-mp-13 -DHAVE_CONFIG_H -I. -I./common -I./include -I/opt/local/include -Wall -W -Wredundant-decls -Wpointer-arith -Wcast-qual -Wcast-align -Wformat-security -fno-gnu-keywords -Wundef -Woverloaded-virtual -Wstrict-null-sentinel -Wshadow -Wstrict-overflow=1 -Wlogical-op -Wmissing-declarations -Wdouble-promotion -Winit-self -Wduplicated-cond -Wduplicated-branches -fvisibility=hidden -fvisibility-inlines-hidden -pipe -Os -D_GLIBCXX_USE_CXX11_ABI=0 -arch ppc -fno-math-errno -MT net/tcpclient.lo -MD -MP -MF net/.deps/tcpclient.Tpo -c net/tcpclient.cc  -fno-common -DPIC -o net/.libs/tcpclient.o
In file included from net/tcpclient.cc:29:
net/resolver.h: In constructor 'Resolver::Resolver(const std::string&, int, int)':
net/resolver.h:99:18: error: 'AI_NUMERICSERV' was not declared in this scope; did you mean 'NI_NUMERICSERV'?
   99 |         flags |= AI_NUMERICSERV;
      |                  ^~~~~~~~~~~~~~
      |                  NI_NUMERICSERV
make[2]: *** [net/tcpclient.lo] Error 1
```

### DIFF
--- a/xapian-core/net/resolver.h
+++ b/xapian-core/net/resolver.h
@@ -143,7 +143,9 @@ class Resolver {
 	if (host != "::1" && host != "127.0.0.1" && host != "localhost") {
 	    flags |= AI_ADDRCONFIG;
 	}
+#ifdef AI_NUMERICSERV
 	flags |= AI_NUMERICSERV;
+#endif
 
 	struct addrinfo hints;
 	std::memset(&hints, 0, sizeof(struct addrinfo));


ojwb:
Thanks.  Is this with macos 10.5?  I guess you're still using that for powerpc support?

barracuda156:
@ojwb Thank you for merging!
Yes, this is primarily for PowerPC (where even 10.6 does not have it, I think: it was added in some of release SDK versions, but those lost ability to boot on PPC).
But this is also relevant for 10.5 on i386.

PR #340, opened 2026-05-09
add fallback UUID generation on Android

x6prl:
Currently I build Xapian for Linux/Web and Android. One of my Android devices has Android 8.1, on which SELinux policy bans the access to /proc/uuid and I get:
```
7505:05-09 22:53:41.631 13614 13614 W SDLThread: type=1400 audit(0.0:24): avc: denied { read } for name="uuid" dev="proc" ino=110567 scontext=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:proc:s0 tclass=file permissive=0
```

Yes, it is an **old** Android version, but supporting a fallback is not so heavy, I suppose.

### DIFF
--- a/xapian-core/backends/uuids.cc
+++ b/xapian-core/backends/uuids.cc
@@ -35,6 +35,9 @@
 
 #ifdef USE_PROC_FOR_UUID
 # include "safesysstat.h"
+#ifdef __ANDROID__
+#  include <stdlib.h> // for arc4random_buf()
+#endif // __ANDROID__
 #elif defined HAVE_UUID_UUID_H
 # include <exception>
 # include <uuid/uuid.h>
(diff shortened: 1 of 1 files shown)


ojwb:
It'd be nice to use `getentropy()` here - it seems to be fairly widely supported, and was standardised by POSIX.1-2024.  It has some advantages (like not requiring a free fd or access to `/proc` or `/dev`).

Unfortunately https://android.googlesource.com/platform/bionic.git/+/HEAD/libc/include/bits/getentropy.h says it's "Available since API level 28" - that seems to be slightly too new for you though according to: https://apilevels.com/

It seems `getrandom()` was added at the same time, so isn't a better equivalent here.

Could you double-check if you have `getentropy()` or `getrandom()` when targetting your Android 8.1 device?

x6prl:
I have checked `getentropy()` and it is not available on my device. But I found in the `sys/random.h` `getrandom()` source code comments that `arc4random_buf()` is available on every API level

---

Also, I didn't tested, but `arc4random_buf()` could be faster than handling `proc/uuid` and may be more preferable as a basic solution for UUID generation. And more portable.

`man` says:
```
These functions give cryptographically-secure pseudorandom numbers.
...
OpenBSD 2.1, FreeBSD 3.0, NetBSD 1.6, DragonFly 1.0, libbsd, glibc 2.36.
```

ojwb:
FWIW, speed really isn't a prime concern here as we only call this code when we create a new writable database, and only once per database created.

However it does seem simpler to just fill a buffer with suitably secure randomness and then set/clear a few bits, and getting UUIDs from the magic file under `/proc` has some annoying corner-case problems (e.g. `/proc` may not be mounted; needs an unused fd; might also be blocked by SELinux or similar on other platforms).  One can avoid using `/proc` by installing libuuid, but that's an extra dependency.  So I think I'm going to investigate using this approach more widely.

I'll merge this shortly as a first step.

x6prl:
Yeah, as you said, it is an advantage not having to free an fd or access anything with `open()`, and in that case, working with something like `arc4random_buf()` or other popular function gives us not only the simplicity of usage, but removes the headache about whether user's machine allows us touching `/proc`. I mean, not only on Android, but, as `man` says, _anywhere_.

---

Also, the initialization speed can be a thing of interest: I use `xapian` in an application, and making firstrun fast enough is important for the first impression) (In my code `xapian` init is not a problem, SDL and my code _IS_)))))))))

ojwb:
> Also, the initialization speed can be a thing of interest: I use `xapian` in an application, and making firstrun fast enough is important for the first impression) (In my code `xapian` init is not a problem, SDL and my code _IS_)))))))))

I'm not saying it's irrelevant, and it's good that avoiding `/proc/sys/kernel/random/uuid` is likely to be faster as well, but realistically any of these ways of generating a UUID is going to be a few milliseconds at worst (unless the system is so starved of entropy that it blocks waiting for entropy, but that will presumably affect any non-flawed way of generating a randomness-based UUID).

I've now merged the changes to generate UUIDs using `arc4random_buf()`, etc where available.

PR #341, opened 2026-06-01
configure.ac: drop deprecated `-Wstrict-overflow=1` `gcc` flag

trofi:
`gcc-17` formally deprecated `-Wstrict-overflow` options in https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=b31a43560e88600d412318f75a43ac8e562ba8c5. Informally the option was depracated since `gcc-8`.

I only noticed it because `gcc-17` deprecated it slightly incorrectly and started to fail `xapian-core` build: https://gcc.gnu.org/PR125558

While it's a `gcc` bug `gcc` devs suggested removing use of the flag. This change does the removal.

### DIFF
--- a/xapian-core/configure.ac
+++ b/xapian-core/configure.ac
@@ -1306,7 +1306,7 @@ if test yes = "$GXX"; then
       dnl -Wduplicated-branches was added in GCC 7.
       dnl
       dnl All the other options were supported by g++ 2.95.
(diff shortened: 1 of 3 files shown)


ojwb:
Thanks.

You've also dropped it for clang, but checking the clang docs it's a no-op there (for GCC compatibility) so that's fine.

Merged via cherry-pick on the command-line (and added a note about clang to the commit message).

PR #342, opened 2026-08-20
Windows: use explicit A-suffixed Win32 APIs (CreateProcessA(), CreateFileA()...)

medoc92:
Rationale: this is mostly a cosmetic change. Using the explicit name makes the code a bit clearer, and prevents a compilation error if UNICODE is defined in the context.

### DIFF
--- a/xapian-core/api/error.cc
+++ b/xapian-core/api/error.cc
@@ -56,7 +56,7 @@ Xapian::Error::get_error_string() const
             int e = abs(my_errno);
             DWORD len;
             char * error;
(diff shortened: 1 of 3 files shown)


ojwb:
Thanks!  Looks good, so cherry-picked via command line.  (I tweaked the indentation of function parameters for a few calls and reformatted the commit message a bit, but the changes should be the same except for whitespace.)

I noticed there are more calls to these in the testsuite and in omega - I've updated those as well and will push once the testsuite has run.

PR #343, opened 2026-08-30
Add Enquire::cancel() to abort a running local match

soloturn:
> **AI-assisted change proposal.** Filed by agent driven by @soloturn via [GDD](https://siliconsaga.github.io/yggdrasil/gdd/).

## Motivation

kiwix-android's cross-book ZIM search (kiwix/kiwix-android#4978) can currently only be cancelled between books, not between documents within a single book's Xapian search, because Xapian itself gives no way to interrupt a match in progress. This closes that gap at the source.

## Summary

- `Enquire::set_time_limit()` only disables the `check_at_least` optimization - the docs already say "There's not currently a way to force the match to end after a certain time." There is currently no way to abort a `get_mset()` call in progress at all.
- New `Xapian::MatchCancelledError` (`RuntimeError` subclass) and `Enquire::cancel()` - callable from another thread while `get_mset()` runs on this `Enquire` (or a copy sharing its state). Sets an `atomic<bool>` flag.
- `Matcher::get_mset()`/`get_local_mset()` check that flag once per candidate document in the match loop and throw `MatchCancelledError` when set - checked between documents, not instantaneous, same granularity `TimeOut` already uses.
- Local databases only. `remoteserver.cc`'s call passes `nullptr` for the new parameter - no client-side cancellation wire protocol for the remote backend, that's separate work.
- New test `matchcancel1` in `api_postingsource.cc`: a thread calls `cancel()` while a deliberately slow `PostingSource`-backed match runs on the main thread, asserts `MatchCancelledError` is thrown. First use of `std::thread` in this test suite.

## Test plan

- [x] Full `xapian-core` test suite via autotools with `--enable-maintainer-mode`: 3455 tests passed, 1 pre-existing expected failure (`checkstatsweight3`, unrelated to this change), 39 skipped.
- [x] `matchcancel1` passes on every applicable local backend (honey, inmemory, glass, singlefile_glass, multi_glass); correctly skipped for remote backends.

## Related

- kiwix/kiwix-android#4978

### DIFF
--- a/xapian-core/api/enquire.cc
+++ b/xapian-core/api/enquire.cc
@@ -196,6 +196,12 @@ Enquire::set_time_limit(double time_limit)
     internal->time_limit = time_limit;
 }
 
+void
+Enquire::cancel() const
+{
+    internal->cancel();
+}
+
 MSet
 Enquire::get_mset(doccount first,
                   doccount maxitems,
@@ -266,6 +272,10 @@ Enquire::Internal::get_mset(doccount first,
         return mset;
     }
 
+    // Each get_mset() call starts fresh - a cancel() from a previous call
+    // (or one requested before this call even started) shouldn't apply here.
+    cancel_requested.store(false, std::memory_order_relaxed);
+
     if (percent_threshold && (sort_by == VAL || sort_by == VAL_REL)) {
         throw Xapian::UnimplementedError("Use of a percentage cutoff while "
                                          "sorting primary by value isn't "
@@ -325,6 +335,7 @@ Enquire::Internal::get_mset(doccount first,
                                sort_by,
                                sort_val_reverse,
                                time_limit,
+                               &cancel_requested,
                                matchspies);
 
     if (first_orig != first) {
--- a/xapian-core/api/enquireinternal.h
+++ b/xapian-core/api/enquireinternal.h
@@ -31,6 +31,7 @@
 #include "xapian/mset.h" // Only needed to forward declare MSet::Internal.
 #include "xapian/query.h"
 
+#include <atomic>
 #include <memory>
 #include <string>
 #include <vector>
@@ -79,6 +80,9 @@ class Enquire::Internal : public Xapian::Internal::intrusive_base {
 
     double time_limit = 0.0;
 
+    /** Set by cancel(), checked in the match loop. See Enquire::cancel(). */
+    mutable std::atomic<bool> cancel_requested{false};
+
     enum { EXPAND_PROB, EXPAND_BO1 } eweight = EXPAND_PROB;
 
     double expand_k = 1.0;
@@ -92,6 +96,10 @@ class Enquire::Internal : public Xapian::Internal::intrusive_base {
                   doccount checkatleast,
                   const RSet* rset,
                   const MatchDecider* mdecider) const;
+
+    void cancel() const {
+        cancel_requested.store(true, std::memory_order_relaxed);
+    }
 
     TermIterator get_matching_terms_begin(docid did) const;
 
--- a/xapian-core/exception_data.pm
+++ b/xapian-core/exception_data.pm
@@ -175,6 +175,15 @@ errorclass(21, 'DatabaseClosedError', 'DatabaseError',
            'Indicates an attempt to access a closed database.',
            '');
 
+errorclass(22, 'MatchCancelledError', 'RuntimeError',
+           'Indicates a match was cancelled before it completed.',
+           <<'DOC');
+Thrown from Xapian::Enquire::get_mset() (or similar) if
+Xapian::Enquire::cancel() was called - on the same Xapian::Enquire object, or
+a copy of it sharing the same underlying state - while the match was still
+running.
+DOC
+
 sub for_each_nothrow {
     my $func = shift @_;
     foreach my $header ('include/xapian.h', <include/xapian/*.h>) {
--- a/xapian-core/include/xapian/enquire.h
+++ b/xapian-core/include/xapian/enquire.h
@@ -357,6 +357,15 @@ class XAPIAN_VISIBILITY_DEFAULT Enquire {
      */
     void set_time_limit(double time_limit);
 
+    /** Cancel a currently running match.
+     *
+     *  Call from another thread while get_mset() is running (on this
+     *  Enquire, or a copy sharing its state) to make it throw
+     *  Xapian::MatchCancelledError. Checked once per candidate document, not
+     *  instantaneous. Resets per get_mset() call. Local databases only.
+     */
+    void cancel() const;
+
     /** Run the query.
      *
      *  Run the query using the settings in this Enquire object and those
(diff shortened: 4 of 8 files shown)


soloturn:
@ojwb Would value your opinion here: does adding cancellable matches like this fit where you want Xapian's API to go, and does the approach (atomic flag checked once per candidate document in the match loop, local backends only) seem right to you? Context on the Kiwix side: full-text/title search in kiwix-android is backed by Xapian via libzim/libkiwix, and a user can currently only cancel a cross-book search between books, not mid-match within a single book's Xapian query - there's no hook to interrupt `get_mset()` once it's running. This PR is meant to close that gap. Happy to adjust the design based on your input before this goes further.

ojwb:
Thanks for the patch.  I should warn you that I'm not sure we want to take LLM-assisted patches at present.  We don't currently have an explicit policy on this though - the situation with LLMs is evolving rapidly and we get very few patches from outside the project.  Coming up with a policy takes time and effort and is not an enjoyable task, but that policy that might be out-dated before it was ever even used.

> Would value your opinion here: does adding cancellable matches like this fit where you want Xapian's API to go, and does the approach (atomic flag checked once per candidate document in the match loop, local backends only) seem right to you? Context on the Kiwix side: full-text/title search in kiwix-android is backed by Xapian via libzim/libkiwix, and a user can currently only cancel a cross-book search between books, not mid-match within a single book's Xapian query - there's no hook to interrupt `get_mset()` once it's running. This PR is meant to close that gap. Happy to adjust the design based on your input before this goes further.

The problem with handling the cancelling in the matcher loop is that we only return there when we find a matching document, which may happen infrequently, or never in the case of a query which doesn't match anything!  (Note that some queries can take a very long time to discover there are no matches.)  The user is most likely to want to cancel when a search is very slow, so the cancelling could sometimes take minutes to actually take effect, whereas really you want Xapian to stop working ASAP.

At another extreme, for a search where there are a lot of matching documents your approach will check the atomic flag for every single one.  The overhead of each individual check is clearly negligible, but the overhead of hundreds of millions of such checks might matter.  This cost is incurred whether or not the `cancel()` method is even used by the application, so would definitely need evaluating if we take this approach.

I think a better approach might be to run the match in its own thread internally, and then we can terminate that thread on a call to `cancel()`.  That should happen pretty much instantly, and the overhead for a match which isn't cancelled is only the overhead of creating and destroying that thread.  I suspect the hard part is avoiding leaking memory.  Perhaps using an arena allocator for everything that has the lifetime of the match process would neatly solve that, and also probably be faster as we can just deallocate the whole arena after the match.

Whatever mechanism is used internally, this feature seems to inherently require accessing the same Xapian object from multiple threads concurrently, which we explicitly document as [not supported](https://getting-started-with-xapian.readthedocs.io/en/latest/concepts/concurrency.html).  That could potentially be addressed by documenting this method as an exception to that policy, but I think that would need some very careful thought.  `Xapian::Enquire` (like most API classes) is a reference counted handle to an internal class, and we don't use thread locks around reference count updates, so a thread invoking `Xapian::Enquire`'s copy constructor or assignment operator while another thread is using the same object could potentially corrupt the reference count.  An exception that allowed calling `enquire.cancel()` could be easy to use in a way that seems to work but is actually not thread-safe.

Perhaps the thread safety issues could be side-stepped by saying the way to handle this situation is for the application to create a thread to run the match in, and then it can cancel that thread to cancel the match.  That might make not leaking memory even harder to pull off though.

soloturn:
Closing this - thanks for the review, @ojwb. Summary of where this landed:

**Your objections hold up.** Checked both independently:

- **Granularity**: correct that the flag is checked in the wrong place for the case that matters most - a slow/non-matching query burns time inside `pltree.next()` itself, which is exactly where this doesn't get a chance to check anything. Fixing that properly means pushing checks into the postlist tree's hot inner loops, not something this PR's shape can grow into.
- **Thread-safety**: also correct, and worse than I'd scoped it - `Enquire` being a ref-counted handle with an unsynchronized refcount means "call cancel() from another thread" invites exactly the unsafe pattern the docstring itself suggested ("a copy sharing its state"). Digging further, even the "run the match on its own thread and terminate it" alternative you floated has a sharper problem than the memory leak you flagged: a thread killed while holding an internal backend lock (glass/honey's block cache, an LRU, a cursor) wedges that lock forever - not just for the cancelled search, but for *any other search reusing that same open Database*, for the life of the process. Worse than the bug being fixed, so I'm not pursuing that direction either.
- **Overhead**: this one I can actually retire with numbers - reverted this commit in place, rebuilt with identical flags, and A/B'd a 500k-document match-everything query against both builds. Difference was ~19.0 ns/candidate either way, smaller than run-to-run jitter within the same binary. Not a real cost, for what that's worth going forward.

**The reason I'm closing rather than reworking the design**: while digging into the actual motivating use case (kiwix's search-as-you-type), I found the underlying need doesn't require this API at all. A `PostingSource` subclass that checks an app-owned flag and throws, combined with the real query via `OP_FILTER`, gets the same mid-match interruption on stock Xapian today - verified empirically, including through an `OP_AND`-combined query, with no `try`/`catch` anywhere in the match path swallowing or converting the exception. That's a viable path entirely at the application layer, with none of the thread-safety surface this PR would add to `Enquire` itself.

Given a working alternative exists outside Xapian's own API, and given the real design cost your review surfaced, this doesn't clear the bar to keep pursuing as a core API addition. Appreciate you taking the time to look at it either way.

ojwb:
> * **Thread-safety**: also correct, and worse than I'd scoped it - `Enquire` being a ref-counted handle with an unsynchronized refcount means "call cancel() from another thread" invites exactly the unsafe pattern the docstring itself suggested ("a copy sharing its state").

It occurred to me that this part could be addressed by having the public API be something like passing an `atomic<bool>*` (`NULL` unless the caller wants to be able to cancel).  Then the caller can safely set the pointed to value from another thread without thread-safety problems.

That doesn't solve where the cancelling is checked for though.

> Digging further, even the "run the match on its own thread and terminate it" alternative you floated has a sharper problem than the memory leak you flagged: a thread killed while holding an internal backend lock (glass/honey's block cache, an LRU, a cursor) wedges that lock forever - not just for the cancelled search, but for _any other search reusing that same open Database_, for the life of the process. Worse than the bug being fixed, so I'm not pursuing that direction either.

FWIW, the only backend lock is for a writer, and that's taken when a `WritableDatabase` is created and released when it is closed or destroyed, so there aren't any such locks to worry about.  I think it really is "only" memory allocations that need to not get leaked.

> A `PostingSource` subclass that checks an app-owned flag and throws, combined with the real query via `OP_FILTER`, gets the same mid-match interruption on stock Xapian today

That seems a good approach.  It should actually stop sooner than the patch here in some cases - e.g. for `OP_FILTER` with `OP_PHRASE` as a subquery, the and-like operations are combined and the positional check hoisted above that, so it can stop even when we don't return to the matcher main loop because the positional check is failing.

